PR #45, opened 2016-01-27
Made get_searchable_lane an external class method and added tests for…

leonardr:
This branch fixes and tests the behavior of the internal method get_searchable_lane() (now a property method of Lane called search_target), the code that determines which lane should actually be searched when the patron initiates a search in a given lane.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1025,17 +1025,33 @@ class Lane(object):
 
         return q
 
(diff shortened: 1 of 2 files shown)


leonardr:
This fixes https://github.com/NYPL-Simplified/circulation/issues/53

PR #46, opened 2016-01-28
Moved the helper method for finding the work associated with a loan o…

leonardr:
This moves the helper method I initially put in circulation into a property of Loan and Hold called .work

### DIFF
--- a/model.py
+++ b/model.py
@@ -423,7 +423,7 @@ class Patron(Base):
         return True
 
 
-class LoanAndCoreMixin(object):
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -427,7 +423,22 @@ def authorization_is_active(self):
         return True
 
 
-class Loan(Base):
+class LoanAndCoreMixin(object):
```
aslagle: Did you mean to name this LoanAndHoldMixin?

PR #49, opened 2016-02-01
Improve coverage provider

leonardr:
Previously, coverage providers were supposed to return a list of successes and a list of failures. This was awkward because there was no way to distinguish between transient failures (which should be retried later) and permanent failures (which should not).

This branch introduces the CoverageFailure class, which contains all relevant information about a failure to provide coverage. Instead of two lists, coverage providers now return one list containing both identifiers or editions (the successes), and CoverageFailure objects (the failures). 

If we ask a CoverageProvider to provide coverage for an identifier, and it does not show up as either a success or a failure, we now treat it as a transient failure. Previously we were liable to immediately try to cover that identifier again.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -69,17 +69,10 @@ class CoverageProvider(object):
             self._db, self.input_sources, self.output_source).order_by(
                 func.random())
 
-    def _identifier(self, o):
-        if isinstance(o, Identifier):
-            return o
-        elif isinstance(o, Edition):
-            return o.primary_identifier
-        else:
-            raise NotImplementedError()
-
     def run(self):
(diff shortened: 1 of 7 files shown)


coverage.py
```diff
@@ -52,47 +71,75 @@ def editions_that_need_coverage(self):
...
+            successes, failures, num_ignored
+        )
+
+        # Finalize this batch before moving on to the next one.
+        self.finalize_batch()
+
+        self.log.info(
+            "Batch processed with %d successes, %d failures, %d ignored.",
+            successes, failures, num_ignored
+        )
```
aslagle: you're logging the same thing before and after finalizing the batch

PR #53, opened 2016-02-03
Coverage scripts

leonardr:
This branch makes some changes to the coverage-provider code so that metadata and circulation can use the same code.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -257,11 +257,13 @@ class BibliographicCoverageProvider(CoverageProvider):
             metadata.apply(edition)
         except Exception as e:
             return CoverageFailure(self, identifier, repr(e), transient=True)
-        return identifier
 
         work = self.work(identifier)
         if isinstance(work, CoverageFailure):
             return work
+
+        return identifier
+
 
     def set_presentation_ready(self, identifier):
         """Set a Work presentation-ready."""
--- a/coverage.py
+++ b/coverage.py
@@ -161,7 +161,7 @@ class CoverageProvider(object):
             on_multiple='interchangeable',
         )
         if force or coverage_record is None:
-            result = self.process_identifier(identifier)
+            result = self.process_item(identifier)
             if isinstance(result, CoverageFailure):
                 return result.to_coverage_record()
             else:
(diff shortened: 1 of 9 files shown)


coverage.py
```diff
@@ -37,20 +37,20 @@ def to_coverage_record(self):
...
-    doesn't get processed next time.
+    """Run Identifiers of certain types (the input_identifier_types)
+    through code associated with a DataSource (the
+    `output_source`). If the code returns success, add a
+    CoverageRecord for the Edition and the output DataSource, so that
+    the record doesn't get processed next time.
     """
 
-    def __init__(self, service_name, input_sources, output_source,
+    def __init__(self, service_name, input_identifier_types, output_source,
```
leonardr: Consistently focus on identifiers, which is what we store in the CoverageRecord table. Instead of mapping a number of data sources to one other data source, we map a number of identifier types to a data source.

coverage.py
```diff
@@ -61,19 +61,24 @@ def log(self):
         return self._log        
 
     @property
-    def editions_that_need_coverage(self):
-        """Find all Editions lacking coverage from this CoverageProvider.
+    def items_that_need_coverage(self):
```
leonardr: A CoverageProvider that operates on Editions should still work, but we don't use 'editions' anymore in method names and by default we operate on Identifiers.

coverage.py
```diff
@@ -179,52 +185,37 @@ def finalize_batch(self):
...
     def process_batch(self):
         """Returns a list of successful identifiers and CoverageFailures"""
         raise NotImplementedError
 
-    def set_presentation_ready(self, identifier, metadata):
-        """Given a metadata object as content, finds an identifier's work
-        and sets it presentation ready.
-
-        Returns either a coverage failure or the processed identifier.
+    def work(self, identifier):
```
leonardr: I split set_presentation_ready into three methods: work(), set_metadata(), and set_presentation_ready(). This is because the metadata wrangler can't set a work as presentation-ready based on data from a single CoverageProvider.

threem.py
```diff
@@ -326,3 +327,19 @@ def value(threem_key):
 
 
 
+class ThreeMBibliographicCoverageProvider(BibliographicCoverageProvider):
```
leonardr: Moved this code in from circulation.

PR #55, opened 2016-02-04
Code to get edition, work, etc. is useful for other kinds of coverage…

leonardr:
… providers, not just BibliographicCoverageProviders.

This branch moves methods like .edition() from BibliographicCoverageProvider into CoverageProvider so that other types of providers can use it if they want.

Arguably this code should be moved into Identifier, but I'm not comfortable with that step yet.


courte:
:+1: on this.

PR #57, opened 2016-02-05
Make it possible to pass in a different Configuration object when loa…

leonardr:
…ding facets.

Otherwise a component that subclasses Configuration can't get load_facets_from_request to use its subclass--it will always use the core Configuration object.


aslagle:
:+1:

PR #60, opened 2016-02-08
Remove a provider from the list if it completes its work or does no w…

leonardr:
…ork during a run. Once all providers have completed, stop the script.


leonardr:
I fixed an error in my last commit, and fixed the real cause of the problem I was seeing: any random CoverageRecord for an ID will prevent other CoverageRecords for being created from that ID. I fixed this by adding some code and a test to Identifier.missing_coverage_from to bring it in line with Edition.missing_coverage_from.

PR #62, opened 2016-02-09
Implemented and tested LicensePool.open_access_links.

leonardr:
This branch fixes https://github.com/NYPL-Simplified/content_server/issues/42 by adding indexes to the materialized views. These indexes optimize a query for all books ordered by date added to the collection.

I also refactored a bit of code to create a method called LicensePool.open_access_links, a generator which yields every open-access link available for a LicensePool. This is necessary because open-access books from unglue.it might have books in multiple formats, and the content server doesn't necessarily know which one is the best.

### DIFF
--- a/model.py
+++ b/model.py
@@ -2474,17 +2474,16 @@ class Edition(Base):
                 similarity = self.similarity_to(candidate)
                 if similarity >= threshold:
                     yield candidate
+
     @property
     def best_open_access_link(self):
-        """Find the best open-access Resource for this LicensePool."""
-        open_access = Hyperlink.OPEN_ACCESS_DOWNLOAD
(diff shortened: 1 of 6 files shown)


model.py
```diff
@@ -5188,14 +5187,28 @@ def calculate_work(self, even_if_no_author=False, known_edition=None,
         return work, created
 
     @property
+    def open_access_links(self):
+        """Yield all open-access Resources for this LicensePool."""
+
+        open_access = Hyperlink.OPEN_ACCESS_DOWNLOAD
+        _db = Session.object_session(self)
+        best = None
+        best_priority = None
```
courte: Do you still need these two variables (`best` and `best_priority`)?

PR #69, opened 2016-02-16
Changed all HTTP Overdrive URLs to HTTPS.

leonardr:
I've tested this by doing starting up the application and looking at my loans, running the overdrive_monitor_full script, etc.


aslagle:
:+1:

PR #78, opened 2016-02-23
Make it possible to set works presentation-ready immediately upon imp…

leonardr:
…ort from an OPDS feed, assuming the OPDS feed provides a license for the book and basic metadata.

The motivating use case is to be allow a circulation manager to import a subset of books from the content server. If the content server publishes enough information to fit the book into an OPDS feed and make it available to the client, the books can be made presentation-ready immediately, without going to the metadata wrangler first. (Besides which, the metadata wrangler currently can't resolve a lot of the books that come from the content server.)

There's also a small unrelated change in opds.py, which handles the case where you try to generate a grouped OPDS feed for the top level of the site, but there aren't enough books on the whole site to generate a grouped feed. Previously the code crashed; now it shows you a paginated list of every book on the whole site. It's still not a great solution, but at least it doesn't crash.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -590,7 +590,7 @@ class Metadata(object):
         for format in self.formats:
             if format.link:
                 link = format.link
-                if self.rights_uri == None and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+                if self.rights_uri in (None, RightsStatus.UNKNOWN) and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
                     # We haven't determined rights from the metadata or the data source, but there's an
                     # open access download link, so we'll consider it generic open access.
                     self.rights_uri = RightsStatus.GENERIC_OPEN_ACCESS
--- a/model.py
+++ b/model.py
@@ -3402,7 +3402,9 @@ class Work(Base):
         self.presentation_ready_attempt = as_of
         self.random = random.random()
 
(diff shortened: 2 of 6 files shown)


metadata_layer.py
```diff
@@ -590,7 +590,7 @@ def set_default_rights_uri(self, data_source):
         for format in self.formats:
             if format.link:
                 link = format.link
-                if self.rights_uri == None and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+                if self.rights_uri in (None, RightsStatus.UNKNOWN) and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
```
aslagle: I might have had a reason for not putting RightsStatus.UNKNOWN here before, though we shouldn't have something with an open access link and unknown rights.

leonardr: I think if it was unknown before, and then you see an open-access link, it becomes generic open access. I can't think of any counterexamples.

aslagle: I think this would only happen if something we're importing has a rights string but we don't recognize it, so then it should be fine to make it generic open access if there's an open access link.

leonardr: The other possibility I can think of is that the source OPDS feed is mistaken or lying about the open-access status of the work. That is, the feed gives a link with rel="open-access" that points to a copy of a book that's actually not distributable. But that's a possibility even when the license is explicitly stated, so I don't think we can address that problem here.

PR #83, opened 2016-02-24
When filtering searches by target age, accept books that have no upper bound or lower bound on age, not just books whose upper or lower bound falls within the given ranges.

leonardr:


aslagle:
:+1:

PR #84, opened 2016-02-24
Google oauth

courte:
Creates models for admin accounts and Oauth credentials. It also creates a problem detail for invalid authorization identifiers. Also there's a small whitespace change.

### DIFF
--- a/model.py
+++ b/model.py
@@ -598,6 +598,7 @@ class DataSource(Base):
     ADOBE = "Adobe DRM"
     PLYMPTON = "Plympton"
     OA_CONTENT_SERVER = "Library Simplified Open Access Content Server"
+    GOOGLE = "Google Oauth"
 
     __tablename__ = 'datasources'
     id = Column(Integer, primary_key=True)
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -598,6 +598,7 @@ class DataSource(Base):
     ADOBE = "Adobe DRM"
     PLYMPTON = "Plympton"
     OA_CONTENT_SERVER = "Library Simplified Open Access Content Server"
+    GOOGLE = "Google Oauth"
```
aslagle: It seems weird for this to be in data sources. It's more like Millenium Patron than the book data sources, but I don't think there's anything in core for that.

aslagle: As I'm thinking about this more I do think it might be better to have it work more like patrons. I think we can say every circulation manager has to have a single admin auth provider, so we don't need to store the data source. We could parse the response from google into optional columns in the admins table, or store the json there.

courte: I was modeling it after the Adobe Vendor ID, but I've also probably overthought it since there's unlikely to be multiple admin authentication processes for a library (even if they're different).

PR #86, opened 2016-02-29
Introduce the concept of 'suppressed' licensepools

leonardr:
A 'suppressed' LicensePool is one that for whatever reason we just don't want to use until we perform a manual intervention to restore it. Maybe the book doesn't render, maybe the metadata is horribly broken, maybe we got a copy of a book that's not technically published yet... whatever. At some point, someone decides we need to temporarily remove a book from the collection. Suppressing the licensepool allows us to hide the book (at least from the bad source) without deleting it or creating inaccurate information about its presentation or licensing.

Up to this point we've been handling this kind of intervention by blanking out the available copies, marking works that are presentation ready as not presentation ready, etc. It's confusing and it causes unwanted side effects. This gives us an officially supported path to hiding individual books.

By default, a Work will not show up in lanes or search results unless it has at least one un-suppressed LicensePool.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -858,6 +858,8 @@ class Lane(object):
 
         * Have a delivery mechanism that can be rendered by the
           default client.
+
+        * Have an unsuppressed license pool.
         """
 
         q = self._db.query(Work).join(Work.primary_edition)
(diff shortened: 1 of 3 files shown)


leonardr:
Since the administrative interface _will_ want to show suppressed license pools, I've left in a path to turning off the filter.

PR #87, opened 2016-02-29
After refreshing the materialized views, vacuum the database.

leonardr:
It turns out the real problem behind https://github.com/NYPL-Simplified/circulation/issues/82 is that we keep changing the database, messing with indexes, refreshing the materialized views, etc. without running the VACUUM command. So I've changed the script that refreshes the materialized views to immediately follow up with a VACUUM.

Since VACUUM can't be executed inside a transaction block, and all our connections set up a transaction block automatically, this requires creating a brand new database connection, but this is at the very end of one script, so I think it's a better solution than changing the whole way we get connections.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -2,6 +2,7 @@ import os
 import logging
 import sys
 from nose.tools import set_trace
+from sqlalchemy import create_engine
 from sqlalchemy.sql.functions import func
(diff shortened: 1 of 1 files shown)


courte:
Fixes NYPL-Simplified/circulation#82.

PR #88, opened 2016-03-01
Improve metadata import

leonardr:
This branch offers a number of small improvements to improve the process of metadata import and debugging problems with metadata:

* By default, a CoverageProvider that calls metadata.apply() will completely replace the old set of subjects, links, and contributors previously obtained from the CoverageProvider. Without this branch, it's additive, so there's no way to get rid of bad data.

* Previously, the no-op when a book's classifications had not changed was achieved by deleting all the classifications and recreating them. This branch speeds things up by only deleting a classification if it's no longer in the list.

* Make it possible for a script to take a number of specific identifiers as command-line arguments and operate just on those Identifiers.

* Create a BibliographicRefreshScript which takes identifiers as command-line arguments and asks   Overdrive or 3M about those Identifiers, refreshing the local metadata to fit.

### DIFF
--- a/model.py
+++ b/model.py
@@ -3617,7 +3617,6 @@ class Work(Base):
         elif ya_score > threshold:
             audience = Classifier.AUDIENCE_YOUNG_ADULT
 
-
         # Remove any genres whose fiction status is inconsistent with the
         # (independently determined) fiction status of the book.
         #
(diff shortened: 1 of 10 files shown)


model.py
```diff
@@ -4515,7 +4514,9 @@ def __repr__(self):
             genre = ' genre="%s"' % self.genre.name
         else:
             genre = ""
-        if self.target_age is not None:
+        if (self.target_age is not None
```
leonardr: This handles the case where target_age is NumberRange(None, None)

PR #90, opened 2016-03-02
Complaints

skomputer:


opds.py
```diff
@@ -664,6 +664,31 @@ def search(cls, _db, title, url, lane, search_engine, query, limit=30,
...
+        facets = Facets.default()
+        pagination = pagination or Pagination.default()
+
+        q = Work.with_complaint(_db)
+        q = pagination.apply(q)
+
+        (works, types, counts) = zip(*q.all())
+        feed = cls(_db, title, url, works, annotator)
+
+        if len(works) > 0:
```
aslagle: Shouldn't this use the pagination limit somehow?

skomputer: i copied the page link logic in page(), where there is always an empty page at the end. let's see whether that should be changed too.

aslagle:
:+1:

Issue #92, opened 2016-03-07
Fix audience classification edge cases


leonardr:
There are a number of edge cases not caught by our existing classification rules:

* If the metadata wrangler classifies a book as 'Adult' but also shows a target age that implies it's a YA book, the content server will ignore the overall classification and classify the book as a YA book.

This is an active problem that affects real books in the NYPL collection.

* It's possible for a YA book to be classified as Adult because the classification 'vote' is split between YA and Children. That is, neither YA nor Children are heavy enough to outweigh Adult+Adults Only, but combined they would be heavy enough. In such a case we should classify a book as YA.

Unlike the first problem, this problem is mostly speculative at this point.

Issue #91, opened 2016-03-03
Tags that mention 'children' are not necessarily "Parenting" tags


leonardr:
The tag "Children: Young Adult (Gr. 7-9)" is correctly classified in terms of audience and target age, but incorrectly classified as the 'Parenting' genre (because it mentions 'Children').

PR #101, opened 2016-03-14
Improve metadata import

leonardr:
This branch fixes a number of issues to improve the classification of books:

https://github.com/NYPL-Simplified/metadata_wrangler/issues/35
https://github.com/NYPL-Simplified/server_core/issues/91
https://github.com/NYPL-Simplified/server_core/issues/92

The centerpiece of this branch is a refactoring of the classification code into a standalone class, WorkClassifier. I did this so I could write unit tests for a lot of code that didn't have tests before. Over the course of this work I found and fixed a number of minor problems.


classifier.py
```diff
@@ -3130,59 +3090,436 @@ def target_age(cls, identifier, name):
...
+            self._db = test_session
+        self.work = work
+        self.fiction_weights = Counter()
+        self.audience_weights = Counter()
+        self.target_age_relevant_classifications = set()
+        self.genre_weights = Counter()
+        self.direct_from_license_source = set()
+
+        # Start things off with information that can be derived from
+        # the Work itself, e.g. the publisher.
```
aslagle: is this comment in the right place?

classifier.py
```diff
@@ -3130,59 +3090,436 @@ def target_age(cls, identifier, name):
...
+            threshold = total_adult_weight * 2
+        else:
+            threshold = 10
+
+        # If both the 'children' weight and the 'YA' weight pass the
+        # threshold, we go with 'YA'.
+        if (children_weight > threshold and children_weight > ya_weight):
+            audience = Classifier.AUDIENCE_CHILDREN
+        elif ya_weight + children_weight > threshold:
+            audience = Classifier.AUDIENCE_YOUNG_ADULT
```
aslagle: this code doesn't match the comments

classifier.py
```diff
@@ -3130,59 +3090,436 @@ def target_age(cls, identifier, name):
...
+        for c in self.target_age_relevant_classifications:
+            score = c.quality_as_indicator_of_target_age
+            if not score:
+                continue
+            if (not highest_quality_score or score > highest_quality_score):
+                # If we gather a bunch of data, then discover a more reliable
+                # type of data, we need to start all over.
+                highest_quality_score = score
+                reliable_classifications = []
+            if score >= highest_quality_score:
```
aslagle: I think it would be clearer to say score == highest_quality_score since it should never be greater when it gets here.

classifier.py
```diff
@@ -3130,59 +3090,436 @@ def target_age(cls, identifier, name):
...
+        target_age_min = None
+        target_age_max = None
+        if target_age_mins:
+            # We have opinions about the proper minimum age.
+            mins = Counter(target_age_mins)
+            [(target_age_min, count)] = mins.most_common(1)
+            if count == 1:
+                # Everyone has a different opinion. Pick the smallest
+                # minimum age.
+                target_age_min = min(target_age_mins)
```
aslagle: what if there's a tie with count > 1?

classifier.py
```diff
@@ -3130,59 +3090,436 @@ def target_age(cls, identifier, name):
...
+
+        if target_age_min is None:
+            target_age_min = target_age_max
+
+        if target_age_max is None:
+            target_age_max = target_age_min
+
+        # If min and max got mixed up somehow, un-mix them.
+        if target_age_min > target_age_max:
+            target_age_min, target_age_max = target_age_max, target_age_min
```
aslagle: wouldn't it be better to do this earlier in the process, for each source, before taking a min or max of all the opinions?

classifier.py
```diff
@@ -3130,59 +3090,436 @@ def target_age(cls, identifier, name):
...
+
+        # Consolidate parent genres into their heaviest subgenre.
+        genres = self.consolidate_genre_weights(genres)
+        total_weight = float(sum(genres.values()))
+
+        # Strip out the stragglers.
+        for g, score in list(genres.items()):
+            affinity = score / total_weight
+            if affinity < cutoff:
+                total_weight -= score
```
aslagle: maybe use a separate variable for the new total weight. changing the total_weight and also using it to calculate affinity in this loop means that changing the order of the genres in the list can affect which ones get removed.

classifier.py
leonardr: Good question. I haven't really seen that come up but it is a possibility. I've refactored the code to take that into account.

classifier.py
leonardr: That's true, but this should never happen; it's just here as a backstop in case something goes wrong earlier in the process. So I'm not too worried about it.

I did change some of the code that deals with Subjects as they come in, so that a weird subject like "Ages 14-12" will be treated as saying the minimum age is 12 and the maximum age is 14. That's the only situation I can think of where this might have happened. I'd still like to keep the backstop though.

classifier.py
leonardr: Maybe I shouldn't be changing total_weight in the first place. What do you think?

aslagle: I guess you're not using it for anything else. I think it would be fine not to change it.

leonardr: I think the idea is that as the big genres are removed from the pool, the smaller genres have a better chance of meeting the cutoff. I'm not convinced that this is a good idea, but here's why I think I came up with it. To take an extreme example let's say it's:

Science Fiction 100000 (99%)
Romance 1000 (0.99%)
Western 1 (0.00099%)
Total: 101001

Clearly the primary classification here is Science Fiction. Western is obviously out. But is it fair to exclude Romance, which has a really high score, just because SF has an even higher score? I think my idea was to recalibrate what "high score" meant after each round.

In retrospect I think this idea needs work, at the very least. But my inclination is to leave it alone for now and have Courteney take a look at it.

aslagle: Maybe you'd want to do multiple rounds then, instead of modifying the total weight in a single round - remove everything below the cutoff using the original total weight, then calculate the new total weight and check again, until you don't modify anything.

leonardr: Yeah... at the same time, if you're going to assign percentage points to the genres, as WorkGenre does, the only logical thing to do here is to assign 99% to Science Fiction and 1% to Romance. So my impulse is to take this code out until we replace it with something better.

PR #107, opened 2016-03-16
Delete WorkGenres that are no longer accurate

leonardr:
Currently a Work that stops being classified under a Genre will retain its WorkGenre object for that genre, but the affinity will be set to zero. This branch removes the WorkGenre object altogether.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -1002,10 +1002,6 @@ class TestWork(DatabaseTest):
     def test_assign_genres_from_weights(self):
         work = self._work()
 
-        fantasy, ignore = Genre.lookup(self._db, Fantasy)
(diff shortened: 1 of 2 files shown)


tests/test_model.py
```diff
@@ -997,6 +999,24 @@ def test_set_presentation_ready(self):
         # TODO: there are some other things you can do to stop a work
         # being presentation ready, and they should all be tested.
 
+    def test_assign_genres_from_weights(self):
+        work = self._work()
+
+        fantasy, ignore = Genre.lookup(self._db, Fantasy)
+        sf, ignore = Genre.lookup(self._db, Science_Fiction)
+        romance, ignore = Genre.lookup(self._db, Romance)
```
courte: It doesn't look like you're using these genres anymore.

PR #111, opened 2016-03-17
Subjects classified under Erotica must also have an audience of "Adults Only"

leonardr:
I noticed that books classified under Erotica were also being classified as audience=Adult. They should be audience=Adults Only, in fact this is the only reason the Adults Only audience exists. But that only happened for Overdrive books, where the classifier's implementation of `.audience` was set up to work this way.

Rather than change all the individual classifiers, I changed the model code to add this restriction to all Subject objects.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -267,7 +267,7 @@ class WorkProcessingScript(IdentifierInputScript):
         works = True
         offset = 0
         while works:
-            works = self.query.offset(offset).limit(self.batch_size)
+            works = self.query.offset(offset).limit(self.batch_size).all()
             for work in works:
                 self.process_work(work)
             offset += self.batch_size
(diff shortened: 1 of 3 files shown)


scripts.py
```diff
@@ -267,7 +267,7 @@ def do_run(self):
         works = True
         offset = 0
         while works:
-            works = self.query.offset(offset).limit(self.batch_size)
+            works = self.query.offset(offset).limit(self.batch_size).all()
```
leonardr: This is an unrelated fix which stops this code from going into an infinite loop. Previously, `if works` always succeeded because `works` was always a `Query` object. This makes `works` into a list, which will eventually be empty.

aslagle:
:+1:

PR #114, opened 2016-03-18
Pass keyword args to monitors

courte:
This allows scripts to pass along keyword arguments when initialized and removes passed authentication values from ThreeMAPI initialization. Config files fo' life.

Related to NYPL-Simplified/circulation#108.


leonardr:
Awesome.

Issue #112, opened 2016-03-17
Clarify when a book becomes "erotica"


leonardr:
Most people understand the question of whether a book is "erotica" to be a binary one. People might disagree on whether a given book is "erotica", but the answer they give is either yes or no. 

In the Library Simplified system, "erotica" is a genre, which means that a book can be 27% or 50% or 33% erotica. It's common for books to be classified as 50% Erotica and 50% Romance. This is useful when displaying "erotic romance" books in both lanes, but it means that a book can have a significant "erotica" component yet be classified as "Adult" instead of "Adults Only".

I think it makes sense in our system for erotica to be treated as a genre, but there should probably be a system where a significant (how significant?) Erotica component in a Work's WorkGenres will force the work as a whole to be Adults Only, similarly to how, on a smaller scale, a Subject assigned to the Erotica genre is always assigned to the Adults Only audience.

leonardr:
Another possibility would be to kill the Erotica genre altogether and treat "Adults Only" and "Erotica" as _identical_. I think this is a bad idea.

PR #118, opened 2016-03-21
Don't boost 'adult' audience if there's an 'adults only' indicator in the data

leonardr:
This branch fixes at least one situation where bug #112 happens. I think it fixes the whole thing but I don't yet have ironclad evidence of this.

Currently, if the distributor of a book gives classifications, and none of them are assigned to the 'children' or 'young adult' audience, we give 'adult' a big boost. This is to handle the situation with 3M, which tells you when a book is a 'juvenile' book but says nothing in particular when a book is for adults.

But consider a book of erotica. 3M classifies the book under the Erotica genre, which is 'adults only'. But since there are no classifications under 'children' or 'young adult', we give 'adult' that big boost--a boost that outweighs the 'adults only' classification 3M explicitly gave us. This is how a book that is clearly erotica can end up being classified as 'adult' instead of 'adults only'.

This branch adds 'adults only' to the list of audiences that will stop the 'adult' boost from happening. The rest of the branch is minor stuff I added to make this scenario easier to debug.

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -3187,7 +3187,7 @@ class WorkClassifier(object):
     nonfiction_publishers = set(["Wiley"])
     fiction_publishers = set([])
 
-    def __init__(self, work, test_session=None):
+    def __init__(self, work, test_session=None, debug=True):
         self._db = Session.object_session(work)
         if test_session:
             self._db = test_session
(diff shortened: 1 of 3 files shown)


leonardr:
Looks like this doesn't fix the whole problem, but it's a good first step.

PR #119, opened 2016-03-21
Draw conclusions about target audience from genre where appropriate

leonardr:
Specifically, the Erotica genre implies the Adults Only audience.

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -3314,7 +3314,7 @@ class WorkClassifier(object):
         # Actually figure out the classifications
         fiction = self.fiction
         genres = self.genres(fiction)
-        audience = self.audience
+        audience = self.audience(genres)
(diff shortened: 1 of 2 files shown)


classifier.py
```diff
@@ -3300,9 +3303,6 @@ def prepare_to_classify(self):
             # distinguished by their _lack_ of childrens/YA
             # classifications.
             self.audience_weights[Classifier.AUDIENCE_ADULT] += 500
-        else:
```
leonardr: I took this out because it was boosting audience info just because the classification came from the distributor. I don't think that's necessary, it's not necessary here, and if we did want to do it we should do it by multiplying the weight that came into the system by 5 (or whatever) rather than just adding a big number to it.

aslagle:
:+1:

PR #120, opened 2016-03-21
Create search index client from calculate_presentation if it's not passed in.

aslagle:
I changed work.calculate_presentation to create an ExternalSearchIndex, so we won't forget to pass one in and end up with an outdated index. 

I also don't want to accidentally mess up a real search index by running tests without passing in the dummy one, so I'm using mock.patch to replace the ExternalSearchIndex class used by model for all tests.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -17,9 +17,6 @@ from model import (
 from metadata_layer import (
     ReplacementPolicy
 )
-from external_search import (
-    ExternalSearchIndex,
-)
 
 import log # This sets the appropriate log format.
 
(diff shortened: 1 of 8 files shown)


leonardr:
This looks good. Since you're adding a new dependency it'll be a little tricky to deploy.

aslagle:
It's only a dependency for running tests. Maybe we could split them up somehow?

leonardr:
I don't think it's a big deal, it's just a thing I want to be aware of.

aslagle:
I think if you deployed the code and ran it without installing the new dependency everything would still work fine. However, everyone will have to install it locally for the tests.

leonardr:
Yeah, you're right. The mock code would never be imported so there'd be no problem.

PR #122, opened 2016-03-22
added complaints_index() to Work

skomputer:


model.py
```diff
@@ -3695,6 +3695,12 @@ def _restrict_to_customlist_subquery_condition(
         qu = qu.filter(condition)
         return qu
 
+    def complaints_index(self):
+        index = dict({})
```
leonardr: You can use the Counter class from the collections module instead; it'll make this code look nicer.

model.py
```diff
@@ -3695,6 +3695,12 @@ def _restrict_to_customlist_subquery_condition(
         qu = qu.filter(condition)
         return qu
 
+    def complaints_index(self):
```
leonardr: I'm not totally opposed to putting this in the model, but it seems like it might fit better inside a controller-type class?

skomputer:
decided to move complaint counting back to a controller in the circulation manager.

PR #124, opened 2016-03-23
Moved in Axis 360 bibliographic coverage provider from circulation.

leonardr:
This branch moves the Axis 360 bibliographic coverage provider from circulation into core. The only reason we need this coverage provider is for the bibliographic repair script, and since that script is in core, this code needs to be in core as well--even though the circ manager is the only component that we'd reasonably expect to have Axis 360 credentials.

I've added a simple test for the helper method create_identifier_strings.

### DIFF
--- a/axis.py
+++ b/axis.py
@@ -10,6 +10,7 @@ import re
 
 from util import LanguageCodes
 from util.xmlparser import XMLParser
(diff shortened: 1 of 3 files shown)


aslagle:
:+1:

PR #125, opened 2016-03-23
Run on identifiers

leonardr:
This branch defines a new method of CoverageProvider, run_on_identifiers, which splits a predefined set of identifiers into batches and runs process_batch() on each batch. I implemented this after trying to  run a CoverageProvider script with a huge list of identifiers, which failed as it tried to process all the identifiers in one huge batch.


aslagle:
:+1:

PR #127, opened 2016-03-23
Control what data gets recalculated with a PresentationCalculationPolicy

leonardr:
This branch introduces PresentationCalculationPolicy, a class similar to the metadata layer's ReplacementPolicy, which lets you specify exactly which parts of a Work's (or Edition's) policy you would like to recalculate.

This branch also changes calculate_presentation() and all the methods it calls so that it keeps track of whether or not any of the presentation _actually changed_. Work.last_update_time and Edition.last_update_time now track the last time the presentation actually changed, not the last time we _tried_ to change something.

The follow-up to this branch will introduce coverage providers that keep track of when we last _tried_ to change something about a Work or Edition.

Don't merge this branch yet; I'm getting it reviewed because it's approaching the maximum reasonable size for a reviewable branch.

Other, smaller changes:
- Choosing a cover is now controlled by a separate policy from setting other types of metadata associated with the Edition such as sort_title and permanent_work_id.
- The output of Classifier.target_age is now always supposed to be a NumericRange object. I created Classifier.nr as a helper method to make this easier.
- I moved some code out of the presentation calculation method and implemented it more fully as Classifier.default_audience_for_target_age.
- I took out some code for adding a measurement of the total number of OCLC editions associated with a work. That code belongs in the metadata wrangler.

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -81,6 +81,11 @@ class Classifier(object):
     classifiers = dict()
 
     @classmethod
+    def nr(cls, lower, upper):
+        """Turn a 2-tuple into an inclusive NumericRange."""
+        return NumericRange(lower, upper, '[]')
+
+    @classmethod
     def lookup(cls, scheme):
         """Look up a classifier for a classification scheme."""
         return cls.classifiers.get(scheme, None)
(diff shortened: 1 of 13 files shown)


courte:
:muscle: Lookin' like a strong start, @leonardr!

PR #129, opened 2016-03-24
Improve audience classification

leonardr:
In addition to minor debug-message changes, this branch makes some improvements to classification based on real-world experience.

1. If there is absolutely no evidence that a book is for adults, then any amount of information that it's for YA / children is enough to determine its audience classification. We had a problem where books were being classified as 'adult' because although the evidence clearly indicated YA or children, the evidence didn't rise to the arbitrary level we set in the code.

The goal of the old system was to err against marking books that are for adults as children's books, and we need to be careful that doesn't happen on other collections, but now that we have better classifications, better data, and easy ways to complain about misclassification, I don't think it's as much of an issue.

2. The BISAC classifier (used by Axis 360) was based on the 3M classifier. It should probably be the other way around, but that's for later. The problem here is that 3M uses BISAC in such a way that both children's and YA books are filed under "Juvenile Fiction". Axis 360 uses BISAC properly, filing children's books under "Juvenile" and YA books under "Young Adult". But since we were using the broken 3M logic, "Young Adult" books had no implied audience and "Juvenile" books had an implied audience of Young ADult.

So I implemented BISACClassifier.audience to classify BISAC classifications properly, assuming they're being used properly.

### DIFF
--- a/axis.py
+++ b/axis.py
@@ -205,7 +205,7 @@ class Axis360BibliographicCoverageProvider(BibliographicCoverageProvider):
                     self._db, Identifier.AXIS_360_ID, identifier_string
                 )
                 result = CoverageFailure(
-                    self, identifier, "Book not in collection", transient=True
+                    self, identifier, "Book not in collection", transient=False
                 )
                 batch_results.append(result)
         return batch_results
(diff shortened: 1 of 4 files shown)


courte:
:+1:

PR #130, opened 2016-03-24
Work coverage record

leonardr:
This branch introduces WorkCoverageRecord, which is like CoverageRecord for works. 

I also change CoverageRecord.date (a Date) to a DateTime called CoverageRecord.timestamp, because it's sometimes important to keep track of exactly when, within a single day, a coverage record was updated.

And I add 'operation' to CoverageRecord, because we're at a point where sometimes there are two types of coverage we might want to give an Identifier, both of which come from the same data source.


courte:
:+1:

leonardr:
This has been merged into master.

PR #132, opened 2016-03-25
Use work coverage records

leonardr:
This branch completes the work-coverage-records saga by registering CoverageRecords and WorkCoverageRecords at various points in the presentation calculation process. This lets us finally keep track of exactly when (let's say) we last chose a cover for this edition or calculated the work's quality.

I also created __repr__ implementations for CoverageRecord and WorkCoverageRecord to make it easier to see what you're looking at in the middle of a pdb session.


courte:
Honestly this is so beautiful :+1:

PR #136, opened 2016-03-28
Tell metadata layer to mirror and scale resources

leonardr:
This branch makes the metadata layer capable of automatically downloading the resources mentioned in LinkData objects and mirroring them to an S3 bucket. Cover images are automatically thumbnailed, and the thumbnails are also mirrored.

Previously, downloading these resources and mirroring them was the responsibility of the code that called Metadata.apply(). This led to duplicate code, since this process--download, scale, mirror--always works the same way. Now all you have to do is pass your S3Uploader (or other mirror) object into your ReplacementPolicy constructor.

There are two smaller changes I made in the interests of having something that works in general:

1. The code for determining the filename to use when mirroring a resource has been made much more robust. It will give the best filename possible, depending on the original URL of the resource, the media type of the resource, or the link relation between the main Metadata object and the resource. If the resource is an open-access book, the filename is the title of the book (assuming we know that).

2. The code now uses CairoSVG to scale SVG cover images and convert them to PNG thumbnails. "But SVGs auto-scale!" you say. This is true, but SVGs can get really huge, and we find it useful to use tiny PNG thumbnails instead.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -1107,7 +1107,7 @@ class Metadata(object):
 
         # The thumbnail and image are different. Make sure there's a
         # separate link to the thumbnail.
-        thumbnail_obj, ignore = identifier.add_link(
+        thumbnail_obj, ignore = link_obj.identifier.add_link(
             rel=thumbnail.rel, href=thumbnail.href, 
             data_source=data_source, 
             license_pool=pool, media_type=thumbnail.media_type,
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -50,6 +50,7 @@ class ReplacementPolicy(object):
             formats=False,
             rights=False,
             mirror=None,
+            http_get=None,
             even_if_not_apparently_updated=False,
     ):
         self.identifiers = identifiers
@@ -60,6 +61,7 @@ class ReplacementPolicy(object):
         self.formats = formats
         self.even_if_not_apparently_updated = even_if_not_apparently_updated
         self.mirror = mirror
+        self.http_get = http_get
 
     @classmethod
     def from_license_source(self, mirror=None, even_if_not_apparently_updated=False):
@@ -959,7 +961,7 @@ class Metadata(object):
                 # We need to mirror this resource. If it's an image, a
                 # thumbnail may be provided as a side effect.
                 self.mirror_link(
-                    pool, data_source, link, link_obj, replace.mirror
+                    pool, data_source, link, link_obj, replace
                 )
             elif link.thumbnail:
                 # We don't need to mirror this image, but we do need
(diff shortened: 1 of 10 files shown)


metadata_layer.py
```diff
@@ -807,7 +811,11 @@ def apply(self, edition, metadata_client=None, replace=None,
             replace_rights=False,
```
aslagle: Looks like most of the arguments to this method should be removed now?

aslagle:
What happens when we import something that has a thumbnail already? We create a thumbnail when we mirror the cover image, but then we also connect the other thumbnail to the main image. Which one gets used?

aslagle:
Or I guess we don't use the original thumbnail at all?

model.py
```diff
@@ -5884,8 +5902,16 @@ def get(cls, _db, url, do_get=None, extra_request_headers=None,
...
         # Do we already have a usable representation?
+        #
+        # 'Usable' means we tried it and either got some data or
+        # received a status code outside the 2xx series.
         usable_representation = (
-            representation and not representation.fetch_exception)
+            representation and not representation.fetch_exception
+            and (
+                representation.content or representation.local_path
+                or representation.status_code / 100 != 2
```
aslagle: shouldn't this be `== 2`

leonardr: `!= 2` is what I meant, but on reflection I think it's wrong.

If the status code is in the 200 range then we get some .content. I'm trying to capture the fact that even status codes outside the 200 range can be cached. 

For example when we look up an ISBN in OCLC+LD we get a 302 redirect. I'd like to reuse that redirect next time instead of making the HTTP request again.

Some OCLC+LD files give a 404 error. I'd like to cache the 404 error on the same terms as I would cache a successful response.

This is what `test_302_creates_cachable_representation` tests.

I think what I actually want here is to verify that the status code was not in the 5xx range, because that's the only time you might reasonable expect that making the request again would give you a different result.

metadata_layer.py
leonardr: Yes, they should be, but I don't want to take them out until I have a chance to change all the calls to Metadata.apply() in circulation, metadata and content.

leonardr:
Re: which thumbnail we use. There are two scenarios.

1. When we are importing from a license source into the content server or the metadata wrangler, we should always generate our own thumbnail. Even if the license source provides a thumbnail, there's no guarantee it's the right size. The only time we should even look at a thumbnail is if the license source provides a thumbnail but no original image. In that case we should treat the thumbnail as though it were an individual image.

2. When we are importing into a circulation manager, we should always use the thumbnail provided by the data source. Either we will get a better thumbnail later (e.g. when we first find out about an Overdrive book) or we can trust the data source to have taken care of the thumbnail process (e.g. when we are talking to content server or metadata wrangler).

Case #1 corresponds to the case where an S3Uploader is provided, and Case #2 corresponds to the case where there is no S3Uploader. Case #1 is tested by TestOPDSImporterWithS3Mirror.test_resources_are_mirrored_on_import. I can make a pretty minor change to the OPDS feed to show that an incoming thumbnail is ignored.

I don't think I have a good test case for Case #2. I'll look into it.

Does this make sense?

aslagle:
:+1: That makes sense, and everything looks good to me as long as the tests pass.

PR #137, opened 2016-03-29
Minor fixes to get content server scripts to work.

leonardr:


model.py
```diff
@@ -2638,6 +2638,7 @@ def best_open_access_link(self):
         for resource in pool.open_access_links:
             if not any(
                     [resource.representation and
+                     resource.representation.media_type and
                      resource.representation.media_type.startswith(x)
```
leonardr: Standard Ebooks has some open-access links that have unrecognized media types, which we ignore. This doesn't solve the problem of what to do with those files but it does stop a crash.

opds_import.py
```diff
@@ -696,7 +696,9 @@ def run_once(self, start, cutoff):
 class OPDSImporterWithS3Mirror(OPDSImporter):
     """OPDS Importer that mirrors content to S3."""
 
-    def __init__(self, _db, **kwargs):
+    def __init__(self, _db, default_data_source, **kwargs):
```
leonardr: TestOPDSImporterWithS3Mirror doesn't actually instantiate this class (it instantiates OPDSImporter with a dummy S3Uploader object), so I didn't see that this constructor had the wrong signature.

aslagle:
:+1:

PR #139, opened 2016-03-30
Fix age range search

leonardr:
This branch fixes bug #138 and adds a test to detect a future reversion, which makes sure that Lane data is properly converted into a search filter. The test isn't great--it would be better if there were a method you could pass the Lane into and get back the search filter--but I think it's good enough for now.


aslagle:
:+1:

PR #141, opened 2016-03-30
Search improvements

aslagle:


tests/test_external_search.py
```diff
@@ -2,10 +2,488 @@
...
+        fiction_summary.summary_text = "A book about a dog"
+        fiction_summary.set_presentation_ready()
+        fiction_summary.update_external_index(self.search)
+
+        nonfiction_summary = self._work(fiction=False)
+        nonfiction_summary.summary_text = "A book about a real dog"
+        nonfiction_summary.set_presentation_ready()
+        nonfiction_summary.update_external_index(self.search)
+
+        time.sleep(1)
```
leonardr: Is there any way to reduce the amount of time we have to wait for the index to get updated? There are a lot of these tests and they all sleep for a second.

Or maybe we could create the entire index in the module setup so that we only have to sleep once? I'm not sure what's the best way to handle this.

external_search.py
```diff
@@ -50,6 +47,69 @@ def __init__(self, url=None, works_index=None):
...
+                                },
+                            }
+                        }
+                    }
+                },
+            )
+        
+            mapping = {"properties": {}}
+            for field in ["title", "series", "subtitle", "summary"]:
+                mapping["properties"][field] = {
```
leonardr: Flagging this as something I don't understand and would like to talk about in person. Why do you use both the en_analyzer and the en_minimal_analyzer here?

external_search.py
```diff
@@ -49,6 +49,15 @@ def __init__(self, url=None, works_index=None):
...
     def setup_index(self):
+        """
+        Create the search index with appropriate mapping.
+
+        This will destroy the search index, and all works will need
+        to be indexed again. In production, don't use this on an
+        existing index. Use it to create a new index, then change the 
+        alias to point to the new index.
+        """
+
```
leonardr: How do you use this to create a new index if it always operates on self.works_index? That seems unreliable. Would it make sense to have a method that orchestrates the whole create-a-new-index and change-the-alias process?

PR #142, opened 2016-03-30
Reduce our dependence on foreign medians

courte:
Numpy is a huge library that takes a really long time to load, lengthening test and deploy times everywhere. And we only use it in one place in our code base. Well, **no more**! Here's a median of our own design, with our *own* best interests in mind.

The expected values in the test were created by running the same sets of data in numpy's median function.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -19,6 +19,7 @@ import csv
 import datetime
 import logging
 from util import LanguageCodes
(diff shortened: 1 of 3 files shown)


courte:
:us: :fireworks: :1234: :us:

leonardr:
:us:

PR #147, opened 2016-03-31
Store Unicode in Work.summary_text

leonardr:
This branch fixes a bug where Work.summary_text, despite being a Unicode field, was being filled with binary data from Representation.content. This caused a crash when we tried to index the binary data in Elasticsearch.

This branch ensures that Work.summary_text is set to a Unicode value. If we can't figure out a Unicode value from Representation.content, we would have no value at all then a value that can't be indexed.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -113,7 +113,7 @@ class CoverageProvider(object):
         batch = self.items_that_need_coverage.limit(
             self.workset_size).offset(offset)
 
-        if not batch:
+        if not batch.count():
             # The batch is empty. We're done.
             return None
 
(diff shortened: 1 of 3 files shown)


coverage.py
```diff
@@ -113,7 +113,7 @@ def run_once(self, offset):
         batch = self.items_that_need_coverage.limit(
             self.workset_size).offset(offset)
 
-        if not batch:
+        if not batch.count():
```
leonardr: This is a Query object, not a list. I noticed that CoverageProvider scripts were running forever even after they were done.

aslagle:
:+1:

PR #148, opened 2016-03-31
If an exception happens during OPDS import, don't let it bring down the entire import script

leonardr:
Instead, treat it as an error that only affects the book being imported.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -8,6 +8,7 @@ import datetime
 import feedparser
 import logging
 import requests
+import traceback
 import urllib
 from urlparse import urlparse, urljoin
 from sqlalchemy.orm.session import Session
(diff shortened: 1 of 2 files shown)


opds_import.py
```diff
@@ -157,60 +158,79 @@ def import_from_feed(self, feed, even_if_no_author=False,
...
+                # Rather than scratch the whole import, treat this as a failure that only applies
+                # to this item.
+                identifier, ignore = metadata.primary_identifier.load(self._db)
+                message = StatusMessage(500, "Local exception during import:\n%s" % traceback.format_exc())
+                messages[identifier.urn] = message
+
+        return imported, messages, next_links
+
+
+    def import_from_metadata(
```
leonardr: All this 'new' code is just the old inner loop from import_from_metadata(), moved into its own method and dedented.

aslagle:
:+1:

PR #149, opened 2016-04-01
Some CoverageProviders can autocreate missing LicensePools

leonardr:
Currently if a CoverageProvider tries to figure out the Edition for an Identifier it's processing, and there is no LicensePool for that Identifier, the CoverageProvider automatically fails, because we can't get from Identifier to Edition unless we also have a DataSource, which we currently get through LicensePool.

This isn't a problem on circulation managers, where we create a LicensePool as soon as we find out about the book. But it's a problem on the metadata wrangler, where we find out about books through UnresolvedIdentifier objects. LicensePools are only necessary on metadata because the data model currently requires them, but making sure they exist is a lot simpler than making the data model not require them.

This branch changes certain CoverageProviders -- in particular, the BibliographicCoverageProviders that ask Overdrive and 3M about books -- so that they can autocreate missing LicensePools for the Identifiers they process. This makes the identifier resolution process work on the metadata wrangler, so long as the BibliographicCoverageProvider is the first one to run. Since only the BibliographicCoverageProviders have this power, there's no risk that we'll create LicensePools for an Identifier with a DataSource that doesn't actually offer licenses.

This branch also changes `Script.parse_identifier_list` so that certain scripts can autocreate _identifiers_ that are mentioned on the command line but not already in the database. For most scripts, an unrecognized identifier on the command line is probably a typo. But I sometimes use a script like metadata/bin/identifiers_resolve to stick a new identifier into the system, so that script needs to autocreate identifiers if they don't exist.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -12,6 +12,7 @@ from model import (
     DataSource,
     Edition,
     Identifier,
+    LicensePool,
     Timestamp,
 )
 from metadata_layer import (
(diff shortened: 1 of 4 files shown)


courte:
:+1:

PR #150, opened 2016-04-01
Fix a bug that occured when a work had measurements of rating and qua…

leonardr:
…lity but not popularity.

The bug was pretty simple: an 'if' instead of an 'elif'. If `popularity` is None but `rating` and `quality` are not, we ended up trying to calculate `(popularity * popularity_weight) + (rating * rating_weight)`, and multiplying None by a number gave a TypeError.

I added a test for this case and a few related cases because I thought the coverage here was a little light.

### DIFF
--- a/model.py
+++ b/model.py
@@ -4084,12 +4084,15 @@ class Measurement(Base):
             # Our idea of the quality depends entirely on the work's quality scores.
             return quality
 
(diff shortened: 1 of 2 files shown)


courte:
Nice catch! :+1:

PR #152, opened 2016-04-01
Improved use of metadata policy

leonardr:
This branch is kind of obnoxious but I was able to come up with decent tests for it. Basically, I was having problems with the CoverageProviders' interactions with the metadata layer.

1. When the circ manager first hears about an Overdrive book it's supposed to start using Overdrive's thumbnail image until it gets our version of the thumbnail from the metadata wrangler. This wasn't happening. The full-size image (which we don't use) was being recorded, but its thumbnail was not. I determined this was because although we identify the thumbnail image as such, we don't record the fact that it's a thumbnail _of the main image_, so the metadata layer ignores it. I fixed this, and I also noticed that Overdrive now provides some larger-size thumbnails that are closer to the size we use, so I check for those before settling for the (greatly inferior) "thumbnail" image.

2. When the metadata wrangler's IdentifierResolutionMonitor hears about cover images from 3M and Overdrive, it's supposed to automatically mirror and scale them. This was not happening because although the IRM creates an OverdriveBibliographicCoverageProvider and a ThreeMBibliographicCoverageProvider, it wasn't passing in the custom MetadataReplacementPolicy that includes the essential information about how to 'mirror' an image. That was because Overdrive and TheeMBibliographicCoverageProvider didn't accept that constructor argument. I added the constructor argument to those classes and made sure the MetadataReplacementPolicy was used in the crucial set_metadata() call.

3. Once I fixed #1 and #2, re-running the scripts would solve the database problem, but the OPDS entries for the broken books wouldn't get updated, because the metadata layer didn't see that anything had changed. So I changed MetadataReplacementPolicy to take a PresentationCalculationPolicy object as an argument, and changed Metadata.apply() to use the PresentationCalculationPolicy when it calls Work.calculate_presentation() or Edition.calculate_presentation().

This lets me set up a repair script that fixes the data for a book and then forces an update for its OPDS entry and search index entry. Previously this would have required two different scripts.

### DIFF
--- a/axis.py
+++ b/axis.py
@@ -177,11 +177,12 @@ class Axis360BibliographicCoverageProvider(BibliographicCoverageProvider):
     not normally necessary because the Axis 360 API combines
     bibliographic and availability data.
     """
-    def __init__(self, _db):
+    def __init__(self, _db, metadata_replacement_policy=None):
         self.parser = BibliographicParser()
         super(Axis360BibliographicCoverageProvider, self).__init__(
             _db, Axis360API(_db), DataSource.AXIS_360,
-            workset_size=25
+            workset_size=25, 
+            metadata_replacement_policy=metadata_replacement_policy
         )
 
     def process_batch(self, identifiers):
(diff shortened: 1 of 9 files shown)


courte:
:+1: Primo stuff here!

PR #153, opened 2016-04-04
Add run script

courte:
Adds a `run` script that can be used to run any Library Simplified script and confirm it's not already in-progress on a server / computer. Scripts run without the `core/bin/run <script name> <script arguments>` format will not be tracked, and thus may still run simultaneously.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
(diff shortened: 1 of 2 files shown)


leonardr:
Looks good. I think the $LIBSIMPLE_DIR variable isn't currently the way we do things, but we'll figure it out as needed.

PR #155, opened 2016-04-04
Target age improvements

leonardr:
This branch makes a large number of minor improvements to the data gathering and classification process, mainly to improve the treatment of children's/YA books obtained from Overdrive.

There are a number of more complicated changes I could make, which would help in some edge cases, but these relatively minor changes seem to solve all the common cases, so I'm going to re-run a bunch of code and see how many edge cases there really are.


classifier.py
```diff
@@ -295,7 +295,9 @@ class GradeLevelClassifier(Classifier):
 
     @classmethod
     def audience(cls, identifier, name, require_explicit_age_marker=False):
-        young, old = cls.target_age(identifier, name, require_explicit_age_marker)
+        target_age = cls.target_age(identifier, name, require_explicit_age_marker)
```
leonardr: This was a bug created when I switched from `target_age` always returning a 2-tuple to making it always return a NumericRange.

classifier.py
```diff
@@ -1685,8 +1693,15 @@ class KeywordBasedClassifier(AgeOrGradeClassifier):
     JUVENILE_INDICATORS = match_kw(
         "for children", "children's", "juvenile",
         "nursery rhymes", "9-12")
-    YOUNG_ADULT_INDICATORS = match_kw("young adult", "ya", "12-Up", 
-                                      "teenage fiction")
+    YOUNG_ADULT_INDICATORS = match_kw(
```
leonardr: I noticed that keywords based on 'teen' language were not being treated as indicating Young Adult even when that was clearly the case.

classifier.py
```diff
@@ -3223,6 +3252,7 @@ class WorkClassifier(object):
         "Harlequin Historical" : Historical_Romance,
         "Harlequin Historical Undone" : Historical_Romance,
         "Frommers" : Travel,
+        "LucasBooks": Media_Tie_in_SF,
```
leonardr: This fixes the Star Wars problem I mentioned in issue 134.

overdrive.py
```diff
@@ -615,6 +615,13 @@ def book_info_to_metadata(cls, book):
                         )
             )
 
+        for grade_level_info in book.get('gradeLevels', []):
```
leonardr: Overdrive provides detailed recommendations about target grade levels for some children's books, but we weren't recording this information.

tests/test_classification.py
```diff
@@ -555,8 +597,18 @@ def a(x, y):
             return v.lower, v.upper
         eq_((0,4), a("Picture Book Nonfiction", None))
         eq_((5,8), a("Beginning Reader", None))
+        eq_((12,17), a("Young Adult Fiction", None))
         eq_((None,None), a("Fiction", None))
 
+    def test_audience(self):
```
leonardr: Just a test I noticed was missing.

courte:
Tests are still failing right now, but otherwise 👍 .

PR #156, opened 2016-04-05
Fix juvenile classifications

leonardr:
This branch fixes a couple of problems we had classifying 3M books. The biggest one is that certain classifications are now treated as generic 'juvenile' classifications. Their weight is divided up between the 'children' and 'YA' audiences (instead of all being assigned to YA). This makes it easier for data from some other source to set the record straight if it is in fact a children's book.

I also added a fair amount of debug code that's useful when tracking down this sort of problem. None of this code runs unless you change debug=False in the WorkClassifier constructor.

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -12,6 +12,7 @@
 # SQL to find commonly used classifications not assigned to a genre 
 # select count(identifiers.id) as c, subjects.type, substr(subjects.identifier, 0, 20) as i, substr(subjects.name, 0, 20) as n from workidentifiers join classifications on workidentifiers.id=classifications.work_identifier_id join subjects on classifications.subject_id=subjects.id where subjects.genre_id is null and subjects.fiction is null group by subjects.type, i, n order by c desc;
 
+import logging
 import json
 import os
 import pkgutil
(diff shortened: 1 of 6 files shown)


classifier.py
```diff
@@ -1523,6 +1533,9 @@ def audience(cls, identifier, name):
         if isinstance(identifier, basestring) and identifier.startswith('Y'):
             return cls.AUDIENCE_YOUNG_ADULT
 
+        if isinstance(identifier, basestring) and identifier=='FIC':
+            # FIC is used for all types of fiction.
+            return None
```
leonardr: Previously FIC was treated as an indicator of adult fiction, which is not how FIC is used in the real world.

classifier.py
```diff
@@ -3305,11 +3318,13 @@ def __init__(self, work, test_session=None, debug=False):
         self.prepared = False
         self.debug = debug
         self.classifications = []
+        self.log = logging.getLogger("Classifier (workid=%d)" % self.work.id)
```
leonardr: This is all that debug code.

overdrive.py
```diff
@@ -619,7 +619,7 @@ def book_info_to_metadata(cls, book):
             grade_level = grade_level_info.get('value')
             subjects.append(
                 SubjectData(type=Subject.GRADE_LEVEL, identifier=grade_level,
-                            weight=1)
+                            weight=100)
```
leonardr: Overdrive grade levels turn out to be about as reliable as other Overdrive data.

threem.py
```diff
@@ -209,7 +209,7 @@ def parse_genre_string(self, s):
             if not i:
                 continue
             i = i.replace("&amp;amp;", "&amp;").replace("&amp;", "&").replace("&#39;", "'")
-            genres.append(SubjectData(Subject.THREEM, i))
+            genres.append(SubjectData(Subject.THREEM, i, weight=15))
```
leonardr: We were giving 3M classifications a weight of 1. 3M classifications aren't great, but they're a lot better than random tags.

courte:
👍

PR #157, opened 2016-04-05
Dont remove classifications from another source

leonardr:
This branch fixes a bug that only showed up when you got classifications from two different sources. Calling metadata.apply() on the second data source would destroy all the classifications created by the first source, and vice versa.


aslagle:
:+1:

PR #158, opened 2016-04-05
Make both implementations of missing_coverage_from take the operation into account

leonardr:
This fixes a bug where an attempt to find a coverage record succeeded when it shouldn't have, because there was a coverage record for the same data source and identifier, but for some other operation.


aslagle:
:+1:

PR #161, opened 2016-04-06
Fix a number of edge cases in target age classification

leonardr:
This branch fixes a number of problems with the algorithms that figures out which age range a string is talking about. 

1. We weren't properly handling cases like "Grades 8-5" or "Ages 14-12", where the older age comes before the younger age. (I had code for handling this but took it out, not understanding where the problem was.)
2. We didn't understand a grade number that included a leading zero, a la "Grade 05"

I've added tests for these cases, and I've run the algorithm on a large number of real-world age range strings without encountering any further problems.


courte:
👍

Issue #163, opened 2016-04-06
Target age and audience sometimes have no overlap


leonardr:
Although our code which calculates a target age for a work takes the predetermined audience into account, sometimes the target age has no overlap whatsoever with the predetermined audience.

For reference, we consider "Young Adult" as covering the years 14-17, and "Children" as covering everything younger. Other sources consider "Young Adult" as starting at 12.

"Captains Courageous" is classified as a children's book for 18-year-olds:
http://metadata.alpha.librarysimplified.org/lookup?urn=urn%3Alibrarysimplified.org%2Fterms%2Fid%2FOverdrive%2520ID%2Feffc5791-dea8-4bc9-86aa-818ea63bf8fd

"Dandelion Wine" is classified as a children's book for 14-16:
http://metadata.alpha.librarysimplified.org/lookup?urn=urn%3Alibrarysimplified.org%2Fterms%2Fid%2FOverdrive%2520ID%2Fe6524ccc-676c-4f25-9351-7f0774c48a84

"Petra K and the Blackhearts" is classified as a children's book for 14-17:
http://metadata.alpha.librarysimplified.org/lookup?urn=urn%3Alibrarysimplified.org%2Fterms%2Fid%2FOverdrive%2520ID%2F40b2cdc4-8ef6-409e-9432-6e12c54153dc

"What My Mother Doesn't Know" is a YA book for ages 9-13:
http://metadata.alpha.librarysimplified.org/lookup?urn=urn%3Alibrarysimplified.org%2Fterms%2Fid%2FOverdrive%2520ID%2Fa401a71d-4bfb-4a2e-92bc-8e50da0122b7

"Riding Invisible" is a YA book for ages 9-12
http://metadata.alpha.librarysimplified.org/lookup?urn=urn%3Alibrarysimplified.org%2Fterms%2Fid%2FOverdrive%2520ID%2F05ee675e-8af8-4e0b-8c39-95b317df38ca

It's okay for a book to have some overlap. For instance, a YA book might be suitable for ages 13-15, or a children's book might be suitable for ages 10-14. But a) we don't have a good system for dealing with that overlap, and b) clearly we sometimes get into situations where there is absolutely no overlap, but we come up with these nonsensical results anyway.

leonardr:
Every time I come back to this issue I fix a bunch of underlying problems and a lot of the weird cases go away. At this point I think the only weird cases are ones where:

* There is a trusted assertion that the book is YA
[...]

Issue #162, opened 2016-04-06
Overdrive "Juvenile Fiction" alone implies ages 9-12


leonardr:
When Overdrive classifies a book as "Juvenile Fiction" or "Juvenile Nonfiction", but does not also have a more specific classification like "Picture Books" or "Beginning Readers", it's pretty good evidence that the book is suitable for ages 9-12. We don't use it as such because our system is not very good at drawing conclusions from a piece of information plus an _absence_ of some other information.

I don't think this is terribly important because most of the time Overdrive provides a specific set of grade levels or an interest level. Pretty soon we should be in a good position to measure how many books would have their classification improved by this code.

leonardr:
It turns out that Overdrive's list of grade levels are a measure of reading level, not interest level, so this became significantly more important.

PR #164, opened 2016-04-06
Improve classifications

leonardr:
Another big batch of improvements to the way target ages are calculated and the target age for a work is determined. This fixes issue #162 and does a lot of work for issue #163.

The biggest change is that when calculating target age we now consider all Classifications that have an opinion about target age, instead of only considering the "most reliable subset" of classifications. However, this doesn't mean we suddenly consider all classifications equally reliable. The numbers associated with the old hierarchy of reliable/unreliable classifications have been turned into coefficients. We consider all classifications, but we multiply the weight of each classification by the coefficient associated with its source. This lets us make the right decision in situations where the "most reliable subset" is wrong about something and all the "less reliable" data points in a different direction.

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -92,45 +92,6 @@ class Classifier(object):
         return NumericRange(lower, upper, '[]')
 
     @classmethod
-    def default_audience_for_target_age(cls, nr):
-        if nr is None:
-            return None
-        lower = nr.lower
-        upper = nr.upper
-        if not lower and not upper:
-            return None
-        if lower and not nr.lower_inc:
-            lower += 1
-        if upper and not nr.upper_inc:
-            upper += 1
-        if not lower:
-            if upper > 18:
-                # e.g. "up to 20 years", though that doesn't
-                # make much sense.
-                return cls.AUDIENCE_ADULT
-            elif upper > cls.YOUNG_ADULT_AGE_CUTOFF:
-                # e.g. "up to 15 years"
-                return cls.AUDIENCE_YOUNG_ADULT
-            else:
-                # e.g. "up to 14 years"
-                return cls.AUDIENCE_CHILDREN
-
-        # At this point we can assume that lower is not None.
-        if lower >= 18:
-            return cls.AUDIENCE_ADULT
-        elif lower >= cls.YOUNG_ADULT_AGE_CUTOFF:
-            return cls.AUDIENCE_YOUNG_ADULT
-        elif lower >= 12 and (not upper or upper >= cls.YOUNG_ADULT_AGE_CUTOFF):
-            # Although we treat "Young Adult" as starting at 14, many
-            # outside sources treat it as starting at 12. As such we
-            # treat "12 and up" or "12-14" as an indicator of a Young
-            # Adult audience, with a target age that overlaps what we
-            # consider a Children audience.
-            return cls.AUDIENCE_YOUNG_ADULT
-        else:
-            return cls.AUDIENCE_CHILDREN
-
-    @classmethod
     def lookup(cls, scheme):
         """Look up a classifier for a classification scheme."""
         return cls.classifiers.get(scheme, None)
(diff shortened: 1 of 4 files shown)


classifier.py
```diff
@@ -202,42 +202,72 @@ def default_target_age_for_audience(cls, audience):
...
 
     @classmethod
-    def default_audience_for_target_age(cls, target_age):
-        """The default audience for a given target age.
-
-        Inverse of default_target_age_for_audience.
-        """
-        if not target_age:
-            # We were not passed a NumericRange
+    def default_audience_for_target_age(cls, nr):
```
leonardr: There were three partial implementations of this code (which turns an age range into an audience) in classifier.py. I consolidated them here.

classifier.py
```diff
@@ -202,42 +202,72 @@ def default_target_age_for_audience(cls, audience):
...
+            # treat "12 and up" or "12-14" as an indicator of a Young
+            # Adult audience, with a target age that overlaps what we
+            # consider a Children audience.
+            return cls.AUDIENCE_YOUNG_ADULT
+        else:
+            return cls.AUDIENCE_CHILDREN
 
-        lower = target_age.lower
-        upper = target_age.upper
+    @classmethod
```
leonardr: There were two partial implementations of this code; I refactored it and made it complete.

classifier.py
```diff
@@ -3330,13 +3332,18 @@ def __init__(self, work, test_session=None, debug=False):
         self.classifications = []
         self.log = logging.getLogger("Classifier (workid=%d)" % self.work.id)
 
+        # Keep track of whether we've seen one of Overdrive's generic
```
leonardr: This is to fix issue 162.

classifier.py
```diff
@@ -3380,12 +3387,32 @@ def add(self, classification):
         else:
             self.audience_weights[subject.audience] += weight
```
leonardr: Now that we use all relevant classifications to calculate target age, instead of just the "most reliable subset", we can build up the target age Counters as we go, instead of waiting until prepare_to_classify().

classifier.py
```diff
@@ -3574,49 +3613,29 @@ def target_age(self, audience):
             return Classifier.default_target_age_for_audience(audience)
 
         # Only consider the most reliable classifications.
-        reliable_classifications = self.most_reliable_target_age_subset
-
+        
         # Try to reach consensus on the lower and upper bounds of the
         # age range.
-        target_age_mins = []
```
leonardr: We already built the Counters for target age min and max, so this code is no longer needed.

model.py
```diff
@@ -4836,24 +4836,55 @@ def scaled_weight(self):
...
-        # These measure reading level, except for TAG, which measures
-        # who-knows-what.
-        (DataSource.OVERDRIVE, Subject.GRADE_LEVEL) : 5,
-        Subject.TAG : 2,
-        Subject.LEXILE_SCORE : 1,
-        Subject.ATOS_SCORE: 1,
-
-        Subject.AGE_RANGE : 10,
-        Subject.GRADE_LEVEL : 10,
+
```
leonardr: Previously these numbers were just relative rankings and had no special significance. Now they are coefficients ranging from 0 to 1 and expressing how confident we are in the reliability of a target age indicator.

model.py
```diff
@@ -4836,24 +4836,55 @@ def scaled_weight(self):
...
+        Subject.AXIS_360_AUDIENCE : 0.9,
+        (DataSource.OVERDRIVE, Subject.INTEREST_LEVEL) : 0.9,
+        (DataSource.OVERDRIVE, Subject.OVERDRIVE) : 0.9, # But see below
+        (DataSource.AMAZON, Subject.AGE_RANGE) : 0.9,
+        (DataSource.AMAZON, Subject.GRADE_LEVEL) : 0.9,
+        
+        # Although Overdrive usually reserves Fiction and Nonfiction
+        # for books for adults, it's not as reliable an indicator as
+        # other Overdrive classifications.
```
leonardr: This is new, and helps fix issue #163.

model.py
```diff
@@ -4836,24 +4836,55 @@ def scaled_weight(self):
...
+        # Tags that come from OCLC Linked Data are of lower quality
+        # because they sometimes talk about completely the wrong book.
+        (DataSource.OCLC_LINKED_DATA, Subject.TAG) : 0.3,
+        
+        # These measure reading level, not age appropriateness.
+        # However, if the book is a remedial work for adults we won't
+        # be calculating a target age in the first place, so it's okay
+        # to use reading level as a proxy for age appropriateness in a
+        # pinch. (But not outside of a pinch.)
+        (DataSource.OVERDRIVE, Subject.GRADE_LEVEL) : 0.35,
```
leonardr: I'm undoing here a change I made recently, which I made because I forgot that Overdrive's "grade level" classifications are measurements of reading level.

overdrive.py
```diff
@@ -552,11 +552,15 @@ def book_info_to_metadata(cls, book):
 
         extra = dict()
         if 'grade_levels' in book:
+            # n.b. Grade levels are measurements of reading level, not
+            # age appropriateness. We can use them as a measure of age
+            # appropriateness in a pinch, but we weight them less
+            # heavily than other information from Overdrive.
```
leonardr: Just a reminder so I don't make the same mistake again.

classifier.py
```diff
@@ -3629,10 +3648,9 @@ def target_age(self, audience):
         if target_age_max is None:
             target_age_max = target_age_min
 
-        # If min and max got mixed up somehow, un-mix them. This should
-        # never happen, but we fix it just in case.
+        # Err on the side of setting the minimum age too high.
         if target_age_min > target_age_max:
-            target_age_min, target_age_max = target_age_max, target_age_min
+            target_age_min = target_age_max
```
aslagle: is this backwards? I think you want both to be the higher one, which is min in this case

aslagle:
:+1:

Issue #144, opened 2016-03-31
Search: Handle subject/genre classifications better


aslagle:
If you search for "basketball", that shouldn't be mapped to the sports genre.

aslagle:
Also "dogs" and "cats" are both mapped to "pets" right now.

aslagle:
"teen" is a similar problem - it's useful for search but may not reliably indicate books for teens vs books about teens

PR #165, opened 2016-04-08
Handle subjects better in search

aslagle:
This connects to #144 

There are two changes on this branch:
- Subjects are now stored in the search index and searched, separately from genres.
- Keywords that are specific examples of a classification rather than an exact identifier are excluded when used from search queries, so a search for "cats" won't match "pets" and return books about all pets.

### DIFF
--- a/model.py
+++ b/model.py
@@ -3881,33 +3881,39 @@ class Work(Base):
                 dict(name=contributor.name, family_name=contributor.family_name,
                      role=contribution.role))
 
-        # identifier_ids = self.all_identifier_ids()
-        # classifications = Identifier.classifications_for_identifier_ids(
-        #     _db, identifier_ids)
-        # by_scheme_and_term = Counter()
-        # classification_total_weight = 0.0
-        # for c in classifications:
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -3881,33 +3881,39 @@ def to_search_document(self):
...
+            if not subject.type in Subject.TYPES_FOR_SEARCH:
+                # This subject type doesn't have terms that are useful for search
+                continue
+            if subject.type in Subject.uri_lookup:
+                scheme = Subject.uri_lookup[subject.type]
+                term = subject.name or subject.identifier
+                if not term:
+                    # There's no text to search for.
+                    continue
+                term = term.replace("/", " ").replace("--", " ")
```
aslagle: Looks like scrub_identifier doesn't remove the slashes or add spaces, it just adds a slash at the end if there isn't one.

I'm going to leave this here for now but we might need to revisit it later.

aslagle:
I got rid of the confusing partial function, but I still need the dictionary.

PR #166, opened 2016-04-08
Add resolved column to complaints

skomputer:
This branch includes a migration that adds a `resolved` column to the `complaints` table, modifies `Complaint.register()` to support creation of resolved complaints (only for testing purposes at this time), adds a `resolve()` method to complaint objects, and modifies `LicensePool.with_complaint()` to support inclusion of license pools with resolved complaints.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5130,7 +5130,7 @@ class LicensePool(Base):
         )
 
     @classmethod
-    def with_complaint(cls, _db):
+    def with_complaint(cls, _db, exclude_resolved=True):
         """Return query for LicensePools that have at least one Complaint."""
         subquery = _db.query(
                 LicensePool.id,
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -5217,16 +5217,21 @@ def deliverable(self):
         )
 
     @classmethod
-    def with_complaint(cls, _db):
+    def with_complaint(cls, _db, exclude_resolved=True):
```
leonardr: This allows two scenarios: all complaints, and all unresolved complaints. Would it make sense to make this a variable called `resolved`? `resolved=True` means resolved complaints, `resolved=False` means unresolved complaints, `resolved=None` means both. I don't know how often you'd need to look at resolved complaints but right now there's no method for that.

skomputer: That makes sense. I made the change.

PR #169, opened 2016-04-11
Remove Edition.no_known_cover

leonardr:
At one point our rules for when a work was 'presentation-ready' (i.e. would be shown to patrons for checkout) were pretty strict. In particular, you either had to have a cover image or Edition.no_known_cover had to be set. no_known_cover meant that we had tried really hard to look for a cover image and come up short.

But recently we changed the rules so that works are made presentation-ready much sooner. If there's a problem completing the metadata for a book, we want a) the book to be loanable during the period when the problem happens, and b) the problem to be visible (e.g. by bad metadata), not invisible (e.g. by a book not showing up in the collection). So we effectively stopped using no_known_cover.

I'm about to do some work that changes the relationship between Edition and Work, and as cleanup I'm getting rid of no_known_cover, which complicates that relationship and is no longer used.


courte:
:+1: This is excitiiing!

PR #171, opened 2016-04-11
More search

aslagle:
Continuing to tweak search weights.

This is mainly to improve searches for an author's name. Stemmed and fuzzy fields were matching too highly.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -242,6 +242,15 @@ class ExternalSearchIndex(object):
             'imprint'
         ]
 
+        # These words will fuzzy match other common words that aren't relevant,
+        # so if they're present and correctly spelled we shouldn't use a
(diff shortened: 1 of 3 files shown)


external_search.py
```diff
@@ -197,24 +197,47 @@ def make_target_age_query(target_age):
...
+            'title.minimal^4',
+            'series.minimal^4',
+            "subtitle.minimal^3",
+            "summary.minimal^2",
+
+            "classifications.term^2",
+            'author^4',
             'publisher',
             'imprint'
         ]
```
aslagle: I broke these up more so I could give the minimal fields higher weight for both non-stemmed and non-fuzzy queries, and so I could give author a higher weight in non-fuzzy queries.

PR #172, opened 2016-04-12
Redo some subjects

leonardr:
This branch makes it possible to get the SubjectAssignmentMonitor to operate on a subset of subjects, rather than all of them. It creates a new script class, SubjectAssignmentScript, which retrieves command line options and uses them to instantiate an appropriate SubjectAssignmentMonitor.


courte:
👍

PR #179, opened 2016-04-14
Handle poetry and YA classifications better.

leonardr:
This branch fixes a number of cases where the ThreeMClassifier ignored the genre of young adult books and/or books of poetry.

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -3356,7 +3356,7 @@ class WorkClassifier(object):
     nonfiction_publishers = set(["Wiley"])
     fiction_publishers = set([])
 
-    def __init__(self, work, test_session=None, debug=True):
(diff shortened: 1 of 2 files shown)


classifier.py
```diff
@@ -3347,7 +3356,7 @@ class WorkClassifier(object):
     nonfiction_publishers = set(["Wiley"])
     fiction_publishers = set([])
 
-    def __init__(self, work, test_session=None, debug=False):
+    def __init__(self, work, test_session=None, debug=True):
```
aslagle: you left debug=True in here again

leonardr: :( I need to figure out a better way to do this.

aslagle:
:+1:

PR #180, opened 2016-04-14
Bible is not a synonym for christianity.

aslagle:
I changed my mind about this one after I tried it.


leonardr:
Ok, I guess this makes sense.

PR #181, opened 2016-04-14
Enable authenticated libraries

courte:
This is a simple way to keep track of libraries for the metadata wrangler. While the client_id and client_secret arrangement mimics OAuth, at this point library authentication is extremely insecure, in part because both the client_id and client_secret are being held in the database as plaintext. We could--and maybe should?!--use Bcrypt to encrypt the client_secret, but that means we'd only have one opportunity to see and share the client secret, at the moment we create the library.

Maybe this is fine. We're responsible people.

Or maybe - since Library's don't currently have any special privileges at all right now - this is temporarily acceptable?

Also, whitespace changes.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -400,7 +400,7 @@ class URNLookupController(object):
 
         return feed_response(opds_feed)
 
-    
+
 class ComplaintController(object):
     """A controller to register complaints against objects."""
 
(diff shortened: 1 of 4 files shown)


courte:
Pay no attention to the code behind the curtain.

PR #182, opened 2016-04-14
Metadata edition tracks updates, removed call to edition.last_update_time (edition doesn't have attribute)

daryanypl:


model.py
```diff
@@ -2842,11 +2842,6 @@ def calculate_presentation(self, policy=None):
         ):
             changed = True
 
-        if changed:
-            # last_update_time tracks the last time the data 
-            # actually changed.
-            self.last_update_time = datetime.datetime.utcnow()
```
aslagle: Does last_update_time work differently in edition and work now?

leonardr: It turns out Edition doesn't have a .last_update_time at all! This code is setting a variable on the Edition object that is not stored in the database or used anywhere.

PR #183, opened 2016-04-15
Enable authenticated collections

courte:
This branch:
- creates collections (née libraries) as user-type objects for the Metadata Wrangler
- encrypts their `client_secret`
- authenticates requests from true-blue collections
- tracks the lookups of authenticated collections in the database

PHEW!

The content server and circulation manager are going to need their `requirements.txt` updated to include bcrypt before they can submodule update past this point.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -400,7 +400,7 @@ class URNLookupController(object):
 
         return feed_response(opds_feed)
 
-    
+
 class ComplaintController(object):
     """A controller to register complaints against objects."""
 
@@ -433,4 +433,4 @@ class ComplaintController(object):
             )
 
         return make_response("Success", 201, {"Content-Type": "text/plain"})
-        
+
--- a/app_server.py
+++ b/app_server.py
@@ -453,7 +453,6 @@ class CollectionController(object):
                 return collection
 
             # If inaccurate authorization details were sent, return error.
-            error = INVALID_CREDENTIALS
-            return problem(error.uri, error.status_code, error.title)
+            return INVALID_CREDENTIALS.response
         return None
 
--- a/app_server.py
+++ b/app_server.py
@@ -235,7 +235,7 @@ class URNLookupController(object):
 
         return (400, self.UNRESOLVABLE_URN)
 
-    def process_urn(self, urn):
+    def process_urn(self, urn, collection=None):
         """Turn a URN into a Work suitable for use in an OPDS feed.
 
         :return: If a Work is found, the return value is None.
@@ -246,6 +246,9 @@ class URNLookupController(object):
         if not isinstance(identifier, Identifier):
             # Error.
             return identifier
+
+        if collection:
+            collection.track_asset(self._db, identifier)
 
         if identifier.licensed_through:
             # There is a LicensePool for this identifier!
@@ -360,14 +363,14 @@ class URNLookupController(object):
         # We made it!
         return entry
 
-    def work_lookup(self, annotator, controller_name='lookup'):
+    def work_lookup(self, annotator, controller_name='lookup', collection=None):
         """Generate an OPDS feed describing works identified by identifier."""
         urns = flask.request.args.getlist('urn')
 
         messages_by_urn = dict()
         this_url = cdn_url_for(controller_name, _external=True, urn=urns)
         for urn in urns:
-            code, message = self.process_urn(urn)
+            code, message = self.process_urn(urn, collection=collection)
             if code:
                 messages_by_urn[urn] = (code, message)
 
--- a/app_server.py
+++ b/app_server.py
@@ -434,3 +434,23 @@ class ComplaintController(object):
 
         return make_response("Success", 201, {"Content-Type": "text/plain"})
 
+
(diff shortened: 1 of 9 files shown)


app_server.py
```diff
@@ -360,14 +363,14 @@ def make_opds_entry_from_metadata_lookups(self, identifier):
         # We made it!
         return entry
 
-    def work_lookup(self, annotator, controller_name='lookup'):
+    def work_lookup(self, annotator, controller_name='lookup', collection=None):
```
courte: For more about why this is weird, see NYPL-Simplified/metadata_wrangler#49.

model.py
```diff
@@ -6821,6 +6824,106 @@ def update_credentials(self, _db, access_token, credential):
...
+        while get_one(_db, cls, client_id=client_id):
+            client_id, plaintext_client_secret = cls.generate_client_details()
+
+        logging.info(
+            ("Creating collection %s. "
+             "RECORD THE FOLLOWING AUTHENTICATION DETAILS. "
+             "The client secret cannot be recovered."), name
+        )
+        logging.info("CLIENT ID: %s", client_id)
+        logging.info("CLIENT SECRET: %s", plaintext_client_secret)
```
leonardr: I don't think it's a good idea to log this.

model.py
```diff
@@ -6821,6 +6824,106 @@ def update_credentials(self, _db, access_token, credential):
...
+    @classmethod
+    def authenticate(cls, _db, client_id, plaintext_client_secret):
+        collection = get_one(_db, cls, client_id=client_id)
+        if (collection and
+            collection._correct_secret(plaintext_client_secret)):
+            return collection
+        return None
+
+    def _correct_secret(self, plaintext_secret):
+        return (bcrypt.hashpw(plaintext_secret, self.client_secret)
```
leonardr: This bcrypt thing is complicated enough and important enough that I'd feel better if it were moved into a helper method so you don't duplicate code between here and line 6847.

model.py
```diff
@@ -1050,6 +1051,9 @@ class Identifier(Base):
     # One Identifier may have many associated CoverageRecords.
     coverage_records = relationship("CoverageRecord", backref="identifier")
 
+    # One Identifier may be registered in many collections.
+    collections = relationship("CollectionAsset", backref="identifier")
+
```
leonardr: This will give you a list of CollectionAsset objects. Either this needs to be called collection_assets (which I don't recommend because CollectionAsset looks like a pure join table) or you need to make this a many-to-many relationship a la http://docs.sqlalchemy.org/en/rel_0_9/orm/basic_relationships.html#many-to-many, so that identifier.collections actually gives you a list of Collection objects.

model.py
```diff
@@ -6830,6 +6836,136 @@ def update_credentials(self, _db, access_token, credential):
...
+        license_pool = identifier.licensed_through
+        if not license_pool:
+            license_pool, new = get_one_or_create(
+                _db, LicensePool, identifier=identifier
+            )
+        if license_pool not in self.license_pools:
+            self.license_pools.append(license_pool)
+
+    def catalog(self, _db):
```
skomputer: something like this might be a more efficient query for the identifiers:
```
qu = _db.query(Identifier).join(LicensePools).join(Collection).filter(Collection.id=self.id)
```

problem_details.py
```diff
@@ -13,3 +13,10 @@
       "Unrecognized data source.",
       "I don't know anything about that data source.",
 )
+
+INVALID_CREDENTIALS = pd(
+      "http://librarysimplified.org/terms/problem/credentials-invalid",
+      401,
+      "Invalid credentials",
+      "A valid library card barcode number and PIN are required.",
```
daryanypl: Is this a valid error for a collection?

model.py
courte: Trying to directly connect LicensePool and Collection like that raises an error because of the way the collection was made, but @aslagle suggested a change that will fix that 🎉 !

tests/test_model.py
```diff
@@ -2435,3 +2437,66 @@ def test_resolve(self):
...
+
+    def test_authenticate(self):
+        collection = self._collection()
+
+        result = Collection.authenticate(self._db, u"abc", u"def")
+        eq_(collection, result)
+
+        result = Collection.authenticate(self._db, u"abc", u"bad_secret")
+        eq_(None, result)
+
```
daryanypl: Could we add 
result = Collection.authenticate(self._db, u"bad_client_id", u"def")
?

tests/test_app_server.py
```diff
@@ -111,6 +104,15 @@ def test_process_urn_unrecognized_identifier(self):
...
 
+    def test_process_urn_with_collection(self):
+        identifier = self._identifier()
+        collection = self._collection()
+
+        eq_([], collection.catalog)
+        self.controller.process_urn(identifier.urn, collection=collection)
+        eq_(1, len(collection.catalog))
+        eq_([identifier], collection.catalog)
+
```
aslagle: maybe process a second identifier too, and make sure you can process one twice

PR #186, opened 2016-04-15
Fix CoverageProvider.items_that_need_coverage() to send the correct arguments, and add some basic tests.

leonardr:


courte:
:+1:

PR #189, opened 2016-04-20
Made it possible to remove a work's summary.

aslagle:
Previously, you could set a work's summary to None, but it would retain the summary_text from the previous summary.

For the admin interface, I think it will be useful to remove an incorrect summary even if you don't have a new summary to replace it with.

I'm not sure how this will change with composite editions, but it works for now.

This connects to https://github.com/NYPL-Simplified/circulation/issues/146

### DIFF
--- a/model.py
+++ b/model.py
@@ -1649,7 +1649,7 @@ class Identifier(Base):
 
     @classmethod
     def evaluate_summary_quality(cls, _db, identifier_ids, 
-                                 privileged_data_source=None):
+                                 privileged_data_sources=None):
         """Evaluate the summaries for the given group of Identifier IDs.
 
         This is an automatic evaluation based solely on the content of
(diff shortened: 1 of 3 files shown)


leonardr:
I've been worried about this for e.g. subtitle (if admin thinks the book has no subtitle but metadata wrangler thinks it does have one), but for description it's not so bad. Instead of editing Edition.summary you can create a description Resource associated with the admin interface data source. That resource will take precedence over other resources, so even if the text associated with the resource is blank, that text will become the Work's description.

aslagle:
@leonardr I'm creating the description in https://github.com/NYPL-Simplified/circulation/pull/149, but if the text of the description is an empty string, resource.representation ends up being None so I needed this change as well.

model.py
```diff
@@ -3250,6 +3250,8 @@ def set_summary(self, resource):
         # TODO: clean up the content
         if resource:
```
aslagle: maybe I should make this `if resource and resource.representation`, so I can set a description with an empty string instead of setting the resource to None?

leonardr: I like this idea.

aslagle: I can put this in, but it won't let you remove a summary entirely if it's from another source - Identifier.evaluate_summary_quality will throw out the empty one. That's probably ok for now. Once composite editions are ready we'll have a better way to fix it.

leonardr:
It'll be a little bit of a headache to merge this with the presentation-edition work, but this looks good.

PR #192, opened 2016-04-21
Fixed test_calculate_work_for_licensepool_creates_new_work.

daryanypl:
All tests/test_model.py:TestWorkConsolidation are now passing.


aslagle:
:+1:

metadata_layer.py
```diff
@@ -1093,6 +1068,43 @@ def _key(classification):
...
+                # is associated with the original image.
+                self.make_thumbnail(pool, data_source, link, link_obj)
+
+        # follows up on the mirror link, and confirms the delivery mechanism.
+        for format in self.formats:
+            if format.link:
+                link = format.link
+                if not format.content_type:
+                    format.content_type = link.media_type
+                # TODO: I think it's always true that this link
```
leonardr: Now that we do keep track of the link objects you can remove this comment.

model.py
```diff
@@ -5303,8 +5298,8 @@ def sort_key(edition):
         return sorted(self.identifier.primarily_identifies, key=sort_key)
 
 
+    # TODO:  policy is not used in this method.  Remove argument?
```
leonardr: I'm pretty sure we should remove this argument. I think my thinking was that when we call Metadata.apply() inside this method we should pass in a ReplacementPolicy object to control exactly what work gets done. But actually if there is a policy question, it's whether a LicensePool's presentation edition should be set/recalculated in the first place, and that policy question should be determined by whoever is calling set_presentation_edition(). 

So we might end up adding a variable to `ReplacementPolicy` which controls whether or not this method is called, but I think by the time we get in here, we're committed to completely recalculating the presentation edition.

test_all.sh
```diff
@@ -0,0 +1,10 @@
+# PYTHONWARNINGS=ignore suppresses SQLAlchemy and other warnings.
+# --nocapture lets you output to stdout, while the tests is still running (arguably more useful while debugging a single test, not running the whole batch).
+# --detailed-errors attempts a more detailed stack trace, but doesn't actually work.  Comment here to serve as warning.
+# Also check out: --failed  (Run the tests that failed in the last test run.)  It's helpful when isolating the few tests you need to pay attention to after a full test suite run.
+
+PYTHONWARNINGS=ignore nosetests --nocapture 
+
+
+
+
```
leonardr: There is already a `test` file that's very similar. Can you change `test` instead of adding a new file?

testing.py
```diff
@@ -184,11 +192,20 @@ def _work(self, title=None, authors=None, genre=None, language=None,
             genre, ignore = Genre.lookup(self._db, genre, autocreate=True)
         work.genres = [genre]
         work.random = 0.5
+
         work.editions = [primary_edition]
         primary_edition.is_primary_for_work = True
-        work.primary_edition = primary_edition
+
+        #work.primary_edition = primary_edition
```
leonardr: Take this line out.

testing.py
```diff
@@ -184,11 +192,20 @@ def _work(self, title=None, authors=None, genre=None, language=None,
...
+        work.set_primary_edition()
+
         if pool != None:
-            work.license_pools.append(pool)
+            # make sure the pool's presentation_edition is set, 
+            # bc loan tests assume that.
+            pool.set_presentation_edition(None)
+
+            if not work.license_pools[0]:
```
leonardr: This looks like work.license_pools is a list where the first item is `None`. I think maybe you meant to check if the list is empty? Like, `if not work.license_pools`.

The fact that this never happens (if the list was empty you'd get a crash here) indicates that it might be okay to remove this bit.

tests/test_metadata.py
```diff
@@ -220,31 +220,39 @@ def test_image_scale_and_mirror(self):
...
 
         # This link will not be mirrored.
-        l2 = LinkData(
+        link_unmirrored = LinkData(
             rel=Hyperlink.SAMPLE, href="http://example.com/2",
             media_type=Representation.TEXT_PLAIN,
             content="i am a tiny (This is a sample. To read the rest of this book, please visit your local library.)"
         )
 
+        #pool.set_presentation_edition(None)
```
leonardr: Take this out.

tests/test_metadata.py
```diff
@@ -262,6 +271,14 @@ def test_open_access_content_mirrored(self):
         )
         assert book.mirror_url.endswith(expect)
 
+        # make sure the mirrored link is safely on edition
+        assert edition.license_pool.identifier.links[0].resource.representation.mirror_url.startswith('http://s3.amazonaws.com/test.content.bucket/')
```
leonardr: This is hard to read. Try something like this up front to make the assertion code shorter:

`mirrored_representation, unmirrored_representation = [l.resource.representation for l in edition.license_pool.identifier.links]`

tests/test_model.py
```diff
@@ -1181,6 +1183,7 @@ def test_set_presentation_edition(self):
         eq_(edition_mw.is_primary_for_work, False)
         eq_(edition_od.is_primary_for_work, False)
         eq_(edition_admin.is_primary_for_work, False)
+        # TODO: does the following line need to be set to work?
```
leonardr: I feel like you should be able to uncomment this test and it would work.

tests/test_model.py
```diff
@@ -1226,21 +1229,21 @@ def test_calculate_presentation(self):
 
         # This Work starts out with a single CoverageRecord reflecting the
         # work done to generate its initial OPDS entry.
-        [record] = work.coverage_records
-        eq_(WorkCoverageRecord.GENERATE_OPDS_OPERATION, record.operation)
+        [record1, record2] = work.coverage_records
+        # we don't know what order the records came back in, but one of them is an opds generation record.
```
leonardr: When I don't know what order items come back in, I usually sort them. In this case `operation` would work as a sort key. Something like this:

`[choose_edition, generate_opds] = sort(work.coverage_records, key=lambda x: x.operation)`

tests/test_model.py
```diff
@@ -1833,8 +1823,12 @@ def test_open_access_loan(self):
         eq_(loan, loan2)
         eq_(False, was_new)
 
+
     def test_work(self):
         """Test the attribute that finds the Work for a Loan or Hold."""
+        # TODO:  determine if using pool.presentation_edition is correct for lending.
+        # Because, while the old pool.edition was tied to something you can lend, the presentation_edition 
+        # is not representing a lendable object, really.
```
leonardr: I don't understand this comment. The `LicensePool` is always something you can lend, and its `.edition` (old) or `.presentation_edition` (new) associated with it never represented anything lendable.

leonardr:
I have a lot of comments but they're all pretty minor and overall this looks very good.

metadata_layer.py

testing.py
daryanypl: done

testing.py
daryanypl: Changed.  I had a problem where the work got two identical license pools, added at different points in the code, and this line makes sure there's no duplication.

tests/test_metadata.py
daryanypl: done

tests/test_metadata.py
daryanypl: How about:
        for edlink in edition.license_pool.identifier.links:
            edmirror = resource.representation.mirror_url
            assert edmirror.startswith('http://s3.amazonaws.com/test.content.bucket/')

leonardr: That's okay for line 275 but then you don't have the other representation handy to check on lines 277 and 279.

test_all.sh
daryanypl: done

tests/test_metadata.py
daryanypl: done

tests/test_model.py
daryanypl: eq_(edition_composite.is_primary_for_work, True)
fails, because we don't tell the composite edition that it's primary for work when we set it on the pool.

daryanypl: done

model.py
daryanypl: I take it back.  If I remove the policy argument, all tests break.  Keeping for now.  To do this in the future, I'd cut a separate branch.

tests/test_model.py
daryanypl: done

tests/test_model.py
daryanypl: I'm just not sure we should be testing lending functionality through the composite edition.  Maybe the overdrive edition would be better?

leonardr: I think I see what you're saying. 

This test is testing the property `LoanAndHoldMixin.work`. If everything is working properly, the `LicensePool` has a `.work` set and that's the end of it.

But there might be a problem such that `LicensePool.work` is not set. In this case (which shouldn't happen) we might still be able to find the appropriate `.work` because of some redundancy in the database. Specifically, the LicensePool's `.presentation_edition` might be associated with some `.work`. This will always be the `Work` that `LicensePool.work` _should_ have been set to, so we can check that and fall back to it.

Currently we can generally rely on `LicensePool.presentation_edition` having an associated `.work`, even if `Edition.is_primary_for_work` is not set. However that might change in the future because we might be getting rid of that redundancy. If we get rid of `Edition.is_primary_for_work`, then an `Edition` will only have a `.work` if it is the primary for its work. 

Most of the time you'll still be able to get from `LicensePool` to `Edition` to `Work`, because most of the time a `Work` will only have one `LicensePool`. But when there are multiple `LicensePool`s, it's always possible that some other `LicensePool` is the one that provides the `.presentation_edition` for the work.

Since this is fallback code and there will eventually be situations where the fallback doesn't work, I'm okay with taking the code out of `LoanAndHoldMixin.work` and taking out this part of the test.

PR #193, opened 2016-04-21
Remove Axis 360 integration dependency

courte:
This mirrors what we're doing with the Overdrive and 3M objects right now and allows the app to load when there's no Axis 360 integration configured.

Fixes NYPL-Simplified/circulation#153.

### DIFF
--- a/overdrive.py
+++ b/overdrive.py
@@ -118,7 +118,6 @@ class OverdriveAPI(object):
     def from_environment(cls, _db):
         # Make sure all environment values are present. If any are missing,
         # return None. Otherwise return an OverdriveAPI object.
-        values = cls.environment_values()
         try:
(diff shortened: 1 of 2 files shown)


overdrive.py
```diff
@@ -118,7 +118,6 @@ def environment_values(cls):
     def from_environment(cls, _db):
         # Make sure all environment values are present. If any are missing,
         # return None. Otherwise return an OverdriveAPI object.
-        values = cls.environment_values()
```
aslagle: why did you take this out?

courte: It's not actually used elsewhere in the `from_environment` method. It's called in the `__init__` method instead, when it's invoked on line 123.

aslagle:
:+1:

PR #195, opened 2016-04-25
Fixed searching from lane with exclude_languages set. (Fixes #190)

### LINKED ISSUE #190: Searching from the other languages lane doesn't work
There's a problem with the filter for exclude_language.

aslagle:
This makes sure exclude_languages is a list before it gets serialized by elasticsearch. languages was already a list, but I did the same just in case.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -383,9 +383,9 @@ class ExternalSearchIndex(object):
 
         clauses = []
         if languages:
-            clauses.append(dict(terms=dict(language=languages)))
+            clauses.append(dict(terms=dict(language=list(languages))))
         if exclude_languages:
-            clauses.append({'not': dict(terms=dict(language=exclude_languages))})
+            clauses.append({'not': dict(terms=dict(language=list(exclude_languages)))})
         if genres:
             genre_ids = [genre.id for genre in genres]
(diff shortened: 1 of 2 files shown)


external_search.py
```diff
@@ -383,9 +383,9 @@ def _f(s):
 
         clauses = []
         if languages:
-            clauses.append(dict(terms=dict(language=languages)))
+            clauses.append(dict(terms=dict(language=list(languages))))
```
leonardr: What was languages before? I'm a little concerned that if it comes in a string 'eng' this will turn it into `['e', 'n', 'g']`.

Or did this just not work at all before?

aslagle: Languages worked before, but exclude_languages didn't because the lane was created with a set. 

If it's a string it gets converted to a list here: https://github.com/NYPL-Simplified/server_core/blob/master/lane.py#L515
But if it's a set it stays a set.

Should I add a test for the string?

aslagle: This is the code that uses a set for exclude_languages but a list for languages:
https://github.com/NYPL-Simplified/circulation/blob/master/api/lanes.py#L56

leonardr: OK, set versus list, got it. :+1:

PR #196, opened 2016-04-25
Show collection updates

courte:
- Moves the CollectionController to the metadata wrangler: NYPL-Simplified/metadata_wrangler#52
- Removes some of the more special characters from the `client_secret`, since they were getting in way of effortlessly `curl`ing the route
- Adds the column `last_checked` to collections, so we can:
- Return(s) works updated since a collection's last visit.
- Creates a datasource for a collection made during testing (like `Collection.register` does all the time)

Also: tests, removal of some commenty cruft, and the use of unicode encoding where it was immediately apparent.

Fixes NYPL-Simplified/metadata_wrangler#51.

### DIFF
--- a/model.py
+++ b/model.py
@@ -6915,7 +6915,7 @@ class Collection(Base):
         client_id_chars = ('abcdefghijklmnopqrstuvwxyz'
                            'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
                            '0123456789')
-        client_secret_chars = client_id_chars + '!"#$%&()*+,-./[]^_`{}|~'
+        client_secret_chars = client_id_chars + '!#$%&*+,-._'
 
         def make_client_string(chars, length):
             return u"".join([random.choice(chars) for x in range(length)])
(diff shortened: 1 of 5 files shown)


aslagle:
Looks good to me. It might also be useful to be able to pass in a date to use instead of last_checked.

aslagle:
I also wonder if this will need to be paginated - it could be a lot of works if people don't check very often.

model.py
```diff
@@ -6937,6 +6925,26 @@ def catalog_identifier(self, _db, identifier):
...
+        """Returns all of a collection's works that have been updated since the
+        last time the collection was checked"""
+        if self.catalog:
+            catalog_identifier_ids = [identifier.id for identifier in self.catalog]
+
+            # Find the works in the catalog that have been created or updated
+            # since the collection last checked.
+            query = _db.query(Work).join(Work.coverage_records)
+            query = query.join(Work.license_pools).join(Identifier)
+            query = query.filter(Identifier.id.in_(catalog_identifier_ids))
```
leonardr: There are going to be hundreds of thousands of Identifiers in a collection, right? Rather than iterate over all of them collecting their IDs, you should join Identifier against collections_identifiers against Collection and let the database do the work.

model.py
```diff
@@ -6937,6 +6925,26 @@ def catalog_identifier(self, _db, identifier):
...
+            if self.last_checked:
+                query = query.filter(
+                    WorkCoverageRecord.timestamp > self.last_checked
+                )
+
+            return query.all()
```
leonardr: Have this return the query itself, rather than every item in the resultset. This will make it easy to use our existing pagination system.

leonardr:
:+1:

aslagle:
:+1:

PR #198, opened 2016-04-26
Added fiction status to opds entries and removed it as a genre. (Fixes #187)

### LINKED ISSUE #187: OPDS feeds do not mention fiction status per se
If a book is classified under a Simplified genre we'll show a `category` tag like this:

`<category term="http://librarysimplified.org/terms/genres/Simplified/Urban%20Fiction" scheme="http://librarysimplified.org/terms/genres/Simplified/" label="Urban Fiction"/>`

But we don't have any indication for whether a book is Fiction or Nonfiction, per se. When we show a book directly to a patron this doesn't matter much, because human beings know that Urban Fiction is a type of Fiction and Biography is a type of Nonfiction.

But when we're presenting an edit interface, and one of the main things you can change is a work's fiction status, we need to know explicitly whether the book is Fiction or Nonfiction, so we can present the edit form properly.

aslagle:
I made up a new scheme for fiction/nonfiction status, and it's always included. Fiction/nonfiction no longer shows up as a genre if there are no other genres, so it will be up to the clients to check for genres and decide whether to display fiction/nonfiction.

### DIFF
--- a/model.py
+++ b/model.py
@@ -4565,9 +4565,11 @@ class Subject(Base):
     PERSON = Classifier.PERSON
     ORGANIZATION = Classifier.ORGANIZATION
     SIMPLIFIED_GENRE = "http://librarysimplified.org/terms/genres/Simplified/"
+    SIMPLIFIED_FICTION_STATUS = "http://librarysimplified.org/terms/fiction/"
 
     by_uri = {
         SIMPLIFIED_GENRE : SIMPLIFIED_GENRE,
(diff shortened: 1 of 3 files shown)


leonardr:
I looked around to see if we could reuse a standard controlled vocabulary's values for "Nonfiction" or "Fiction" per se, but strangely enough I didn't see anything (Dewey has the unofficial 'F' for fiction but nothing for nonfiction), so I think this is good. :+1:

PR #200, opened 2016-04-27
Remove ids from collections

courte:
Supports NYPL-Simplified/metadata_wrangler#55 by:
- Pulling the `returns_problem_detail` wrapper from the Circulation Manager (to be updated accordingly in a soon-to-come submodule update), and
- factoring the `INVALID_URN` response into a problem detail.

Also a teensy refactoring in the `TestCollection` class. It's a grab bag, people! 👜

Fixes NYPL-Simplified/metadata_wrangler#54.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -192,15 +192,12 @@ class HeartbeatController(object):
 
 class URNLookupController(object):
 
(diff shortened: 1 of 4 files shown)


leonardr:
:+1:

PR #201, opened 2016-04-28
Made links from metadata layer use license data source if present.

aslagle:
This connects to https://github.com/NYPL-Simplified/content_server/issues/59


aslagle:
Not sure if I should change the data_source passed to mirror_link and make_thumbnail too.

leonardr:
Nice. :nail_care:

I do think you should change the data_source passed to mirror_link and make_thumbnail. Once you do that I'll take another look!

Hopefully this will just go away once we bring in the presentation-edition code.

aslagle:
@leonardr changed them

leonardr:
:+1:

PR #202, opened 2016-04-28
Better uplinks

skomputer:
This branch fixes a few minor uplink issues, such as uplinks pointing to invisible lanes and search feed uplinks pointing to group feed urls when the search lane isn't at a group feed url.


leonardr:
:+1:

PR #203, opened 2016-04-28
Added pool supercession logic to work.calculate_presentation

daryanypl:


model.py
```diff
@@ -3481,6 +3481,14 @@ def calculate_presentation(self, policy=None, search_index_client=None):
         if new_primary_edition:
             self.primary_edition = new_primary_edition
 
+        # go through the loser pools, and tell them they lost
+        for pool in self.license_pools:
+            if pool.presentation_edition is self.primary_edition:
+                # make sure edition knows it's primary
+                pool.presentation_edition.is_primary_for_work = True
```
leonardr: This shouldn't be an issue, but... all of this pool's editions _except_ for this one should also be marked nonprimary.

You might replace `LicensePool.mark_editions_nonprimary()` with `mark_editions(primary_for_work=None)`. That would let you use the same code for the lucky pool that has the primary edition and the unlucky pool that doesn't.

leonardr:
Looks great with one suggestion.

model.py
daryanypl: done

aslagle:
Looks good

PR #204, opened 2016-04-29
Update rights status when importing to an existing license pool from …

aslagle:
…OPDS.


leonardr:
:+1: 

I'm a little concerned about cases where this might do the wrong thing, but I think we can tackle those cases once we get the presentation-edition stuff integrated.

PR #205, opened 2016-04-29
Stop storing session-specific information in global spaces

leonardr:
This branch makes two changes in preparation for giving each incoming request to an app server its own database session (allowing multithreaded use).

First, it stops pre-caching DataSource, Genre, and LicensepoolDeliveryMechanism objects globally. These objects are associated with a specific database session, which works as long as the entire app is running against the same session, but falls apart as soon as session #2 tries to use an object created by session #1.

I don't know what the performance implications of this will be, but I really don't see any way around this, and we've seen that sharing a single database session has terrible implications for scalability.

Second, it refactors the process of setting up database sessions and rolling them back to use the standard `unittest` setup/teardown mechanisms, rather than the old (very hacky) mechanism that I made because I didn't know what I was doing.

I think this second bit is not strictly necessary for the scoped-session code. It turns out it's nearly impossible to test app servers with scoped sessions, so we're mostly going to be using the same old technique where we use a single database connection for the whole test. But either way this is an improvement over the old system.

### DIFF
--- a/model.py
+++ b/model.py
@@ -4509,27 +4509,17 @@ class Genre(Base):
 
     @classmethod
     def lookup(cls, _db, name, autocreate=False):
-        if not hasattr(_db, '_genre_cache'):
-            _db._genre_cache = dict()
-        if isinstance(name, Genre):
-            return name, False
         if isinstance(name, GenreData):
             name = name.name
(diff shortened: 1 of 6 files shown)


aslagle:
:+1:

PR #207, opened 2016-05-02
Changed OPDS import media type handling to work with standard ebooks

aslagle:


model.py
```diff
@@ -6069,14 +6069,12 @@ def get(cls, _db, url, do_get=None, extra_request_headers=None,
...
-        # actually got from the server differs from what we thought
-        # we had.
+        # we don't have one already, or if the URL we actually got from 
+        # the server differs from what we thought we had.
         if (not usable_representation 
-            or media_type != representation.media_type
             or url != representation.url):
             representation, is_new = get_one_or_create(
-                _db, Representation, url=url, media_type=media_type)
+                _db, Representation, url=url)
```
aslagle: For standard ebooks, we don't get a media type back at all when we fetch book files, but we can set it later based on the media type in the link. If we created a new representation instead of using the existing one, we might not be able to set the media type correctly later due to the unique constraint on url + media type. 

In general, if we fetch a resource and get a media type that's not what we thought, it seems like we should update the existing representation for that url instead of creating a new one, since the url may have changed. I can't think of a case where we'd want to have multiple media types for the same url. The unique constraint would make more sense on url only instead of url + media type, but I'm afraid to make that change in case we already have things that violate it.

model.py
```diff
@@ -6248,10 +6246,35 @@ def clean_media_type(self):
...
+            return known_match.group()
+
+        else:
+            any_extension_re = re.compile("\.[\w\d]*$", re.I)
+        
+            any_match = any_extension_re.search(self.url)
+
+            if any_match:
+                return any_match.group()
+        return None
```
aslagle: These don't handle query parameters after the extension, eg "file.epub?version=3"

tests/test_opds_import.py
```diff
@@ -519,10 +519,10 @@ def test_resources_are_mirrored_on_import(self):
         # Each resource was 'mirrored' to an Amazon S3 bucket.
         # The first resource has no bibframe provider in OPDS so it uses the importer's data source.
         eq_(
-            ['http://s3.amazonaws.com/test.content.bucket/Library%20Simplified%20Open%20Access%20Content%20Server/Gutenberg%20ID/10557.epub',
-             'http://s3.amazonaws.com/test.cover.bucket/Gutenberg/Gutenberg%20ID/10441/cover_10441_9.png.svg', 
+            ['http://s3.amazonaws.com/test.content.bucket/Library%20Simplified%20Open%20Access%20Content%20Server/Gutenberg%20ID/10557/Johnny%20Crow%27s%20Party.epub.images',
+             'http://s3.amazonaws.com/test.cover.bucket/Gutenberg/Gutenberg%20ID/10441/cover_10441_9.png',
```
aslagle: I'm not sure I like the way the changes affect this case. I'm preserving the original extension to handle something like "epub3" where multiple extensions can match a media type. But here the existing extension matches _a different extension_.

opds.py
```diff
@@ -653,7 +653,7 @@ def page(cls, _db, title, url, lane, annotator=None,
             title = visible_parent.display_name
         else:
             title = top_level_title
-        up_uri = annotator.groups_url(visible_parent)
+        up_uri = annotator.lane_url(visible_parent)
```
aslagle: The content_server doesn't have any grouped feeds, so it makes more sense to use lane_url. In the circulation manager, lane_url can return a grouped feed url or a page feed url depending on whether the lane has visible sublanes.

model.py
leonardr: I think it would be better to pass in a `presumed_media_type` to `get()` to handle the case where the server doesn't specify a media type but we know ahead of time what it probably is. That way we don't have a situation where a `Resource` is (temporarily) without a media type.

The url+media type constraint is designed to handle a case where a data source distinguishes between formats using content negotiation rather than distinctive URLs. So far this hasn't been an issue but it's theoretically possible and I wanted to account for it in core.

model.py
leonardr: If it's a matter of stripping them, you could use `urlparse.urlparse` on `self.url` and run the regular expression against the `.path` attribute instead of the full URL. To derive useful information out of the query parameter we'd have to write site-specific code.

tests/test_opds_import.py
leonardr: Agreed that this isn't great. Could you make it work for .epub3 and leave the .images/.noimages thing alone to be handled specially by the Gutenberg-specific code?

metadata_layer.py
```diff
@@ -1093,7 +1102,7 @@ def mirror_link(self, pool, data_source, link, link_obj, policy):
             if pool and pool.edition and pool.edition.title:
                 title = pool.edition.title
             else:
-                title = None
+                title = self.title or None
```
leonardr: I like this in theory but it feels like this is going to rename all our Project Gutenberg EPUBs that we've already stored on S3, and it's going to be a pain to rename or re-upload all of them. Do you have any ideas for this?

aslagle: Will we be running this code on the books we've already imported? I was thinking those would continue to work, but anything new that we imported would be different.

model.py
aslagle: Changed

leonardr:
:+1:

PR #209, opened 2016-05-03
Made it possible to deepcopy a metadata object with circulation data.

aslagle:
I use deepcopy in the unglue.it importer to make separate metadata objects for each book link, in case they have different rights. 

deepcopy uses pickle internally and the logger contains a thread.lock object which can't be pickled, so I moved it to the class instead of the object's state.


leonardr:
:+1:

PR #210, opened 2016-05-03
Don't store DataSource or CustomList objects in Lane objects

leonardr:
This branch changes the Lane setup so that a lane based on a data source or a set of custom lists (such as a list of best-sellers) stores the ID of the data source or the IDs of the lists instead of the database objects themselves. If we store objects, they can't be used outside the database session that created them.

As with https://github.com/NYPL-Simplified/circulation/pull/168, the existing tests verify that we maintain functionality, but I don't have a solution for testing the specific case where database sessions go in and out of scope.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -535,7 +535,7 @@ class Lane(object):
 
         # The lane may be restricted to books that are on a list
         # from a given data source.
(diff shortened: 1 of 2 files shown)


aslagle:
:+1:

PR #211, opened 2016-05-03
Dont store database objects in lanes

leonardr:
Same problem as before but with Genre objects. Same solution as before: store the IDs (and in this case the human-readable names), not the objects.


aslagle:
:+1:

PR #214, opened 2016-05-04
Staff classifications

skomputer:
This branch adds a couple things used by Circulation Manager's admin genre editing: SimplifiedGenreClassifier and  work.classifications_with_genre().

### DIFF
--- a/testing.py
+++ b/testing.py
@@ -2,6 +2,7 @@ from datetime import (
     datetime,
     timedelta,
 )
+import logging
 import os
 import shutil
 import tempfile
@@ -25,11 +26,13 @@ from model import (
     Representation,
     Resource,
     Identifier,
+    SessionManager,
     Edition,
     Work,
     WorkCoverageRecord,
     UnresolvedIdentifier,
-    get_one_or_create
+    get_one_or_create,
+    production_session
 )
 from classifier import Classifier
 from coverage import (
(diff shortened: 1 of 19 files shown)


testing.py
```diff
@@ -398,6 +400,22 @@ def _collection(self, name=u"Faketown Public Library"):
...
+    def _subject(self, type, identifier, genre=None):
+        if genre == None:
+            return get_one_or_create(
+                self._db, Subject, type=type, identifier=identifier
+            )[0]
+
+        return get_one_or_create(
+            self._db, Subject, type=type, identifier=identifier, genre=genre
+        )[0]
+
```
leonardr: I don't think you get to specify genre when creating a subject. That's up to the classifier. I see what you're doing in the test but I would take this out of the _subject() helper function because that genre is going to get immeidately overridden by whatever the classifier says.

Within the test below, it's okay to set Subject.genre to a dummy value after creating the Subject object, but I think putting it in the helper function makes a promise that can't be kept.

tests/test_model.py
```diff
@@ -1267,6 +1268,33 @@ def test_assign_genres_from_weights(self):
...
+        source = DataSource.lookup(self._db, DataSource.AXIS_360)        
+        classification1 = self._classification(
+            identifier=identifier, subject=subject1, 
+            data_source=source, weight=1)
+        classification2 = self._classification(
+            identifier=identifier, subject=subject2, 
+            data_source=source, weight=2)
+                
+        results = work.classifications_with_genre().all()
+
```
leonardr: Are these two items `classification1` and `classification2`? I think they might be. If so, you can check them for equality instead of checking that all the attributes match:

```
eq_([classification1, classification2], work.classifications_with_genre().all())
```

However you'll also need to make sure to sort the query, since sometimes databases return things in a weird order. Something like this should work:

```
classification_q = work.classifications_with_genre().order_by(Classification.id)
eq_([classification1, classification2], classification_q.all())
```

skomputer: ah yes, i'd love to check for object equality instead of checking their attributes one by one. the method being tested already sorts by weight so i can rely on the order without sorting them in the test.

testing.py
skomputer: agreed

PR #215, opened 2016-05-04
Improved exception handling when mirroring open access content.

aslagle:


model.py
```diff
@@ -3722,6 +3722,7 @@ def set_presentation_ready_based_on_content(self):
         """
         if (not self.primary_edition
             or not self.license_pools
+            or all([lp.suppressed for lp in self.license_pools])
```
leonardr: I don't agree with this. 

presentation-ready is an issue of metadata: roughly, whether we have good enough metadata to generate an OPDS <entry> that doesn't look embarassing. The question of whether we can fulfill loan requests is separate. There are a number of reasons we might not be able to fulfill loan requests: no unsuppressed license pools, no license pools at all, or empty license pools. A work that's not presentation-ready won't make it into the materialized view; a work that's suppressed will be in the materialized view but will be filtered out by Lane.works() or equivalent.

I would like to continue to keep the question "do we have enough metadata" separate from the question "do we have licenses". If we don't have enough metadata, we need to talk to the metadata wrangler or to the license provider, and until we do, nothing will change. If we don't have licenses then there's absolutely nothing we can do, but the situation might change on its own at any moment.

leonardr:
Looks great apart from the change to presentation-ready, which I disagree with.

model.py
aslagle: You're right, I'll take this out. For some reason I was thinking I needed it but works with suppressed license pools are already left out by other code.

leonardr:
:+1:

PR #216, opened 2016-05-04
testing genre fix

skomputer:
This branch fixes _work() in testing so that a genre with an invalid name isn't created by default.


leonardr:
:+1:

PR #218, opened 2016-05-05
made minor changes, with calculate_presentation now

daryanypl:


testing.py
```diff
@@ -438,6 +438,10 @@ def test_name(self):
             self.print_database_instance()  # TODO: remove before prod
             [code...]
         """
+        if not 'TESTING' in os.environ:
```
leonardr: I suggest you use logging.error() or logging.warn() here to print out a blatant warning so that we notice the code is in place and know to remove it.

leonardr:
Looks good with one minor change.

model.py
```diff
@@ -3488,15 +3467,60 @@ def calculate_presentation(self, policy=None, search_index_client=None):
...
+            for edition in self.editions:
+                if edition != self.primary_edition:
+                    edition.is_primary_for_work = False
+                else:
+                    edition.is_primary_for_work = True
+                    # let the edition know it's attached to this work now
+                    edition.work = self
+                    # let the edition's pool know they have a work
+                    if edition.is_presentation_for:
+                        edition.is_presentation_for.work = self
```
aslagle: I think we wanted this section to be part of set_primary_edition. You never want to set the primary edition without also updating is_primary_for_work on all the editions.

leonardr: Agreed. Also, without this there's not enough code in set_primary_edition to justify its existence.

model.py
```diff
@@ -3396,69 +3396,47 @@ def all_descriptions(self):
...
-        
-        WorkCoverageRecord.add_for(
-            self, operation=WorkCoverageRecord.CHOOSE_EDITION_OPERATION
-        )
 
-
-    def calculate_presentation(self, policy=None, search_index_client=None):
-        """Make a Work ready to show to patrons.
+    def calculate_primary_edition(self, policy=None, pool_to_not_check=None):
+        """ Which of this Work's Editions should be used as the default?
```
leonardr: pool_to_not_check doesn't seem to be used and I'm not sure why you would exempt one and only one license pool from consideration.

testing.py
daryanypl: done

model.py
daryanypl: ok

leonardr:
:+1: but have Amy look at it as well so she knows what changed.

model.py
```diff
@@ -3480,29 +3484,50 @@ def calculate_presentation(self, policy=None, search_index_client=None):
                 # edition is still an option, choose it.
                 new_primary_edition = potential_primary_edition
 
-        if new_primary_edition:
-            self.primary_edition = new_primary_edition
+        # Note: policy.choose_edition is true in default PresentationCalculationPolicy.
+        # If we don't have a self.primary_edition by now, this will just set all the editions' 
+        # is_primary_for_work attributes to False.
```
aslagle: I don't think the second part of this comment is true anymore - what should happen if we don't have a primary edition by now?

PR #220, opened 2016-05-05
Use NoveList API

### LINKED ISSUE #197: Add NoveList as a data source & API

courte:
Sets up configuration and a DataSource for the NoveList Select service. Allows Representations to be created (or caches to be recalled) with POST requests.

Fixes #197. Supports NYPL-Simplified/metadata_wrangler#59.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5971,8 +5971,8 @@ class Representation(Base):
 
     @classmethod
     def get(cls, _db, url, do_get=None, extra_request_headers=None,
-            params=None, accept=None, max_age=None, pause_before=0,
-            allow_redirects=True, presumed_media_type=None, debug=True):
+            accept=None, max_age=None, pause_before=0, allow_redirects=True,
+            presumed_media_type=None, debug=True):
         """Retrieve a representation from the cache if possible.
         
         If not possible, retrieve it from the web and store it in the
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -5989,9 +5989,12 @@ def get(cls, _db, url, do_get=None, extra_request_headers=None,
         :return: A 2-tuple (representation, obtained_from_cache)
 
         """
-        do_get = do_get or cls.simple_http_get
-
         representation = None
+        request_kwargs = {}
+        if params:
+            do_get = do_get or cls.simple_http_post
```
leonardr: This works, but using a POST to simulate GET because we know the result is cacheable is weird enough that I'd prefer to do everything in the do_post() method.

model.py
```diff
@@ -6147,6 +6150,12 @@ def get(cls, _db, url, do_get=None, extra_request_headers=None,
         representation.content = content
         return representation, False
 
+    @classmethod
+    def post(cls, _db, url, params, max_age=None):
```
leonardr: First, this should be called `cachable_post` or something. Most of the time the responses to POST requests are not cachable, and I don't want someone to call this thinking it's just a shortcut for making a POST request.

Second, by currying the `simple_http_post` method here you can take unsightly POST-specific code out of `get()`. Something like this:

```
def do_post(*args, *kwargs):
    request_kwargs = {'data' : params}
    kwargs.update(request_kwargs=request_kwargs)
    cls.simple_http_post(*args, **kwargs)

return cls.get(_db, url, do_get=do_post, max_age=max_age)
```

Does this make sense?

leonardr:
:postbox: :+1:

PR #221, opened 2016-05-05
Update changed books on OPDS import.

aslagle:
This branch changes the way the cutoff date is used in OPDS import - now it's compared with the time the book was updated instead of the time it was first added to the collection. 

Since the content of an open access link may have been updated as well as the metadata, I also had to make Representation.get always fetch again to see if it changed and we need to update our mirror. I made a policy for this, but maybe policy.mirror + policy.links should imply that we fetch the links again. If we fetch the content and get a 304, we don't mirror it again.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -1082,7 +1082,7 @@ class Metadata(object):
 
         if policy.link_content:
             # We want to fetch the representation again, even if we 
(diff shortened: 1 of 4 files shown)


leonardr:
Awesome. :tada: :balloon:

PR #222, opened 2016-05-06
Don't roll back database session if database session hasn't been created yet

leonardr:
This branch fixes the display of errors that happen early on in the startup process, by removing code that can mask that error with another, more stupid error--the attempt to roll back a database session that hasn't been initialized yet.

This branch also changes the ElasticSearch client configuration to increase the default timeout and number of simultaneous connections.


aslagle:
:+1:

PR #223, opened 2016-05-10
Start using `argparse` to gather command-line arguments instead of per-script hacks

leonardr:
This branch overhauls script.py to use the Python standard library's `argparse` module to parse command-line arguments. The alternative was an increasingly hacky arrangement which made it very difficult to add features to scripts.

There are basically three types of scripts:

* `IdentifierInputScript`s, which accept a named --identifier-type argument and an optional list of IDENTIFIER positional arguments. If you specify an identifier-type but no IDENTIFIERs, it's assumed that you want to operate on all identifiers of a given type. Once you parse the command line, the corresponding Identifier objects will be automatically loaded from the database and be made available as `parsed.identifiers`.

Previously we only took positional arguments and the first positional argument was equivalent to --identifier-type, unless it was the _only_ positional argument, in which case it identified the data source. I took out the data source code as not currently useful.

* `RunCoverageProviderScript`s, which are like `IdentifierInputScript`s except that they also take a --cutoff-date argument.

* `SubjectInputScript`s, which take arguments --subject-type (to restrict their operation to subjects of a specific type) and --subject-filter (to restrict their operation to subjects that match a substring filter).

We should also convert `CustomListManagementScript`, which currently doesn't really interact with the command line (it expects positional arguments in a very specific order), but that can wait til later.

I've added some front-to-back tests of the command-line parsing functionality to show how these things work.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -172,7 +172,7 @@ class IdentifierInputScript(Script):
         parser = argparse.ArgumentParser()
         parser.add_argument(
             '--identifier-type', 
-            help='Type of the identifiers to process'
+            help='Process identifiers of this type. If IDENTIFIER is not specified, all identifiers of this type will be processed. If IDENTIFIER is specified, this argument is required.'
         )
         parser.add_argument(
             'identifier_strings',
--- a/scripts.py
+++ b/scripts.py
@@ -188,22 +188,19 @@ class RunCoverageProvidersScript(Script):
 class IdentifierInputScript(Script):
     """A script that takes identifiers as command line inputs."""
 
-    def parse_identifiers(self):
-        potential_identifiers = sys.argv[1:]
-        identifiers = self.parse_identifier_list(
-            self._db, potential_identifiers
+    def arg_parser(cls):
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -300,37 +337,35 @@ class WorkProcessingScript(IdentifierInputScript):
...
-                    )
+            query = query.filter(
+                LicensePool.identifier_id.in_([x.id for x in identifiers])
             )
+        if identifier_type:
             self.log.info(
-                "Processing %d specific works." % query.count()
+                'Restricted to identifier type "%s".' % identifier_type
             )
+            query = query.filter(Identifier.type==identifier_type)
```
aslagle: Does this work if you haven't joined Identifier?

aslagle:
Looks great except my one question

scripts.py
leonardr: I think it won't work. I think I did it that way because I was planning on filtering LicensePool.data_source_id but then I got rid of the data source stuff altogether.

aslagle:
:+1:

PR #224, opened 2016-05-10
Stop opds import on error

aslagle:
This is a tiny change that stops the OPDS import monitor when it gets an error code from the link it's trying to follow, instead of trying to import the error page and failing to parse it.

Unfortunately I haven't been able to think of a reliable way to automatically resume the import later, since the books' positions in the feed may have changed. This will log the error so someone can investigate and fix problems or manually restart from an appropriate page.


leonardr:
:+1:

PR #227, opened 2016-05-10
Updated WorkConsolidationScript

aslagle:
This updates the existing WorkConsolidationScript to use the new argument parsing in WorkProcessingScript so it can use a list of identifiers as well as an identifier type. This script will delete the works for those identifiers and all references to them if force=True, and then call LicensePool.consolidate_works to generate new works.

I think this will fix https://github.com/NYPL-Simplified/operations/issues/5. When I ran it with one of the problematic ids on qa it created two new, separate works. Maybe there was a problem with permanent work ids and it's been fixed already.

I'm adding bin/repair/work_consolidation to circulation, and I came up with this query to get the ids to pass to the script:
```select string_agg(identifiers.identifier, ' ') from (select * from editions) edition1 inner join (select * from editions) edition2 on edition1.work_id=edition2.work_id and edition1.permanent_work_id!=edition2.permanent_work_id left join identifiers on edition1.primary_identifier_id=identifiers.id where edition1.data_source_id in (1) and edition2.data_source_id in (2,3,4);```

Data source 1 is Gutenberg - on QA, all the other open access editions with the problem shared a work with a Gutenberg edition, so fixing the works for the Gutenberg IDs should be enough.

There are a few cases (6 on QA) where an Overdrive or 3M edition correctly has the same permanent work id as the Gutenberg edition and I'm not sure what to do with those so I excluded them from the query for now.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -24,6 +24,7 @@ from model import (
     PresentationCalculationPolicy,
     Subject,
     Work,
+    WorkCoverageRecord,
     WorkGenre,
 )
 from external_search import (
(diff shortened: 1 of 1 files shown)


leonardr:
This looks good. I was initially worried about deleting works, because I had this whole system set up around merging works to avoid deleting them, but a) the work deleting code was already there and b) i don't think the work merging code is necessary anymore, so it's all good.

PR #229, opened 2016-05-11
Coverage providers should add coverage

leonardr:
This branch fixes two problems with coverage records not being added:

* When ensure_coverage() checked for an existing coverage record, it didn't take the operation into account. So if your coverage provider ran operation 'foo' and there was an existing record for operation 'bar', it would seem as though the 'foo' operation had completed. No work would be done and no new coverage record would be created.

* The run_on_identifiers() method was doing the work, but not creating coverage records reflecting the work done. This is because it called process_batch(), which does the work, but not running the coverage record code, which was duplicated across run_once() and ensure_coverage(). I refactored this code into a new method, process_batch_and_handle_results(). I also added tests for the code in process_batch_and_handle_results, which was previously untested.

I normalized the return value of all of these methods so they return a 2-tuple (counts, records), where `counts` is a 3-tuple (# of successes, # of transient failures, # of persistent failures) and `records` is a mixed list of `CoverageRecord` and `CoverageFailure` objects.

I tested a previously untested code path using a CoverageProvider that simply ignores all the work given to it. (This is treated as a transient failure.)

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -217,8 +217,10 @@ class CoverageProvider(object):
     def ensure_coverage(self, item, force=False):
         """Ensure coverage for one specific item.
 
-        :return: If coverage was already present, None. Otherwise,
-            either a CoverageRecord or a CoverageFailure.
+        :param force: Run the coverage code even if a CoverageRecord already
+        exists for this item.
+
+        :return: Either a CoverageRecord or a CoverageFailure.
         """
         if isinstance(item, Identifier):
             identifier = item
@@ -240,7 +242,7 @@ class CoverageProvider(object):
             else:
                 record = None
             return record
-        return None
+        return coverage_record
 
     def license_pool(self, identifier):
         """Finds or creates the LicensePool for a given Identifier."""
--- a/coverage.py
+++ b/coverage.py
@@ -111,7 +111,7 @@ class CoverageProvider(object):
         index = 0
         while index < len(identifiers):
             batch = identifiers[index:index+self.workset_size]
-            self.process_batch(batch)
+            ignore = self.process_batch_and_handle_results(batch)
             self._db.commit()
             index += self.workset_size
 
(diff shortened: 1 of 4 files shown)


coverage.py
```diff
@@ -122,49 +135,77 @@ def run_once(self, offset):
...
+        )
+
+        # Ignore transient failures.
+        return offset + transient_failures
 
+    def process_batch_and_handle_results(self, batch):
+        """:return: A 2-tuple (counts, records). 
+
+        `counts` is a 3-tuple (successes, transient failures,
+        persistent_failures).
```
aslagle: be consistent with calling it persistent vs permanent throughout this code

monitor.py
```diff
@@ -155,17 +155,17 @@ def resolve(self, unresolved_identifier):
         for provider in self.required_coverage_providers:
             if not identifier.type in provider.input_identifier_types:
                 continue
-            record = provider.ensure_coverage(identifier, force=True)
-            if record.exception:
-                raise ResolutionFailed(500, record.exception)                
+            counts, records = provider.ensure_coverage(identifier, force=True)
+            if records and records[0].exception:
+                raise ResolutionFailed(500, records[0].exception)
```
aslagle: this needs to be changed back

monitor.py
```diff
@@ -155,17 +155,17 @@ def resolve(self, unresolved_identifier):
...
 
         # Now go through the optional providers. It's the same deal,
         # but a CoverageFailure doesn't cause the entire identifier
         # resolution process to fail.
         for provider in self.optional_coverage_providers:
             if not identifier.type in provider.input_identifier_types:
                 continue
-            record = provider.ensure_coverage(identifier, force=True)
+            counts, records = provider.ensure_coverage(identifier, force=True)
```
aslagle: this too

monitor.py
leonardr: Sorry that this is a moving target. :sweat_smile:

leonardr:
OK, I've changed ensure_coverage() so that it always returns either a CoverageRecord or a CoverageFailure, even if it ends up not doing any work. This should remove the need to make changes in other branches, and I think it also removes some bugs where some code was really expecting to get something back from ensure_coverage(), and receiving None would have crashed it.

aslagle:
:+1:

PR #230, opened 2016-05-11
Coverage providers should add coverage

leonardr:
This branch fixes two _more_ coverage provider problems.

* CoverageProvider.missing_coverage_from was counting an identifier as covered if it had _any_ coverage record from this data source, even if the operation didn't match. This made it think it had covered identifiers it hadn't actually covered.

* CoverageProvider.run_on_identifiers wasn't checking the identifiers it was given to see if they were covered already. It just went ahead and ran the provider no matter what. Now it does check, and any identifiers that have up-to-date coverage are treated as automatic successes.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -88,7 +88,7 @@ class CoverageProvider(object):
         """
         return Identifier.missing_coverage_from(
             self._db, self.input_identifier_types, self.output_source,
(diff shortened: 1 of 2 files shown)


aslagle:
👍

PR #231, opened 2016-05-11
Add series position to editions

courte:
This veritable smörgåsbord includes:
- An annoying number of whitespace changes courtesy of my home computer! (Sorry!)
- Adding the `series_position` to Editions, including factoring basic edition fields out in the Metadata class
- Creating NoveList identifiers
- Moving the subtitle extractor into core/util, as suggested in NYPL-Simplified/metadata_wrangler#59
- Keeping identifiers from being equivalent to themselves

Supports NYPL-Simplified/metadata_wrangler#59


leonardr:
I have one question and would like to see a little more test coverage.

* What should happen in metadata.apply() if series_position is present but series is not? Is that okay?
* Change TestMetadata.test_from_edition to show that series_position is copied over.
* Change TestMetadata.test_apply to show that series_position gets applied to the Edition. (I realize that test_apply only tests the title now, but we basically know that all the existing stuff works, so I'm not too worried about it. But if you want to put in the work to have it test other fields that would also be fine.)
* Add a test to TestMetadata that shows that an IdentifierData that's the same as the primary identifier is ignored and not turned into an equivalency.

leonardr:
To answer my own question, it's fine if series_position is present but series is not. That will happen in real life when someone uses the admin interface to change series_position but leaves series alone.

metadata_layer.py
```diff
@@ -945,8 +945,11 @@ def apply(self, edition, metadata_client=None, replace=None,
         # TODO: remove equivalencies when replace.identifiers is True.
         if self.identifiers is not None:
             for identifier_data in self.identifiers:
-                if (identifier==identifier_data or
-                    not identifier_data.identifier):
+                if not identifier_data.identifier:
+                    continue
+                if (identifier_data.identifier==identifier.identifier and
+                    identifier_data.type==identifier.type):
+                    # These are the same identifier.
```
leonardr: Good catch.

model.py
```diff
@@ -6377,11 +6377,18 @@ def get(cls, _db, url, do_get=None, extra_request_headers=None,
...
         """Transforms cacheable POST request into a Representation"""
 
         def do_post(url, headers, **kwargs):
             kwargs.update({'data' : params})
+            if response_reviewer:
+                # An optional function passed to raise errors if the
+                # post response isn't worth caching.
+                return response_reviewer(
+                    cls.simple_http_post(url, headers, **kwargs)
+                )
```
leonardr: I think response_reviewer is useful enough to go into get(). I think that @aslagle will need something similar to make opds_import.py more reliable--sometimes you try to mirror a URL listed in an OPDS feed and the thing on the other end is just wrong.

It's up to you and @aslagle whether it's better to make this change in this branch or in a future branch.

leonardr:
:+1: with the comment that response_reviewer should eventaully end up in the core get method.

PR #232, opened 2016-05-11
Possible fix for calculate_work loop

aslagle:
This is one way we could fix the loop in LicensePool.calculate_work.

There are two changes:
- metadata.apply(edition) no longer updates the work, it only calls edition.calculate_presentation (which already no longer calls work.calculate_presentation).
- license_pool.calculate_work always calls work.calculate_presentation, even if it returns a work that already existed.

I looked at all the calls to metadata.apply in core, and there's only one where we don't also call either work.calculate_presentation or license_pool.calculate_work, in TitleFromExternalList.to_edition. I"m not sure, but I don't think we're using metadata from the external lists in the works' presentation. If we are, we could also call work.calculate_presentation there.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5565,7 +5565,9 @@ class LicensePool(Base):
             if primary_edition:
                 primary_edition.work = self.work
             
-            # The work has already been done.
+            # The work has already been done. Make sure the work's
+            # display is up to date.
+            self.work.calculate_presentation()
             return self.work, False
 
 
@@ -5577,7 +5579,7 @@ class LicensePool(Base):
                      self.identifier)
             
             return None, False
-        if primary_edition.license_pool != self:
+        if primary_edition.is_presentation_for != self:
             raise ValueError(
                 "Primary edition's license pool is not the license pool for which work is being calculated!")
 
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -5577,7 +5579,7 @@ def calculate_work(self, even_if_no_author=False, known_edition=None):
                      self.identifier)
             
             return None, False
-        if primary_edition.license_pool != self:
+        if primary_edition.is_presentation_for != self:
```
aslagle: This is to fix Edition.license_pool not working for a presentation edition.

opds_import.py
```diff
@@ -273,7 +273,6 @@ def import_from_metadata(
                 even_if_no_author=even_if_no_author,
             )
             if work:
-                work.calculate_presentation()
```
aslagle: This is no longer necessary since LicensePool.calculate_work always calls work.calculate_presentation. Before, there were cases when we would call calculate_presentation twice.

aslagle:
An alternate way to fix this would be to move line https://github.com/NYPL-Simplified/server_core/blob/master/model.py#L5640 so it sets the work on the edition after it calls work.calculate_presentation.

PR #234, opened 2016-05-12
SimplifiedGenreClassifier returns correct fiction status

skomputer:
This branch has SimplifiedGenreClassifier look up a genre name in the canonical genre data to determine whether it's fiction or nonfiction, instead of the default method of looking for "fiction" or "nonfiction" in the genre name.


leonardr:
:+1:

PR #235, opened 2016-05-13
Don't give books the same Work just because we have equally little information about any of them

leonardr:
If we have no information about two Editions, they end up with the same permanent work ID and they can end up grouped together. This branch puts a stop to that in as many ways as I can think of:

* It stops a permanent work ID from being generated for an Edition that has no title.
* It stops a Work from being generated for an Edition that has no title. (This should have been happening before, but I made it more explicit.)

### DIFF
--- a/model.py
+++ b/model.py
@@ -5592,13 +5592,18 @@ class LicensePool(Base):
             primary_edition.calculate_presentation()
 
         if not primary_edition.title:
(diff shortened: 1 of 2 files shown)


aslagle:
This looks good. To be safe we could call primary_edition.calculate_permanent_work_id(), even if it already exists.

aslagle:
:+1:

PR #236, opened 2016-05-13
Make bibliographic coverage providers runnable from scripts

leonardr:
This branch upgrades the various BibliographicCoverageProvider classes so they can be run with the RunCoverageProviderScript. This means accepting (and ignoring) the standard `input_identifier_types` argument, and also accepting (and passing on) general `**kwargs` such as the `cutoff_time` now gathered from the command line and passed into the `CoverageProvder` constructor.


courte:
:+1:

PR #238, opened 2016-05-13
Make BibliographicRefreshScript work without specific identifiers

leonardr:
This branch makes BibliographicRefreshScript a script you can just run, without specifying any identifiers, and it will take care of bibliographic coverage for all of your licensed books. This gives you a single script you can run instead of having to individually run `overdrive_bibliographic_coverage` and similar scripts.

It also incidentally adds `--cutoff-date` support to BibliographicRefreshScript.


aslagle:
:+1:

PR #239, opened 2016-05-13
Lookups without licensepools

### LINKED ISSUE #217: LookupAcquisitionFeed silently drops works that have no active license pool
LookupAcquisitionFeed takes a list of works to turn into OPDS entires and a list of messages regarding identifiers that cannot be turned into OPDS entries.

But there's a third case that currently slips through the cracks: works that are presentation-ready but have no active LicensePool. This includes open-access works that have no download link. These works get filtered out by LookupAcquisitionFeed.create_entry() and the client that was trying to look something up never hears about them at all.

There are two problems here. First, LookupAcquisitionFeed is primarily used by the metadata wrangler, and whether or not a book has an active license pool is not really a concern for the metadata wrangler--no one's actually trying to borrow the book, someone asked for the data and if we have it we should give it to them.

Second, even a real error happens, the book shouldn't just disappear from the OPDS feed. It should be turned into an appropriate message like 'I decided not to show you this book because there's no active license pool.'

courte:
This branch adjusts the LookupAcquisitionFeed so that it doesn't lose entries for works without license pools, for all the reasons detailed in #217. 

I have a little confusion around whether a work's primary_edition is still a viable way to get edition information in a post-presentation_edition world. I'm also not sure `UNRESOLVABLE_URN` is the right error message here -- would something new, like `ANOMALOUS_IDENTIFIER` be more useful? Particularly given this is something we _shouldn't_ see happening?

Fixes #217.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -372,12 +372,13 @@ class URNLookupController(object):
         # We made it!
         return entry
 
-    def work_lookup(self, annotator, controller_name='lookup', collection=None):
+    def work_lookup(self, annotator, route_name='lookup',
+                    require_active_licensepool=True, collection=None):
         """Generate an OPDS feed describing works identified by identifier."""
         urns = flask.request.args.getlist('urn')
 
         messages_by_urn = dict()
-        this_url = cdn_url_for(controller_name, _external=True, urn=urns)
+        this_url = cdn_url_for(route_name, _external=True, urn=urns)
         for urn in urns:
             code, message = self.process_urn(urn, collection=collection)
             if code:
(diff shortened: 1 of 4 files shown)


leonardr:
You're right, UNRESOLVABLE_URN is the wrong one to use. I would create a new problem detail that's the equivalent of 'internal server error', Its status code would be 500 and it would basically say "something went wrong, I don't know what but it's my fault."

You're in a position to make some more specific error messages. For instance, if the work exists but has no title, you could return a message saying 'I know what work you're talking about but I don't have enough information to service a lookup request.' 

Work.primary_edition is still the way to go--it's a presentation edition associated with one of the work's LicensePools.

I'm a little uncomfortable with the circulation manager processing lookups for works that aren't in the collection anymore. How would you feel about adding an argument to the constructor `require_active_licensepool`? The metadata wrangler would pass in False for that argument and the circ manager/content server would pass in True. If you pass in True, then lookup fails for works not currently in the collection (either with a generic 'not found' message or with a message like 'I know what work you're talking about but it's not currently in the collection'). 

This last one isn't a big deal--I would :+1: if you just changed UNRESOLVABLE_URN to INTERNAL_SERVER_ERROR.

courte:
I added INTERNAL_SERVER_ERROR and started returning 404's when an active licensepool is required (the default behavior). I think it's good to go.

leonardr:
:+1:

PR #244, opened 2016-05-16
Removed Work.merge_into

### LINKED ISSUE #241: Get rid of Work.was_merged_into_id
Our work merging code currently keeps the old work around, even though we no longer need it. We should be able to destroy the old work once it's merged--just move over its license pools to the new work and recalculate the new work.

It might be worth it to also take another look at Work.merge_into and Work.similarity_to. This code contains very old code for determining if two Works are "the same", and I'm not really comfortable with it now that we use permanent work ID to make that decision instead.

aslagle:
Work.merge_into isn't used anymore, and Work.similarity_to is only used in one other obsolete piece of code that I will also remove.

Fixes #241

### DIFF
--- a/migration/20160516-remove-work-was-merged-into.sql
+++ b/migration/20160516-remove-work-was-merged-into.sql
@@ -1,2 +1,219 @@
 DELETE FROM works where was_merged_into_id is not null;
 ALTER TABLE works DROP COLUMN was_merged_into_id CASCADE;
+
+create materialized view mv_works_editions_datasources_identifiers
+as
+ SELECT 
+    distinct works.id AS works_id,
+    editions.id AS editions_id,
+    editions.data_source_id,
+    editions.primary_identifier_id,
+    editions.sort_title,
+    editions.permanent_work_id,
+    editions.sort_author,
+    editions.medium,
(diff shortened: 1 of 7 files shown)


migration/20160516-remove-work-was-merged-into.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM works where was_merged_into_id is not null;
+ALTER TABLE works DROP COLUMN was_merged_into_id CASCADE;
```
aslagle: This drops the materialized views - what's the right way to create them again?

leonardr: Recreate the new views by copying your new versions of materialized_view_works.sql and materialized_view_works_workgenres.sql to the bottom of the migration script.

leonardr:
:+1: once the migration script recreates the materialized views.

migration/20160516-remove-work-was-merged-into.sql
aslagle: Done

leonardr:
Looks good!

PR #245, opened 2016-05-16
Improve work consolidation

leonardr:
This branch fixes a number of problems with the work consolidation script and the workflow it triggers. It was driven by my attempt to fix bad works in a real database.

The biggest change is that when you invoke the work consolidation script on a specific identifier, it deletes any preexisting Work for that identifier, and it makes sure to call calculate_work() on _every_ `LicensePool` that used to be associated with that work. Basically, it deletes one `Work` and creates two new `Work`s in its place. Without the delete code, `LicensePool`s that are in bad `Work`s never leave those `Work`s. Without the loop over `LicensePool`s, deleting a bad `Work` creates orphan `LicensePool`s.

I think the code that deletes a work is becoming too important to sit untested in a script class, but this is good enough for now.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5532,7 +5532,8 @@ class LicensePool(Base):
         return hold, new
 
     @classmethod
-    def consolidate_works(cls, _db, calculate_work_even_if_no_author=False):
+    def consolidate_works(cls, _db, calculate_work_even_if_no_author=False,
+                          batch_size=10):
         """Assign a (possibly new) Work to every unassigned LicensePool."""
         a = 0
         lps = cls.with_no_work(_db)
@@ -5549,9 +5550,9 @@ class LicensePool(Base):
                 continue
             a += 1
             logging.info("When consolidating works, created %r", etext)
-            if a and not a % 10:
+            if a and not a % batch_size:
                 _db.commit()
-
+        _db.commit()
 
 
     def calculate_work(self, even_if_no_author=False, known_edition=None):
--- a/model.py
+++ b/model.py
@@ -3256,7 +3256,9 @@ class Work(Base):
     def all_identifier_ids(self, recursion_level=5):
         _db = Session.object_session(self)
         primary_identifier_ids = [
-            x.primary_identifier.id for x in self.editions]
+            x.primary_identifier.id for x in self.editions
+            if x.primary_identifier
+        ]
         identifier_ids = Identifier.recursively_equivalent_identifier_ids_flat(
             _db, primary_identifier_ids, recursion_level)
         return identifier_ids
@@ -3557,6 +3559,7 @@ class Work(Base):
 
             primary_identifier_ids = [
                 x.primary_identifier.id for x in self.editions
+                if x.primary_identifier
             ]
             data = Identifier.recursively_equivalent_identifier_ids(
                 _db, primary_identifier_ids, 5, threshold=0.5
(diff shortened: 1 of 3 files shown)


model.py
```diff
@@ -3256,7 +3256,9 @@ def all_editions(self, recursion_level=5):
     def all_identifier_ids(self, recursion_level=5):
         _db = Session.object_session(self)
         primary_identifier_ids = [
-            x.primary_identifier.id for x in self.editions]
+            x.primary_identifier.id for x in self.editions
+            if x.primary_identifier
+        ]
```
leonardr: If a Work is associated with an Edition that has no primary identifier, that's weird, but that in itself shouldn't prevent `all_identifier_ids` from working.

model.py
```diff
@@ -5542,9 +5550,9 @@ def consolidate_works(cls, _db, calculate_work_even_if_no_author=False):
                 continue
             a += 1
             logging.info("When consolidating works, created %r", etext)
-            if a and not a % 100:
+            if a and not a % batch_size:
                 _db.commit()
-
+        _db.commit()
```
leonardr: Was missing a final commit, which left the final batch forever unprocessed.

scripts.py
```diff
@@ -417,23 +418,48 @@ def do_run(self):
         work_ids_to_delete = set()
         unset_work_id = dict(work_id=None)
 
-        if self.force:
-            self.clear_existing_works()                  
+        if self.identifiers:
```
leonardr: This is the core of my change. If you specify a set of identifiers on the command line, you're asking that the Works for the LicensePools corresponding to those identifiers be recalculated from scratch.

scripts.py
```diff
@@ -417,23 +418,48 @@ def do_run(self):
...
-        LicensePool.consolidate_works(self._db)
+                if pool.work:
+                    # We're about to delete a preexisting work. If the
+                    # problem is that this LicensePool is incorrectly
+                    # grouped together with some other LicensePool,
+                    # then that LicensePool must also have
+                    # calculate_work() called on it, so that we can
+                    # create two Works where there used to be one.
+                    all_pools = pool.work.license_pools
+                    self.clear_works(pool.work.id)
```
leonardr: This requires calling `clear_works`.

scripts.py
```diff
@@ -417,23 +418,48 @@ def do_run(self):
...
+                else:
+                    all_pools = [pool]
+                for pool in all_pools:
```
leonardr: And it also has ripple effects on other LicensePools associated (probably incorrectly) with this Work.

scripts.py
```diff
@@ -417,23 +418,48 @@ def do_run(self):
...
-        for i in self._db.query(Work).filter(Work.primary_edition==None):
-            self._db.delete(i)            
-        self._db.commit()
+            qu = self._db.query(Work).filter(Work.primary_edition==None)
+            logging.info("Deleting %d Works that lack Editions." % qu.count())
+            for i in qu:
+                self._db.delete(i)            
+            self._db.commit()
 
     def clear_existing_works(self):
```
aslagle: I think you can delete this function

aslagle:
:+1:

scripts.py
leonardr: The problem was I'd forgotten to add back the code that calls the function.

PR #246, opened 2016-05-17
Immediate bibliographic coverage upon discovery of book

leonardr:
This wide-ranging branch fixes a number of problems that stand between me and Sensible Behavior.

This is Sensible Behavior:

* When the Overdrive or 3M circulation monitor hears about a book it hasn't encountered before, it immediately provides that book with bibliographic coverage, creates a Work, and sets the Work to presentation ready. 

* A 3M book will have no cover image until we hear from the Metadata wrangler, because 3M doesn't provide thumbnail images. An Overdrive book will start using the Overdrive thumbnail image immediately, but once we hear from the metadata wrangler, we'll switch to using the MW thumbnail image.

Here are the problems:

* The code that sets newly created Overdrive and 3M works to presentation-ready was located in custom subclasses located in `circulation`. But the script that provides bibliographic coverage was located in `core` and was using the parent classes located in `core`, which doesn't run this code. I moved the code into `core` and will be removing it from `circulation`.

* The metadata wrangler serves OPDS categories that classify a book under Simplified genres:

`<category term="http://librarysimplified.org/terms/genres/Simplified/Urban%20Fiction" scheme="http://librarysimplified.org/terms/genres/Simplified/" label="Urban Fiction"/>`

The OPDS import process runs these classifications through the `SimplifiedGenreClassifier`, but the `SimplifiedGenreClassifier` can't handle an identifier like "http://librarysimplified.org/terms/genres/Simplified/Urban%20Fiction". It's expecting "Urban Fiction". I added a `scrub_identifier` implementation to make `SimplifiedGenreClassifier` handle both Simplified genre URIs and normal genre names.

* There's no way to say "All else being equal, a book cover from the metadata wrangler is better than one that comes from Overdrive." `Identifier.best_cover_for()` is the natural place to put this code, but that method is a mess of code that has no unit tests and currently sits mostly unused. It was designed for a no-longer-used system where we would give a Gutenberg book a decent book cover by gathering covers from commercially published editions of those works, and picking the 'best' one. It's still potentially useful for resolving conflicts, but most of the time, if there are two covers for the same book, it's the case I'm trying to fix--one cover comes from the license source and one comes from the metadata wrangler, and the data source, not the aspect ratios, are the deciding factor in which one we display to patrons.

Anyway, I had to refactor all that code and then, to make sure I didn't break anything, write tests for as much of it as I could. (I think I fixed several preexisting bugs in the code as I did this.) My current implementation gives a cover image a big quality bump if it comes from one of the data sources in `DataSource.PRESENTATION_EDITION_PRIORITY` -- basically the metadata wrangler or the admin interface.

* Here's the problem I can't fix yet. Because opds_import still uses `license_data_source`, when we import a cover image for an Overdrive book from the metadata wrangler, it's inserted into the database as though it came from Overdrive. This used to be desirable behavior, but now that we have presentation edition code it's not necessary, and now that we _also_ get a cover image direct from Overdrive it's downright harmful. It means that there's no way to _stop_ using the Overdrive cover image. Even with my code in place, we're talking about comparing two images that are approximately the same aspect ratio and both (supposedly) come from Overdrive. 

Once we stop using `license_data_source` and start being honest about where the metadata wrangler data comes from, this problem should be fixed automatically.

### DIFF
--- a/model.py
+++ b/model.py
@@ -4168,7 +4168,7 @@ class Resource(Base):
 
     # The point at which a generic geometric image is better
     # than a lousy cover we got from the Internet.
-    MINIMUM_IMAGE_QUALITY = 1.0/8
+    MINIMUM_IMAGE_QUALITY = 0.25
 
     id = Column(Integer, primary_key=True)
 
--- a/model.py
+++ b/model.py
@@ -1584,10 +1584,6 @@ class Identifier(Base):
     IDEAL_IMAGE_HEIGHT = 240
     IDEAL_IMAGE_WIDTH = 160
 
-    # The point at which a generic geometric image is better
-    # than some other image.
-    MINIMUM_IMAGE_QUALITY = 0.25
-
     @classmethod
     def best_cover_for(cls, _db, identifier_ids):
         # Find all image resources associated with any of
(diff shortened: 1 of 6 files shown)


model.py
```diff
@@ -4355,16 +4292,62 @@ def update_quality(self):
                          ((self.voted_quality or 0) * votes_for_quality))
         self.quality = total_quality / float(total_weight)
 
-    def set_representation(self, media_type, content, uri=None,
```
leonardr: I took this method out. Nobody calls it and it's clearly broken. It refers to a `_db` variable that doesn't exist and calls a nonexistent `.generic_uri` method.

tests/test_model.py
```diff
@@ -2518,6 +2521,165 @@ def test_book_smaller_than_thumbnail_size(self):
...
+    def test_best_covers_among(self):
+        # Here's a book with a thumbnail image.
+        edition, pool = self._edition(with_license_pool=True)
+
+        hyperlink, ignore = pool.add_link(
+            Hyperlink.THUMBNAIL_IMAGE, self._url, pool.data_source
+        )
+        resource_with_no_representation = hyperlink.resource
+
+        link1, ignore = pool.add_link(
```
courte: Did you mean to do this twice?

courte:
👍

PR #247, opened 2016-05-17
Remove license data source

leonardr:
This is Darya's branch to change the way we do OPDS imports. An OPDS import no longer edits the 'real' `Edition` obtained from the license source--it creates a new `Edition` representing the provider of the OPDS feed's view of certain `Identifier`s and `LicenseSource`s. As such, there is no need for `license_data_source`, which  keeps track of which one is the 'real' `Edition`.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -342,7 +342,7 @@ class CoverageProvider(object):
         return work
 
 
-    def set_meta_circ_data(self, identifier, metadata, circulationdata, 
+    def set_metadata_and_circulation_data(self, identifier, metadata, circulationdata, 
         metadata_replacement_policy=None, 
         circulationdata_replacement_policy=None, 
     ):
@@ -353,6 +353,7 @@ class CoverageProvider(object):
 
         TODO:  Makes assumption of one license pool per identifier.  In a 
         later branch, this will change.
+        TODO:  Update doc string removing reference to past function.
 
         :return: The Identifier (if successful) or an appropriate
         CoverageFailure (if not).
--- a/coverage.py
+++ b/coverage.py
@@ -487,7 +487,8 @@ class BibliographicCoverageProvider(CoverageProvider):
     CAN_CREATE_LICENSE_POOLS = True
 
     def __init__(self, _db, api, datasource, workset_size=10,
-                 metadata_replacement_policy=None, cutoff_time=None
+                 metadata_replacement_policy=None, circulationdata_replacement_policy=None, 
+                 cutoff_time=None
     ):
         self._db = _db
         self.api = api
@@ -497,7 +498,11 @@ class BibliographicCoverageProvider(CoverageProvider):
         metadata_replacement_policy = (
             metadata_replacement_policy or ReplacementPolicy.from_metadata_source()
         )
+        circulationdata_replacement_policy = (
+            circulationdata_replacement_policy or ReplacementPolicy.from_license_source()
+        )
         self.metadata_replacement_policy = metadata_replacement_policy
+        self.circulationdata_replacement_policy = circulationdata_replacement_policy
         super(BibliographicCoverageProvider, self).__init__(
             service_name,
             input_identifier_types, output_source,
--- a/coverage.py
+++ b/coverage.py
@@ -341,11 +341,45 @@ class CoverageProvider(object):
             return CoverageFailure(self, identifier, e, transient=True)
         return work
 
-    def set_metadata(self, identifier, metadata, 
+
+    def set_meta_circ_data(self, identifier, metadata, circulationdata, 
+        metadata_replacement_policy=None, 
+        circulationdata_replacement_policy=None, 
+    ):
+        """
+        Performs the function of the old set_metadata.  Finds or creates the Edition 
+        and the LicensePool for the passed-in Identifier, updates them, 
+        then finds or creates a Work for them.
+
+        TODO:  Makes assumption of one license pool per identifier.  In a 
+        later branch, this will change.
+
+        :return: The Identifier (if successful) or an appropriate
+        CoverageFailure (if not).
+        """
+
+        result = self._set_metadata(identifier, metadata, metadata_replacement_policy)
+        if isinstance(result, CoverageFailure):
+            return result
+        
+        result = self._set_circulationdata(identifier, circulationdata, circulationdata_replacement_policy)
+        if isinstance(result, CoverageFailure):
+            return result
+
+        # now that made sure that have an edition and a pool on the identifier, 
+        # can try to make work
+        work = self.work(identifier)
+        if isinstance(work, CoverageFailure):
+            return work
+
+        return identifier
+
+
+    def _set_metadata(self, identifier, metadata, 
                      metadata_replacement_policy=None
     ):
         """Finds or creates the Edition for an Identifier, updates it
-        with the given metadata, then creates a Work for the book.
+        with the given metadata.
 
         :return: The Identifier (if successful) or an appropriate
         CoverageFailure (if not).
@@ -373,11 +407,46 @@ class CoverageProvider(object):
             )
             return CoverageFailure(self, identifier, repr(e), transient=True)
 
-        work = self.work(identifier)
-        if isinstance(work, CoverageFailure):
-            return work
+        return identifier
+
+
+    def _set_circulationdata(self, identifier, circulationdata, 
+                     circulationdata_replacement_policy=None
+    ):
+        """Finds or creates the LicensePool for an Identifier, updates it
+        with the given circulationdata, then creates a Work for the book.
+
+        TODO:  Makes assumption of one license pool per identifier.  In a 
+        later branch, this will change.
+
+        :return: The Identifier (if successful) or an appropriate
+        CoverageFailure (if not).
+        """
+        circulationdata_replacement_policy = circulationdata_replacement_policy or (
+            ReplacementPolicy.from_license_source()
+        )
+
+        pool = self.license_pool(identifier)
+        if isinstance(pool, CoverageFailure):
+            return pool
+
+        if not circulationdata:
+            e = "Did not receive circulationdata from input source"
+            return CoverageFailure(self, identifier, e, transient=True)
+
+        try:
+            circulationdata.apply(
+                pool, replace=circulationdata_replacement_policy,
+            )
+        except Exception as e:
+            self.log.warn(
+                "Error applying circulationdata to pool %d: %s",
+                pool.id, e, exc_info=e
+            )
+            return CoverageFailure(self, identifier, repr(e), transient=True)
 
         return identifier
+
 
     def set_presentation_ready(self, identifier):
         """Set a Work presentation-ready."""
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -446,6 +446,7 @@ class CirculationData(object):
                 # If a link has a rights_uri, make that the overall rights_uri. 
                 # TODO: If there are multiple links with a rights_uri, make them go into individual delivery mechanisms 
                 # (first need to move rights_uri onto delivery mechanisms). 
+                # TODO: make sure am not writing anything to DB here, only on apply.
                 if link.rights_uri and not self.default_rights_uri:
                     self.set_default_rights_uri(data_source_name=self.data_source_name, default_rights_uri=link.rights_uri)
 
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -672,7 +672,7 @@ class CirculationData(MetaToModelUtility):
                 _db, RightsStatus, uri=self.default_rights_uri
             )
 
-            last_checked = datetime.datetime.utcnow()
+            last_checked = self.last_checked or datetime.datetime.utcnow()
             license_pool, is_new = LicensePool.for_foreign_id(
                 _db, data_source=self.data_source_obj,
                 foreign_id_type=self.primary_identifier.type, 
@@ -738,7 +738,7 @@ class CirculationData(MetaToModelUtility):
         """  Update the passed-in license pool with this CirculationData's information.
         """
         made_changes = False
-
+        
         if pool is None:
             raise ValueError("CirculationData.apply needs a pool.")
 
@@ -815,7 +815,7 @@ class CirculationData(MetaToModelUtility):
                 # CirculationData object in the bibliographic parser for a 3m log, then 
                 # check if that log update value is after the license pool's current knowledge, 
                 # and only update license infos if new information is really new.
-                if (pool.last_checked and (self.last_checked > pool.last_checked)):
+                if (pool.last_checked and (self.last_checked >= pool.last_checked)):
                     # Update availabily information. This may result in the issuance
                     # of additional events.
                     changed_licenses = pool.update_availability(
@@ -1107,89 +1107,6 @@ class Metadata(MetaToModelUtility):
             self.primary_identifier.identifier,
             create_if_not_exists=create_if_not_exists
         )
-
-
-    def set_default_rights_uri(self, data_source):
-        if self.rights_uri == None and data_source:
-            # We haven't been able to determine rights from the metadata, so use the default rights
-            # for the data source if any.
-            default = RightsStatus.DATA_SOURCE_DEFAULT_RIGHTS_STATUS.get(data_source.name, None)
-            if default:
-                self.rights_uri = default
-
-        for format in self.formats:
-            if format.link:
-                link = format.link
-                if self.rights_uri in (None, RightsStatus.UNKNOWN) and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
-                    # We haven't determined rights from the metadata or the data source, but there's an
-                    # open access download link, so we'll consider it generic open access.
-                    self.rights_uri = RightsStatus.GENERIC_OPEN_ACCESS
-
-        if self.rights_uri == None:
-            # We still haven't determined rights, so it's unknown.
-            self.rights_uri = RightsStatus.UNKNOWN
-
-    def license_pool(self, _db):
-        if not self.primary_identifier:
-            raise ValueError(
-                "Cannot find license pool: metadata has no primary identifier."
-            )
-
-        license_pool = None
-        is_new = False
-
-        identifier_obj, ignore = self.primary_identifier.load(_db)
-
-        metadata_data_source = self.data_source(_db)
-        license_data_source = self.license_data_source(_db)
-
-        self.set_default_rights_uri(metadata_data_source)
-        if license_data_source:
-            can_create_new_pool = True
-            check_for_licenses_from = [license_data_source]
-        else:
-            check_for_licenses_from = DataSource.license_sources_for(
-                _db, identifier_obj
-            ).all()
-            if len(check_for_licenses_from) == 1:
-                # Since there is only one source for this kind of book,
-                # we can create a new license pool if necessary.
-                can_create_new_pool = True
-                self.license_data_source_obj = check_for_licenses_from[0]
-            elif metadata_data_source in check_for_licenses_from:
-                # We can assume that the license comes from the same
-                # source as the metadata.
-                self.license_data_source_obj = metadata_data_source
-                can_create_new_pool = True
-            else:
-                # We might be able to find an existing license pool
-                # for this book, but we won't be able to create a new
-                # one, because we don't know who's responsible for the
-                # book.
-                can_create_new_pool = False
-
-        license_data_source = self.license_data_source(_db)
-        for potential_data_source in check_for_licenses_from:
-            license_pool = get_one(
-                _db, LicensePool, data_source=potential_data_source,
-                identifier=identifier_obj
-            )
-            if license_pool:
-                break
-
-        if not license_pool and can_create_new_pool:
-            rights_status = get_one(_db, RightsStatus, uri=self.rights_uri)
-            license_pool, is_new = LicensePool.for_foreign_id(
-                _db, self.license_data_source_obj,
-                self.primary_identifier.type,
-                self.primary_identifier.identifier,
-                rights_status=rights_status,
-            )
-            if self.has_open_access_link:
-                license_pool.open_access = True
-            if self.rights_uri:
-                license_pool.set_rights_status(self.rights_uri)
-        return license_pool, is_new
 
 
     def consolidate_identifiers(self):
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -442,7 +442,22 @@ class CirculationData(object):
             )
 
 
+    def __repr__(self):
+        description_string = '<CirculationData primary_identifier="%s" licenses_owned="%s" licenses_available="%s" licenses_reserved=%d' % (
+            self.primary_identifier, self.licenses_owned, self.licenses_available, self.licenses_reserved
+        )
+        description_string += 'patrons_in_hold_queue="%s" default_rights_uri="%s" first_appearance="%s" last_checked="%s" >' % (
+            patrons_in_hold_queue, default_rights_uri, first_appearance, last_checked
+        )
+        description_string += 'links="%r" formats="%r" data_source_obj="%r">' % (
+            links, formats, data_source_obj
+        )
+
+        return description_string
+
+
     def data_source(self, _db):
+        set_trace()
         if not self.data_source_obj:
             if self._data_source:
                 obj = DataSource.lookup(_db, self._data_source)
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -380,9 +380,7 @@ class MetaToModelUtility(object):
         The model_object can be either a pool or an edition.
         """
 
-        if link_obj.rel not in (
-            Hyperlink.IMAGE, Hyperlink.OPEN_ACCESS_DOWNLOAD
-        ):
+        if link_obj.rel not in Hyperlink.MIRRORED:
             # we only host locally open-source epubs and cover images
             return
 
@@ -582,6 +580,7 @@ class CirculationData(MetaToModelUtility):
             pools (the rights-related links).
         """
         # start by deleting any old links
+        set_trace()
         self.__links = []
 
         if not arg_links:
@@ -590,7 +589,7 @@ class CirculationData(MetaToModelUtility):
         self.set_default_rights_uri(data_source_name=self.data_source_name)
 
         for link in arg_links:
-            if link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+            if link.rel in Hyperlink.CIRCULATION_ALLOWED:
                 # TODO:  what about Hyperlink.SAMPLE?
                 # only accept the types of links relevant to pools
                 self.__links.append(link)
@@ -603,8 +602,7 @@ class CirculationData(MetaToModelUtility):
                     self.set_default_rights_uri(data_source_name=self.data_source_name, default_rights_uri=link.rights_uri)
 
                 # An open-access link or open-access rights implies a FormatData object.
-                open_access_link = (link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD
-                                    and link.href)
+                open_access_link = (link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD and link.href)
                 # try to deduce if the link is open-access, even if it doesn't explicitly say it is
                 open_access_rights_link = (link.media_type in Representation.BOOK_MEDIA_TYPES 
                                            and link.href
@@ -759,7 +757,7 @@ class CirculationData(MetaToModelUtility):
         link_objects = {}
 
         for link in self.links:
-            if link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+            if link.rel in Hyperlink.CIRCULATION_ALLOWED:
                 link_obj, ignore = identifier.add_link(
                     rel=link.rel, href=link.href, data_source=data_source, 
                     license_pool=pool, media_type=link.media_type,
@@ -768,7 +766,7 @@ class CirculationData(MetaToModelUtility):
                 link_objects[link] = link_obj
 
         for link in self.links:
-            if link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+            if link.rel in Hyperlink.CIRCULATION_ALLOWED:
                 link_obj = link_objects[link]
                 if replace.mirror:
                     # We need to mirror this resource. If it's an image, a
@@ -949,7 +947,7 @@ class Metadata(MetaToModelUtility):
             return
 
         for link in arg_links:
-            if link.rel not in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+            if link.rel in Hyperlink.METADATA_ALLOWED:
                 # only accept the types of links relevant to editions
                 self.__links.append(link)
                 
@@ -1320,17 +1318,11 @@ class Metadata(MetaToModelUtility):
                     surviving_hyperlinks.append(hyperlink)
             if dirty:
                 identifier.links = surviving_hyperlinks
-
-
-        # TODO:  also, when we call apply() from test_metadata.py:TestMetadataImporter.test_open_access_content_mirrored, 
-        # pool gets set, but when we call apply() from test_metadata.py:TestMetadataImporter.test_measurements, 
-        # pool is not set.  If we're going to rely on the pool getting its edition set here, then 
-        # it's problematic that pool isn't always being set in this method.
         
         link_objects = {}
 
         for link in self.links:
-            if link.rel not in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+            if link.rel in Hyperlink.METADATA_ALLOWED:
                 link_obj, ignore = identifier.add_link(
                     rel=link.rel, href=link.href, data_source=data_source, 
                     license_pool=None, media_type=link.media_type,
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -363,7 +363,153 @@ class FormatData(object):
 
 
 
-class CirculationData(object):
+class MetaToModelUtility(object):
+    """
+    TODO
+    """
+
+    #def mirror_link(self, pool, data_source, link, link_obj, policy):
+    def mirror_link(self, model_object, data_source, link, link_obj, policy):
+        """Retrieve a copy of the given link and make sure it gets
+        mirrored. If it's a full-size image, create a thumbnail and
+        mirror that too.
+
+        The model_object can be either a pool or an edition.
+        """
+
+        if link_obj.rel not in (
+                Hyperlink.IMAGE, Hyperlink.OPEN_ACCESS_DOWNLOAD
+        ):
+            # we only host locally open-source epubs and cover images
+            return
+
+        mirror = policy.mirror
+        http_get = policy.http_get
+
+        _db = Session.object_session(link_obj)
+        original_url = link.href
+        identifier = link_obj.identifier
+
+        self.log.debug("About to mirror %s" % original_url)
+
+        pool = None
+        edition = None
+        title = None
+        identifier = None
+        if model_object:
+            if isinstance(model_object, LicensePool):
+                pool = model_object
+                identifier = model_object.identifier
+
+                if (identifier and identifier.primarily_identifies and identifier.primarily_identifies[0]): 
+                    edition = identifier.primarily_identifies[0]
+            elif isinstance(model_object, Edition):
+                pool = model_object.license_pool
+                identifier = model_object.primary_identifier
+                edition = model_object
+        if edition and edition.title:
+            title = edition.title
+        else:
+            title = self.title or None
+
+        max_age = None
+        if policy.link_content:
+            # We want to fetch the representation again, even if we
+            # already have a recent usable copy. If we fetch it and it
+            # hasn't changed, we'll keep using the one we have.
+            max_age = 0
+
+        # This will fetch a representation of the original and 
+        # store it in the database.
+        representation, is_new = Representation.get(
+            _db, link.href, do_get=http_get,
+            presumed_media_type=link.media_type,
+            max_age=max_age,
+        )
+
+        # Make sure the (potentially newly-fetched) representation is
+        # associated with the resource.
+        link_obj.resource.representation = representation
+
+        # If we couldn't fetch this representation, don't mirror it,
+        # and if this was an open access link, then suppress the associated 
+        # license pool until someone fixes it manually.
+        # The license pool to suppress will be either the passed-in model_object (if it's of type pool), 
+        # or the license pool associated with the passed-in model object (if it's of type edition).
+        if representation.fetch_exception:
+            if pool and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+                pool.suppressed = True
+                pool.license_exception = "Fetch exception: %s" % representation.fetch_exception
+            return
+
+        # If we fetched the representation and it hasn't changed,
+        # the previously mirrored version is fine. Don't mirror it
+        # again.
+        if representation.status_code == 304:
+            return
+
+        # Determine the best URL to use when mirroring this
+        # representation.
+        if title and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+            extension = representation.extension()
+            mirror_url = mirror.book_url(
+                identifier, data_source=data_source, title=title,
+                extension=extension
+            )
+        else:
+            filename = representation.default_filename(link_obj)
+            mirror_url = mirror.cover_image_url(
+                data_source, identifier, filename
+            )
+
+        representation.mirror_url = mirror_url
+        mirror.mirror_one(representation)
+
+        # If we couldn't mirror an open access link representation, suppress
+        # the license pool until someone fixes it manually.
+        if representation.mirror_exception and pool and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+            pool.suppressed = True
+            pool.license_exception = "Mirror exception: %s" % representation.mirror_exception
+
+        # The metadata may have some idea about the media type for this
+        # LinkObject, but the media type we actually just saw takes 
+        # precedence.
+        if representation.media_type:
+            link.media_type = representation.media_type
+
+        if link_obj.rel == Hyperlink.IMAGE:
+            # Create and mirror a thumbnail.
+            thumbnail_filename = representation.default_filename(
+                link_obj, Representation.PNG_MEDIA_TYPE
+            )
+            thumbnail_url = mirror.cover_image_url(
+                data_source, identifier, thumbnail_filename,
+                Edition.MAX_THUMBNAIL_HEIGHT
+            )
+            thumbnail, is_new = representation.scale(
+                max_height=Edition.MAX_THUMBNAIL_HEIGHT,
+                max_width=Edition.MAX_THUMBNAIL_WIDTH,
+                destination_url=thumbnail_url,
+                destination_media_type=Representation.PNG_MEDIA_TYPE,
+                force=True
+            )
+            if is_new:
+                # A thumbnail was created distinct from the original
+                # image. Mirror it as well.
+                mirror.mirror_one(thumbnail)
+
+        if link_obj.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+            # If we mirrored book content successfully, don't keep it in
+            # the database to save space. We do keep images in case we
+            # ever need to resize them.
+            if representation.mirrored_at and not representation.mirror_exception:
+                representation.content = None
+
+
+
+
+
+class CirculationData(MetaToModelUtility):
     """Information about actual copies of a book that can be delivered to
     patrons.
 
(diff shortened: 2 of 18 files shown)


metadata_layer.py
```diff
@@ -351,36 +351,188 @@ def __repr__(self):
...
+            return
+
+        for link in arg_links:
+            if link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+                # only accept the types of links relevant to pools
+                self.__links.append(link)
+
+                # If a link has a rights_uri, make that the overall rights_uri. 
+                # If there are multiple links with a rights_uri, they should have gotten 
+                # split into separate metadata objects in the content server or whatever code called this.
```
leonardr: Is this still true? I'd like to be able to handle one LicensePool with multiple DeliveryMechanisms by creating one Metadata object and a number of CirculationData objects.

metadata_layer.py
```diff
@@ -351,36 +351,188 @@ def __repr__(self):
...
+        for link in arg_links:
+            if link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+                # only accept the types of links relevant to pools
+                self.__links.append(link)
+
+                # If a link has a rights_uri, make that the overall rights_uri. 
+                # If there are multiple links with a rights_uri, they should have gotten 
+                # split into separate metadata objects in the content server or whatever code called this. 
+                if link.rights_uri:
+                    self.set_default_rights_uri(data_source=self.data_source, default_rights_uri=link.rights_uri)
```
leonardr: Since set_default_rights_uri overwrites any previous setting, you should only call it if default_rights_uri is not already set.

metadata_layer.py
```diff
@@ -351,36 +351,188 @@ def __repr__(self):
...
+
+                # An open-access link or open-access rights implies a FormatData object.
+                open_access_link = (link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD
+                                    and link.href)
+                
+                open_access_rights_link = (link.media_type in Representation.BOOK_MEDIA_TYPES
```
leonardr: Is this used?

metadata_layer.py
```diff
@@ -351,36 +351,188 @@ def __repr__(self):
...
+                            link=link
+                        )
+                )
+
+
+    def __repr__(self):
+        description_string = '<CirculationData primary_identifier="%r" licenses_owned="%s" licenses_available="%s" licenses_reserved=%s' % (
+            self.primary_identifier, self.licenses_owned, self.licenses_available, self.licenses_reserved
+        )
+        description_string += 'patrons_in_hold_queue="%s" default_rights_uri="%s" first_appearance="%s" last_checked="%s" >' % (
```
leonardr: This is kind of a mess. I recommend you define the entire template at once:

'<CirculationData primary_identifier="%(primary_identifier)r" licenses_owned="%(licenses_owned)d" licenses_available="%(licesnses_available)d" licenses_reserved=%(licenses_reserved)d patrons_in_hold_queue="%(patrons_in_hold_queue)d" default_rights_uri="%(default_rights_uri)s" first_appearance="%(first_appearance)s" last_checked="%(last_checked)s" links="%(links)r" formats="%(formats)r" data_source="%(data_source_name)s">'

Then you can create a dictionary with the values for each interpolation variable, and then format the whole thing at once:

`return description_template % variables`

metadata_layer.py
```diff
@@ -351,36 +351,188 @@ def __repr__(self):
...
+        if not self.primary_identifier:
+            raise ValueError(
+                "Cannot find license pool: CirculationData has no primary identifier."
+            )
+
+        license_pool = None
+        is_new = False
+
+        identifier_obj, ignore = self.primary_identifier.load(_db)
+        data_source = self.data_source(_db)
```
leonardr: Looking at these two lines right next to each other it occurs to me we might be able to save some ugly code by creating a DataSourceData object which has a load() method. In the constructors we would just ensure that whatever was passed in we end up with a DataSourceData object, and then when we want the actual DataSource we call load(). Just something to think about.

metadata_layer.py
```diff
@@ -437,6 +589,192 @@ def update(self, license_pool, license_pool_is_new):
...
+        if replace is None:
+            replace = ReplacementPolicy()
+
+        _db = Session.object_session(pool)
+        data_source = self.data_source(_db)
+
+        identifier = pool.identifier
+
+        link_objects = {}
+        for link in self.links:
```
leonardr: This makes me realize that we need to be able to handle the case where the URL to a link changes or a link disappears. But we don't deal with that now, so it can wait til later.

metadata_layer.py
```diff
@@ -437,6 +589,192 @@ def update(self, license_pool, license_pool_is_new):
...
+                _db.delete(lpdm)
+            pool.delivery_mechanisms = []
+
+        # TODO: bring back
+        #self.set_default_rights_uri(data_source)
+
+        for link in self.links:
+            if link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+                link_obj = link_objects[link]
+                # TODO: We do not properly handle the (unlikely) case
```
leonardr: This copied TODO is not relevant here.

metadata_layer.py
```diff
@@ -437,6 +589,192 @@ def update(self, license_pool, license_pool_is_new):
...
+            if pool:
+                pool.set_delivery_mechanism(
+                    format.content_type, format.drm_scheme, resource
+                )
+
+        pool.last_checked = datetime.datetime.utcnow()
+        return pool, made_changes
+
+
+    def mirror_link(self, pool, data_source, link, link_obj, policy):
```
leonardr: You should not need two copies of this method -- put it in a superclass shared by CirculationData and Metadata.

metadata_layer.py
```diff
@@ -463,32 +800,24 @@ def __init__(
...
             links=None,
-            formats=None,
-            rights_uri=None,
             last_update_time=None,
-            circulation=None,
+            #circulation=None,
     ):
-        # data_source is where the data comes from.
+        # data_source is where the data comes from (e.g. overdrive, metadata wrangler, admin interface), 
+        # and not where the lending licenses are coming from.
```
leonardr: This is no longer accurate. If there is any CirculationData to be had here, it comes from the same DataSource as the Metadata.

opds_import.py
```diff
@@ -176,17 +180,21 @@ def __repr__(self):
...
+    Should be used when a circulation server asks for data from 
+    our internal content server, and also when our content server asks for data 
+    from external content servers. 
 
-    This importer should be used when a circulation server
-    communicates with a content server. It ignores author and subject
-    information, under the assumption that it will get better author
-    and subject information from the metadata wrangler.
+    Note: Ignores author and subject information, under the assumption that it will 
+    get better author and subject information from the metadata wrangler.
```
leonardr: I think with the advent of the presentation edition stuff, we can relax this restriction. There's no risk that 'bad' data will overwrite 'good' data.

opds_import.py
```diff
@@ -202,42 +210,95 @@ def __init__(self, _db, data_source_name=DataSource.METADATA_WRANGLER,
...
+                    circulationdata
+                )
 
-        return imported, messages, next_links
 
-    def import_from_metadata(
+        return imported_editions.values(), imported_pools.values(), imported_works.values(), status_messages.values(), next_links
+
+
+    def import_editions_from_metadata(
```
leonardr: Why is 'editions' plural?

opds_import.py
```diff
@@ -260,53 +321,135 @@ def import_from_metadata(
             edition, self.metadata_client, replace=policy
         )
 
+        return edition
+
+
+    def import_pools_works_from_circulationdata(
```
leonardr: I think it would be easier to understand (and your error handling would be cleaner) if you split out the code that creates the Work into a separate method.

opds_import.py
```diff
@@ -260,53 +321,135 @@ def import_from_metadata(
...
+        with the new license pool and work.  
+        """
+
+        # Locate or create a LicensePool for this book.
+        license_pool, is_new_license_pool = circulationdata.license_pool(self._db)
+        work = None
+
+        if (not (cutoff_date 
+            and not is_new_license_pool 
+            and license_pool.last_checked < cutoff_date)):
```
leonardr: I think this is wrong. The right thing to check here is `metadata.last_update_time`. We're trying to avoid doing work if the OPDS entry hasn't changed since the last time we checked. In this case `cutoff_date` stands in for "the last time we checked". You could argue that `license_pool.last_checked` is a better proxy for 'the last time we checked', but right now you're comparing two values that both basically mean 'the last time we checked'. You want to compare 'the last time we checked' to 'the last time the OPDS entry changed server-side'.

This means adding a last_update_time to CirculationData that is the same value you give to Metadata. It might also make sense to not use `cutoff_date` here, and instead compare `license_pool.last_checked` to `last_update_time`.

opds_import.py
```diff
@@ -260,53 +321,135 @@ def import_from_metadata(
...
+            # We've already imported this book, we've been told
+            # not to bother with books that appeared before a
+            # certain date, and this book did in fact appear
+            # before that date. There's no reason to do anything.
```
leonardr: This copied comment is no longer relevant.

opds_import.py
```diff
@@ -321,26 +464,39 @@ def combine(self, d1, d2):
...
                     status_messages[identifier] = status_message
+                else:
+                    if detail_meta:
+                        values_meta[identifier] = detail_meta
+                    if detail_circ:
+                        values_circ[identifier] = detail_circ
+            else:
+                # That's bad. Can't make an item-specific error message, but write to 
+                # log that something very wrong happened.
+                logging.error("Tried to parse an element without a valid identifier.")
```
leonardr: I don't think this is bad enough to warrant an `.error`. It looks like this happens if someone serves an OPDS entry that lacks an <id> tag, which is annoying but not fatal to us. I think warn() is sufficient, and it would also be good to give a string representation of the entry.

metadata_layer.py

metadata_layer.py
daryanypl: it is now

metadata_layer.py

metadata_layer.py
daryanypl: added a todo

metadata_layer.py
daryanypl: I actually uncommented the set_default_rights_uri call, so it gets reset before the links start getting calculated.

coverage.py
```diff
@@ -341,11 +341,45 @@ def work(self, identifier):
             return CoverageFailure(self, identifier, e, transient=True)
         return work
 
-    def set_metadata(self, identifier, metadata, 
+
+    def set_meta_circ_data(self, identifier, metadata, circulationdata,
```
aslagle: how about set_metadata_and_circulation_data()?

coverage.py
```diff
@@ -341,11 +341,45 @@ def work(self, identifier):
...
         return work
 
-    def set_metadata(self, identifier, metadata, 
+
+    def set_meta_circ_data(self, identifier, metadata, circulationdata, 
+        metadata_replacement_policy=None, 
+        circulationdata_replacement_policy=None, 
+    ):
+        """
+        Performs the function of the old set_metadata.  Finds or creates the Edition
```
aslagle: don't talk about the way code used to be - take out this first sentence

metadata_layer.py
```diff
@@ -352,93 +352,492 @@ def __repr__(self):
...
+        ):
+            # we only host locally open-source epubs and cover images
+            return
+
+        mirror = policy.mirror
+        http_get = policy.http_get
+
+        _db = Session.object_session(link_obj)
+        original_url = link.href
+        identifier = link_obj.identifier
```
aslagle: We set identifier here and then immediately unset it on line 399. Either we need a sanity check to make sure the link's identifier is the same as the identifier of the model object, or it shouldn't be set here in the first place.

metadata_layer.py
```diff
@@ -352,93 +352,492 @@ def __repr__(self):
...
         if link and not isinstance(link, LinkData):
             raise TypeError(
                 "Expected LinkData object, got %s" % type(link)
             )
         self.link = link
+        self.rights_uri = rights_uri
+
+
+
+class MetaToModelUtility(object):
```
aslagle: how about MirrorMixin?

metadata_layer.py
```diff
@@ -352,93 +352,492 @@ def __repr__(self):
...
-    def __init__(self, content_type, drm_scheme, link=None):
+    def __init__(self, content_type, drm_scheme, link=None, rights_uri=None):
         self.content_type = content_type
         self.drm_scheme = drm_scheme
         if link and not isinstance(link, LinkData):
             raise TypeError(
                 "Expected LinkData object, got %s" % type(link)
             )
         self.link = link
+        self.rights_uri = rights_uri
```
aslagle: if no rights_uri is passed in, try to get it from link.rights_uri

metadata_layer.py
```diff
@@ -352,93 +352,492 @@ def __repr__(self):
...
+    
+
+    def data_source(self, _db):
+        if not self.data_source_obj:
+            if self._data_source:
+                obj = DataSource.lookup(_db, self._data_source)
+                if not obj:
+                    raise ValueError("Data source %s not found!" % self._data_source)
+                if not obj.offers_licenses:
+                    raise ValueError("Data source %s does not offer licenses and cannot be used as a LicenseData data source." % self._data_source)
```
aslagle: this message should say CirculationData not LicenseData

metadata_layer.py
```diff
@@ -352,93 +352,492 @@ def __repr__(self):
...
+            rights_status = get_one(
+                _db, RightsStatus, uri=self.default_rights_uri
             )
-            # TODO: Also put this in the log.
 
-        changed = (license_pool.licenses_owned != self.licenses_owned or
-                   license_pool.licenses_available != self.licenses_available or
-                   license_pool.patrons_in_hold_queue != self.patrons_in_hold_queue or
-                   license_pool.licenses_reserved != self.licenses_reserved)
+            last_checked = datetime.datetime.utcnow()
```
aslagle: we need to support taking last_checked as an argument to circulation data for the 3m case

metadata_layer.py
```diff
@@ -352,93 +352,492 @@ def __repr__(self):
...
+    def has_open_access_link(self):
+        """Does this Circulation object have an associated open-access link?"""
+        return any(
+            [x for x in self.links 
+             if x.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD and x.href]
+        )
+
+
+    def set_default_rights_uri(self, data_source_name, default_rights_uri=None):
+        """ TODO: should there be a way to delete default_rights_uri?
```
aslagle: No - if you delete it it should go back to the one for the data source

metadata_layer.py
```diff
@@ -352,93 +352,492 @@ def __repr__(self):
...
+                    # open access download link, so we'll consider it generic open access.
+                    self.default_rights_uri = RightsStatus.GENERIC_OPEN_ACCESS
+
+        if self.default_rights_uri == None:
+            # We still haven't determined rights, so it's unknown.
+            self.default_rights_uri = RightsStatus.UNKNOWN
+
+
+    def apply(self, pool, replace=None):
+        """  TODO: apply name provisional, doc string follows.
```
aslagle: apply is good since it's the same as metadata

metadata_layer.py
```diff
@@ -352,93 +352,492 @@ def __repr__(self):
...
+        # now that pool uses a composite edition, we must create that edition before 
+        # calculating any links
+        # TODO:  if we do call pool.set_presentation_edition from here, watch out for circular logic.
+
+        self.set_default_rights_uri(data_source)
+        # TODO: be able to handle the case where the URL to a link changes or a link disappears.
+        link_objects = {}
+
+        for link in self.links:
+            if link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
```
aslagle: if the self.links setter is already filtering them, we shouldn't need this check

metadata_layer.py
```diff
@@ -471,32 +869,25 @@ def __init__(
             contributors=None,
             measurements=None,
             links=None,
-            formats=None,
-            rights_uri=None,
-            last_update_time=None,
-            circulation=None,
+            provider_entry_updated=None,
```
aslagle: leonard and i vote for data_source_last_updated

metadata_layer.py
```diff
@@ -513,37 +904,38 @@ def __init__(
...
-            if open_access_link or open_access_rights_link:
-                self.formats.append(
-                    FormatData(
-                        content_type=link.media_type,
-                        drm_scheme=DeliveryMechanism.NO_DRM,
-                        link=link
-                    )
-            )
+        # circulation-metadata connection has been removed, but might come back
+        #self.circulation = circulation
```
aslagle: i thought this did come back?

metadata_layer.py
```diff
@@ -513,37 +904,38 @@ def __init__(
...
+
+        if not arg_links:
+            return
+
+        for link in arg_links:
+            if link.rel not in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
+                # only accept the types of links relevant to editions
+                self.__links.append(link)
+                
+    # TODO: perhaps should move to CirculationData, as Metadata will no longer ever have open-access links.
```
aslagle: yes, this should be removed from metadata now

metadata_layer.py
```diff
@@ -709,14 +1088,16 @@ def edition(self, _db, create_if_not_exists=True):
                 "Cannot find edition: metadata has no primary identifier."
             )
 
-        data_source = self.license_data_source(_db) or self.data_source(_db)
+        #data_source = self.license_data_source(_db) or self.data_source(_db)
```
aslagle: take out this comment

metadata_layer.py
```diff
@@ -709,14 +1088,16 @@ def edition(self, _db, create_if_not_exists=True):
...
+        data_source = self.data_source(_db)
 
         return Edition.for_foreign_id(
             _db, data_source, self.primary_identifier.type,
             self.primary_identifier.identifier,
             create_if_not_exists=create_if_not_exists
         )
 
+    # TODO: move method to CirculationData (compare first, see if something changed in master version)
```
aslagle: make sure this is done and then the method is removed from metadata

metadata_layer.py
```diff
@@ -864,6 +1183,16 @@ def _run_query(self, qu, potentials, confidence):
...
+    '''
+    def apply(self, edition, metadata_client=None, replace=None,
+              replace_identifiers=False,
+              replace_subjects=False,
+              replace_contributions=False,
+              replace_links=False,
+              replace_formats=False,
+              replace_rights=False,
+              force=False,
+    '''
```
aslagle: take out this comment

metadata_layer.py
```diff
@@ -1012,34 +1342,23 @@ def _key(classification):
...
-        # TODO:  remove below comment.
-        # now that pool uses a composite edition, we must create that edition before 
-        # calculating any links
-        # TODO:  we're calling pool.set_presentation_edition from a bunch of places, 
-        # and it's possible there's a better way.
+
         # TODO:  also, when we call apply() from test_metadata.py:TestMetadataImporter.test_open_access_content_mirrored, 
         # pool gets set, but when we call apply() from test_metadata.py:TestMetadataImporter.test_measurements, 
         # pool is not set.  If we're going to rely on the pool getting its edition set here, then 
         # it's problematic that pool isn't always being set in this method.
```
aslagle: pool should never get set here anymore, let's take out this comment too

metadata_layer.py
```diff
@@ -1012,34 +1342,23 @@ def _key(classification):
...
+        
         link_objects = {}
 
         for link in self.links:
-            link_obj, ignore = identifier.add_link(
-                rel=link.rel, href=link.href, data_source=link_data_source,
-                license_pool=pool, media_type=link.media_type,
-                content=link.content
-            )
+            if link.rel not in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
```
aslagle: these are already filtered out of self.links

metadata_layer.py
```diff
@@ -1069,6 +1388,7 @@ def _key(classification):
 
         # obtains a presentation_edition for the title, which will later be used to get a mirror link.
         for link in self.links:
+            #if link.rel not in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.DRM_ENCRYPTED_DOWNLOAD]:
```
aslagle: take out this comment

metadata_layer.py
```diff
@@ -1244,9 +1430,9 @@ def make_thumbnail(self, pool, data_source, link, link_obj):
         # The thumbnail and image are different. Make sure there's a
         # separate link to the thumbnail.
         thumbnail_obj, ignore = link_obj.identifier.add_link(
-            rel=thumbnail.rel, href=thumbnail.href,
-            data_source=data_source,
-            license_pool=pool, media_type=thumbnail.media_type,
+            rel=thumbnail.rel, href=thumbnail.href, 
+            data_source=data_source, 
+            license_pool=None, media_type=thumbnail.media_type,
```
aslagle: we need to make sure that once the pool is set, all the links associated with the identifier get the pool set

opds_import.py
```diff
@@ -231,48 +240,111 @@ def __init__(self, _db, data_source_name=DataSource.METADATA_WRANGLER,
...
+        imported_pools = {}
+        imported_works = {}
+        # status_messages notes business logic errors and non-success download statuses
+        status_messages = {}
+        next_links = []
+        try:
+            metadata_objs, circulation_objs, status_messages, next_links = self.extract_feed_data(feed)
+        except Exception, e:
+            message = StatusMessage(500, "Local exception during import:\n%s" % traceback.format_exc())
+            return imported_editions.values(), imported_pools.values(), imported_works.values(), status_messages, next_links
```
aslagle: there are some problems with this, but I'll take care of it in the next branch since I'll need to change a lot of this anyway.

aslagle: basically this function should never return until the end, it should only add to status_messages

opds_import.py
```diff
@@ -288,53 +360,148 @@ def import_from_metadata(
...
+            contributions=True,
+            rights=True,
+            link_content=True,
+            even_if_not_apparently_updated=True,
+            mirror=self.mirror,
+            http_get=self.http_get,
+        )
+        circulationdata.apply(
+            license_pool, replace=policy
+        )
```
aslagle: this should also be inside the if license_pool is None below

opds_import.py
```diff
@@ -288,53 +360,148 @@ def import_from_metadata(
...
-            if work:
-                if immediately_presentation_ready:
-                    # We want this book to be presentation-ready
-                    # immediately upon import. As long as no crucial
-                    # information is missing (like language or title),
-                    # this will do it.
-                    work.set_presentation_ready_based_on_content()
+            error_message = "No LicensePool created for CirculationData %r, will not attempt to continue to create Work." % circulationdata
+            self.log.warn(error_message)
+            raise DBImportException(error_message)
```
aslagle: not sure if we need a specific exception if it's not caught anywhere

opds_import.py
```diff
@@ -288,53 +360,148 @@ def import_from_metadata(
...
+    ):
+        """ Making work is not based on pool's cutoff logic, but based on whether have a pool and and edition, period.
+        Don't mind if they're new or not.
+        """
+
+        # Note: pool.calculate_work will call self.set_presentation_edition(), 
+        # which will find editions attached to same Identifier.
+        # If the pool has no edition, pool.calculate_work is responsible for not allowing the work to be created.
+        work = None
+        work, is_new_work = license_pool.calculate_work(even_if_no_author=even_if_no_author,)
```
aslagle: take out the trailing comma just to be safe

opds_import.py
```diff
@@ -288,53 +360,148 @@ def import_from_metadata(
...
+        # if pool.calculate_work made a new work, it already called calculate_presentation
```
aslagle: license_pool.calculate_work always calls work.calculate_presentation, even if the work isn't new

opds_import.py
```diff
@@ -288,53 +360,148 @@ def import_from_metadata(
...
+            raise DBImportException(error_message)
+
+        return license_pool
 
-        return edition
 
-    def extract_metadata(self, feed):
-        """Turn an OPDS feed into a list of Metadata objects and a list of
-        messages.
+    def import_work_from_circulationdata(
```
aslagle: this doesn't take circulation data as an argument, it's more of a cleanup method after the license pool is imported.

opds_import.py
```diff
@@ -288,53 +360,148 @@ def import_from_metadata(
...
+        if (work and not is_new_work):
+            work.calculate_presentation()
+        if (work):
+            if immediately_presentation_ready:
+                # We want this book to be presentation-ready
+                # immediately upon import. As long as no crucial
+                # information is missing (like language or title),
+                # this will do it.
+                work.set_presentation_ready_based_on_content()
+        return license_pool, work
```
aslagle: it doesn't need to return the license_pool since it didn't change it at all

opds_import.py
```diff
@@ -288,53 +360,148 @@ def import_from_metadata(
...
+
+
+    def extract_feed_data(self, feed):
+        """Turn an OPDS feed into lists of Metadata and CirculationData objects, 
+        with associated messages and next_links.
```
aslagle: This should be changed to create the CirculationData first, put it inside the Metadata, and return one list like it used to.

opds_import.py
```diff
@@ -288,53 +360,148 @@ def import_from_metadata(
...
+            if combined_circ.get('data_source') is None:
+                combined_circ['data_source'] = self.data_source_name
+            
+            combined_circ['primary_identifier'] = identifier_obj
+
+            circulationdata[internal_identifier.urn] = CirculationData(**combined_circ)
+            if id in fp_data_circ:
+                del fp_data_circ[id]
+
+        for id, fp_data_dict in fp_data_circ.items():
```
aslagle: if you put the circulation data in the metadata, you shouldn't need to maintain fp_data_circ and you can take out this part

overdrive.py
```diff
@@ -530,6 +530,9 @@ def parse_roles(cls, id, rolestring):
 
     @classmethod
     def book_info_to_circulation(cls, book):
+        """ Note:  The json data passed into this method is from a different file 
+        from the json data that goes into the book_info_to_circulation() method.
```
aslagle: this comment is talking about a different method: book_info_to_metadata or something like that

aslagle: also, it's only a file in tests - it comes from a different source

tests/test_circulation_data.py
```diff
@@ -0,0 +1,265 @@
...
+from . import (
+    DatabaseTest,
+    DummyHTTPClient,
+)
+
+from s3 import DummyS3Uploader
+
+class TestCirculationData(DatabaseTest):
+
+    def test_metadata_can_be_deepcopied(self):
```
aslagle: this is testing circulation data

tests/test_circulation_data.py
```diff
@@ -0,0 +1,265 @@
...
+        filtered_links = sorted(circulation_data.links, key=lambda x:x.rel)
+
+        eq_([link1], filtered_links)
+
+
+    def test_open_access_content_mirrored(self):
+        # Make sure that open access material links are translated to our S3 buckets, and that 
+        # commercial material links are left as is.
+        # Note: Mirroring tests passing does not guarantee that all code now 
+        # correctly calls on CirculationData, as well as Metadata.  This is a risk.
```
aslagle: would this be addressed if metadata.apply also called circulationdata.apply?

tests/test_circulation_data.py
```diff
@@ -0,0 +1,265 @@
...
+        circulation_data = CirculationData(
+        	data_source=edition.data_source, 
+        	primary_identifier=edition.primary_identifier,
+        	links=[link_mirrored, link_unmirrored],
+        )
+        circulation_data.apply(pool, replace=policy)
+        
+
+        # Only the open-access link has been 'mirrored'.
+        # TODO: make sure the refactor is done right, and metadata does not upload
```
aslagle: this comment needs to be addressed now

tests/test_circulation_data.py
```diff
@@ -0,0 +1,265 @@
...
+        #eq_(0, len(mirror.uploaded))
+        [book] = mirror.uploaded
+
+        # TODO: make sure the refactor is done right, and circulation does upload
```
aslagle: this comment needs to be addressed now

aslagle: these should also be separate tests

aslagle: actually we don't need this, it's already handled by this test, but the first one still needs to be addressed

tests/test_circulation_data.py
```diff
@@ -0,0 +1,265 @@
...
+        unmirrored_representation, mirrored_representation = [edlink.resource.representation for edlink in sorted_edition_links]
+        assert mirrored_representation.mirror_url.startswith('http://s3.amazonaws.com/test.content.bucket/')
+
+        # make sure the unmirrored link is safely on edition
+        eq_('http://example.com/2', unmirrored_representation.url)
+        # make sure the unmirrored link has not been translated to an S3 URL
+        eq_(None, unmirrored_representation.mirror_url)
+
+
+    def test_mirror_open_access_link_fetch_failure(self):
```
aslagle: there should be a separate class that tests the mirror mixin class - it can still use circulation data as the example class

tests/test_circulation_data.py
```diff
@@ -0,0 +1,265 @@
...
+)
+
+from . import (
+    DatabaseTest,
+    DummyHTTPClient,
+)
+
+from s3 import DummyS3Uploader
+
+class TestCirculationData(DatabaseTest):
```
aslagle: it would be more consistent with the rest of the code base to put this class in test_metadata.py, since CirculationData is in metadata.py

tests/test_metadata.py
```diff
@@ -321,9 +270,11 @@ def test_mirror_open_access_link_fetch_failure(self):
         assert representation.fetched_at != None
         eq_(None, representation.mirrored_at)
 
-        # The license pool is suppressed when fetch fails.
-        eq_(True, pool.suppressed)
-        assert representation.fetch_exception in pool.license_exception
+        # TODO:  should the edition's identifier-associated license pool be 
+        # suppressed when fetch fails on getting image?
```
aslagle: No

tests/test_metadata.py
```diff
@@ -366,14 +319,19 @@ def test_mirror_open_access_link_mirror_failure(self):
 
         # The mirror url should still be set.
         assert "Gutenberg" in representation.mirror_url
-        assert representation.mirror_url.endswith("%s.epub" % edition.title)
+        # TODO:  changed from edition.title to edition.primary_identifier.identifier.  make sure 
+        # the change is warranted.
```
aslagle: this is right, we weren't using titles for mirrored images.

tests/test_metadata.py
```diff
@@ -366,14 +319,19 @@ def test_mirror_open_access_link_mirror_failure(self):
...
+        assert representation.mirror_url.endswith("%s/cover.jpg" % edition.primary_identifier.identifier)
 
         # Book content is still there since it wasn't mirrored.
         assert representation.content != None
 
         # The license pool is suppressed when mirroring fails.
-        eq_(True, pool.suppressed)
-        assert representation.mirror_exception in pool.license_exception
+        # TODO:  should the edition's identifier-associated license pool be 
+        # suppressed when link fails on getting image?
```
aslagle: No

tests/test_metadata.py
```diff
@@ -398,9 +356,10 @@ def test_explicit_formatdata(self):
             drm_scheme=DeliveryMechanism.ADOBE_DRM,
         )
 
-        metadata = Metadata(formats=[drm_format],
-                            data_source=edition.data_source)
-        metadata.apply(edition)
+        circulation_data = CirculationData(formats=[drm_format],
+                            data_source=edition.data_source, 
+                            primary_identifier=edition.primary_identifier)
+        circulation_data.apply(pool)
```
aslagle: this test (and others that primarily use CirculationData) should be moved to the TestCirculationData class

tests/test_opds_import.py
```diff
@@ -232,21 +240,23 @@ def test_extract_metadata_from_elementtree(self):
...
-        # book because the metadat wrangler doesn't know who actually
-        # provides copies of this book.
+        # metadata wrangler. No Work has been created.
         eq_(DataSource.METADATA_WRANGLER, crow.data_source.name)
         eq_(None, crow.work)
+        eq_(None, crow.license_pool)
         eq_(Edition.BOOK_MEDIUM, crow.medium)
 
         # But the 'mouse' book is known to come from Project Gutenberg,
         # so a Work has been created for that book.
```
aslagle: this comment is wrong now

tests/test_opds_import.py
```diff
@@ -286,59 +296,135 @@ def test_import(self):
...
+        )
+
+        [crow, mouse] = sorted(imported_editions, key=lambda x: x.title)
+
+        # Work was created for both books.
+        assert crow.license_pool.work is not None
+        eq_(Edition.BOOK_MEDIUM, crow.medium)
+
+        # But the 'mouse' book is known to come from Project Gutenberg,
+        # so a Work has been created for that book.
```
aslagle: this comment is still inaccurate

tests/test_opds_import.py
```diff
@@ -286,59 +296,135 @@ def test_import(self):
...
+            OPDSImporter(self._db, data_source_name=DataSource.OA_CONTENT_SERVER).import_from_feed(feed)
+        )
+
+        [crow, mouse] = sorted(imported_editions, key=lambda x: x.title)
+
+        # Work was created for both books.
+        assert crow.license_pool.work is not None
+        eq_(Edition.BOOK_MEDIUM, crow.medium)
+
+        # But the 'mouse' book is known to come from Project Gutenberg,
```
aslagle: that's not why the work was created - we no longer know where it came from

tests/test_opds_import.py
```diff
@@ -286,59 +296,135 @@ def test_import(self):
...
-        # If we import the same file again, we get the same list of Editions.
-        imported2, messages, next_links = OPDSImporter(self._db).import_from_feed(feed)
-        eq_(imported2, imported)
+
+
+    def test_import_with_wrangler_data_source(self):
+        # Tests that will create Edition, LicensePool, and Work objects, when appropriate.
+        # For example, on a Metadata_Wrangler data source, it is only appropriate to create 
+        # editions, but not pools or works.  On a lendable data source, should create 
+        # pools and works as well as editions.
```
aslagle: this seems to duplicate some of the previous test - the cutoff is the only new thing that's being tested here

aslagle: everything below 357

aslagle: and the cutoff part os duplicative of test_import_with_cutoff below

tests/test_opds_import.py
```diff
@@ -286,59 +296,135 @@ def test_import(self):
...
             importer.import_from_feed(feed, cutoff_date=cutoff)
         )
 
         # Both books were imported again.
-        eq_(2, len(imported))
+        eq_(2, len(imported_editions))
+        eq_(2, len(imported_pools))
+        eq_(2, len(imported_works))
+
+        # TODO: last_checked isn't getting set.  correct behavior or should fix?
```
aslagle: last_checked should be getting set - this needs to be investigated

tests/test_opds_import.py
```diff
@@ -416,50 +510,58 @@ def test_import_and_make_presentation_ready(self):
...
                 else:
                     # Any other import fails.
                     raise Exception("Utter failure!")
         path = os.path.join(self.resource_path, "content_server_mini.opds")
         feed = open(path).read()
-        imported, messages, next_links = DoomedOPDSImporter(self._db).import_from_feed(feed)
-        
-        # One book was imported successfully.
-        [success] = imported
-        eq_("Johnny Crow's Party", success.title)
```
aslagle: this should still succeed, there's probably an exception in import_from_metadata that needs to be looked at.

tests/test_opds_import.py
```diff
@@ -550,25 +659,21 @@ def test_resources_are_mirrored_on_import(self):
...
-            s3.uploaded
-        )
+        imported_representations = [e2_image_link.resource.representation,e2_image_link.resource.representation.thumbnails[0],e2_oa_link.resource.representation,e1_oa_link.resource.representation,]
+        eq_(imported_representations, s3.uploaded)
+
+
+        eq_(4, len(s3.uploaded))
 
         # Each resource was 'mirrored' to an Amazon S3 bucket.
         # The first resource has no bibframe provider in OPDS so it uses the importer's data source.
```
aslagle: this comment is no longer relevant

tests/test_overdrive.py
```diff
@@ -215,14 +229,6 @@ def test_book_info_with_metadata(self):
             sorted(ids)
         )
 
-        # Available formats.
-        [kindle, pdf] = sorted(metadata.formats, key=lambda x: x.content_type)
-        eq_(DeliveryMechanism.KINDLE_CONTENT_TYPE, kindle.content_type)
-        eq_(DeliveryMechanism.KINDLE_DRM, kindle.drm_scheme)
-
-        eq_(Representation.PDF_MEDIA_TYPE, pdf.content_type)
-        eq_(DeliveryMechanism.ADOBE_DRM, pdf.drm_scheme)
```
aslagle: this should go into test_book_info_with_circulationdata

threem.py
```diff
@@ -286,11 +287,14 @@ def parse_genre_string(self, s):
             genres.append(SubjectData(Subject.THREEM, i, weight=15))
         return genres
 
+
     def process_one(self, tag, namespaces):
-        """Turn an <item> tag into a Metadata object."""
+        """Turn an <item> tag into a Metadata and a CirculationData objects, 
+        and return them as a tuple."""
```
aslagle: this comment is inaccurate

coverage.py
daryanypl: Sure!  I was trying to curb my long-method-name enthusiasm for the team's sake :).

coverage.py
daryanypl: This is on purpose.  I'm concerned there may be code not covered by the unit tests that would break.  In that case, it would be helpful to remember what the broken code was expecting.  I plan to remove these comments in a couple of months, once I know it's safe.

metadata_layer.py
daryanypl: I asked Winnie if she thought "DataHorse" would be an apt name, because it's like a workhorse that shuttles data, right?  She didn't think it descriptive :).
I ended up going with "utility" because I'm expecting to put more common code in there in the future.

metadata_layer.py
daryanypl: Leonard, is this yours?

leonardr: It's currently not used at all. I was planning for what has become the next branch after this one, in which RightsStatus adheres to the DeliveryMechanism, not to the LicensePool. (or not _just_ to the LicensePool)

metadata_layer.py

metadata_layer.py

metadata_layer.py
leonardr: We decided to remove the TODO.

metadata_layer.py
daryanypl: some changes made

metadata_layer.py

metadata_layer.py

metadata_layer.py
daryanypl: Theoretically, no, but python doesn't completely prevent accidental setting of class properties.  Better be triple-safe.

metadata_layer.py
daryanypl: but data source can be something that's local to us, like the MetadataWrangler.  doesn't this property refer only to the last time we went over and checked the provider's remote server's idea of what the metadata should be?

metadata_layer.py
daryanypl: it did.  which makes me wonder why the unit tests passed with that line commented out in the init.

metadata_layer.py
aslagle: in this case we would consider the metadata wrangler a remote data source as well. we manage it, but it's remote to the circulation manager. it would be from the last time we checked the metadata wrangler, not the last time the metadata wrangler checked its remote sources.

metadata_layer.py
aslagle: well if you want to use the list of link types in a bunch of places, it should at least be in a constant

metadata_layer.py

metadata_layer.py

metadata_layer.py

metadata_layer.py

metadata_layer.py
daryanypl: shall we take a look at the unit tests and see what needs to change?

metadata_layer.py
daryanypl: agreed, but python safety...

metadata_layer.py

metadata_layer.py
daryanypl: I thought in future branch we'll have more than one pool per identifier.  how would we handle that?

aslagle: In the future branch, the link won't have a pool anymore. But until that happens, we should make sure it has the right pool in case something's using it.

opds_import.py
daryanypl: done

opds_import.py
daryanypl: done

opds_import.py
daryanypl: It might help filter logs.

opds_import.py
daryanypl: renamed to "MagicMirror" :)

opds_import.py
daryanypl: true

opds_import.py
daryanypl: it might have.  the pool gets updated with the new work.  having said that, the pool is passed in, so doesn't need to be returned.  having said _that_, I like returning it, to make it a bit more clear that stuff possibly happened to the pool.

opds_import.py
daryanypl: I've been trying to keep circ and meta separate, as much as possible, to help with any future refactoring.  Tell me why you'd like to see this change?

opds_import.py
daryanypl: see response above.

opds_import.py
daryanypl: adding feed to log, as discussed

overdrive.py
daryanypl: done

tests/test_circulation_data.py
daryanypl: You're right.  But that file is already quite long.  Maybe we should start breaking those up into thematic blocks?

tests/test_circulation_data.py
daryanypl: done

tests/test_circulation_data.py
daryanypl: we don't require metadata to have a circulation, so it wouldn't be a complete guarantee.

tests/test_circulation_data.py
daryanypl: done

tests/test_circulation_data.py
daryanypl: done

tests/test_circulation_data.py
daryanypl: cool

opds_import.py
leonardr: Right now you have some code that treats CirculationData as a property of Metadata, and some code (such as here) that treats Metadata and CirculationData as totally separate things that need to be kept in separate lists. It's good that CirculationData can be used separately, but given that `Metadata.circulation` is allowed, turning one OPDS feed into a list of Metadata (each of which may or may not have a `.circulation`) yields one list, which is a lot easier to deal with than two.

If you see future refactoring that gets rid of `Metadata.circulation` then it would be appropriate to leave this alone, since we're partway there, but like having `Metadata.circulation` and I think you do as well. What kind of refactoring are you thinking of?

tests/test_circulation_data.py
leonardr: I think it makes sense for Metadata.apply to call self.circulation.apply() if that object is present.

tests/test_circulation_data.py
aslagle: What would you think of moving CirculationData to its own file too, then? I also prefer smaller files but I like consistency so I know where to look for things.

tests/test_metadata.py
daryanypl: 'k

tests/test_metadata.py
daryanypl: ok

tests/test_metadata.py
daryanypl: done

tests/test_metadata.py
daryanypl: sure

tests/test_opds_import.py
daryanypl: done

tests/test_opds_import.py
daryanypl: done

tests/test_opds_import.py
daryanypl: changed test name.  the goal of the test is to make sure that lendable and non-lendable data_sources are recognized.

tests/test_opds_import.py
daryanypl: fixed now

tests/test_opds_import.py
daryanypl: let's talk

tests/test_opds_import.py
daryanypl: ok

tests/test_overdrive.py
daryanypl: couldn't put it there, because overdrive_availability_information.json doesn't contain formats, but returned the code to the metadata test, now that have meta.circ back.

threem.py
daryanypl: ok, let's talk

daryanypl: done

tests/test_circulation_data.py
daryanypl: ok, added code.

metadata_layer.py

metadata_layer.py
daryanypl: changed to "tiny_dancer" :)

metadata_layer.py

metadata_layer.py

tests/test_opds_import.py
daryanypl: done

tests/test_opds_import.py
```diff
@@ -416,50 +507,59 @@ def test_import_and_make_presentation_ready(self):
...
-        
-        # One book was imported successfully.
-        [success] = imported
-        eq_("Johnny Crow's Party", success.title)
+
+        imported_editions, imported_pools, imported_works, error_messages, next_links = (
+            DoomedOPDSImporter(self._db).import_from_feed(feed)
+        )
+
+        # No books were imported.
```
aslagle: this comment needs to be fixed

daryanypl: ok

coverage.py
```diff
@@ -426,7 +497,11 @@ def __init__(self, _db, api, datasource, workset_size=10,
         metadata_replacement_policy = (
             metadata_replacement_policy or ReplacementPolicy.from_metadata_source()
         )
+        circulationdata_replacement_policy = (
```
leonardr: I see what's happening here and I'm okay with it for now, but I designed ReplacementPolicy to cover both the metadata and the circulation data. If we're passing in two objects, then they should be different objects -- a CirculationDataReplacementPolicy which covers links, rights, and formats, and a MetadataReplacementPolicy which covers links, identifiers, subjects, etc.

One change I do think needs to be made this branch is the default value for circulationdata_replacement_policy should be the same as metadata_replacement_policy, not the (potentially more permissive) ReplacementPolicy.from_license_source()

metadata_layer.py
```diff
@@ -1073,6 +1359,18 @@ def _key(classification):
                 edition.sort_author = primary_author.sort_name
                 edition.display_author = primary_author.display_name
 
+        # we updated the links.  but does the associated pool know?
+        pool = None
+        if self.circulation:
+            pool, is_new = self.circulation.license_pool(_db)
+            if (pool and not is_new):
```
leonardr: Why skip this step if the pool is new?

leonardr:
Amy and I have decided to merge this branch despite my quibbles, so that she can proceed with her work on the content server. Darya, I'd like you to address my quibbles (assuming they need addressing) in a separate branch.

PR #249, opened 2016-05-17
Add series details to feed

### LINKED ISSUE #248: Add series details to OPDS feed

courte:
This branch adds the series name and a work's position in the volumes of a series to the OPDS feed. It also removes some bits of code that weren't actually being used.

Fixes #248.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -78,11 +78,7 @@ Format a date the way Atom likes it (RFC3339?)
     return d.strftime(AtomFeed.TIME_FORMAT)
 
 default_typemap = {datetime: lambda e, v: _strftime(v)}
-
 E = builder.ElementMaker(typemap=default_typemap, nsmap=nsmap)
-SCHEMA = builder.ElementMaker(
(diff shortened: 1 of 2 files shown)


opds.py
```diff
@@ -222,6 +218,18 @@ def authors(cls, work, license_pool, edition, identifier):
...
 
     @classmethod
+    def series(cls, series_name, series_position):
+        """Generate a schema:Series tag for the given name and position."""
+        if not series_name:
+            return None
+        series_details = dict()
+        series_details['name'] = series_name
+        if series_position:
+            series_details['{%s}position' % schema_ns] = unicode(series_position)
```
aslagle: what's the reason for converting to unicode? it seems like it would be more useful as an integer.

courte: The element maker expects a string. Trying to pass an integer raises an error. 😞

aslagle: oh well :(

aslagle:
:+1:

PR #250, opened 2016-05-18
Dont store datasource in coverageprovider

leonardr:
This branch stops `CoverageProvider` from storing the `DataSource` object it's provided with. Instead, it stores `DataSource.name` and then looks up the `DataSource` object whenever necessary. This avoids problems where we're in a multithreaded environment with a scoped session and the `DataSource` is accessed outside of the session that created it. This is mainly a hypothetical worry at this point but I believe it does actually happen in the admin interface for the circulation manager.

This branch also allows you to pass in a custom OverdriveAPI (e.g. a DummyOverdriveAPI for use in testing) into the OverdriveBibliographicCoverageProvider constructor. Similarly for the ThreeMBibliographicCoverageProvider.

Both of these changes are used by https://github.com/NYPL-Simplified/circulation/pull/192

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -63,10 +63,22 @@ class CoverageProvider(object):
         if not isinstance(input_identifier_types, list):
             input_identifier_types = [input_identifier_types]
         self.input_identifier_types = input_identifier_types
-        self.output_source = output_source
+        self.output_source_name = output_source.name
         self.workset_size = workset_size
         self.cutoff_time = cutoff_time
         self.operation = operation
+
+    @property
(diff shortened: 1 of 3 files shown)


aslagle:
We're not storing the data source, but we're still storing self._db?

leonardr:
Yeah, self._db is fine. In a scoped environment _db isn't a specific database connection--it's the scoped session itself, and it transparently dispatches what you want to do to your thread's connection.

aslagle:
:+1:

PR #251, opened 2016-05-18
Removed Edition.is_primary_for_work. (Fixes #242)

### LINKED ISSUE #242: Get rid of Edition.is_primary_for_work
We should be able to have a simpler setup where Edition.work_id is called something like Edition.presentation_for_work_id. This field would be null for all editions except the ones that are the primary edition for some work. I don't think we gain anything by keeping work_id in sync for all the Editions that are not the presentation edition for their work.

I could be wrong, but we should be able to simplify the code considerably by getting rid of this field.

aslagle:
Instead of multiple Editions sharing a work id, Work now has a presentation edition id, and only one edition is directly associated with a work. This matches how license pools are linked to their presentation editions.

### DIFF
--- a/migration/20160517-remove-edition-is-primary-for-work.sql
+++ b/migration/20160517-remove-edition-is-primary-for-work.sql
@@ -1,0 +1,223 @@
+ALTER TABLE works ADD COLUMN presentation_edition_id integer;
+UPDATE works as w SET presentation_edition_id = e.id from editions as e where e.work_id = w.id and e.is_primary_for_work = true;
+ALTER TABLE editions DROP COLUMN work_id CASCADE;
+ALTER TABLE editions DROP COLUMN is_primary_for_work CASCADE;
+
+
+create materialized view mv_works_editions_datasources_identifiers
+as
+ SELECT 
+    distinct works.id AS works_id,
+    editions.id AS editions_id,
+    editions.data_source_id,
+    editions.primary_identifier_id,
+    editions.sort_title,
+    editions.permanent_work_id,
+    editions.sort_author,
+    editions.medium,
+    editions.language,
+    editions.cover_full_url,
+    editions.cover_thumbnail_url,
+    editions.open_access_download_url,
+    datasources.name,
+    identifiers.type,
+    identifiers.identifier,
+    works.audience,
+    works.target_age,
+    works.fiction,
+    works.quality,
+    works.rating,
+    works.popularity,
+    works.random,
+    works.last_update_time,
+    works.simple_opds_entry,
+    works.verbose_opds_entry,
+    licensepools.id AS license_pool_id,
+    licensepools.availability_time
+
+   FROM works
+     JOIN editions ON editions.id = works.presentation_edition_id
+     JOIN licensepools ON editions.data_source_id = licensepools.data_source_id AND editions.primary_identifier_id = licensepools.identifier_id
+     JOIN datasources ON editions.data_source_id = datasources.id
+     JOIN identifiers on editions.primary_identifier_id = identifiers.id
+  WHERE works.presentation_ready = true
+    AND works.simple_opds_entry IS NOT NULL
+  
+  ORDER BY editions.sort_title, editions.sort_author, licensepools.availability_time;
+
+-- Create a unique index so that searches can look up books by work ID.
+
+create unique index mv_works_editions_work_id on mv_works_editions_datasources_identifiers (works_id);
+
+-- Create an index on everything, sorted by descending availability time, so that sync feeds are fast.
+
+create index mv_works_editions_by_availability on mv_works_editions_datasources_identifiers (availability_time DESC, sort_author, sort_title, works_id);
+
+-- Similarly, an index on everything, sorted by descending update time.
+
+create index mv_works_editions_by_modification on mv_works_editions_datasources_identifiers (last_update_time DESC, sort_author, sort_title, works_id);
+
+-- We need three versions of each index:
+--- One that orders by sort_author, sort_title, and works_id
+--- One that orders by sort_title, sort_author, and works_id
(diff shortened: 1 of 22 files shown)


model.py
```diff
@@ -2239,13 +2239,9 @@ class Edition(Base):
     primary_identifier_id = Column(
         Integer, ForeignKey('identifiers.id'), index=True)
 
-    # A Edition may be associated with a single Work.
-    work_id = Column(Integer, ForeignKey('works.id'), index=True)
+    # An Edition may be the presentation edition for a single Work.
+    work = relationship("Work", uselist=False, backref="presentation_edition")
```
aslagle: I'm not sure what to call this. For LicensePool, it's is_presentation_for, but that's confusing now that it can also be the presentation for a Work.

leonardr: I think it's okay to just call it `work` with the knowledge that `Edition.work` can now be None if the Edition isn't a presentation edition.

migration/20160517-remove-edition-is-primary-for-work.sql
```diff
@@ -0,0 +1,223 @@
+ALTER TABLE works ADD COLUMN presentation_edition_id integer;
```
leonardr: I think you also need to explicitly add the index on Work.presentation_edition_id -- it won't be created automatically unless you create the database from scratch.

model.py
```diff
@@ -3237,7 +3215,7 @@ def with_no_genres(self, q):
 
     def all_editions(self, recursion_level=5):
         """All Editions identified by a Identifier equivalent to 
-        any of the primary identifiers of this Work's Editions.
+        the primary identifiers of this Work's presentation edition.
```
leonardr: "identifier" is now singular

model.py
```diff
@@ -3250,12 +3228,9 @@ def all_editions(self, recursion_level=5):
 
     def all_identifier_ids(self, recursion_level=5):
         _db = Session.object_session(self)
-        primary_identifier_ids = [
-            x.primary_identifier.id for x in self.editions
-            if x.primary_identifier
-        ]
+        primary_identifier_id = self.presentation_edition.primary_identifier.id
```
leonardr: The old code has to go but I don't think this is right. I think you now want to iterate over all the LicensePools associated with the Work and use their identifiers as your initial set.

If a Work has multiple LicensePools, we've made a determination (based on permanent work ID) that the Identifiers of those LicensePools identify "the same book". This is the sort of thing we usually track in `equivalencies`, and that's what `Identifier.recursively_equivalent_identifier_ids_flat` handles. But when we decide that two LicensePools identify the same Work, we don't actually create an equivalency between their Identifiers. So picking one of those Identifiers and using it as the initial set will tend to leave Identifiers out.

Maybe we should create those equivalencies! If we did that, the presentation edition's identifier ID would be equivalent to the identifier IDs of all the other LicensePools associated with that work. In that case, the only change you'd have to make here is to change `recursion_level` from 5 to 6 (because there's now an extra step between the identifier of the presentation edition and the identifier associated with one of the other LicensePools).

But I think it would be simpler to not create that equivalency, and generate our initial set of `primary_identifier_ids` from the identifiers of all the LicensePools grouped together in this Work.

model.py
```diff
@@ -3270,10 +3245,9 @@ def language_code(self):
 
     def all_cover_images(self):
         _db = Session.object_session(self)
-        primary_identifier_ids = [
-            x.primary_identifier.id for x in self.editions]
+        primary_identifier_id = self.presentation_edition.primary_identifier.id
```
leonardr: Same deal here, and below. Maybe create an internal method that gets the starter set of identifier IDs.

model.py
```diff
@@ -3284,74 +3258,55 @@ def all_cover_images(self):
...
-            if pool.presentation_edition is self.primary_edition:
-                # make sure edition knows it's primary
-                pool.presentation_edition.is_primary_for_work = True
-            else:
-                pool.mark_edition_primarity(primary_for_work_edition=None)
+        self.presentation_edition = new_presentation_edition
 
+        # Let the edition's license pool know it has a work.
+        if self.presentation_edition.is_presentation_for:
+            self.presentation_edition.is_presentation_for.work = self
```
leonardr: Am I right in thinking there's no need to notify the losers because this actually sets Work.presentation_edition_id?

aslagle: Right, loser editions won't have a work already. The old code had to set is_primary_for_work separately.

model.py
```diff
@@ -3448,12 +3400,9 @@ def calculate_presentation(self, policy=None, search_index_client=None):
             # classifications, or measurements.
             _db = Session.object_session(self)
 
-            primary_identifier_ids = [
-                x.primary_identifier.id for x in self.editions
-                if x.primary_identifier
-            ]
+            primary_identifier_id = self.presentation_edition.primary_identifier.id
```
leonardr: Once again, I think you want all the identifiers associated with all the LicensePools.

tests/test_model.py
```diff
@@ -1877,18 +1849,17 @@ def test_calculate_work_does_nothing_unless_edition_has_title_and_author(self):
...
+        # work's presentation edition.
         eq_(work, edition.work)
-        eq_(True, edition.is_primary_for_work)
-        eq_(edition, work.primary_edition)
+        eq_(edition, work.presentation_edition)
 
-        # But without this commit, the join for the .primary_edition
+        # But without this commit, the join for the .presentation_edition
         # won't succeed and work.title won't work.
         self._db.commit()
```
leonardr: This might not be necessary anymore; I remember the problem had something to do with the foreign key relationships and you've just simplified that.

model.py
leonardr: I guess I'm worried about the case where a former winner Edition becomes a loser, but we're editing Work, not Edition, so it's okay.

leonardr:
:+1:

PR #252, opened 2016-05-18
Find the main library simplified directory dynamically

courte:
Because this script came from our circulation-docker, this directory in the code was hardcoded up until now as /var/www/circulation. Now that `run` is in core/bin, we know we can find the main directory if we shed the filename and go up two levels.

I've tested this on the BPL server, and it's working great. Just trust me. It's fine. 😉👌


leonardr:
:+1:

PR #253, opened 2016-05-18
Allow BibliographicRefreshScript to run even on a site where not every single license source is configured

leonardr:
If we encounter CannotLoadConfiguration when running bibliographic_coverage, it's more likely because the library doesn't have a certain license source than that there's a real problem. So just mention what happened and move on.


courte:
:+1: Thanks!

PR #254, opened 2016-05-18
Don't add breadcrumbs to a feed with no visible parent.

aslagle:
This fixes the content server, which has page feeds with no parent or ancestors, and doesn't have an implementation of lane_url in its annotator.

I don't think we'd ever have a lane with ancestors but no visible parent.


leonardr:
:+1:

PR #259, opened 2016-05-19
If an Exception tells you how to turn it into a problem detail document, use that document when handling the error

leonardr:
This branch makes it so that an Exception subclass can have an associated ProblemDetail associated with it. If that exception is raised in the operation of the app server, and not caught, the ErrorHandler will use the ProblemDetail to generate the response instead of returning a generic plain-text response.


aslagle:
:+1:

PR #260, opened 2016-05-20
Pass around identifier object instead of string during OPDS creation.

aslagle:
This tiny change to core is to support https://github.com/NYPL-Simplified/circulation/issues/187, since we need the identifier type as well the identifier when creating URLs now.


leonardr:
:+1:

PR #261, opened 2016-05-20
Staff classifications rule

skomputer:
This branch modifies WorkClassifier to ensure that classifications entered by staff override other classifications.


aslagle:
:+1:

PR #262, opened 2016-05-20
Target age max not inclusive fix

skomputer:
This branch fixes a bug in WorkClassifier that incorrectly transformed a max target age from non-inclusive to inclusive, resulting in a max target age that was 2 years higher than intended.


leonardr:
:+1:

PR #265, opened 2016-05-23
Make problem detail documents out of network errors and timeouts

leonardr:
This branch makes it clear to the patron when their request failed because of the failure of an external service accessible via HTTP.

Every HTTP interaction that might involve a patron is routed through the new helper method `HTTP.request_with_timeout`. All `requests` exceptions are caught and rethrown as new exception subclasses that support `to_problem_detail`.


aslagle:
:+1:

PR #266, opened 2016-05-23
Turn 5xx response codes into RemoteIntegrationExceptions.

leonardr:
This branch adds a new type of exception, RemoteIntegrationException, for errors that are caused by the failure of a remote integration rather than something on our side. RequestNetworkException is now a `requests`-specific subclass of this exception, but it is also raised if the HTTP request completes normally but with a 5xx response code. I also changed the one place in core where `RemoteIntegrationException` should be raised even on a non-5xx response code: the code that acquires the Axis 360 access token, where anything other than a 200 response is treated as an error.

### DIFF
--- a/axis.py
+++ b/axis.py
@@ -9,7 +9,10 @@ import re
 
 from util import LanguageCodes
 from util.xmlparser import XMLParser
-from util.http import HTTP
+from util.http import (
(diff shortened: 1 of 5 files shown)


aslagle:
:+1:

PR #267, opened 2016-05-23
using schema namespace for entry alternativeHeadline elements

skomputer:
This branch adds a `schema:` namespace to [alternativeHeadline](https://schema.org/alternativeHeadline) elements in entry xml.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -893,7 +893,9 @@ class AcquisitionFeed(OPDSFeed):
             **kw
         )
         if edition.subtitle:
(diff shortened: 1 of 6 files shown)


leonardr:
This is the right thing to do, but it's going to make opds_import.py not know where to look for subtitles. You need to also change detail_for_feedparser_entry to look in the right place, and change `content_server.opds`, `content_server_mini.opds`, and `metadata_wrangler_overdrive.opds` in `tests/files/opds` to reflect what our OPDS servers would actually generate.

leonardr:
:+1:

PR #268, opened 2016-05-24
Raise a subclass of RemoteIntegrationException when the request completes but gives a bad result

leonardr:
This branch introduces `BadResponseException`, which is used in cases where there was no network-level problem communicating with an external HTTP server, but the response was bad. What is meant by "bad" is highly variable, but it generally means "we got a 5xx response code".

However, there are cases where anything other than a 200 response code is unacceptable, or where the only unacceptable response code is a 401. I've gotten all these cases working and tested, and it lets me save little bits of code here and there.

Since `BadResponseException` has a real HTTP response to work with, I changed its `debug_message` to include the text of the response, which can sometimes be helpful.

### DIFF
--- a/axis.py
+++ b/axis.py
@@ -107,7 +107,9 @@ class Axis360API(object):
     def refresh_bearer_token(self):
         url = self.base_url + self.access_token_endpoint
         headers = self.authorization_headers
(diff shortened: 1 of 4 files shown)


aslagle:
:+1:

PR #269, opened 2016-05-24
Overdrive uses http helper

leonardr:
This branch fixes a problem where the Overdrive API was not properly using the util.http.HTTP class to make POST requests. To make sure the problem has been fixed I created a MockOverdriveAPI class and for the first time wrote tests for the procedure by which Overdrive creates and maintains its bearer token, as well as the various failures that might happen.

I also created a MockThreeMAPI class and wrote some basic tests for it. The real tests of the 3M API will be in the circulation manager.

### DIFF
--- a/threem.py
+++ b/threem.py
@@ -131,13 +131,16 @@ class ThreeMAPI(object):
         signature = base64.b64encode(digest)
         return signature, now
 
-    def request(self, path, body=None, method="GET", identifier=None,
-                max_age=None):
+    def full_url(self, path):
         if not path.startswith("/"):
             path = "/" + path
         if not path.startswith("/cirrus"):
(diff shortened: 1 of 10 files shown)


threem.py
```diff
@@ -131,13 +131,16 @@ def signature(self, method, path):
         signature = base64.b64encode(digest)
         return signature, now
 
-    def request(self, path, body=None, method="GET", identifier=None,
-                max_age=None):
+    def full_url(self, path):
```
leonardr: This is unrelated--it's needed by the circulation branch I'm working on to get 3M using this system.

aslagle:
👍

aslagle:
:+1:

PR #270, opened 2016-05-24
Metadata.apply() clears fields with 'none' values

skomputer:


leonardr:
:+1: 

I'm not terribly happy with the use of -1 to indicate the lack of a value, but it's semi-standard and I don't have a better idea.

PR #271, opened 2016-05-24
Fix 3m signature

leonardr:
This branch fixes the inputs into the code that calculates the cryptographic signature for a 3M HTTP request. I screwed them up when I created `ThreeMAPI.full_url`. The thing that gets signed is not the URL, it's just the path part of the URL, and my implementation of `full_url` changed the path in a way that wasn't propagated back to the caller.


aslagle:
:+1:

PR #272, opened 2016-05-25
Prepare to replace circulation RemoteInitiatedServerError with core RemoteIntegrationException

leonardr:
This branch makes it clear that you can create a RemoteIntegrationException and specify only the service name, not the exact URL that caused the problem. This will make it possible to replace the circ manager's RemoteInitiatedServerError class with RemoteIntegrationException.

After looking over the code in circulation I've decided not to actually replace those classes right now. RemoteInitiatedServerError does the important part of the work--it turns remote errors into 502 errors instead of 500 errors. If I change it I'm going to have to write a bunch of unit tests for  FirstBook and 3M to make sure I didn't break anything. Those unit tests ought to be written but it's not my priority.

I also changed some of the wording so that a network-level error shows up as a problem "contacting" the third-party server, not a problem "accessing" it.

Finally, I changed the ErrorHandler so that an exception that results in a 502 status code (which includes both RemoteInitiatedServerError and RemoteIntegrationException) is logged at log level WARN instead of ERROR. These exceptions indicate serious problems, but they're generally problems that we can't fix. Either they're transient, in which case they're just randomly occuring problems, or they're persistent, in which case we need to bother the upstream source of the problem.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -173,8 +173,6 @@ class ErrorHandler(object):
         self.debug = debug
 
     def handle(self, exception):
-        logging.error(
-            "Exception in web app: %s", exception, exc_info=exception)
         if hasattr(self.app, 'manager') and hasattr(self.app.manager, '_db'):
             # There is an active database session. Roll it back.
             self.app.manager._db.rollback()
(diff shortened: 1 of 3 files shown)


aslagle:
:+1:

PR #274, opened 2016-05-25
Filter out books for which we no longer have licenses.

leonardr:
This branch gets `Lane.only_show_ready_deliverable_works` to filter out non-open-access books for which we currently have no `licenses_owned`. This stops such books from showing up in lanes as well as search results.

This fixes https://github.com/NYPL-Simplified/circulation/issues/213


aslagle:
:+1:

PR #276, opened 2016-05-26
Fixed URNLookupController adding 500 message for existing works.

aslagle:
This fixes a problem where existing works would be counted as missing and show an entry with a 500 error as well as the correct entry.

process_urn handles error messages as well as works and precomposed entries now.


leonardr:
:+1: once the tests complete.

PR #277, opened 2016-05-27
Refactor and mock OPDS lookup class

leonardr:
This branch refactors the class that implements the Simplified OPDS lookup protocol so that it uses the HTTP helper class. I also created a mock version of `SimplifiedOPDSLookup`, though I ended up not needing it.


aslagle:
:+1:

PR #278, opened 2016-05-27
Add has_next_page to pagination

courte:
This does some surgery on SQLAlchemy's Query object to allow for a faster count, and then uses this faster method to determine if there are any more objects to be paginated.


tests/test_lane.py
```diff
@@ -1005,5 +1006,36 @@ def test_only_show_ready_deliverable_works(self):
...
+
+    def test_has_next_page(self):
+        query = self._db.query(Work)
+        pagination = Pagination(size=2)
+
+        # When the query is empty, pagination doesn't have a next page.
+        pagination.apply(query)
+        eq_(False, pagination.has_next_page)
+
+        # When there are more works in the query, it doesn't.
```
leonardr: The comment is backwards.

tests/test_lane.py
```diff
@@ -1005,5 +1006,36 @@ def test_only_show_ready_deliverable_works(self):
...
+        for num in range(3):
+            # Create three works.
+            self._work()
+        pagination.apply(query)
+        eq_(True, pagination.has_next_page)
+
+        # When there aren't, pagination has_next_page.
```
leonardr: This one too.

leonardr:
I know I told you earlier that this could be "a new method on Query", but Amy pointed out that it would be better as a normal Python function that takes a query as an argument. Try that out and see how it works.

leonardr:
:+1:

PR #279, opened 2016-05-27
Fixed the mock OPDS lookup client now that we're using it in a test.

leonardr:
This branch makes some very small changes to mock classes used in circulation tests. The mock OPDS lookup client was never used before, which is how the constructor error slipped through.


aslagle:
:+1:

PR #280, opened 2016-05-27
Don't crash if metadata wrangler is unconfigured.

leonardr:
This should really be done with generic auth being provided to the constructor, and then a metadata wrangler-specific subclass, but this is good enough for now.


aslagle:
:+1:

PR #281, opened 2016-05-27
Create recommendation data

courte:
This branch translates between various Recommendation APIs (which is to say, NoveList, for now) and the works that we have available.

Also `Work.feed_query` wasn't being used anywhere, but it seems pretty useful, and looks like it overlaps with the work being done by `Lane.only_show_ready_deliverable_works`. I made some slight tweaks so it could work on a broader scale than just the feed, since I needed it to make sure the that print and audiobook ISBNs that NoveList sends back don't count as a recommendation.

Supports NYPL-Simplified/circulation#227.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -400,8 +400,8 @@ class RecommendationData(object):
 
         equivalent_identifier = aliased(Identifier)
         works_q = Work.feed_query(_db)
-        works_q = works_q.join(Identifier.equivalencies).\
-            join(equivalent_identifier, Equivalency.output).\
+        works_q = works_q.outerjoin(Identifier.equivalencies).\
+            outerjoin(equivalent_identifier, Equivalency.output).\
             filter(or_(
                 Identifier.id.in_(identifier_ids),
                 equivalent_identifier.id.in_(identifier_ids)
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -377,6 +377,9 @@ class RecommendationData(object):
 
     @property
     def recommended_works(self):
+        """:return: a query of all of the recommended works in self.identifiers
+        or None
+        """
         identifier_ids = []
         _db = Session.object_session(self.data_source)
         for identifier_data in self.identifiers[:]:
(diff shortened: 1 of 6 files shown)


metadata_layer.py
```diff
@@ -361,6 +368,46 @@ def __init__(self, content_type, drm_scheme, link=None):
             )
         self.link = link
 
+
+class RecommendationData(object):
+
+    def __init__(self, data_source, identifiers=None):
+        self.data_source = data_source
+        self.identifiers = identifiers or []
```
leonardr: I think you should move this class into model.py, even though there's currently no database model object for recommendations. All the code you've got here deals with the database, and the metadata layer is supposed to be independent of the database. You start here by taking a DataSource object (as opposed to the name of a data source) as an argument, and if that was all it was, I'd just suggest you change that part. But...

metadata_layer.py
```diff
@@ -361,6 +368,46 @@ def __init__(self, content_type, drm_scheme, link=None):
...
+
+    @property
+    def recommended_works(self):
```
leonardr: Then you've got this thing which runs a query, and really ought to be in model.py. It feels like a method called `Work.from_identifiers`, which takes a list of Identifier objects and returns a list of Works.

metadata_layer.py
```diff
@@ -361,6 +368,46 @@ def __init__(self, content_type, drm_scheme, link=None):
...
+            if not identifier:
+                self.identifiers.remove(identifier_data)
+                continue
+            identifier_ids.append(identifier.id)
+
+        if not identifier_ids:
+            return None
+
+        equivalent_identifier = aliased(Identifier)
+        works_q = Work.feed_query(_db)
```
leonardr: Wow, I'd forgotten about `feed_query`. I stopped using it when I got all the "get a list of Works" code into Lanes. Since this is a "list of Works" and it's going to eventually get turned into an OPDS feed, I think it makes sense as a type of `Lane`. Maybe a subclass? Then you'd be able to get rid of `feed_query` altogether.

PR #285, opened 2016-06-02
analytics providers initialized in config

skomputer:
This branch initializes analytics providers from modules specified in the configuration file and collects analytics events whenever the availability of license pools change.

### DIFF
--- a/config.py
+++ b/config.py
@@ -309,8 +309,9 @@ class Configuration(object):
         return cls.policy(cls.SHOW_STAFF_PICKS_ON_TOP_LEVEL, default=True)
 
     @classmethod
-    def collect_analytics_event(cls, *args, **kwargs):
-        cls.instance[cls.POLICIES][cls.ANALYTICS_POLICY].collect_event(*args, **kwargs)
+    def collect_analytics_event(cls, _db, license_pool, event_type, time, **kwargs):
+        cls.instance[cls.POLICIES][cls.ANALYTICS_POLICY].collect_event(
+            _db, license_pool, event_type, time, **kwargs)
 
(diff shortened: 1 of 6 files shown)


mock_analytics_provider.py
```diff
@@ -0,0 +1,18 @@
...
+    @classmethod
+    def from_config(cls, config):        
+        return cls(config.get('option'))
+
+    def __init__(self, option=None):
+        self.option = option
+        self.count = 0
+        self.event = None
+
+    def collect_event(self, _db, lp, event_type, time, **kwargs):
```
leonardr: This is fine for now, but if testing gets more complicated you might like to store the events in a list rather thanjust store a count and the type of the most recent event.

config.py
```diff
@@ -327,6 +334,11 @@ def load(cls):
...
     @classmethod
-    def _load(cls, fh):
-        lines = [x for x in fh if not x.strip().startswith("#")]
-        return json.loads("".join(lines))
+    def _load(cls, str):
+        lines = [x for x in str.split("\n") if not x.strip().startswith("#")]
+        config = json.loads("\n".join(lines))
+        if not config.get(cls.POLICIES):
+            config[cls.POLICIES] = {}
+        providers = config[cls.POLICIES].get(cls.ANALYTICS_POLICY, [])
```
leonardr: I'd like to see the default analytics policy (i.e. if you specify nothing in the config) to be the current default (i.e. analytics events get logged as CirculationEvent objects). This seems like the appropriate place to put this information.

leonardr:
Looks good apart from my confusion about the default behavior, which you should be able to resolve in the circulation manager.

PR #286, opened 2016-06-02
Calling LicensePool.calculate_work() will automatically merge and split LicensePools to enforce the rules on Works

leonardr:
Our production data has a lot of Works that contain LicensePools they shouldn't contain. Rather than try to tease out all the past bugs that led to this situation, this branch makes LicensePool.calculate_work() create new Works, destroy existing Works, and move LicensePools between works to enforce these two rules:

1. Every commercially-licensed LicensePool has its own unique Work.
2. Every open-access LicensePool associated with a given permanent work ID is grouped together in a single Work, which contains _only_ LicensePools associated with that specific PWID.

### DIFF
--- a/model.py
+++ b/model.py
@@ -3161,7 +3161,6 @@ class Work(Base):
         """
         _db = Session.object_session(self)
         for pool in list(self.license_pools):
-            print pool.presentation_edition.permanent_work_id
             other_work = is_new = None
             if not pool.open_access:
                 # This needs to have its own Work--we don't mix
@@ -3175,7 +3174,6 @@ class Work(Base):
                 this_pwid = pool.presentation_edition.permanent_work_id
                 if not this_pwid:
                     continue
-                print this_pwid, pwid
                 if this_pwid != pwid:
                     # This LicensePool should not belong to this Work.
                     # Make sure it gets its own Work, creating a new one
--- a/model.py
+++ b/model.py
@@ -3215,8 +3215,6 @@ class Work(Base):
             _db.delete(wg)
         for cr in self.coverage_records:
             _db.delete(cr)
-        for contribution in self.contributions:
-            _db.delete(contribution)
         _db.delete(self)
 
     def set_summary(self, resource):
--- a/model.py
+++ b/model.py
@@ -3090,14 +3090,14 @@ class Work(Base):
                 len(self.license_pools))).encode("utf8")
 
     @classmethod
-    def for_permanent_work_id(cls, pwid):
+    def for_permanent_work_id(cls, _db, pwid):
         """Find or create the Work encompassing all open-access LicensePools
         whose presentation Editions have the given permanent work ID.
 
-        This may result in the consolidation of Works, if a book's
-        permanent work ID has changed without calculate_work() being
-        called, or if the data is in an inconsistent state for any
-        other reason.
+        This may result in the consolidation or splitting of Works, if
+        a book's permanent work ID has changed without
+        calculate_work() being called, or if the data is in an
+        inconsistent state for any other reason.
         """
         is_new = False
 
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -5362,39 +5525,33 @@ def consolidate_works(cls, _db, calculate_work_even_if_no_author=False,
 
 
     def calculate_work(self, even_if_no_author=False, known_edition=None):
```
leonardr: known_edition has started to cause problems. I think with the advent of presentation_edition it's no longer necessary. There used to be situations surrounding OPDS import where the LicensePool's presentation edition was missing and had to be passed in, but now I think we can consistently rely on LicensePool.presentation_edition working properly.

I'm not changing the use of known_edition in this branch but hopefully this explains a couple changes I made later on.

model.py
```diff
@@ -5362,39 +5525,33 @@ def consolidate_works(cls, _db, calculate_work_even_if_no_author=False,
...
 
         :param even_if_no_author: Ordinarily this method will refuse
         to create a Work for a LicensePool whose Edition has no title
         or author. But sometimes a book just has no known author. If
         that's really the case, pass in even_if_no_author=True and the
         Work will be created.
         """
-        self.set_presentation_edition(None)
-
-        presentation_edition = known_edition or self.presentation_edition
```
leonardr: Specifically, there's frequently a problem right here. If we always call set_presentation_edition(), there's a good chance that `self.presentation_edition` will be set to a different edition than `known_edition`. This causes the check `presentation_edition.is_presentation_for != self` to fail, because only one Edition can be the presentation edition for a LicensePool. Even if that magic Edition used to be `known_edition`, it might not be `known_edition` by the time we finish calling `set_presentation_edition().`

So in the future I want to get rid of `known_edition` and always rely on `LicensePool.presentation_edition`. But for now, I'm skipping the `set_presentation_edition` call if `known_edition` is provided.

scripts.py
```diff
@@ -419,100 +419,45 @@ def process_work(self, work):
         raise NotImplementedError()      
 
 class WorkConsolidationScript(WorkProcessingScript):
+    """Given an Identifier, make sure all the LicensePools for that
+    Identifier are in Works that follow these rules:
 
-    name = "Work consolidation script"
```
leonardr: This script tried to do a lot of the work now done by LicensePool.calculate_work(), so I was able to rip out most of it and just get it to call calculate_work() over and over again.

model.py
```diff
@@ -3085,10 +3085,161 @@ def has_open_access_license(self):
...
+        # Work.
+        licensepools_for_work = Counter()
+        for lp in qu:
+            if lp.work and not licensepools_for_work[lp.work]:
+                licensepools_for_work[lp.work] = len(lp.work.license_pools)
+
+        work = None
+        if len(licensepools_for_work) == 0:
+            # None of these LicensePools have a Work. Create a new one.
+            for pool in licensepools:
```
aslagle: I don't think you need a loop here, you only want one work.

model.py
```diff
@@ -5424,50 +5579,96 @@ def calculate_work(self, even_if_no_author=False, known_edition=None):
...
+            #
+            # A commercially-licensed book should have a Work to
+            # itself. All other LicensePools need to be kicked out and
+            # associated with some other work.
+            #
+            # This won't cause an infinite recursion because we're
+            # setting pool.work to None before calling
+            # pool.calculate_work(), and the recursive call only
+            # happens if self.work is set.
+            my_pwid = self.presentation_edition
```
aslagle: this isn't used, and it's also wrong

model.py
```diff
@@ -5424,50 +5579,96 @@ def calculate_work(self, even_if_no_author=False, known_edition=None):
...
+            work.license_pools.append(self)
+            licensepools_changed = True
 
         # Recalculate the display information for the Work, since the
-        # associated Editions have changed.
-        work.calculate_presentation()
+        # associated LicensePools have changed, which may have caused
+        # the Work's presentation Edition to change.
+        if licensepools_changed:
+            work.calculate_presentation()
```
aslagle: I've been assuming calculate_work always calls calculate_presentation, so there may be some places we need to add a call to calculate_presentation now.

leonardr: OK, I took that out for now since it was only there for efficiency.

aslagle:
:+1:

PR #287, opened 2016-06-02
Support recommendation lanes

courte:
Replacing #281, this branch supports NYPL-Simplified/circulation#227. As suggested in that PR, this branch:

- creates `Work.from_identifiers`, a schweet method to return all of the works related to a list of identifiers
- tosses a list of recommendations (Identifiers or IdentifierDatas) into the metadata layer and gives Metadata the ability to filter out identifiers that aren't available in circulation
- Makes some slightly-awkward caveats around CachedFeeds, including connecting them to LicensePools and letting Lanes indicate their own `MAX_CACHE_AGE`
- Also let's Lane subclasses set an `apply_filters()` method that just cancels out the whole dang thing.

In general, lanes are on the come up. I think this is moving towards a `WorkBasedLane` or `LicensePoolBasedLane` parent class (for books in a series, other works by an author, etc.), in which case `Lane` will likely change a bit more. I wanted to do minimal damage here, though.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -400,8 +400,8 @@ class RecommendationData(object):
 
         equivalent_identifier = aliased(Identifier)
         works_q = Work.feed_query(_db)
-        works_q = works_q.join(Identifier.equivalencies).\
-            join(equivalent_identifier, Equivalency.output).\
+        works_q = works_q.outerjoin(Identifier.equivalencies).\
+            outerjoin(equivalent_identifier, Equivalency.output).\
             filter(or_(
                 Identifier.id.in_(identifier_ids),
                 equivalent_identifier.id.in_(identifier_ids)
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -377,6 +377,9 @@ class RecommendationData(object):
 
     @property
     def recommended_works(self):
+        """:return: a query of all of the recommended works in self.identifiers
+        or None
+        """
         identifier_ids = []
         _db = Session.object_session(self.data_source)
         for identifier_data in self.identifiers[:]:
@@ -392,6 +395,9 @@ class RecommendationData(object):
                 continue
             identifier_ids.append(identifier.id)
 
+        if not identifier_ids:
+            return None
+
         equivalent_identifier = aliased(Identifier)
         works_q = Work.feed_query(_db)
         works_q = works_q.join(Identifier.equivalencies).\
@@ -400,9 +406,7 @@ class RecommendationData(object):
                 Identifier.id.in_(identifier_ids),
                 equivalent_identifier.id.in_(identifier_ids)
             ))
-
         return works_q
-
 
 class CirculationData(object):
 
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -11,10 +11,11 @@ from collections import defaultdict
 from sqlalchemy.orm.session import Session
 from nose.tools import set_trace
 from dateutil.parser import parse
-from sqlalchemy.sql.expression import and_
+from sqlalchemy.sql.expression import and_, or_
 from sqlalchemy.orm.exc import (
     NoResultFound,
 )
+from sqlalchemy.orm import aliased
 import csv
 import datetime
 import logging
@@ -29,6 +30,7 @@ from model import (
     DataSource,
     DeliveryMechanism,
     Edition,
+    Equivalency,
     Hyperlink,
     Identifier,
     LicensePool,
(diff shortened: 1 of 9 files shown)


model.py
```diff
@@ -3151,6 +3115,29 @@ def with_no_genres(self, q):
         q = q.filter(WorkGenre.genre==None)
         return q
 
+    @classmethod
+    def from_identifiers(cls, _db, identifiers, base_query=None):
+        """Finds all of the works indicated by a list of identifiers"""
```
leonardr: Make the rules clear in the docstring. A work is indicated by an identifier if one of its LicensePools is directly associated with the identifier, or if one of its LicensePools is directly equivalent to the identifier.

model.py
```diff
@@ -3151,6 +3115,29 @@ def with_no_genres(self, q):
...
+            base_query = _db.query(Work).join(Work.license_pools).\
+                join(LicensePool.identifier)
+
+        equivalent_identifier = aliased(Identifier)
+        query = base_query.outerjoin(Identifier.equivalencies).\
+                outerjoin(equivalent_identifier, Equivalency.output).\
+                filter(or_(
+                    Identifier.id.in_(identifier_ids),
+                    equivalent_identifier.id.in_(identifier_ids)
+                ))
```
leonardr: I think you should put a minimum threshold on Equivalency.strength so that someone's 2% opinion that two books are the same doesn't trigger a recommendation. Right now this won't matter because ISBN equivalencies directly associated with a commercial ID (such as an Overdrive ID) are very strong, but once we start getting LicensePools sold by ISBN, bad equivalencies will become a much bigger risk.

I would set the threshold at maybe 0.6 or 0.7.

model.py
```diff
@@ -4795,24 +4776,34 @@ class CachedFeed(Base):
     # The content of the feed.
     content = Column(Unicode, nullable=True)
 
+    # A LicensePool for feeds associated with a certain work.
```
leonardr: I would say 'literary work' to avoid confusion with the Work data model object. Or just say 'LicensePool'.

opds.py
```diff
@@ -488,16 +488,17 @@ class AcquisitionFeed(OPDSFeed):
 
     @classmethod
     def groups(cls, _db, title, url, lane, annotator, 
-               force_refresh=False,
+               force_refresh=False, cache_type=None,
```
leonardr: don't understand why cache_type is now an argument to `groups`. I get `page` but not `groups`. Is it just to keep the signatures the same or do you anticipating making both page-type and group-type recommendation feeds? (If you do plan on making both, then 'recommendations' isn't a new type of CachedFeed, it's a separate piece of information that needs to be stored separately from whether the feed is `page` or `groups`.

metadata_layer.py
```diff
@@ -1292,10 +1294,29 @@ def update_contributions(self, _db, edition, metadata_client=None,
                     contributor_data.display_name
                 )
 
+    def filter_recommendations(self, _db):
+        """Filters out  recommendations that don't currently exist in the db
```
leonardr: This is a little misleading. We're filtering out recommended _Identifiers_ that don't exist in the database.

I was wondering if it would also make sense to filter out Identifiers that are not associated with any LicensePool, but you're dealing with that in circulation's RecommendationLane, and I think that makes sense.

Looking up Identifiers one at a time is pretty slow, and you could improve this by doing one lookup for each type of identifier, i.e. gather all the ISBNs and do a query like `_db.query(Identifier).filter(Identifier.type==ISBN).filter(Identifier.identifier.in_(...))`

That will use only one query and tell you which of the ISBNs are actually in the database.

tests/test_opds.py
```diff
@@ -826,13 +826,13 @@ def test_groups_feed(self):
...
                 self._db, "test", self._url, fantasy_lane, annotator, 
-                False, False
+                force_refresh=False, use_materialized_works=False
             )
             eq_(CachedFeed.PAGE_TYPE, feed.type)
 
             cached_groups = AcquisitionFeed.groups(
                 self._db, "test", self._url, fantasy_lane, annotator, 
-                True, False
+                force_refresh=True, use_materialized_works=False
```
leonardr: FWIW, you can update the materialized views with `SessionManager.refresh_materialized_views(self._db)`. There are some examples in test_opds.py.

opds.py
courte: Nope, just a signature choice, and it can go.

leonardr:
:+1:

PR #289, opened 2016-06-06
Localization

aslagle:
This branch wraps error strings in a function so they can be extracted with a script and localized with Flask-Babel. 

Unfortunately, using Flask-Babel in core like this will mean that all the flask apps need to initialize Babel, even if they don't have any translations. I haven't been able to figure out a way around this.

I also removed a few emacs temp files I noticed were in the repo.

This connects to https://github.com/NYPL-Simplified/circulation/issues/244


app_server.py
```diff
@@ -149,13 +150,13 @@ def load_pagination(size, offset):
     try:
         size = int(size)
     except ValueError:
-        return INVALID_INPUT.detailed("Invalid size: %s" % size)
+        return INVALID_INPUT.detailed(_("Invalid size: %(size)s", size=size))
```
leonardr: "Invalid page size" would be more helpful and easier to translate.

config.py
```diff
@@ -309,6 +311,10 @@ def show_staff_picks_on_top_level(cls):
         return cls.policy(cls.SHOW_STAFF_PICKS_ON_TOP_LEVEL, default=True)
 
     @classmethod
+    def localization_languages(cls):
+        return cls.policy(cls.LOCALIZATION_LANGUAGES, default=["en"])
```
leonardr: We generally use the 3-letter language codes in the config file but I see why you're using the 2-letter codes here--that's what Flask-Babel's request.accept_languages_best_match takes. Is it worth taking 3-letter codes for the sake of consistency and using LanguageCodes.three_to_two to convert them?

leonardr:
In private conversation we decided it would be good to use lazy_gettext consistently and change `ProblemDocument.detailed` or possibly `problem_detail.json()` to issue a warning if it is given a title or description that is not a `LazyString`. This will help us spot future localization problems.

util/problem_detail.py
```diff
@@ -53,6 +52,14 @@ def detailed(self, detail, status_code=None, title=None, instance=None,
 
         The detailed error message will be shown to patrons.
         """
+
+        # Title and detail must be LazyStrings from Flask-Babel that are
+        # localized when they are first used as strings.
+        if title and not isinstance(title, LazyString):
+            raise Exception("\"%s\" has not been internationalized" % title)
```
leonardr: Raising an exception here will create a 500 error and make the patron think something is seriously wrong. I'd prefer to just log a warning. We can set up an alert for "has not been internationalized".

leonardr:
Now even more :+1:!

leonardr:
:+1: once again

PR #290, opened 2016-06-06
Fix fast query count

leonardr:
fast_query_count crashes when asked to calculate the size of a query that includes a DISTINCT statement applied to a subset of the columns. This branch stops the crash and adds some tests but as yet Courteney and I have not been able to find a way to always make fast_query_count return the correct values.

As a stopgap measure I've changed fast_query_count so that if it detects the kind of query that's going to cause problems, it falls back to the slower .count() behavior.

### DIFF
--- a/tests/test_opds.py
+++ b/tests/test_opds.py
@@ -762,9 +762,10 @@ class TestOPDS(DatabaseTest):
         pagination = Pagination(size=1)
 
         def make_page(pagination):
+            SessionManager.refresh_materialized_views(self._db)
             return AcquisitionFeed.page(
                 self._db, "test", self._url, fantasy_lane, TestAnnotator, 
-                pagination=pagination, use_materialized_works=False
+                pagination=pagination, use_materialized_works=True
             )
         cached_works = make_page(pagination)
         parsed = feedparser.parse(unicode(cached_works.content))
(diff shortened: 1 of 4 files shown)


util/__init__.py
```diff
@@ -22,8 +23,20 @@ def batch(iterable, size=1):
...
+    if isinstance(distinct_columns, list):
+        # TODO: We currently can't create a reliable count for queries
+        # that include specific distinct columns.  Just calculate the
+        # count the normal way.
+        return query.count()
+
+        # We didn't need GROUP BY when we were selecting rows made
+        # distinct by a list of columns, but now that we're selecting
+        # an aggregate function we do need to GROUP BY those columns.
+        # count_q = count_q.group_by(*distinct_columns)
```
courte: Did you want to keep this comment of our proposed changes?

util/__init__.py
```diff
@@ -10,6 +10,7 @@
 import re
 import string
 from sqlalchemy.sql.functions import func
+from sqlalchemy import distinct
```
courte: This isn't being used anymore.

tests/test_opds.py
```diff
@@ -762,9 +762,10 @@ def test_page_feed(self):
         pagination = Pagination(size=1)
 
         def make_page(pagination):
+            SessionManager.refresh_materialized_views(self._db)
```
leonardr: I made these changes while diagnosing the problem--I thought it was a problem with the materialized view and that making more code use the materialized view would flush it out. Since we use the materialized view everywhere in the real app I decided it was worthwhile to keep this part of the code.

courte:
:+1: once the tests pass

aslagle:
:+1:

util/__init__.py
leonardr: Yes, I want you to be able to use this as a starting point.

tests/test_opds.py
leonardr: ...and this was a mistake because it broke the tests. I think because the materialized view doesn't get wiped at the end of a test? I'm not sure, but I reverted the code.

PR #292, opened 2016-06-07
More robust merges

leonardr:
This branch improves the robustness of Work.open_access_for_permanent_work_id. It correctly unifies works even if they both contain LicensePools for totally unrelated works.

There's one case I couldn't fix, in which Work A has a LicensePool that belongs in Work B, and Work B has a LicensePool that belongs in Work A. I wrote a test to verify that this case results in an exception, rather than a (much worse) infinite loop. Because fixing this is a lot of work and this particular use case is incredibly unlikely, I'm not addressing it now. If you have suggestions about how to address this I'd be interested in hearing them.

I also made a very minor, unrelated bugfix to the CoverageProvider constructor, and added a tiny test for it.

### DIFF
--- a/model.py
+++ b/model.py
@@ -3138,8 +3138,19 @@ class Work(Base):
                 # earlier.  (This is why we chose the work with the
                 # most LicensePools--it minimizes the disruption
                 # here.)
+
+                # First, make sure this Work is the exclusive
+                # open-access work for its permanent work ID. 
+                # Otherwise the merge may fail.
+                work.make_exclusive_open_access_for_permanent_work_id(pwid)
                 for needs_merge in licensepools_for_work.keys():
                     if needs_merge != work:
+
+                        # Make sure that Work we're about to merge has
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -3208,14 +3224,15 @@ def merge_into(self, other_work):
...
                         )
 
         my_pwids = self.pwids
         other_pwids = other_work.pwids
-        if not my_pwids.issubset(other_pwids):
-            difference = my_pwids.difference(other_pwids)
+        if not my_pwids == other_pwids:
+            difference = my_pwids.symmetric_difference(other_pwids)
             raise ValueError(
                 "Refusing to merge %r into %r because it has permanent work IDs not present in the target work: %s" % (
```
aslagle: maybe update this message

aslagle:
:+1:

model.py
leonardr: I was wrong about how symmetric_difference works; I made the error message more general.

PR #293, opened 2016-06-08
Move rights status to license pool delivery mechanism

### LINKED ISSUE #275: Each LicensePoolDeliveryMechanism can have its own rights status
Currently we track the rights status of a LicensePool with LicensePool.rights_status. This works most of the time, but it falls apart when we consider unglue.it, an aggregator which provides multiple editions of books available under different licenses. Our own content server might eventually look like this to a circulation manager, since we also carry multiple editions of books.

Basically, we need to move RightsStatus so that it applies to the delivery mechanism, not to the license pool.

We might also need to consider a case in which a single data source provides commercial and open-access versions of the same book. (This might happen in a consortium upstream.)

We might want to keep LicensePool.rights_status, but I'm not sure what it would be. Maybe it would be the RightsStatus associated with its open_access_download_url.

aslagle:
Fixes #275

### DIFF
--- a/model.py
+++ b/model.py
@@ -4024,6 +4024,21 @@ class LicensePoolDeliveryMechanism(Base):
     # One LicensePoolDeliveryMechanism may fulfill many Loans.
     fulfills = relationship("Loan", backref="fulfillment")
 
+    # One LicensePoolDeliveryMechanism may be associated with one RightsStatus.
+    rightsstatus_id = Column(
+        Integer, ForeignKey('rightsstatus.id'), index=True)
+
+
+    def set_rights_status(self, uri, name=None):
+        _db = Session.object_session(self)
+        status, ignore = get_one_or_create(
+            _db, RightsStatus, uri=uri,
+            create_method_kwargs=dict(name=name))
+        self.rights_status = status
+        if status.uri in RightsStatus.OPEN_ACCESS:
+            self.license_pool.open_access = True
+        return status
+
     def __repr__(self):
         return "%r %r" % (self.license_pool, self.delivery_mechanism)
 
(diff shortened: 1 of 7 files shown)


model.py
```diff
@@ -4178,6 +4178,21 @@ class LicensePoolDeliveryMechanism(Base):
...
+
+
+    def set_rights_status(self, uri, name=None):
+        _db = Session.object_session(self)
+        status, ignore = get_one_or_create(
+            _db, RightsStatus, uri=uri,
+            create_method_kwargs=dict(name=name))
+        self.rights_status = status
+        if status.uri in RightsStatus.OPEN_ACCESS:
+            self.license_pool.open_access = True
```
leonardr: Presumably we should also set open_access to False if we just destroyed the last remaining open-access delivery mechanism, but I'm not too worried about that case.

tests/test_model.py
```diff
@@ -1227,6 +1210,33 @@ def test_set_presentation_edition(self):
...
+        assert status != status3
+        eq_(uri2, status3.uri)
+
+        open_access_uri = RightsStatus.GENERIC_OPEN_ACCESS
+        open_access_status = lpdm.set_rights_status(open_access_uri)
+        eq_(open_access_uri, open_access_status.uri)
+        eq_(True, pool.open_access)
+
+        non_open_access_status = lpdm.set_rights_status(uri)
+        eq_(True, pool.open_access)
```
leonardr: It looks like you're verifying the case I don't like--that changing a delivery mechanism's rights from open access to non-open access leaves the pool's open-access status alone. Is this right?

aslagle: My thinking was if you set it to something that's not open access, it shouldn't change, because there might still be other open access delivery mechanisms. But you're right, it would be better to actually check if there are.

leonardr: OK, make that change and then :+1:

model.py
aslagle: fixed

leonardr:
:+1:

PR #294, opened 2016-06-08
Fix bibliographic coverage

leonardr:
This branch makes the 3MBibliographicCoverageProvider testable, and in doing so fixes some minor problems introduced by the big CirculationData branch.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -364,6 +364,11 @@ class CoverageProvider(object):
         CoverageFailure (if not).
         """
 
+        if not metadata and not circulationdata:
+            e = "Received neither metadata nor circulation data from input source"
+            return CoverageFailure(self, identifier, e, transient=True)
+
+
         if metadata:
             result = self._set_metadata(identifier, metadata, metadata_replacement_policy)
             if isinstance(result, CoverageFailure):
--- a/coverage.py
+++ b/coverage.py
@@ -517,6 +517,18 @@ class BibliographicCoverageProvider(CoverageProvider):
             cutoff_time=cutoff_time
         )
 
-    def process_batch(self):
+    def process_batch(self, identifiers):
         """Returns a list of successful identifiers and CoverageFailures"""
-        raise NotImplementedError
+        results = []
+        for identifier in identifiers:
+            result = self.process_item(identifier)
+            if not isinstance(result, CoverageFailure):
+                self.handle_success(identifier)
+            results.append(result)
+        return results
+
+    def handle_success(self, identifier):
+        self.set_presentation_ready(identifier)
+
+    def process_item(self, identifier):
+        raise NotImplementedError()
--- a/coverage.py
+++ b/coverage.py
@@ -341,6 +341,12 @@ class CoverageProvider(object):
         return work
 
 
+    def set_metadata(self, identifier, metadata, 
+                     metadata_replacement_policy=None):
+        return self.set_metadata_and_circulation_data(
+            identifier, metadata, None, metadata_replacement_policy,
+        )
+
     def set_metadata_and_circulation_data(self, identifier, metadata, circulationdata, 
         metadata_replacement_policy=None, 
         circulationdata_replacement_policy=None, 
@@ -358,13 +364,15 @@ class CoverageProvider(object):
         CoverageFailure (if not).
         """
 
-        result = self._set_metadata(identifier, metadata, metadata_replacement_policy)
-        if isinstance(result, CoverageFailure):
-            return result
-        
-        result = self._set_circulationdata(identifier, circulationdata, circulationdata_replacement_policy)
-        if isinstance(result, CoverageFailure):
-            return result
+        if metadata:
+            result = self._set_metadata(identifier, metadata, metadata_replacement_policy)
+            if isinstance(result, CoverageFailure):
+                return result
+
+        if circulationdata:
+            result = self._set_circulationdata(identifier, circulationdata, circulationdata_replacement_policy)
+            if isinstance(result, CoverageFailure):
+                return result
 
         # now that made sure that have an edition and a pool on the identifier, 
         # can try to make work
--- a/coverage.py
+++ b/coverage.py
@@ -221,7 +221,7 @@ class CoverageProvider(object):
         num_ignored = max(0, batch_size - len(results))
 
         self.log.info(
-            "Batch processed with %d successes, %d transient failures, %d ignored, %d persistent failures.",
+            "Batch processed with %d successes, %d transient failures, %d persistent failures, %d ignored.",
             successes, transient_failures, persistent_failures, num_ignored
         )
 
--- a/model.py
+++ b/model.py
@@ -5209,6 +5209,15 @@ class LicensePool(Base):
         if was_new and not license_pool.availability_time:
             now = datetime.datetime.utcnow()
             license_pool.availability_time = now
+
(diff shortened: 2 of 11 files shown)


coverage.py
```diff
@@ -341,6 +341,12 @@ def work(self, identifier):
         return work
 
 
+    def set_metadata(self, identifier, metadata,
```
leonardr: I brought this method back because it's very popular. Many metadata sources only get Metadata, or they incorporate the CirculationData as metadata.circulation.

coverage.py
```diff
@@ -358,13 +364,15 @@ def set_metadata_and_circulation_data(self, identifier, metadata, circulationdat
...
-        
-        result = self._set_circulationdata(identifier, circulationdata, circulationdata_replacement_policy)
-        if isinstance(result, CoverageFailure):
-            return result
+        if metadata:
+            result = self._set_metadata(identifier, metadata, metadata_replacement_policy)
+            if isinstance(result, CoverageFailure):
+                return result
+
+        if circulationdata:
```
leonardr: Calling set_circulationdata when no CirculationData was provided was causing an error, which resulted in an incorrect CoverageFailure.

coverage.py
```diff
@@ -509,6 +517,18 @@ def __init__(self, _db, api, datasource, workset_size=10,
             cutoff_time=cutoff_time
         )
 
-    def process_batch(self):
+    def process_batch(self, identifiers):
```
leonardr: This is the refactoring. I wanted to make sure that set_presentation_ready() is always called, so I implemented the most common version of process_batch(), moved the NotImplementedError to process_item(), and put the set_presentation_ready() call in handle_success (which can itself be overridden if you want to do something extra or _don't_ want to call set_presentation_ready()).

model.py
```diff
@@ -5209,6 +5209,15 @@ def for_foreign_id(self, _db, data_source, foreign_id_type, foreign_id, rights_s
...
             license_pool.availability_time = now
+
+        if was_new:
+            # Set the LicensePool's initial values to indicate
+            # that we don't actually know how many copies we own.
+            license_pool.licenses_owned = 0
+            license_pool.licenses_available = 0
+            license_pool.licenses_reserved = 0
+            license_pool.patrons_in_hold_queue = 0
+
```
leonardr: Following a discussion with Amy and Darya we decided that it's important to have LicensePools even when we don't know how many licenses we own, because LicensePools also contain format information. Rather than risk the existence of LicensePools whose circulation information is None, we start with the conservative estimate that we know about the book but own no copies.

model.py
```diff
@@ -5482,14 +5490,25 @@ def update_availability(
...
         # Update the license pool with the latest information.
-        self.licenses_owned = new_licenses_owned
-        self.licenses_available = new_licenses_available
-        self.licenses_reserved = new_licenses_reserved
-        self.patrons_in_hold_queue = new_patrons_in_hold_queue
-        self.last_checked = as_of
+        any_data = False
+        if new_licenses_owned is not None:
+            self.licenses_owned = new_licenses_owned
+            any_data = True
```
leonardr: Unfortunately setting these values to 0 upon creation of the LicensePool was not sufficient. If you have a CirculationData object that has format information but no licensing information, calling update_availability() was replacing the zero values with null values. This puts a stop to that. If the CirculationData doesn't know something, we leave the preexisting value in place.

overdrive.py
```diff
@@ -871,26 +871,28 @@ def __init__(self, _db, input_identifier_types=None,
...
-
     def process_item(self, identifier):
         info = self.api.metadata_lookup(identifier)
+        error = None
         if info.get('errorCode') == 'NotFound':
-            e = "ID not recognized by Overdrive"
-            return CoverageFailure(self, identifier, e, transient=False)
+            error = "ID not recognized by Overdrive: %s" % identifier.identifier
+        elif info.get('errorCode') == 'InvalidGuid':
+            error = "Invalid Overdrive ID: %s" % identifier.identifier
```
leonardr: This is just to turn an annoying Overdrive error from a transient failure to a permanent failure.

threem.py
```diff
@@ -426,16 +426,11 @@ def __init__(self, _db, input_identifier_types=None,
             workset_size=25, metadata_replacement_policy=metadata_replacement_policy, **kwargs
         )
 
-    def process_batch(self, identifiers):
```
leonardr: Don't need this method anymore since we use the refactored version.

threem.py
```diff
@@ -426,16 +426,11 @@ def __init__(self, _db, input_identifier_types=None,
...
-        batch_results = []
-        for identifier in identifiers:
-            metadata = self.api.bibliographic_lookup(identifier)
-            result = self.set_metadata(identifier, metadata)
-            if not isinstance(result, CoverageFailure):
-                # Success!
-                result = self.set_presentation_ready(result)
```
leonardr: Don't need to explicitly call set_presentation_ready() anymore since that happens in handle_success().

aslagle:
there's a failing test, but otherwise looks good

coverage.py
daryanypl: OK with me.

coverage.py
daryanypl: self._set_metadata, not the returned self.set_metadata?

model.py
```diff
@@ -5504,14 +5512,25 @@ def update_availability(
...
+        if new_patrons_in_hold_queue is not None:
+            self.patrons_in_hold_queue = new_patrons_in_hold_queue
+            any_data = True
+
+        if changes_made or any_data:
+            self.last_checked = as_of
 
         # Update the last update time of the Work.
-        if self.work:
+        if self.work and (any_data or changes_made):
```
daryanypl: minor quibble: checking any_data and changes_made twice.  would it be detrimental to set last_checked outside of a conditional?

tests/test_3m.py
```diff
@@ -190,3 +193,40 @@ def test_item_list(cls):
 
         eq_(Hyperlink.DESCRIPTION, description.rel)
         assert description.content.startswith("<b>Winner")
+
+
+class TestBibliographicCoverageProvider(TestThreeMAPI):
+
+    def test_process_item_creates_presentation_ready_work(self):
+        data = self.get_data("item_metadata_single.xml")
```
daryanypl: could I have a small, one-liner doc string on the class and method?

tests/test_model.py
```diff
@@ -955,6 +959,36 @@ def test_update_availability_triggers_analytics(self):
...
+        # Set up a LicensePool.
+        [pool] = work.license_pools
+        pool.last_checked = None
+        pool.licenses_owned = 10
+        pool.licenses_available = 20
+        pool.licenses_reserved = 30
+        pool.patrons_in_hold_queue = 40
+
+        # Pass empty values into update_availability.
+        pool.update_availability(None, None, None, None)
```
daryanypl: what should happen if just one is None?

tests/test_overdrive.py
```diff
@@ -294,3 +299,35 @@ def test_book_info_with_awards(self):
         ]
         eq_(1, awards.value)
         eq_(1, awards.weight)
+
+
+class TestOverdriveBibliographicCoverageProvider(DatabaseTest):
+
```
daryanypl: I'd love to have a quick doc string.

PR #295, opened 2016-06-08
More coverage provider tests

leonardr:
This branch adds basic tests for the Overdrive and Axis 360 bibliographic coverage providers. There were no real surprises but it's good to have the coverage.


overdrive.py
```diff
@@ -890,12 +890,8 @@ def process_item(self, identifier):
...
-        result = self.set_metadata(
+        return self.set_metadata(
             identifier, metadata, 
             metadata_replacement_policy=self.metadata_replacement_policy
         )
 
-        if not isinstance(result, CoverageFailure):
-            # Success!
-            result = self.set_presentation_ready(result)
-        return result
```
leonardr: This is no longer necessary since the superclass process_batch() takes care of setting the work as presentation-ready.

aslagle:
:+1:

PR #296, opened 2016-06-08
Add series lanes

courte:
A few small changes to support work on NYPL-Simplified/circulation#245 and a grouped recommendation feed for works.

This also adds the migration for CachedFeed.license_pool_id that I forgot in my last bit of work on LicensePoolBasedLanes.


opds.py
```diff
@@ -506,6 +506,10 @@ def groups(cls, _db, title, url, lane, annotator,
             return cached
 
         feed_size = Configuration.featured_lane_size()
+        if ignore_feed_size:
+            # This groups feed should return groups no matter how many
+            # books are found.
+            feed_size = 1
```
leonardr: I don't understand how this solves the problem. It seems like you'd want to check this around line 541?

model.py
```diff
@@ -4990,6 +4990,7 @@ class CachedFeed(Base):
     GROUPS_TYPE = 'groups'
     PAGE_TYPE = 'page'
     RECOMMENDATIONS_TYPE = 'recommendations'
+    SERIES_TYPE = 'series'
```
leonardr: If you're going to make the recommendation feed a grouped feed, I don't think you need a separate SERIES_TYPE. The entire feed--recommendations, same author, and same series--would be cached as RECOMMENDATIONS_TYPE.

opds.py
courte: We're chatting about this elsewhere, but note to self: the way that `feed_size` is being used here actually limits the number of featured works that are shown to 1, in addition to allowing them to be shown in groups at all. Definitely better to address this down below. Thanks, @leonardr.

lane.py
```diff
@@ -1201,8 +1197,22 @@ def _defer_unused_opds_entry(self, query, work_model=Work):
...
+            )
+            for work in works:
+                works_and_lanes.append((work, sublane))
+        return works_and_lanes
+
+    def featured_works(self, use_materialized_works=True):
+        """Find a random sample of featured books.
 
         It's semi-okay for this to be slow, since it will only be run to
         create cached feeds.
```
leonardr: This is no longer true. It's true for the default implementation but not necessarily true for the subclass implementation.

model.py
leonardr: However, the grouped feed will link to individual page-type feeds, and those feeds will need different types, so leave this here.

leonardr:
:+1:

PR #298, opened 2016-06-09
Move works in and out of the search index as they become presentation-ready or stop being presentation-ready

leonardr:
Without this code, a work that stops being presentation ready will stay in the index. More importantly, there are cases where a work will become presentation-ready without any of its metadata changing, and when that happens it will not be added to the search index.


aslagle:
:+1:

PR #299, opened 2016-06-10
Make WorkCoverageProvider work on similar terms to CoverageProvider

leonardr:
This branch refactors the CoverageProvider class into BaseCoverageProvider, with two subclasses, CoverageProvider (which takes Identifiers or Editions, runs an algorithm, and produces CoverageRecords) and WorkCoverageProvider (which takes Works, runs an algorithm, and produces WorkCoverageRecords). In both cases, most of the time you only have to implement a method called `process_item()` which handles one item. CoverageProvider/WorkCoverageProvider takes care of splitting the work into batches, handling the results, and interfacing with IdentifierInputScript.

The first WorkCoverageProvider is coming in circulation--it's a provider that updates the search index. Because of this I've also made some minor changes to the code surrounding updating the search index.

### DIFF
--- a/axis.py
+++ b/axis.py
@@ -237,7 +237,7 @@ class Axis360BibliographicCoverageProvider(BibliographicCoverageProvider):
         axis_360_api = axis_360_api or Axis360API(_db)
         super(Axis360BibliographicCoverageProvider, self).__init__(
             _db, axis_360_api, DataSource.AXIS_360,
-            workset_size=25, 
+            batch_size=25, 
             metadata_replacement_policy=metadata_replacement_policy,
             **kwargs
         )
--- a/coverage.py
+++ b/coverage.py
@@ -325,10 +325,10 @@ class CoverageProvider(BaseCoverageProvider):
 
     def __init__(self, service_name, input_identifier_types, output_source,
                  batch_size=100, cutoff_time=None, operation=None):
+        _db = Session.object_session(output_source)
         super(CoverageProvider, self).__init__(
-            service_name, operation, batch_size, cutoff_time
+            _db, service_name, operation, batch_size, cutoff_time
         )
-        self._db = Session.object_session(output_source)
         if input_identifier_types and not isinstance(input_identifier_types, list):
             input_identifier_types = [input_identifier_types]
         self.input_identifier_types = input_identifier_types
--- a/coverage.py
+++ b/coverage.py
@@ -94,7 +94,7 @@ class BaseCoverageProvider(object):
 
     def run(self):
         self.log.info("%d items need coverage.", (
-            self.items_that_need_coverage.count())
+            self.items_that_need_coverage().count())
         )
         offset = 0
         while offset is not None:
@@ -106,9 +106,11 @@ class BaseCoverageProvider(object):
         self._db.commit()
         return offset
 
-    def run_on_specific_items(self, items):
-        """Split a specific set of items into batches and process one batch at
-        a time.
+    def run_on_specific_identifiers(self, identifiers):
+        """Split a specific set of Identifiers into batches and process one
+        batch at a time.
+
+        This is for use by IdentifierInputScript.
 
         :return: The same (counts, records) 2-tuple as
             process_batch_and_handle_results.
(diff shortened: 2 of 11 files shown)


axis.py
```diff
@@ -237,7 +237,7 @@ def __init__(self, _db, input_identifier_types=None,
         axis_360_api = axis_360_api or Axis360API(_db)
         super(Axis360BibliographicCoverageProvider, self).__init__(
             _db, axis_360_api, DataSource.AXIS_360,
-            workset_size=25, 
+            batch_size=25,
```
leonardr: This requires a bunch of changes but I think it has to happen. Every other piece of code refers to this as a 'batch', not a 'workset'.

coverage.py
```diff
@@ -26,11 +28,16 @@ class CoverageFailure(object):
 
     def __init__(self, provider, obj, exception, transient=True):
         self.obj = obj
-        self.output_source = provider.output_source
+        self.output_source = getattr(provider, 'output_source', None)
```
leonardr: WorkCoverageProvider doesn't have an output_source; this lets CoverageFailure work with both CoverageProvider and WorkCoverageProvider.

coverage.py
```diff
@@ -262,14 +262,88 @@ def should_update(self, coverage_record):
...
+        
+        e.g. uploading a bunch of assets to S3.
+        """
+        pass
+
+    #
+    # Subclasses must implement these virtual methods.
+    #
+
+    def items_that_need_coverage(self, identifiers):
```
leonardr: This used to be a property, and any identifier restriction was tacked on afterwards. But the CoverageProvider query is a query directly on Identifier, so it can filter Identifier.id, whereas the WorkCoverageProvider query gets to Identifier via Work and LicensePool, so it needs to filter on LicensePool.identifier_id. The simplest way to implement this difference was in the subclasses.

coverage.py
```diff
@@ -262,14 +262,88 @@ def should_update(self, coverage_record):
...
+        """Do the work necessary to give coverage to one specific item.
+
+        Since this is where the actual work happens, this is not
+        implemented in CoverageProvider or WorkCoverageProvider, and
+        must be handled in a subclass.
+        """
+        raise NotImplementedError()
+
+
+class CoverageProvider(BaseCoverageProvider):
```
leonardr: Pretty much all of this is code moved from what is now BaseCoverageProvider.

coverage.py
```diff
@@ -469,21 +555,76 @@ def set_presentation_ready(self, identifier):
...
+
+        :param: Only Works connected with one of the given identifiers
+        are chosen.
         """
-        pass
+        qu = Work.missing_coverage_from(
+            self._db, operation=self.operation, 
+            count_as_missing_before=self.cutoff_time
+        )
+        if identifiers:
```
leonardr: If no identifiers are provided we don't need to do the join against LicensePool at all.

model.py
```diff
@@ -3626,7 +3645,7 @@ def calculate_opds_entries(self, verbose=True):
         )
 
 
-    def update_external_index(self, client):
+    def update_external_index(self, client, add_coverage_record=True):
```
leonardr: The SearchIndexCoverageProvider (found in circulation) is going to add a coverage record for every book it indexes, so there's no need to also do the work here.

scripts.py
```diff
@@ -153,20 +153,38 @@ class IdentifierInputScript(Script):
     """A script that takes identifiers as command line inputs."""
 
     @classmethod
-    def parse_command_line(cls, _db=None, cmd_args=None, *args, **kwargs):
+    def read_stdin_lines(self, stdin):
```
leonardr: I had about 6000 identifiers that needed reindexing, way too many to put on the command line, so I changed IdentifierInputScript so that you can pass those identifiers in on standard input. This was tricky because a) it's impossible to test the 'real' standard input, and b) if you run the script without providing anything on standard input, I don't want the script to hang waiting for your input.

This method is not tested but I have tested it manually.

scripts.py
```diff
@@ -153,20 +153,38 @@ class IdentifierInputScript(Script):
...
+        """Read lines from a (possibly mocked, possibly empty) standard input."""
+        if stdin is not sys.stdin or not os.isatty(0):
+            # A file has been redirected into standard input. Grab its
```
leonardr: The os.isatty(0) part of this can't be automatically tested but I have tested it manually.

aslagle:
:+1:

PR #300, opened 2016-06-10
OPDS import refactoring

aslagle:
This branch refactors the OPDS import process to be more robust. 

The monitor fetches next links until it reaches a page with nothing new, and then imports starting from the last page. If something fails, it start following the next links from the beginning again, and start importing from where it left off.


metadata_layer.py
```diff
@@ -606,14 +606,22 @@ def links(self, arg_links):
...
+                            format_found = True
+                            break
+                    if not format_found:
+                        self.formats.append(
+                            FormatData(
+                                content_type=link.media_type,
+                                drm_scheme=DeliveryMechanism.NO_DRM,
+                                link=link,
+                                rights_uri=rights_uri,
+                            )
```
aslagle: There are some places in the content server where we explicitly create a format for an open access link. If a format for the link was already passed in, we don't need to create a new one.

opds_import.py
```diff
@@ -356,86 +354,40 @@ def import_edition_from_metadata(
...
+        ]
 
 
     def extract_feed_data(self, feed):
         """Turn an OPDS feed into lists of Metadata and CirculationData objects, 
         with associated messages and next_links.
         """
-        fp_data_meta, fp_data_circ, status_messages, next_links = self.extract_data_from_feedparser(feed=feed, data_source=self.data_source_name)
+        data_source = DataSource.lookup(self._db, self.data_source_name)
+        fp_metadata, fp_status_messages = self.extract_data_from_feedparser(feed=feed, data_source=data_source)
```
aslagle: This now checks if the data source offers licenses, and doesn't add circulation data to the metadata if not.

opds_import.py
```diff
@@ -356,86 +354,40 @@ def import_edition_from_metadata(
...
         # gets: medium, measurements, links, contributors, etc.
-        xml_data_meta = self.extract_metadata_from_elementtree(feed)
+        xml_data_meta, xml_status_messages = self.extract_metadata_from_elementtree(feed)
```
aslagle: Both of these methods catch any exceptions parsing individual entries and convert them to status messages. This way, unexpected problems with third-party OPDS data will be recorded but won't break the whole import. If we're unable to even extract the entries, we throw an exception and stop the import, since we can't record the identifiers that failed.

opds_import.py
```diff
@@ -513,34 +462,23 @@ def combine(self, d1, d2):
...
-        feed = feedparser_parsed['feed']
-        next_links = []
-        if feed and 'links' in feed:
-            next_links = [
-                link['href'] for link in feed['links'] 
-                if link['rel'] == 'next'
-            ]
-
-        return values_meta, values_circ, status_messages, next_links
+        return values, status_messages
```
aslagle: Next links are pulled out in a separate step, before import, so they're not needed here anymore.

opds_import.py
```diff
@@ -925,50 +879,143 @@ def __init__(self, _db, feed_url, default_data_source, import_class,
...
-    def _get(self, url):
+    def _get(self, url, headers):
         """Make the sort of HTTP request that's normal for an OPDS feed.
 
         Long timeout, raise error on anything but 2xx or 3xx.
         """
         kwargs = dict(timeout=120, allowed_response_codes=['2xx', '3xx'])
-        return HTTP.get_with_timeout(url, **kwargs)
+        response = HTTP.get_with_timeout(url, headers=headers, **kwargs)
+        return response.status_code, response.content_type, response.content
```
aslagle: We only need the content, but this matches the DummyHTTPClient so it's easier to test.

opds_import.py
```diff
@@ -925,50 +879,143 @@ def __init__(self, _db, feed_url, default_data_source, import_class,
...
+
+            # There's no record of a failure for this book. Has it been imported from this source already?
+            existing_edition = get_one(
+                self._db, Edition,
+                primary_identifier=identifier,
+                data_source=data_source,
+            )
+            if not existing_edition:
+                new_data = True
+                break
```
aslagle: Checking if there's a record would be enough, except that we don't have records for anything that was imported before this and I don't want to have to reimport them all. I think we'd be able to simplify this a lot if we could write a migration to create the missing CoverageRecords.

opds_import.py
leonardr: :heart_eyes_cat:

opds_import.py
```diff
@@ -593,89 +547,84 @@ def data_detail_for_feedparser_entry(cls, entry, data_source):
...
-
-            content = detail['value']
-            media_type = detail.get('type', 'text/plain')
-            return LinkData(
-                rel=Hyperlink.DESCRIPTION,
-                media_type=media_type,
-                content=content
-            )
+        try:
+            title = entry.get('title', None)
```
leonardr: Again, a big try/catch block is a great opportunity to create a testable helper method.

opds_import.py
```diff
@@ -691,46 +640,51 @@ def detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None):
...
-        data = dict()
-
-        alternate_identifiers = []
-        for id_tag in parser._xpath(entry_tag, "dcterms:identifier"):
-            v = cls.extract_identifier(id_tag)
-            if v:
-                alternate_identifiers.append(v)
-        data['identifiers'] = alternate_identifiers
+        try:
+            # We will fill this dictionary with all the information
```
leonardr: And again here. Although actually writing those tests is likely to be a pain because of the etree code.

opds_import.py
leonardr: I think you're going to be reimporting them all anyway, because those old Editions will have their original DataSources, and the new ones will all be associated with the content server, right? Because we did those imports before we had the presentation-edition code. So I would expect existing_edition to always fail on our existing production site.

aslagle: Yeah, I guess you're right. It would work when the content server is importing from unglue.it, but maybe that's the only case.

opds_import.py
```diff
@@ -925,50 +879,143 @@ def __init__(self, _db, feed_url, default_data_source, import_class,
...
+            CoverageRecord.add_for(
+                edition, data_source, CoverageRecord.IMPORT_OPERATION
             )
-            return []
-        else:
-            return next_links
 
+        # Create CoverageRecords for the failures.
+        for urn, message in messages.items():
+            if message.status_code == 500:
```
leonardr: I could be wrong but I think you're not distinguishing between the local 500 errors which get turned into StatusMessages when there's an exception, and 500 errors from the remote side. Local 500 errors are permanent because they require a change to the code, but remote 500 errors should be treated as transient, because they could be fixed at any time.

What if `messages` was allowed to contain `CoverageFailure` objects as well as `StatusMessage` objects? This would let you differentiate between local and remote errors.

tests/test_opds_import.py
```diff
@@ -616,16 +645,16 @@ def test_resources_are_mirrored_on_import(self):
...
             200, content='I am 10441.epub.images',
             media_type=Representation.EPUB_MEDIA_TYPE
         )
         http.queue_response(
             200, content=svg, media_type=Representation.SVG_MEDIA_TYPE
         )
+        http.queue_response(
+            200, content='I am 10557.epub.images',
+            media_type=Representation.EPUB_MEDIA_TYPE,
+        )
```
leonardr: I don't understand how your branch changed it so 10557 gets mirrored last. It looks like you switched from mirroring all the circulation resources, then mirroring all the other resources, to mirroring one book at a time. I like this change but I can't find the specific bit where you changed it. Can you just point it out?

aslagle: Actually, I have this wrong. Things get popped from the queue in reverse, so 10441 and 10557 are flipped here, but the test doesn't check the content, only the media type. What actually changed is that the epub for 10441 comes before the cover now. This is due to putting the circulation data inside the metadata, so the epub gets mirrored as part of metadata.apply instead of afterwards. metadata.apply does the circulation data first so when it processes links it can associate them with the pool.

leonardr: OK at least we were both confused... :bird: Just verify the content and we'll be fine here.

opds_import.py
```diff
@@ -268,55 +275,46 @@ def import_from_feed(self, feed, even_if_no_author=False,
...
-                except Exception, e:
-                    message = StatusMessage(500, "Local exception during import:\n%s" % traceback.format_exc())
-                    status_messages[key] = message
-            else:
-                self.log.debug(
-                    "DataSource does not offer licenses.  No LicensePool created for CirculationData %r, not attempting to create Work.", 
-                    circulationdata
-                )
-
+            try:
```
aslagle: I moved these to helper methods, and wrote tests that override the helper methods to throw exceptions. I didn't write any additional tests for the helper methods themselves, are there any you think we need?

opds_import.py
aslagle: What if it only contained `CoverageFailure` objects, and transient was set appropriately? I think the `StatusMessage` objects are only used to create `CoverageFailure`s.

leonardr: Sounds great!

opds_import.py
```diff
@@ -266,57 +247,31 @@ def import_from_feed(self, feed, even_if_no_author=False,
...
                 # to this item.
-                message = StatusMessage(500, "Local exception during import:\n%s" % traceback.format_exc())
-                status_messages[key] = message
-                # go into circulation_obs, and remove any on the same identifier, 
-                # so we don't try to make pools and works from it
-                if key in circulation_objs:
-                    del circulation_objs[key]
+                identifier, ignore = Identifier.parse_urn(self._db, key)
+                data_source = DataSource.lookup(self._db, self.data_source_name)
+                failure = CoverageFailure(identifier, traceback.format_exc(), data_source=data_source, transient=False)
```
aslagle: I've replaced StatusMessage with CoverageFailure, but CoverageFailure didn't have status codes, so I took them out. Would it be better to add an optional status code to CoverageFailure, even though it's not in CoverageRecord?

aslagle: The OPDSImportCoverageProvider in circulation is using status code : message so maybe I'll do that, and remove it from circulation.

aslagle: I don't think I need the 500's for local failures, but I'll keep the status codes from the lookup protocol.

leonardr: Making the status code part of the message sounds good to me.

tests/test_opds_import.py
```diff
@@ -561,33 +609,48 @@ def test_import_and_make_presentation_ready(self):
...
 
         # Only one book was imported, the other failed.
         eq_(1, len(imported_editions))
 
-        # The other failed to import, and became a StatusMessage
-        message = error_messages['http://www.gutenberg.org/ebooks/10441']
-        eq_(500, message.status_code)
-        assert "Utter failure!" in message.message
+        # The other failed to import, and became a CoverageFailure
+        failure = failures['http://www.gutenberg.org/ebooks/10441']
```
leonardr: Do a quick check to verify that this is a CoverageFailure object. Just checking `.exception` leaves it a little ambiguous. Also check `.status`.

tests/test_opds_import.py
```diff
@@ -561,33 +609,48 @@ def test_import_and_make_presentation_ready(self):
...
+
+        imported_editions, pools, works, failures = (
+            importer.import_from_feed(feed)
+        )
+
+        # One work was created, the other failed.
+        eq_(1, len(works))
+
+        # There's an error message for the work that failed. 
+        failure = failures['http://www.gutenberg.org/ebooks/10441']
```
leonardr: Same here.

leonardr:
:+1: overall, just needs a couple small additions to the tests.

PR #301, opened 2016-06-14
Cover items with no coverage attempts first

leonardr:
This branch introduces the notion of storing transient failures in `CoverageRecord`s. Previously only successes and persistent failures were stored.

This branch also changes CoverageProvider so that `run()` runs in two steps. First, it tries to cover items that have no CoverageRecord at all. Then, it tries to cover items that already have CoverageRecords that ended in transient failure.

The goal is to fix https://github.com/NYPL-Simplified/circulation/issues/251, where a relatively small number of transient failures stacked up at the beginning of the work and were always processed first.

Finally, I fixed a bug I discovered which meant that a CoverageProvider could never _change_ `CoverageRecord.status` (e.g. from transient failure to success), it could only create a new `CoverageRecord`. We didn't notice this before because successful coverage records don't get retried, and we haven't yet tried to resolve any persistent failures.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -8,6 +8,7 @@ from sqlalchemy.sql.functions import func
 from model import (
     get_one,
     get_one_or_create,
+    BaseCoverageRecord,
     CoverageRecord,
     DataSource,
     Edition,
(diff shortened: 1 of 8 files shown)


coverage.py
```diff
@@ -100,18 +102,31 @@ def log(self):
...
+            offset = self.run_once(
+                offset, covered_statuses=[
+                    BaseCoverageRecord.SUCCESS, 
+                    BaseCoverageRecord.PERSISTENT_FAILURE
+                ]
+            )
+        
         Timestamp.stamp(self._db, self.service_name)
         self._db.commit()
-        return offset
```
leonardr: The concept of a singular 'offset' no longer makes sense since we are actually running two different sets of batches. Fortunately I don't think this was being used by run().

coverage.py
```diff
@@ -156,9 +171,16 @@ def run_on_specific_identifiers(self, identifiers):
             index += self.batch_size
         return (successes, transient_failures, persistent_failures), records
 
-    def run_once(self, offset):
-        batch = self.items_that_need_coverage().limit(
-            self.batch_size).offset(offset)
+    def run_once(self, offset, covered_statuses=None):
+        covered_statuses = covered_statuses or BaseCoverageRecord.DEFAULT_COUNTS_AS_COVERED
+        # Make it clear which class of items we're covering on this
+        # run.
```
aslagle: I understand what this is doing, but the comment sounds wrong to me - covered_statuses is the class of items we're _not_ covering on this run, because they are already considered covered. Maybe it would help to think of a less ambiguous variable name?

aslagle:
Looks good, but I think you should rename covered_statuses in the provider. In CoverageRecord it's clear that we're looking for things that aren't covered and saying what is, but it's confusing in the run methods.

aslagle:
:+1:

PR #302, opened 2016-06-14
Fix fast_query_count to work with distinct-specific queries without s…

### LINKED ISSUE #291: fast_query_count returns incorrect results in some cases
You can test this by uncommenting the premature `return query.count()` in fast_query_count() and running the tests found in test_util.py.

The main real case is when building a lane out of the items in one or more custom lists.

courte:
…ubquery.

Fixes #291.


leonardr:
:+1: :gem:

PR #304, opened 2016-06-14
Native language names

aslagle:
This branch changes LanguageCodes.name_for_languageset to use language names in the native language if possible. I couldn't find an official source of native names. Wikipedia has some, but they're not easy to extract. I went with a list I found on stackoverflow. 

I didn't capitalize the native names on this branch, and I'm not sure a) if language names should be capitalized when they're lane names and b) if that should happen here or when the lanes are created.


tests/test_util.py
```diff
@@ -33,20 +33,27 @@ def test_lookups(self):
...
         eq_(["Spanish", "Castilian"], c.english_names['spa'])
+        eq_([unicode("español", "utf-8"), "castellano"], c.native_names['es'])
+        eq_([unicode("español", "utf-8"), "castellano"], c.native_names['spa'])
 
         eq_("chi", c.two_to_three['zh'])
         eq_("zh", c.three_to_two['chi'])
         eq_(["Chinese"], c.english_names['zh'])
         eq_(["Chinese"], c.english_names['chi'])
+        eq_([unicode("äæ (Zhōngwén)", "utf-8"), unicode("æè", "utf-8"), unicode("æè", "utf-8")], c.native_names['zh'])
+        eq_([unicode("äæ (Zhōngwén)", "utf-8"), unicode("æè", "utf-8"), unicode("æè", "utf-8")], c.native_names['chi'])
```
leonardr: This really doesn't look right. I think there's an encoding problem here. It looks like you took the UTF-8 encoding of the Chinese characters and encoded it as Windows-1252. I can get something similar with 中文:

```
>>> print u"中文".encode("utf8").decode("windows-1252")
ä¸­æ–‡
```

At a glance, Japanese and Nuosu seem to have similar problems.

"Zhōngwén" looks like a Romanization of "中文" and I wouldn't expect to see it in the name of a lane.

There's also the problem that there are two forms of Chinese, Simplified and Traditional. I don't know which one we use in our collection, or whether we can use one or the other to stand in for Chinese generically. We need to check with a librarian about this before we deploy anything.

In fact I think we need to add this to our checklist when we ask people to do translations.

* What does the name of the language look like when used as a header?
* What does it look like as a header when paired with other languages? (e.g. "English & [your language]", "[your language] & English")

Also translate these sublane names:
* "Adult Fiction"
* "Adult Nonfiction"
* "Children & Young Adult"

This should solve the capitalization question as well.

util/__init__.py
```diff
@@ -537,6 +541,191 @@ class LanguageCodes(object):
 zxx|||No linguistic content; Not applicable|pas de contenu linguistique; non applicable
 zza|||Zaza; Dimili; Dimli; Kirdki; Kirmanjki; Zazaki|zaza; dimili; dimli; kirdki; kirmanjki; zazaki"""
 
+    NATIVE_NAMES_RAW_DATA =  [
+        {"code":"ab","name":"Abkhaz","nativeName":"аҧсуа"},
```
leonardr: Any time there's a string like this I would like to use `u''` to turn it into a Unicode string. I think that reduces the risk of encoding errors.

tests/test_util.py
aslagle: Maybe I should get rid of the full list and we can get individual translations for the languages we support.

util/__init__.py
```diff
@@ -537,6 +541,191 @@ class LanguageCodes(object):
...
+        {"code":"ht","name":"Haitian; Haitian Creole","nativeName":"Kreyòl ayisyen"},
+        {"code":"ha","name":"Hausa","nativeName":"Hausa, هَوُسَ"},
+        {"code":"he","name":"Hebrew (modern)","nativeName":"עברית"},
+        {"code":"hz","name":"Herero","nativeName":"Otjiherero"},
+        {"code":"hi","name":"Hindi","nativeName":"हिन्दी, हिंदी"},
+        {"code":"ho","name":"Hiri Motu","nativeName":"Hiri Motu"},
+        {"code":"hu","name":"Hungarian","nativeName":"Magyar"},
+        {"code":"ia","name":"Interlingua","nativeName":"Interlingua"},
+        {"code":"id","name":"Indonesian","nativeName":"Bahasa Indonesia"},
+        {"code":"ie","name":"Interlingue","nativeName":"Originally called Occidental; then Interlingue after WWII"},
```
leonardr: It's after WWII so we should just use "Interlingue".

util/__init__.py
```diff
@@ -537,6 +541,191 @@ class LanguageCodes(object):
...
+        {"code":"ga","name":"Irish","nativeName":"Gaeilge"},
+        {"code":"ig","name":"Igbo","nativeName":"Asụsụ Igbo"},
+        {"code":"ik","name":"Inupiaq","nativeName":"Iñupiaq, Iñupiatun"},
+        {"code":"io","name":"Ido","nativeName":"Ido"},
+        {"code":"is","name":"Icelandic","nativeName":"Íslenska"},
+        {"code":"it","name":"Italian","nativeName":"Italiano"},
+        {"code":"iu","name":"Inuktitut","nativeName":"ᐃᓄᒃᑎᑐᑦ"},
+        {"code":"ja","name":"Japanese","nativeName":"ææè (にほんご／にっぽんご)"},
```
leonardr: Encoding error + extra text.

util/__init__.py
```diff
@@ -537,6 +541,191 @@ class LanguageCodes(object):
...
+        {"code":"kr","name":"Kanuri","nativeName":"Kanuri"},
+        {"code":"ks","name":"Kashmiri","nativeName":"कश्मीरी, كشميري‎"},
+        {"code":"kk","name":"Kazakh","nativeName":"Қазақ тілі"},
+        {"code":"km","name":"Khmer","nativeName":"ភាសាខ្មែរ"},
+        {"code":"ki","name":"Kikuyu, Gikuyu","nativeName":"Gĩkũyũ"},
+        {"code":"rw","name":"Kinyarwanda","nativeName":"Ikinyarwanda"},
+        {"code":"ky","name":"Kirghiz, Kyrgyz","nativeName":"кыргыз тили"},
+        {"code":"kv","name":"Komi","nativeName":"коми кыв"},
+        {"code":"kg","name":"Kongo","nativeName":"KiKongo"},
+        {"code":"ko","name":"Korean","nativeName":"íêì (éåè), ììë (æéè)"},
```
leonardr: Encoding error, extra text.

util/__init__.py
```diff
@@ -537,6 +541,191 @@ class LanguageCodes(object):
...
+        {"code":"mn","name":"Mongolian","nativeName":"монгол"},
+        {"code":"na","name":"Nauru","nativeName":"Ekakairũ Naoero"},
+        {"code":"nv","name":"Navajo, Navaho","nativeName":"Diné bizaad, Dinékʼehǰí"},
+        {"code":"nb","name":"Norwegian Bokmål","nativeName":"Norsk bokmål"},
+        {"code":"nd","name":"North Ndebele","nativeName":"isiNdebele"},
+        {"code":"ne","name":"Nepali","nativeName":"नेपाली"},
+        {"code":"ng","name":"Ndonga","nativeName":"Owambo"},
+        {"code":"nn","name":"Norwegian Nynorsk","nativeName":"Norsk nynorsk"},
+        {"code":"no","name":"Norwegian","nativeName":"Norsk"},
+        {"code":"ii","name":"Nuosu","nativeName":"êêê Nuosuhxop"},
```
leonardr: Encoding error

leonardr:
The problems with this on a technical level are pretty minor but since we ultimately don't understand these languages, I don't think we can just plug in language names we got from Wikipedia. Sometimes the name we use for a language or the script we use to write the name has political implications. Maybe in some languages the name of the language is different in a "list of books" context than in a "title of Wikipedia page" context. We just don't know.

I think the infrastructure you have is fine, but I wouldn't go with this data. We need to fill in the data gradually with help from people who understand the context. We should be able to get this pretty quickly for NYPL's major collections. Since most of the minor collections are Romance or Germanic languages we should be able to get along pretty well with knowledge of English and Spanish. We can also poll everyone we asked for error message translations and get the answers for their languages. 

I'd prefer to fill in this list gradually and leave some language names in English than say something incorrect or disrespectful.

tests/test_util.py
```diff
@@ -78,10 +79,10 @@ def test_name_for_languageset(self):
...
         eq_("English", m(["en"]))
         eq_("English", m(["eng"]))
-        eq_(unicode("español", "utf-8"), m(['es']))
-        eq_(unicode("English & español", "utf-8"), m(["eng", "spa"]))
-        eq_(unicode("español & English", "utf-8"), m("spa,eng"))
-        eq_(unicode("español, English, & äæ (Zhōngwén)", "utf-8"), m(["spa","eng","chi"]))
+        eq_(u"español", m(['es']))
+        eq_(u"English & español", m(["eng", "spa"]))
+        eq_(u"español & English", m("spa,eng"))
+        eq_(u"español, English, & Chinese", m(["spa","eng","chi"]))
```
leonardr: I asked Mauricio whether capitalization was appropriate here and he wasn't sure. He did mention that using a slash to separate the languages would make it seem less like a sentence that needs a consistent set of capitalization rules.

Since we currently have no mixed-language lanes like this I think it's okay to land this and come back to it later.

leonardr:
Mauricio has confirmed that no capitalization is the way to go.

PR #305, opened 2016-06-14
Removed existing edition check from OPDSImportMonitor.

aslagle:
This is a followup to #300.

Now that we create CoverageRecords for imported editions, we don't need to check if there are existing editions for the identifiers in a feed.

I'm not sure if I should be doing something to handle transient vs persistent failures. Should it retry a transient failure even if the feed says it hasn't been updated? Should it not retry a persistent failure even if it has?

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -907,6 +907,15 @@ class OPDSImportMonitor(Monitor):
                     identifier, data_source, operation=CoverageRecord.IMPORT_OPERATION
                 )
 
+            # If there was a transient failure last time we tried to import this book,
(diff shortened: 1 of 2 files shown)


leonardr:
I would retry any kind of failure if the feed has been updated. Even if the problem was caused by a client-side bug, the feed might have been updated so that it no longer triggers the bug.

I would retry a transient failure even if the feed hasn't been updated. It probably won't help, but it won't hurt too much, and most feeds will never be updated, so if it's ever going to succeed it's going to be because we retried the same data against new import code.

leonardr:
:+1:

PR #307, opened 2016-06-15
Scrape sample epubs

daryanypl:


opds.py
```diff
@@ -109,11 +86,11 @@ def group_uri(cls, work, license_pool, identifier):
     @classmethod
     def rating_tag(cls, type_uri, value):
         """Generate a schema:Rating tag for the given type and value."""
-        rating_tag = E._makeelement("{%s}Rating" % schema_ns)
-        value_key = '{%s}ratingValue' % schema_ns
+        rating_tag = AtomFeed.E._makeelement("{%s}Rating" % AtomFeed.SCHEMA_NS)
```
leonardr: This change makes the code uglier than before, but it would become less ugly than before if you defined `AtomFeed.makeelement()` as a shortcut for this.

opds.py
```diff
@@ -192,14 +169,14 @@ def categories(cls, work):
         ):
             if value:
                 appeal = dict(term=schema_url + name, label=name)
-                weight_field = "{%s}ratingValue" % schema_ns
+                weight_field = "{%s}ratingValue" % AtomFeed.SCHEMA_NS
```
leonardr: Might also be nice to shortcut this sort of thing as `AtomFeed.schema_("ratingValue")`

leonardr:
As we discussed in person, use OPDSFeed.add_link_to_feed(feed.feed) instead of feed.add_link_to_feed(feed.feed).

The script itself should go into the Simplified `operations` repository because it's not generally useful--if someone else wants to use this feed we can just point them to the URL of the finished product. This requires setting up `operations` to use `server_core` as a submodule, as we've done with circulation/metadata/content.

opds.py
daryanypl: done

opds.py
daryanypl: done

leonardr:
This looks good once you remove the code that you moved to `operations`.

If you want to keep working on this you could add namespacing methods like bibframe_, dc_, and simplified_ to go along with schema_. It's not necessary though and we can always add it later.

daryanypl:
Made changes -- took script out, some minor cleanup.

PR #308, opened 2016-06-15
Tiny OPDS import fix

aslagle:


opds_import.py
```diff
@@ -887,7 +887,7 @@ def _get(self, url, headers):
         """
         kwargs = dict(timeout=120, allowed_response_codes=['2xx', '3xx'])
         response = HTTP.get_with_timeout(url, headers=headers, **kwargs)
-        return response.status_code, response.content_type, response.content
+        return response.status_code, response.headers, response.content
```
aslagle: This was failing because response.content_type doesn't exist.

I don't actually use this value, and this method isn't used in tests, but returning headers matches the dummy http class.

courte:
:+1:

PR #309, opened 2016-06-15
Removed the materialized views' assumption that the work's presentation edition is connected to the license pool by data source and identifier.

leonardr:
This assumption doesn't hold in a world where the work's presentation edition is likely to come from the presentation edition generator, where the licensepool comes from Overdrive or a source like that.

Instead of joining Work -> Edition -> LicensePool, we join Work -> Edition to get bibliographic information and Work -> LicensePool to get circulation information.


aslagle:
:+1:

PR #311, opened 2016-06-16
Go back to joining Work -> Edition -> LicensePool, but do it the right way and test it.

leonardr:
This branch corrects an error I made yesterday in fixing the materialized view query. The original query joined Work -> Edition -> LicensePool in a way that no longer worked because the database schema had changed. I changed it to join Work to Edition and LicensePool separately. This failed in the case where a work had more than one LicensePool, because works_id is supposed to be unique.

This branch makes the materialized view work like other parts of the app. A work has a presentation edition, and the license pool for that presentation edition is treated as 'the' license pool for the work, to be used whenever 'a' license pool for the work is needed.

I also added a test so that we keep this constraint in any future changes to the materialized view.


aslagle:
:+1:

aslagle:
still :+1:

PR #312, opened 2016-06-16
Changed default pagination size for search to 10.

aslagle:
For me locally, it takes about 2.5 seconds to create 50 OPDS entries in a search request - which can be longer than the time elasticsearch takes. A smaller page size will be much faster, and hopefully the first results will often be enough anyway.


leonardr:
:+1:

PR #313, opened 2016-06-17
Make it possible to test that RunCoverageProviderScript correctly instantiates a CoverageProvider.

leonardr:
Every once in a while we add something to the CoverageProvider constructor signature. Then we have to go through all our CoverageProviders and change their constructors, even if all they do is pass the new variable through to the superclass. Tests don't help because tests only verify that the old signature still works. They don't test what happens when RunCoverageProviderScript passes in the new, unsupported signature.

What we _really_ want to test is that RunCoverageProviderScript correctly instantiates every CoverageProvider. That's the test that will break when we change something. This branch makes it possible to instantiate a RunCoverageProviderScript from a test and investigate the CoverageProvider it instantiates. I demonstrate this by adding tests for the coverage providers for Axis 360, Overdrive and 3M.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -287,8 +287,10 @@ class RunCoverageProviderScript(IdentifierInputScript):
             parsed.cutoff_time = cls.parse_time(parsed.cutoff_time)
         return parsed
 
-    def __init__(self, provider):
-        args = self.parse_command_line(self._db)
+    def __init__(self, provider, _db=None, cmd_args=None, **provider_arguments):
+        if _db:
+            self._session = _db
+        args = self.parse_command_line(self._db, cmd_args)
         if callable(provider):
             if args.identifier_type:
                 self.identifier_type = args.identifier_type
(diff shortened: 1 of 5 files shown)


scripts.py
```diff
@@ -287,8 +287,10 @@ def parse_command_line(cls, _db, cmd_args=None, stdin=sys.stdin,
             parsed.cutoff_time = cls.parse_time(parsed.cutoff_time)
         return parsed
 
-    def __init__(self, provider):
-        args = self.parse_command_line(self._db)
+    def __init__(self, provider, _db=None, cmd_args=None, **provider_arguments):
+        if _db:
+            self._session = _db
```
aslagle: why is this self._session instead of self._db?

leonardr: Script._db is a @property that checks self._session and instantiates a new database session (storing it as ._session) if none is present. Since we're in the middle of a test we don't want to instantiate a new session, so we make sure ._session is set before we do anything.

aslagle:
:+1:

PR #314, opened 2016-06-17
Textual and audio versions of the same book need different works

leonardr:
A textual book and an audio version of the same book have the same permanent work ID (this is part of the permanent work ID standard), but they need to be stored in different Works. This branch extends the code to split and consolidate works, so that both permanent work ID and medium are considered when deciding whether two LicensePools can share a Work.


migration/20160614-add-coverage-status.sql
```diff
@@ -10,8 +10,8 @@ alter table workcoveragerecords add column status coverage_status;
 CREATE INDEX ix_coveragerecords_status ON coveragerecords USING btree (status);
```
leonardr: This is an unrelated change I decided I might as well commit. I had to rerun this migration script and this change makes it run a lot faster the second time.

aslagle:
:+1:

Issue #316, opened 2016-06-20
OPDS import can recreate LicensePools, leaving old LicensePool without identifier


leonardr:
This is kind of a vague issue but I don't have time to investigate it in depth.

After importing books multiple times from the open-access content server, I noticed a number of LicensePools that had no associated Identifiers. I think these two things are related. Maybe the old LicensePool gets destroyed and a new one created? 

This was causing crashes, but I've fixed them.
So apart from cluttering up the database, this isn't a big problem. The main problem is that the old license_pool_id is still stored in the materialized view. Until the materialized view is refreshed, the "changed" book won't show up in search results. 

I would still like to figure out what's really going on here because there might be a more subtle problem I haven't fixed.

leonardr:
Amy speculates that the old LicensePool might be the one with a data source of "Gutenberg", and the new LicensePool might be the one with a data source of "Library Simplified Open Access Content Server". The bug would happen if we don't allow two different data sources to provide a LicensePool for the same Identifier.

PR #317, opened 2016-06-20
Handle book that loses identifier

leonardr:
I've noticed that when doing OPDS imports from an open-access source, sometimes a Work's LicensePool is replaced with a new LicensePool, leaving the old pool with no identifier. (https://github.com/NYPL-Simplified/server_core/issues/316). This branch stops this undesirable case from causing patron-visible errors.


aslagle:
:+1:

PR #318, opened 2016-06-21
analytics events collected through Analytics singleton instead of Configuration

skomputer:
Also added top_genre() to Work as needed by NYPL-simplified/circulation#276.

### DIFF
--- a/analytics.py
+++ b/analytics.py
@@ -24,7 +24,7 @@ class Analytics(object):
         for provider in self.providers:
             provider.collect_event(_db, license_pool, event_type, time, **kwargs)
 
-def format_range(r):
+def format_age_range(r):
     if not r or not r.lower:
         return None
     min = r.lower if r.lower_inc else r.lower + 1
--- a/analytics.py
+++ b/analytics.py
@@ -23,3 +23,13 @@ class Analytics(object):
             time = datetime.datetime.utcnow()
         for provider in self.providers:
             provider.collect_event(_db, license_pool, event_type, time, **kwargs)
+
+def format_range(r):
+    if not r or not r.lower:
+        return None
+    min = r.lower if r.lower_inc else r.lower + 1
+    if r.upper:
+        max = r.upper + 1 if r.upper_inc else r.upper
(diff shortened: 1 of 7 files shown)


analytics.py
```diff
@@ -22,4 +22,14 @@ def collect_event(self, _db, license_pool, event_type, time=None, **kwargs):
         if not time:
             time = datetime.datetime.utcnow()
         for provider in self.providers:
-            provider.collect_event(_db, license_pool, event_type, time, **kwargs)
\ No newline at end of file
+            provider.collect_event(_db, license_pool, event_type, time, **kwargs)
+
+def format_age_range(r):
```
leonardr: This is very similar to Work.target_age_string. Is there a way to reuse/improve that code?

leonardr:
:+1: unless Work.target_age_string can replace format_age_range.

leonardr:
:+1:

analytics.py
```diff
@@ -1,25 +1,48 @@
...
             provider_class = getattr(provider_module, "Provider")
             analytics_providers.append(provider_class.from_config(config))
-        return cls(analytics_providers)
+        cls.__instance = cls(analytics_providers, providers)
+        return cls.__instance
 
-    def __init__(self, providers=[]):
+    def __init__(self, providers=[], provider_strings=[]):
         self.providers = providers
+        self.provider_strings = provider_strings
```
leonardr: We agreed on Slack that you can don't need this and can take it out.

model.py
```diff
@@ -3132,6 +3131,8 @@ def cover_thumbnail_url(self):
 
     @property
     def target_age_string(self):
+        if not self.target_age:
+            return None
```
leonardr: This needs to be the empty string, not None. There are a number of places where we append .target_age_string onto another string.

skomputer: @leonardr should it also return `""` instead of `None` a few lines below?

leonardr: Yes.

leonardr:
A final :+1:

PR #320, opened 2016-06-24
Remove shadowcat

leonardr:
This branch makes some behind-the-scenes improvements to the way we import the NYPL staff picks list.

First, the API for Shadowcat (a NYPL internal API) has been moved into a NYPL-specific private repository.

Second, the metadata layer code that creates equivalencies between Identifiers whose Editions have the same permanent work ID has been changed to only work on Editions that have the same medium. This is to stop audio book IDs from being made equivalent to print book IDs.


courte:
:+1:

PR #322, opened 2016-06-24
Set up CDNs based on the hostname, not the type of resource

leonardr:
Currently we configure CDNs based on the type of thing being CDNed: cover images, OPDS feeds, open-access books. This was a mistake. Our CDNs are set up to cache access to specific hostnames, not types of things. The hostname https://d3pqhns20516vc.cloudfront.net/ is the CDN for the hostname book-covers.nypl.org, not the CDN for "book covers".

Under most circumstances this mistake isn't visible. All the book covers are on book-covers.nypl.org so the CDN always works. But sometimes the book cover isn't on book-covers.nypl.org. Sometimes it's a book we haven't run through the metadata wrangler yet, and its cover is on contentreserve.com. The `cdnify` function just goes ahead and changes a URL that would work fine (http://contentreserve.com/foo/bar.jpg)  to a URL that doesn't work (https://d3pqhns20516vc.cloudfront.net/foo/bar.jpg).

This branch changes CDN configuration to work by hostname. The empty string is used to CDN-ify relative URLs such as those generated by app servers.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -45,10 +45,10 @@ from lane import (
 )
 from problem_details import *
 
-opds_cdn_host = Configuration.cdn_host(Configuration.CDN_OPDS_FEEDS)
+cdns = Configuration.cdns()
 def cdn_url_for(*args, **kwargs):
     base_url = url_for(*args, **kwargs)
-    return cdnify(base_url, opds_cdn_host)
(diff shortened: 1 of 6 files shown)


courte:
👍

PR #323, opened 2016-06-28
Correctly log circulation changes

leonardr:
This branch fixes a problem I noticed with the log messages triggered when circulation information changes. Because the message was triggered after the LicensePool's fields had been changed, it was always saying "[new value] => [new value]".

This branch also refactors the log message generation into its own method, and adds a unit test for it.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5670,7 +5670,6 @@ class LicensePool(Base):
         """
         edition = self.presentation_edition
         message = 'CHANGED '
-        blah = object()
         args = []
         if edition:
             message += '%s "%s" %s (%s)'
--- a/model.py
+++ b/model.py
@@ -5683,30 +5683,30 @@ class LicensePool(Base):
             message += '%s'
             args.append(self.identifier)
 
-        def _part(message, string, old_value, new_value):
-            if old_value == new_value:
-                # Nothing has changed.
-                return "", []
-            return ' %s: %s=>%s', [string, old_value, new_value]
+        def _part(message, args, string, old_value, new_value):
+            if old_value != new_value:
+                args.extend([string, old_value, new_value])
+                message += ' %s: %s=>%s'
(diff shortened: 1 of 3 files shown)


model.py
```diff
@@ -5646,8 +5652,63 @@ def update_availability(
...
+            args.extend([edition.medium, 
+                         edition.title or "[NO TITLE]",
+                         edition.author or "[NO AUTHOR]",
+                         self.identifier]
+                    )
+        else:
+            message += '%s'
+            args.append(self.identifier)
+
+        def _part(message, string, old_value, new_value):
```
courte: You're not using message right now in this method, but you probably could be. You could also pass in `args` as well, and then get rid of lines 5693-4, 5698-9, 5703-4, and 5708-9. Either way works, though; it's up to you.

model.py
```diff
@@ -5646,8 +5652,63 @@ def update_availability(
...
+    def circulation_changelog(self, old_licenses_owned, old_licenses_available,
+                              old_licenses_reserved, old_patrons_in_hold_queue):
+        """Generate a log message describing a change to the circulation.
+
+        :return: a 2-tuple (message, args) suitable for passing into 
+        logging.info or a similar method
+        """
+        edition = self.presentation_edition
+        message = 'CHANGED '
+        blah = object()
```
courte: You're not using this here.

courte:
Couple small things, but otherwise it looks fly.

model.py
leonardr: This is sort of how I originally wrote it, but I had scope problems with `message`. I've redone it as per your suggestion.

model.py
leonardr: This is left over from my attempts to figure out the scope problem.

courte:
:+1:

Issue #315, opened 2016-06-20
If there's a problem generating an OPDS entry it shouldn't bring down the whole feed


leonardr:
We had a problem where a LicensePool was being used in an OPDS feed despite the fact that it had no `.identifier`. This raised an exception which stopped the whole feed from being shown.

First, this specific problem ought to be fixed. This can happen if a Work is removed from the collection but the materialized view has not been updated.

Second, if there is any kind of exception when generating the OPDS entry for a single book, it shouldn't stop the entire feed from being generated. We should log the exception with log level ERROR and proceed to the next entry. It's almost never worthwhile to prevent a feed from being shown because one entry doesn't work.

PR #324, opened 2016-06-28
Bad entry shouldnt break feed

leonardr:
This branch does two things:

1. Makes the instantiation of analytics providers more robust, so that the core tests will run under a wider variety of configurations.
2. Fixes https://github.com/NYPL-Simplified/server_core/issues/315 by making create_entry return None if _create_entry raises an exception.

### DIFF
--- a/tests/test_opds.py
+++ b/tests/test_opds.py
@@ -1092,6 +1092,21 @@ class TestAcquisitionFeed(DatabaseTest):
         eq_(entry_string, work.simple_opds_entry)
 
 
+    def test_exception_during_entry_creation_is_not_reraised(self):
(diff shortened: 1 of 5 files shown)


tests/test_opds.py
```diff
@@ -1092,6 +1092,21 @@ def test_cache_usage(self):
         eq_(entry_string, work.simple_opds_entry)
 
 
+    def test_exception_during_entry_creation_is_not_reraised(self):
+        # This feed will raise an exception whenever it's asked
+        # to create an entry.
+        class DoomedFeed(AcquisitionFeed):
```
skomputer: 🎱

skomputer:
👍

PR #325, opened 2016-06-28
Metadata source may differ from circulation source

leonardr:
This branch reintroduces the idea that, when importing data from an OPDS feed, the actual book might come from a different source than the OPDS feed, and that the <bibframe:distribution> tag says where the actual book comes from.

Currently this only happens on the open-access content server. The Metadata created as part of the import is associated with the source of the OPDS feed (this includes any image links), but the CirculationData is associated with the source of the book. This means that you can end up with a Work whose LicensePool comes from Gutenberg, but the presentation Edition of that LicensePool comes from the open-access content server.

The motivating case here is that if a circ manager finds six editions of a book from the open-access content server, and every one of them has the same data source ("content server"), the circ manager has no way of knowing which edition it should use. When the data sources get passed through, the circ manager can make a rough judgement that certain data sources produce higher-quality books than others.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -533,10 +533,6 @@ class OPDSImporter(object):
 
         # At this point we can assume that we successfully got some
         # metadata, and possibly a link to the actual book.
-
-        # Note: will ignore bibframe_distribution from feed, and use data source passed into the 
-        # importer as license_data_source, too.
-
         try:
             kwargs_meta = cls._data_detail_for_feedparser_entry(entry, data_source)
             return identifier, kwargs_meta, failure
@@ -547,7 +543,7 @@ class OPDSImporter(object):
             return identifier, None, failure
 
     @classmethod
-    def _data_detail_for_feedparser_entry(cls, entry, data_source):
+    def _data_detail_for_feedparser_entry(cls, entry, metadata_data_source):
         """Helper method that extracts metadata and circulation data from a feedparser
         entry. This method can be overridden in tests to check that callers handle things
         properly when it throws an exception.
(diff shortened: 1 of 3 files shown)


opds_import.py
```diff
@@ -556,7 +552,40 @@ def _data_detail_for_feedparser_entry(cls, entry, data_source):
...
+                _db = Session.object_session(metadata_data_source)
+                circulation_data_source = DataSource.lookup(
+                    _db, circulation_data_source_name
+                )
+                if not circulation_data_source:
+                    raise ValueError(
+                        "Unrecognized circulation data source: %s" % (
+                            circulation_data_source_name
+                        )
+                    )
```
leonardr: If a data source is added to the content server, but the circ manager doesn't know about it yet, this code stops a crash.

opds_import.py
```diff
@@ -607,10 +636,12 @@ def summary_to_linkdata(detail):
             data_source_last_updated=last_opds_update,
         )
             
-        # Only add circulation data if the data source is lendable.
-        if data_source.offers_licenses:
+        # Only add circulation data if both the book's distributor *and*
+        # the source of the OPDS feed are lendable data sources.
+        if (circulation_data_source and circulation_data_source.offers_licenses
+            and metadata_data_source.offers_licenses):
```
leonardr: This means that even if the metadata wrangler purports to offer licenses to books, we won't take it up on the offer.

I think this only happens in the tests, but it's good to fix the tests in a way that will stop a bad thing from happening in real life.

aslagle: Should we change the content server data source back to not offering licenses? And then we wouldn't need to check the metadata_data_source here.

daryanypl:
👍

opds_import.py
leonardr: I don't think we really need it either, it's just a safety measure.

aslagle:
:+1:

PR #327, opened 2016-06-29
When importing from an OPDS feed, don't create a LicensePool for a book that is not available in any formats.

leonardr:
This branch comes out of my attempts to fix the import of unglue.it books. A more useful circulation branch is also forthcoming; this one's kind of a sideshow.

If we're importing an OPDS feed from a data source that provides LicensePools, but the feed doesn't actually include any links to the book, this branch stops the import process from actually creating a LicensePool. Only an Edition with the metadata will be created.


tests/files/opds/book_with_license.opds
```diff
@@ -0,0 +1,30 @@
+<feed xmlns:bibframe="http://bibframe.org/vocab/" xmlns:simplified="http://librarysimplified.org/terms/" xmlns:app="http://www.w3.org/2007/app" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:opds="http://opds-spec.org/2010/catalog" xmlns:schema="http://schema.org/" xmlns="http://www.w3.org/2005/Atom">
```
leonardr: Turns out I didn't need this file, but it seems useful. I'm happy to delete it though.

aslagle:
:+1:

PR #328, opened 2016-06-30
Add a missing import

leonardr:
This branch adds a missing import of `time` which was stopping the materialized view script from running.

I tried pretty hard to add a test of the materialized view script but it seems impossible. The CONCURRENTLY part of the refresh command only works on certain versions of Postgres, and a failure invalidates the entire database session. Even if the materialised view refresh _succeeds_, it then closes the database session to vacuum the database, ruining every subsequent test.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -3,6 +3,7 @@ import datetime
 import logging
 import os
 import requests
+import time
 from requests.exceptions import (
     ConnectionError, 
     HTTPError,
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -94,6 +95,15 @@ def parse_time(cls, time_string):
                     continue
         raise ValueError("Could not parse time: %s" % time_string)
 
+    def __init__(self, _db=None):
```
leonardr: I refactored this from RunCoverageProviderScript, making it generally possible to pass in the test database session to a Script constructor when testing it.

courte:
:+1:

PR #329, opened 2016-06-30
New OPDS authentication document format

aslagle:
This connects to https://github.com/NYPL-Simplified/circulation/issues/274

We've been discussing a new OPDS authentication format, and this branch changes the fill_in method to organize the document by provider as we decided to do for Open eBooks. 

It's not clear yet what we'll do for NYPL, which only has one default authentication method. Current clients use the auth headers and don't look at the document at all, so putting in providers for NYPL too won't break anything.


leonardr:
Makes sense. :+1:

PR #330, opened 2016-07-01
Added postgres function to compute recursively equivalent identifiers.

aslagle:


model.py
```diff
@@ -197,6 +202,17 @@ def initialize(cls, url):
...
+
+        # Create the recursive equivalents function. If the function
+        # already exists, it will be replaced.
+        if not connection:
+            connection = engine.connect()
+        resource_file = os.path.join(resource_path, cls.RECURSIVE_EQUIVALENTS_FUNCTION)
+        if not os.path.exists(resource_file):
+            raise IOError("Could not load recursive equivalents function from %s: file does not exist." % resource_file)
+        sql = open(resource_file).read()
+        connection.execute(sql)
```
aslagle: I _think_ this will run when the app is restarted, so it won't be necessary to have a migration.

model.py
```diff
@@ -2671,13 +2561,12 @@ def apply_similarity_threshold(self, candidates, threshold=0.5):
...
+        identifier_ids = [self.primary_identifier.id]
         if distance > 0:
-            data = Identifier.recursively_equivalent_identifier_ids(
-                _db, flattened_data, distance, threshold=threshold)
-            flattened_data = Identifier.flatten_identifier_ids(data)
+            identifier_ids = Identifier.recursively_equivalent_identifier_ids(
+                _db, identifier_ids, distance, threshold=threshold)
 
-        return Identifier.best_cover_for(_db, flattened_data)
+        return Identifier.best_cover_for(_db, identifier_ids)
```
aslagle: There are a number of places like this where we get a list of identifier ids and pass it to something else. It would be faster to use a subquery, but that can be a future branch.

tests/test_model.py
```diff
@@ -708,6 +863,9 @@ def test_recursive_edition_equivalence(self):
         work = Work()
         work.license_pools.extend([gutenberg2_pool])
 
+        # This is necessary for the work to have an id in the db.
+        self._db.flush()
```
aslagle: I don't really understand this :(

model.py
leonardr: Yes, it runs whenever the app starts.

model.py
```diff
@@ -3389,14 +3278,11 @@ def from_identifiers(cls, _db, identifiers, base_query=None):
...
-        equivalent_identifier = aliased(Identifier)
-        query = base_query.outerjoin(Identifier.equivalencies).\
-                outerjoin(equivalent_identifier, Equivalency.output).\
-                filter(or_(
-                    Identifier.id.in_(identifier_ids),
-                    and_(equivalent_identifier.id.in_(identifier_ids),
-                    Equivalency.strength>=1)
-                ))
+        identifier_ids_subquery = Identifier.recursively_equivalent_identifier_ids_query(
+            Identifier.id, levels=1, threshold=0.999)
```
leonardr: Why 0.999?

aslagle: Before it was using >= 1, but the function only does >. 

I wanted this function to behave the same way as before, so > 0.999 is pretty close to >= 1.

aslagle: I tried changing the function to >= but there are other parts of the code that were using >, so I had to pick one.

tests/test_model.py
leonardr: If the ID is the issue, I think it's because the ID is an auto-increment and Postgresql is in charge of it.

Take a look at create() in model.py. That calls db.add(new_object) and then db.flush(). So there's no shame in doing this, but you might want to use create(Work) to remove a bit of code.

leonardr:
This is an incredible branch. I had a question about your use of the power operator in the function, but the tests explained it. :+1:

model.py
leonardr: OK, > 0.999 is fine. I was worried it was >= 0.999 and we would miss identities that we're absolutely sure about, like Overdrive ID -> ISBN.

leonardr:
Actually my initial misgivings were accurate. This system gives approximately the same results in common cases but it doesn't work the same way as the old system.

In the old system, the threshold stays the same but the strength of the signal decreases with every hop. If identifier A is linked to identifier B with confidence 0.8, and B is linked to C with confidence 0.7, then effectively A is linked to C with confidence 0.7*0.8=0.56. This means that C counts if the threshold is 0.5 but not if the threshold is 0.6.

In this system, the strength of the signal stays the same but the threshold increases with each hop. A is linked to C with confidence 0.7, but 0.7 might not be high enough anymore. I don't think this corresponds to a real-world phenomenon we want to track.

See if you can get the strength of the signal to decrease with each hop.

leonardr:
The code looks good but I'm worried that the tests didn't need to change. Can you write a test that varies the threshold to show the cutoff happening?

aslagle:
The exponential threshold increase was similar enough that my tests for that cutoff still worked with decaying strengths. I added another test that works with decaying strengths but wouldn't work with increasing thresholds.

leonardr:
Nice. :+1:

PR #331, opened 2016-07-05
Opds import monitor is coverage provider

leonardr:
This branch changes the OPDS import monitor to work more like a CoverageProvider. Instead of using a Timestamp (as most monitors do) to track which OPDS entries have been imported, the OPDS import monitor now checks the `updated` date on each individual OPDS entry against the date on that entry's CoverageRecord.

I also changed the code so that only certain media types get mirrored. This is so "text/html" open-access editions don't get mirrored. At least on unglue.it, these text/html open-access links tend to link to web pages explaining how you can get a copy of the book, or to a Google Books web reader, not to an HTML copy of the book as it's generally understood. I did add MOBI to the list of media types that get mirrored.


metadata_layer.py
```diff
@@ -456,7 +455,18 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
...
-        if representation.status_code == 304:
+        if representation.status_code == 304 and representation.mirror_url:
+            return
+
+        # The metadata may have some idea about the media type for this
+        # LinkObject, but the media type we actually just saw takes 
+        # precedence.
+        if representation.media_type:
+            link.media_type = representation.media_type
+
```
leonardr: I moved this up so that we store the media type in the database even if the representation turns out not to be mirrorable.

opds_import.py
```diff
@@ -919,10 +906,17 @@ class OPDSImportMonitor(Monitor):
...
     def __init__(self, _db, feed_url, default_data_source, import_class, 
-                 interval_seconds=3600, keep_timestamp=True,
-                 immediately_presentation_ready=False):
+                 interval_seconds=0, keep_timestamp=False,
+                 immediately_presentation_ready=False, force_reimport=False):
+
+        # We never keep a timestamp because we can track when the OPDS
+        # archive feed was last updated.
+        keep_timestamp = False
+        interval_seconds = 0
```
aslagle: These are already the default arguments, do we need to prevent subclasses from changing them?

aslagle:
Will it be a problem that the editions we've already imported don't have CoverageRecords? I was thinking we'd need a migration to create them, but it probably wouldn't be that bad to import them all again once.

Otherwise, looks good :)

opds_import.py
leonardr: If a subclass passes in a value for these arguments I think it's a mistake. Although if you really know what you're doing I guess we should let you do it. I'll change it.

leonardr:
I need to import them all again once anyway, to make sure that all the resources were properly mirrored, so it's not a big deal.

PR #332, opened 2016-07-05
Remove duplicate contributions

leonardr:
This branch does several different things to fix a problem where Editions are listed as having the same author twice. Example: https://circulation.librarysimplified.org/works/Gutenberg/Gutenberg%20ID/35536

This doesn't do the one thing which would really solve the problem: change the model to prevent someone from being listed as both the PRIMARY_AUTHOR and an AUTHOR of a given book. I didn't change this because we're going to be redoing the contributor system pretty soon and that whole thing might change anyway.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -943,7 +943,7 @@ class Metadata(MetaToModelUtility):
         for contribution in edition.contributions:
             contributor = ContributorData.from_contribution(contribution)
             contributors.append(contributor)
-        else:
+        if not edition.contributions:
             # This should only happen for low-quality data sources such as
             # the NYT best-seller API.
             if edition.sort_author:
--- a/model.py
+++ b/model.py
@@ -2240,13 +2240,14 @@ class Edition(Base):
 
     @property
     def contributors(self):
-        return [x.contributor for x in self.contributions]
+        return set([x.contributor for x in self.contributions])
 
     @property
     def author_contributors(self):
-        """All 'author'-type contributors, with the primary author first,
-        other authors sorted by sort name.
+        """All distinct 'author'-type contributors, with the primary author
+        first, other authors sorted by sort name.
         """
+        seen_authors = set()
         primary_author = None
         other_authors = []
         acceptable_substitutes = defaultdict(list)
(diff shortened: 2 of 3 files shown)


metadata_layer.py
```diff
@@ -943,7 +943,7 @@ def from_edition(cls, edition):
         for contribution in edition.contributions:
             contributor = ContributorData.from_contribution(contribution)
             contributors.append(contributor)
-        else:
+        if not edition.contributions:
```
leonardr: This is the main one. That `else` clause was always being triggered because the `for` loop didn't break. In many cases this caused two `ContributorData` objects to be created for the same person: one with AUTHOR_ROLE (the code in the `for` loop) and one with PRIMARY_AUTHOR_ROLE (the code in the `else` statement).

model.py
```diff
@@ -2240,13 +2240,14 @@ def language_code(self):
 
     @property
     def contributors(self):
-        return [x.contributor for x in self.contributions]
+        return set([x.contributor for x in self.contributions])
```
leonardr: This is just for display purposes.

model.py
```diff
@@ -2272,18 +2273,33 @@ def author_contributors(self):
...
+            for i in l:
+                if i in seen:
+                    continue
+                deduped.append(i)
+                seen.add(i)
+            return deduped
+
         if primary_author:
-            return [primary_author] + sorted(other_authors, key=lambda x: x.name)
+            return dedupe([primary_author] + sorted(other_authors, key=lambda x: x.name))
```
leonardr: Even if the same person is "supposed" to be listed as both primary author and a regular author (ie. both those facts are in the data, I'm not saying that actually makes sense), this will stop them from showing up twice in the author list we display to the patron.

model.py
```diff
@@ -5429,7 +5447,10 @@ def set_presentation_edition(self, policy=None):
             metadata.license_data_source_obj = None
             edition, is_new = metadata.edition(_db)
 
-            self.presentation_edition, edition_core_changed = metadata.apply(edition)
+            policy = ReplacementPolicy.from_metadata_source()
+            self.presentation_edition, edition_core_changed = metadata.apply(
+                edition, replace=policy
+            )
```
leonardr: This bit ensures that calling set_presentation_edition multiple times associates a fresh set of contributions with the presentation edition every time. Previously the default ReplacementPolicy was used, which adds to the contribution list instead of replacing it. This means that once you had incorrectly created duplicate Contributions, you could never get rid of them.

tests/test_model.py
```diff
@@ -1389,26 +1415,17 @@ def test_set_presentation_edition(self):
...
         edition_mw = self._edition(data_source_name=DataSource.METADATA_WRANGLER, with_license_pool=False)
         edition_od, pool = self._edition(data_source_name=DataSource.OVERDRIVE, with_license_pool=True)
- 
+
         edition_mw.primary_identifier = pool.identifier
         edition_admin.primary_identifier = pool.identifier
 
         # set overlapping fields on editions
         edition_od.title = u"OverdriveTitle1"
-        [joe], ignore = Contributor.lookup(self._db, u"Sloppy, Joe")
```
leonardr: I took these contributors out because they weren't changing anything about the test.

courte:
Bravo! 👍

metadata_layer.py
daryanypl: Is there ever a case when a duplicate contribution makes sense, like maybe when an author is also the translator?

model.py
daryanypl: Are other_authors allowed to have dupes within themselves and with the primary_author?

metadata_layer.py
leonardr: Yes, that can still happen. This branch is only concerned with stopping a contributor from being listed as both PRIMARY_AUTHOR and AUTHOR; and if a contributor _does_ have multiple authorship roles in a book, stopping them from showing up twice in the list of contributors.

model.py
leonardr: In this case we're trying to figure out what would go on the book cover. The primary author should go first, and be followed by non-primary authors in alphabetical order. People whose role does not rise to the level of "authorship" (e.g. illustrator) do not show up. 

The list as a whole should contain no duplicates. This might happen because someone is erroneously listed twice in the same role, someone is listed as both primary author and regular author, someone is listed as both author and translator, etc. However it happens, your name only shows up once on the front of the book.

daryanypl: This explanation was very helpful to me.  Would you mind including it in a code comment?

leonardr: Sure.

PR #333, opened 2016-07-06
Fixup script that removes ambiguous and duplicate contributors

leonardr:
This branch creates a one-use script that fixes two problems with Editions:

1. Sometimes a person is listed in an 'Unknown' role and also in a more specific role. I think this is fallout from a very old problem with Overdrive.
2. Sometimes a person is listed both as 'Author' and 'Primary Author'. They should only be listed once. This was fixed by https://github.com/NYPL-Simplified/server_core/pull/332. This is mainly a problem with open-access books.

I'm running this on the SimplyE QA server and it's working fine.


courte:
👍 😹

PR #335, opened 2016-07-06
Bulk search index update

aslagle:
It's very hard to work on the search index when creating an index takes days, so this branch is my first attempt to speed it up.

Most of the time was spent individually creating a JSON document for each work, so I changed it to create a single SQL query that generates JSON for a batch of works all together. On my laptop this cuts the time for creating 100 documents from about 14 seconds to 0.14 seconds.

I also use elasticsearch's bulk method to upload the batch, which helped a little but not nearly as much as the other change.

With this branch it's taking about 3 seconds to index 500 documents on my laptop, which is much better but could still use improvement. Most of the time is from CoverageProvider itself now.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -218,6 +218,11 @@ class BaseCoverageProvider(object):
         (for transient failures).
         """
 
+        # Batch is a query that may not be ordered, so it may return
+        # different results when executed multiple times. Converting to
+        # a list ensures that all subsequent code will run on the same items.
+        batch = list(batch)
+
         offset_increment = 0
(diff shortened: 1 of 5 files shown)


aslagle:
I didn't make any changes to the search document as part of this - I tried to reimplement Work.to_search_document exactly as it was in the SQL query. There are some things in there that might be better handled with elasticsearch analyzers or removed entirely. Being able to create a search index quickly will make it possible to try those things. :)

leonardr:
:+1:

PR #336, opened 2016-07-07
Only create the recursive equivalents function in the database if it doesn't already exist.

aslagle:
This fixes an error that could happen if two apps using the same database start at the same time.

### DIFF
--- a/model.py
+++ b/model.py
@@ -75,6 +75,7 @@ from sqlalchemy.sql.expression import (
     join,
     literal_column,
     case,
+    table,
 )
 from sqlalchemy.exc import (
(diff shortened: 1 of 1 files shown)


leonardr:
I think there's still a chance of a race condition. How bad would it be if that happened? One of the scripts would fail but it would just be run again later, right?

leonardr:
Actually the race condition would only happen the very first time you ran something, so it's fine. :+1:

aslagle:
Yeah, there's a race condition if the function doesn't exist yet and you start two apps, but once it exists you're fine.

PR #338, opened 2016-07-07
Fix OPDS import edge cases

leonardr:
This branch fixes some problems I discovered while trying to import books from unglue.it.

1. Some ebooks are hosted at Github, which serves them as `application/octet-stream`. Previous to this branch we had a blanket policy of replacing the `type` of a link in an OPDS feed with the `Content-Type` we got once we followed the link. This branch makes sure that doesn't happen for `application/octet-stream`. 

2. Some OPDS entries (such as https://unglue.it/api/opds/all/?work=158143) include relative URLs. We had code to handle relative URLs but it requires a `self` link to start with, and unlike with Standard Ebooks there's no `self` link here. I changed it so that the original feed URL is passed all the way down into the code that parses the relative URLs, and we use the original feed URL to qualify relative URLs into absolute URLs.

I don't know what the 'right' thing to do is in #2. RFC 4287 says that the `<xml:base>` tag can be used to quality relative URLs, but nobody uses that. I definitely think using the feed URL is better than using the `self` link but I'm torn on whether it's worth getting rid of the code that checks the `self` link--it's definitely not right, but it seems like it will work pretty often, but it's not necessary if you just pass in the feed URL.

### DIFF
--- a/model.py
+++ b/model.py
@@ -2246,6 +2246,19 @@ class Edition(Base):
     def author_contributors(self):
         """All distinct 'author'-type contributors, with the primary author
         first, other authors sorted by sort name.
+
+        Basically, we're trying to figure out what would go on the
+        book cover. The primary author should go first, and be
+        followed by non-primary authors in alphabetical order. People
+        whose role does not rise to the level of "authorship"
+        (e.g. author of afterword) do not show up.
+
+        The list as a whole should contain no duplicates. This might
+        happen because someone is erroneously listed twice in the same
+        role, someone is listed as both primary author and regular
+        author, someone is listed as both author and translator,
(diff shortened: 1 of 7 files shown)


model.py
```diff
@@ -2246,6 +2246,19 @@ def contributors(self):
     def author_contributors(self):
         """All distinct 'author'-type contributors, with the primary author
         first, other authors sorted by sort name.
+
```
leonardr: Darya requested I add this explanation from an earlier PR.

aslagle:
:+1: I agree that passing around the feed URL is better than using the self link. I think the self link can be relative too, so sometimes it won't even help. We also use the feed URL in our client-side OPDS parsing.

PR #340, opened 2016-07-11
Minor opds import cleanup

leonardr:
This branch fixes a number of edge cases that showed up while importing OPDS feeds from unglue.it.

1. If the HTTP request to download a resource before mirroring it results in an HTTP error code, we no longer act as though we successfully downloaded something.
2. An Annotator is allowed (but not required) to raise `UnfulfillableWork` if it's asked to annotate an OPDS entry but it turns out there's no way to acquire that work. If this happens, the `<entry>` tag for the Work will contain only an explanatory error message.
3. In some cases where we go in expecting to generate an `<entry>` for a Work but it turns out we can't, we will now return an explanatory error message instead of None.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -825,6 +825,10 @@ class OPDSImporter(object):
         rel = attr.get('rel')
         media_type = attr.get('type')
         href = attr.get('href')
+        if not href or not rel:
+            # The link exists but has no destination, or no specified
+            # relationship to the entry.
+            return None
         rights = attr.get('{%s}rights' % OPDSXMLParser.NAMESPACES["dcterms"])
         if rights:
             rights_uri = RightsStatus.rights_uri_from_string(rights)
@@ -834,6 +838,7 @@ class OPDSImporter(object):
             # This link is relative, so we need to get the absolute url
             href = urljoin(feed_url, href)
         return LinkData(rel=rel, href=href, media_type=media_type, rights_uri=rights_uri)
+        return foo
 
     @classmethod
     def consolidate_links(cls, links):
@@ -844,7 +849,8 @@ class OPDSImporter(object):
 
         Similarly if link n is a thumbnail and link n+1 is an image.
         """
-        new_links = list(links)
+        new_links = [x for x in links if x]
+        links = list(new_links)
         next_link_already_handled = False
         for i, link in enumerate(links):
 
@@ -966,6 +972,10 @@ class OPDSImportMonitor(Monitor):
             # feed has changed.
             if record and record.status == CoverageRecord.TRANSIENT_FAILURE:
                 new_data = True
+                self.log.info(
+                    "Counting %s as new because previous attempt resulted in transient failure: %s", 
+                    record.identifier, record.exception
+                )
                 break
 
             # If our last attempt was a success or a persistent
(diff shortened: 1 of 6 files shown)


opds_import.py
```diff
@@ -825,6 +825,10 @@ def extract_link(cls, link_tag, feed_url=None):
         rel = attr.get('rel')
         media_type = attr.get('type')
         href = attr.get('href')
+        if not href or not rel:
+            # The link exists but has no destination, or no specified
+            # relationship to the entry.
+            return None
```
leonardr: This stops `<link>` tags with empty `href` or `rel` fields from showing up as LinkData objects.

opds_import.py
```diff
@@ -844,7 +849,8 @@ def consolidate_links(cls, links):
 
         Similarly if link n is a thumbnail and link n+1 is an image.
         """
-        new_links = list(links)
+        new_links = [x for x in links if x]
```
leonardr: This strips out any `None` elements (from `<link>` tags that were missing their `href`s).

opds_import.py
```diff
@@ -966,6 +972,10 @@ def check_for_new_data(self, feed):
             # feed has changed.
             if record and record.status == CoverageRecord.TRANSIENT_FAILURE:
                 new_data = True
+                self.log.info(
```
leonardr: I added these log statements because it was difficult to understand why certain pages were being processed or not.

tests/test_opds_import.py
```diff
@@ -171,6 +172,22 @@ def test_extract_metadata(self):
...
+        eq_(None, OPDSImporter.extract_link(no_href))
+
+        good = E.link(href="http://foo", rel="bar")
+        link = OPDSImporter.extract_link(good)
+        eq_("http://foo", link.href)
+        eq_("bar", link.rel)
+
+        relative = E.link(href="/foo/bar", rel="self")
+        link = OPDSImporter.extract_link(relative, "http://server")
+        eq_("http://server/foo/bar", link.href)
```
leonardr: This is a test for code I committed earlier.

opds_import.py
```diff
@@ -266,7 +266,7 @@ def import_from_feed(self, feed, even_if_no_author=False,
             except Exception, e:
                 identifier, ignore = Identifier.parse_urn(self._db, key)
                 data_source = DataSource.lookup(self._db, self.data_source_name)
-                failure = CoverageFailure(identifier, traceback.format_exc(), data_source=data_source, transient=False)
+                failure = CoverageFailure(identifier, traceback.format_exc(), data_source=data_source, transient=True)
```
aslagle: But this means we'll be retrying imports with problems that aren't fixed yet. It seems like there are really three kinds of failure:
- Something that will go away by itself if we try again later
- Something that will go away once we do something to fix it
- Something that will never go away

opds_import.py
```diff
@@ -844,7 +848,8 @@ def consolidate_links(cls, links):
 
         Similarly if link n is a thumbnail and link n+1 is an image.
         """
-        new_links = list(links)
+        new_links = [x for x in links if x]
+        links = list(new_links)
```
aslagle: it's already a list now

leonardr: It's not that I'm converting it into a list, it's that I'm making another copy of the list. I need `new_links` to be a copy of `links`, because I'm going to go through `links`, consolidate thumbnails, and remove the thumbnails from `new_links`. I can't iterate over `new_links` because I'm going to be changing the list.

Do you have any ideas about how to convey this better?

aslagle: That makes sense. It would probably be clearer looking at all the code instead of this diff, but maybe just add a comment.

opds_import.py
leonardr: I don't think that would be too bad but it's true that my change was not terribly well thought out--there's code a little bit earlier that treats a similar exception as a transient error. So I've changed it back for the time being.

aslagle: I don't see this change yet

PR #344, opened 2016-07-13
Superceded licensepool cannot be active in opds feed

leonardr:
This branch changes the rules for which LicensePool can be the 'active' pool when generating a work's OPDS entry. In particular, a superceded pool can no longer be the active pool. This solves an issue where a better version of a public domain book was coming in, but the circulation manager was still using the old version.

Note that fixing this completely requires updating the materialized view, since the view will still contain a cached version of the entry's old OPDS entry.

I added a number of tests for the previously untested `active_licensepool_for` method. I also added verbosity to the `Work.detailed_representation` and the `Explain` script, which were useful in debugging this issue.

### DIFF
--- a/tests/test_opds.py
+++ b/tests/test_opds.py
@@ -181,7 +181,7 @@ class TestBaseAnnotator(DatabaseTest):
         eq_(pool2, Annotator.active_licensepool_for(work))
 
         # pool2 is superceded and pool1 is not. The active licensepool
-        # is pool2.
+        # is pool1.
         pool1.superceded = False
         pool2.superceded = True
         eq_(pool1, Annotator.active_licensepool_for(work))
--- a/tests/test_opds.py
+++ b/tests/test_opds.py
@@ -152,6 +152,73 @@ class UnfulfillableAnnotator(TestAnnotator):
         raise UnfulfillableWork()
 
 
+class TestBaseAnnotator(DatabaseTest):
+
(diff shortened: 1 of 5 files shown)


tests/test_opds.py
```diff
@@ -152,6 +152,73 @@ def annotate_work_entry(self, *args, **kwargs):
...
+        # though we don't really know or care which one.
+        assert Annotator.active_licensepool_for(work) is not None
+
+        # Neither license pool is open-access, and pool1 is superceded.
+        # The active license pool is pool2.
+        pool1.superceded = True
+        eq_(pool2, Annotator.active_licensepool_for(work))
+
+        # pool2 is superceded and pool1 is not. The active licensepool
+        # is pool2.
```
aslagle: This comment is wrong.

aslagle:
I'd like to know how we ended up with a work's presentation edition coming from a superceded license pool - maybe there's a bug where we're not updating the work's presentation edition and we should be.

leonardr:
That is worrisome, I'll try to figure it out.

leonardr:
I haven't been able to reproduce this problem so now I think of it more as a safety measure. I've also added a new script and a number of useful improvements I created while trying to reproduce the problem. Some of the new code ties into my circulation branch at:

https://github.com/NYPL-Simplified/circulation/pull/314

aslagle:
:+1:

PR #345, opened 2016-07-15
License pool in search index

### LINKED ISSUE #339: Search stops if elasticsearch returns a page of works that aren't ready or deliverable
When we do a search, we get results from elasticsearch, then query the database to get actual ready deliverable works from the materialized view. So it's possible that some of the results from elasticsearch won't be included. However, the pagination parameters go to elasticsearch, so a request for 10 items could return fewer results, even 0. If the client sees a page with 0 results, it won't ask for another page, but it's possible that there are more results on the next page.

This is currently happening with "Olivia" on open ebooks. We have about 40 Olivia books in the database and search index, but only 6 are have licenses and aren't suppressed. The first page of search results has one or two results, and the second page has 0, but the third page has more.

This was always a problem but it became worse when we switched the pagination size from 50 to 10, because it's more likely that an intermediate page will be completely empty.

aslagle:
This branch adds license pool availability info to the search index. Searches are now filtered in elasticsearch to only return deliverable results, instead of returning all results and filtering in the materialized view query afterwards.

This will require building a new search index, so I'd like to hold off on merging this pending performance discussion with Ashok.

This fixes #339.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1133,6 +1133,7 @@ class Lane(object):
                     fields=["_id", "title", "author", "license_pool_id"],
                     size=pagination.size,
                     offset=pagination.offset,
+                    only_deliverable=True,
                 )
             except elasticsearch.exceptions.ConnectionError, e:
                 logging.error(
@@ -1157,7 +1158,6 @@ class Lane(object):
                         lazyload(mw.license_pool, LicensePool.identifier),
                         lazyload(mw.license_pool, LicensePool.presentation_edition),
                     )
-                    q = self.only_show_ready_deliverable_works(q, mw)
                     q = self._defer_unused_opds_entry(q, work_model=mw)
                     work_by_id = dict()
                     a = time.time()
--- a/external_search.py
+++ b/external_search.py
@@ -135,13 +135,14 @@ class ExternalSearchIndex(object):
                 
 
     def query_works(self, query_string, media, languages, exclude_languages, fiction, audience,
-                    age_range, in_any_of_these_genres=[], fields=None, size=30, offset=0):
+                    age_range, in_any_of_these_genres=[], fields=None, size=30, offset=0, only_deliverable=True):
         if not self.works_index:
             return []
 
         filter = self.make_filter(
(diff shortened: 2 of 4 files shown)


lane.py
```diff
@@ -1157,7 +1158,6 @@ def search(self, query, search_client, pagination=None):
                         lazyload(mw.license_pool, LicensePool.identifier),
                         lazyload(mw.license_pool, LicensePool.presentation_edition),
                     )
-                    q = self.only_show_ready_deliverable_works(q, mw)
```
leonardr: I think it would be valuable to apply at least the filter against `LicensePool.suppressed`. When someone suppresses a license pool it's a sign of a serious problem and they want it to be taken off the shelf _immediately_, not at 3 AM the next morning.

leonardr:
Looks good apart from my one comment.

lane.py
aslagle: This reminds me that I need to update the search index when someone suppresses a book through the admin interface. Once I do that, the search index shouldn't return suppressed books anyway. We could still apply the filter just in case. The only part of it that I left out of the search index filter was DeliveryMechanism.default_client_can_fulfill. Clients should be able to figure out that there's another page of search results if they see a page with only books they can't fulfill.

leonardr: Are we going to build the featured lanes from the search index, or will that still use the database? Featured lanes aren't paginated, so I think it's really important that they only include links the default client can fulfill. I'm okay with paginated lists including works with no delivery mechanism fulfillable by the default client, assuming that our clients know enough to hide those entries. We'll need to test that before we deploy to production.

aslagle: This branch only does the paginated lists for search results, but I agree about the featured lanes. We can add that to the search filter if we start using it for that.

leonardr:
It looks like we're having the same rdflib -> html5lib -> setuptools problem we were having on Friday (in some other component? I don't remember). :+1: once you take care of that.

leonardr:
It took a long time, but we finally resolved this issue with a different branch.

PR #346, opened 2016-07-15
Arthurian romances are not romance novels.

leonardr:
This branch classifies "Arthurian romances" as historical fiction instead of romance fiction.


aslagle:
:+1:

PR #348, opened 2016-07-18
Auto convert to svg

leonardr:
This branch fixes https://github.com/NYPL-Simplified/content_server/issues/68 by ensuring that SVG images are converted to PNG when it's time to mirror them to S3. This is as opposed to thumbnail images, which are converted to PNG when the thumbnail is created.

There's also a minor change made to the code that fetches OPDS feds for import purposes, which was necessary to get the Standard Ebooks import to work.

### DIFF
--- a/model.py
+++ b/model.py
@@ -7026,14 +7026,14 @@ class Representation(Base):
         """
         if not self.is_image or self.clean_media_type != self.SVG_MEDIA_TYPE:
             # Passthrough
-            return self.media_type, self.content_fh()
+            return self.content_fh()
 
         # This representation is an SVG image. We want to mirror it as
         # PNG.
         image = self.as_image()
         output = StringIO()
         image.save(output, format='PNG')
-        return self.PNG_MEDIA_TYPE, StringIO(output.getvalue())
+        return StringIO(output.getvalue())
 
     def content_fh(self):
         """Return an open filehandle to the representation's contents.
--- a/model.py
+++ b/model.py
@@ -6969,8 +6969,9 @@ class Representation(Base):
         """Try to come up with a good file extension for this representation."""
         if destination_type:
             return self._extension(destination_type)
-        else:
-            return self.url_extension or self._extension(self.clean_media_type)
+        return self._extension(
+            self._clean_media_type(self.external_media_type)
+        ) or self.url_extension
 
     @classmethod
     def _clean_media_type(cls, media_type):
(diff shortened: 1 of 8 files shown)


model.py
```diff
@@ -6969,8 +6969,9 @@ def extension(self, destination_type=None):
         """Try to come up with a good file extension for this representation."""
         if destination_type:
             return self._extension(destination_type)
-        else:
-            return self.url_extension or self._extension(self.clean_media_type)
+        return self._extension(
+            self._clean_media_type(self.external_media_type)
+        ) or self.url_extension
```
leonardr: If I don't switch the order of these two, then the url_extension for foo.svg returns ".svg" and we don't get a chance to override that result.

opds_import.py
```diff
@@ -943,7 +944,14 @@ def _get(self, url, headers):
...
         headers = dict(headers or {})
-        headers['Accept'] = OPDSFeed.ACQUISITION_FEED_TYPE
+
+        types = dict(
+            opds_acquisition=OPDSFeed.ACQUISITION_FEED_TYPE,
+            atom="application/atom+xml",
+            xml="application/xml",
+            anything="*/*",
+        )
+        headers['Accept'] = "%(opds_acquisition)s, %(atom)s;q=0.9, %(xml)s;q=0.8, %(anything)s;q=0.1" % types
```
leonardr: Standard Ebooks incorrectly serves its OPDS feed as XHTML. At the same time, we can't just get rid of the Accept header because without it, unglue.it will serve HTML instead of OPDS.

s3.py
```diff
@@ -212,3 +217,35 @@ def mirror_batch(self, representations):
                 if not representation.mirror_url:
                     representation.mirror_url = representation.url
                 representation.set_as_mirrored()
+
+class MockS3Response(object):
+    def __init__(self, url):
+        self.url = url
+
+class MockS3Pool(object):
```
leonardr: We have a DummyS3Uploader that mocks `mirror_batch()`. That's useful when we need a fake uploader for some other purpose, but it's useless when we need to test that S3Uploader works as designed.

model.py
```diff
@@ -7012,6 +7031,28 @@ def default_filename(self, link=None, destination_type=None):
...
+        if not self.is_image or self.clean_media_type != self.SVG_MEDIA_TYPE:
+            # Passthrough
+            return self.content_fh()
+
+        # This representation is an SVG image. We want to mirror it as
+        # PNG.
+        image = self.as_image()
+        output = StringIO()
+        image.save(output, format='PNG')
+        return StringIO(output.getvalue())
```
aslagle: Why do you need to create a new StringIO instead of returning output?

leonardr: The S3 code calls read() on the filehandle and if you write to a StringIO then .read() returns the empty string. But calling .seek(0) first does the same thing more efficiently so I'll change it.

aslagle:
:+1:

PR #349, opened 2016-07-19
If a string can be parsed as both UTF-8 and Windows-1252, prefer UTF-8 -- it's almost always the right answer.

leonardr:
This is what was _supposed_ to happen before but a missing `break` statement ensured that if a string could be interpreted as either UTF-8 or Windows-1252, we always chose Windows-1252. Now it's a big problem that needs a migration script and everything.

I've tested the migration script on the OA content server and it seems to work fine.

### DIFF
--- a/migration/20160719-fix-incorrectly-encoded-work-descriptions.py
+++ b/migration/20160719-fix-incorrectly-encoded-work-descriptions.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python
+"""Fix work descriptions that were originally UTF-8 but were incorrectly
+encoded as Windows-1252.
+"""
+import os
+import sys
+import logging
+from pdb import set_trace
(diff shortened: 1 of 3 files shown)


migration/20160719-fix-incorrectly-encoded-work-descriptions.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python
```
leonardr: Work.summary_text is the only place we set work_description, so for simplicity's sake I wrote the migration script to focus on summary_text. This means we don't have to call Work.calculate_presentation() or anything.

leonardr:
This branch fixes https://github.com/NYPL-Simplified/circulation/issues/217, and good riddance.

aslagle:
😂 👍

PR #350, opened 2016-07-21
OverdriveRepresentationExtractor.book_info_to_metadata can exclude bibliographic data or formats

aslagle:
This is for an upcoming circulation pr that updates overdrive formats when we have a delivery mechanism that's no longer available. It makes it possible to extract only bibliographic metadata or only format metadata from an overdrive response.


leonardr:
:+1:

PR #351, opened 2016-07-21
Licensepool with no pwid gets kicked out of work

leonardr:
A previous comment I wrote in Work.make_exclusive_open_access_for_permanent_work_id():

"LicensePools with no presentation edition and no PWID are left alone. (TODO: although maybe they should be kicked out.)"

Yes, they should be kicked out. There's no reason for a LicensePool that lacks a presentation edition, or whose presentation edition lacks a PWID, to belong to a Work. Our SimplyE production server has some LicensePools like this and they're blocking access to books that do have metadata and that people want to read.


aslagle:
:+1:

leonardr:
Running the migration script will need to wait for tomorrow because if it's run while the scripts are running it deadlocks.

PR #354, opened 2016-07-22
Licensepool with no pwid gets kicked out of work

leonardr:
This branch improves the previously committed migration script so that it also investigates a LicensePool with PWID "A" that is grouped together in a Work with another LicensePool with PWID "B". Those LPs should always be in separate works.

Also a LicensePool whose presentation edition has no title, which shouldn't have a work either.


aslagle:
:+1:

PR #357, opened 2016-07-27
Change search labels to match the lane that's actually searched

aslagle:
This is the easy part of https://github.com/NYPL-Simplified/circulation/issues/318

Changing the language lane names in circulation will be tricky because of localization.


leonardr:
:+1:

PR #360, opened 2016-07-27
Added names and lookup method to RightsStatus.

aslagle:
This branch fixes a problem where updating formats through the metadata layer would remove an LicensePoolDeliveryMechanism with null RightsStatus from the pool, and add one with a RightsStatus with a null URI instead. Since it changed, the old LicensePoolDeliveryMechanism would be deleted from any loans that were using it.

First, I changed RightsStatus to use a lookup method like DataSource and DeliveryMechanism since it has known values. If you lookup an unknown URI, it will return the "unknown" RightsStatus instead of creating a new one.

Since we now have LPDMs with both null RightsStatus and null URI RightsStatus, I needed a migration to make them consistent so we won't replace any of them. I changed both to "unknown". We could have a default of "in copyright" for overdrive books, but I think "unknown" is safer. I know Overdrive used to distribute public domain content although I don't see any now.

Finally, I noticed we had an empty name column so I added names for each of the URIs.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -689,10 +689,6 @@ class CirculationData(MetaToModelUtility):
             identifier=identifier_obj
         )
         if not license_pool:
-            rights_status = get_one(
-                _db, RightsStatus, uri=self.default_rights_uri
-            )
-
             last_checked = self.last_checked or datetime.datetime.utcnow()
             license_pool, is_new = LicensePool.for_foreign_id(
                 _db, data_source=self.data_source_obj,
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -797,7 +797,7 @@ class CirculationData(MetaToModelUtility):
             else:
                 resource = None
             lpdm = pool.set_delivery_mechanism(
-                format.content_type, format.drm_scheme, format.rights_uri, resource
+                format.content_type, format.drm_scheme, format.rights_uri or self.default_rights_uri, resource
             )
             new_lpdms.append(lpdm)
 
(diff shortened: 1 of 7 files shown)


metadata_layer.py
```diff
@@ -689,10 +689,6 @@ def license_pool(self, _db):
             identifier=identifier_obj
         )
         if not license_pool:
-            rights_status = get_one(
-                _db, RightsStatus, uri=self.default_rights_uri
-            )
```
aslagle: This wasn't being used. I think it was from before we moved things to CirculationData.

model.py
```diff
@@ -6227,6 +6238,17 @@ class RightsStatus(Base):
...
 
     @classmethod
+    def lookup(cls, _db, uri):
+        if not uri in cls.NAMES.keys():
+            uri = cls.UNKNOWN
+        status, ignore = get_one_or_create(
+            _db, RightsStatus, uri=uri
+        )
+        name = cls.NAMES.get(uri)
+        status.name = name
```
aslagle: Not sure where I should be setting the name. It could just be in the migration, but this will work for a new db as well.

leonardr: Can it go into create_kwargs as passed into get_one_or_create?

leonardr:
Overall this looks good. 

Overdrive's public domain section was a separate section, a little Gutenberg mirror whose books were downloadable but not exposed through their API or their normal checkout process. I think it's safe to classify every Axis 360, 3M, or Overdrive book as in-copyright by default. Even the "public domain" stuff they have is repackaged and recopyrighted versions of public domain books.

model.py
aslagle: Okay, I put it there and added it to the migration.

aslagle:
I changed Overdrive, 3M, and Axis 360 to in copyright. I haven't actually tried running these migrations yet.

leonardr:
Great. The only thing that comes to mind is that since we've switched to versioned releases, I'd like to start organizing the migration scripts by version number rather than date. So rename `20160727-add-rights-status-names.py` to `1.0.5/1-add-rights-status-names.py` and then :+1:.

aslagle:
Okay, I moved the migrations and now I've tested them too.

PR #362, opened 2016-07-29
Added media type for streaming text.

aslagle:


model.py
```diff
@@ -7356,6 +7357,11 @@ class DeliveryMechanism(Base):
     STREAMING_DRM = "Streaming"
     OVERDRIVE_DRM = "Overdrive DRM"
 
+    STREAMING_PROFILE = ";profile=http://librarysimplified.org/streaming"
```
leonardr: I realize I told you to call it this but it was in the way of a generic example. I think `http://librarysimplified.org/terms/profiles/streaming-media` would be better for the long term. It fits in with our link relations, genre names, etc.

aslagle: Fixed.

leonardr:
:+1:

PR #363, opened 2016-07-29
When updating a Contributor from a ContributorData, update VIAF, LC, and Wikipedia name, not just sort and display names.

leonardr:
I'm not sure why we didn't do this before--probably I didn't write the code because we didn't need it.

This makes an ugly situation a little uglier, as there's no guarantee that the Contributor object we're writing to comes from the same DataSource as the data we're using to modify it. But that's a known issue that we will resolve separately. If you write a ContributorData object to a Contributor it should write _all_ the data.

### DIFF
--- a/threem.py
+++ b/threem.py
@@ -431,7 +431,12 @@ class ThreeMBibliographicCoverageProvider(BibliographicCoverageProvider):
         )
 
     def process_item(self, identifier):
-        metadata = self.api.bibliographic_lookup(identifier)
+        # We don't accept a representation from the cache because
+        # either this is being run for the first time (in which case
+        # there is nothing in the cache) or it's being run to correct
+        # for an earlier failure (in which case the representation
+        # in the cache might be wrong).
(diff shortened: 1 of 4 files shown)


threem.py
```diff
@@ -431,7 +431,12 @@ def __init__(self, _db, input_identifier_types=None,
         )
 
     def process_item(self, identifier):
-        metadata = self.api.bibliographic_lookup(identifier)
+        # We don't accept a representation from the cache because
+        # either this is being run for the first time (in which case
+        # there is nothing in the cache) or it's being run to correct
+        # for an earlier failure (in which case the representation
+        # in the cache might be wrong).
+        metadata = self.api.bibliographic_lookup(identifier, max_age=0)
```
courte: Does this also need to be updated in the Overdrive & Axis 360 BibliographicCoverageProviders?

courte:
👍  🙌

PR #364, opened 2016-08-01
Change the base URL to the Cloud Library API.

leonardr:
Bibliotheca bought 3M and changed the URL. The old URL still works but let's start using the new one just to be safe.


aslagle:
:+1:

PR #365, opened 2016-08-02
Seriously allow persistent tokens, now that it's a real use case.

leonardr:
This branch adds a method called `persistent_token_create` which creates a Credential object whose `.credential` does not expire. Previously you could look up persistent tokens but there was no easy way of creating them.

I changed some language that referred to 'permanent' tokens to refer to 'persistent' tokens instead.


aslagle:
:+1:

PR #367, opened 2016-08-04
Added OPDS entry type to streaming delivery mechanisms' format types.

aslagle:


leonardr:
:+1:, add a test if it's easy, but it's OK if not, because this is tested in circulation.

PR #368, opened 2016-08-05
Move complex URNLookupController code into metadata

leonardr:
This branch removes a lot of the complex code in URNLookupController, leaving only a controller that can create an OPDS entry based on an Identifier if it can associate that Identifier with a presentation-ready Work. This is the lookup code used by circulation and content. All the complicated code is specific to the metadata wrangler, and I've created a separate class for the metadata wrangler's URNLookupController, which will go in a metadata branch.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -255,13 +255,13 @@ class URNLookupController(object):
         self.precomposed_entries = []
         self.unresolved_identifiers = []
 
-    def work_lookup(self, annotator, route_name='lookup'):
+    def work_lookup(self, annotator, route_name='lookup', urns=[]):
         """Generate an OPDS feed describing works identified by identifier."""
         urns = flask.request.args.getlist('urn')
 
         this_url = cdn_url_for(route_name, _external=True, urn=urns)
         for urn in urns:
-            self.handle_result(self.process_urn(urn))
+            self.process_urn(urn)
         self.post_lookup_hook()
 
         opds_feed = LookupAcquisitionFeed(
@@ -274,11 +274,15 @@ class URNLookupController(object):
     def permalink(self, urn, annotator, route_name='work'):
         """Look up a single identifier and generate an OPDS feed."""
         this_url = cdn_url_for(route_name, _external=True, urn=urn)
-        self.handle_result(self.process_urn(urn))
+        self.process_urn(urn)
         self.post_lookup_hook()
 
+        # A LookupAcquisitionFeed's .works is a list of (identifier,
+        # work) tuples, but an AcquisitionFeed's .works is just a
+        # list of works.
+        works = [work for (identifier, work) in self.works]
         opds_feed = AcquisitionFeed(
-            self._db, urn, this_url, self.works, annotator,
+            self._db, urn, this_url, works, annotator,
             messages_by_urn=self.messages_by_urn
         )
 
--- a/app_server.py
+++ b/app_server.py
@@ -288,11 +288,14 @@ class URNLookupController(object):
         """Turn a URN into a Work suitable for use in an OPDS feed.
 
         """
-        identifier, is_new = Identifier.parse_urn(_db, urn)
+        try:
+            identifier, is_new = Identifier.parse_urn(self._db, urn)
+        except ValueError, e:
+            identifier = None
 
         if not identifier:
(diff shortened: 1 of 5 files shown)


app_server.py
```diff
@@ -239,218 +239,108 @@ def heartbeat(self):
...
 
         this_url = cdn_url_for(route_name, _external=True, urn=urns)
         for urn in urns:
-            self.process_urn(urn, collection=collection)
-
-        # The commit is necessary because we may have registered new
-        # Identifier or UnresolvedIdentifier objects.
-        self._db.commit()
+            self.process_urn(urn, collection)
+        self.post_lookup_hook()
```
leonardr: Metadata needs to commit the database session after a lookup, but the other implementations don't.

app_server.py
```diff
@@ -239,218 +239,108 @@ def heartbeat(self):
...
 
         opds_feed = LookupAcquisitionFeed(
             self._db, "Lookup results", this_url, self.works, annotator,
             messages_by_urn=self.messages_by_urn, 
             precomposed_entries=self.precomposed_entries,
-            require_active_licensepool=require_active_licensepool
```
leonardr: Content and circulation always need an active LicensePool for the Work. Metadata always needs an active LicensePool for the Work _except_ when doing a lookup based on ISBN, which we do with a precomposed entry instead of a Work.

opds.py
```diff
@@ -1136,15 +1135,6 @@ class LookupAcquisitionFeed(AcquisitionFeed):
     which may be different from the identifier used by the Work's
     default LicensePool.
     """
-    def __init__(self, _db, title, url, works, annotator=None,
-                 messages_by_urn={}, precomposed_entries=[],
-                 require_active_licensepool=True):
-        self.require_active_licensepool=require_active_licensepool
```
leonardr: Since require_active_licensepool is gone there's no need to override the constructor.

opds.py
```diff
@@ -1196,3 +1186,93 @@ def create_entry(self, work, lane_link):
                 "I know about this work but can offer no way of fulfilling it."
             )
 
+# Mock annotators for use in unit tests.
```
leonardr: I moved these into opds.py from a test class because TestAnnotator is now used by multiple test classes.

app_server.py
```diff
@@ -239,218 +239,108 @@ def heartbeat(self):
 
 
 class URNLookupController(object):
+    """A generic controller that takes URNs as input and looks up their
+    OPDS entries.
+    """
 
-    UNRECOGNIZED_IDENTIFIER = "I've never heard of this work."
+    UNRECOGNIZED_IDENTIFIER = "This work is not in the collection."
     UNRESOLVABLE_URN = "I don't know how to get metadata for this kind of identifier."
```
courte: You can delete this constant.

app_server.py
```diff
@@ -239,218 +239,108 @@ def heartbeat(self):
...
+        """Look up a single identifier and generate an OPDS feed."""
+        this_url = cdn_url_for(route_name, _external=True, urn=urn)
         self.process_urn(urn)
+        self.post_lookup_hook()
 
-        # The commit is necessary because we may have registered new
-        # Identifier or UnresolvedIdentifier objects.
-        self._db.commit()
-
+        # A LookupAcquisitionFeed's .works is a list of (identifier,
```
courte: Based on this mixup, I looked around, and `URNLookupController#permalink` isn't being used anywhere. It's fine to leave it, but it's prime for the chopping block.

app_server.py
```diff
@@ -239,218 +239,108 @@ def heartbeat(self):
...
+            self._db, urn, this_url, works, annotator,
+            messages_by_urn=self.messages_by_urn
+        )
 
         return feed_response(opds_feed)
+    
+    def process_urn(self, urn, collection=None):
+        """Turn a URN into a Work suitable for use in an OPDS feed.
 
+        :param collection: Not used in this base class.
```
courte: Can we remove it, or do you want to leave it for potential consortial use?

leonardr: I want to keep it in the method signature because otherwise it causes a big problem in metadata. I'll change this method to take **kwargs instead.

courte:
:+1:

PR #369, opened 2016-08-05
Author matching

daryanypl:
A new apply method for ContributorData, and some explanatory docs for older code.


metadata_layer.py
```diff
@@ -182,6 +185,54 @@ def from_contribution(cls, contribution):
...
+        new_aliases = list(destination.aliases)
+        for name in [self.name] + self.aliases:
+            if name != destination.name and name not in existing_aliases:
+                new_aliases.append(name)
+                made_changes = True
+        if new_aliases != destination.aliases:
+            destination.aliases = new_aliases
+            made_changes = True
+
+        if not destination.lc:
```
leonardr: I would expect any new non-null value to overwrite any old value. Anything more detailed than 'always make the change' would be the responsibility of the replacement policy. Is it OK to change this code so it always makes the change or does that cause a problem?

util/personal_names.py
```diff
@@ -1,14 +1,36 @@
...
+from permanent_work_id import WorkIDCalculator;
+
+
 """Fallback algorithms for dealing with personal names when VIAF fails us."""
 
 def is_corporate_name(display_name):
     """Does this display name look like a corporate name?"""
+    corporations = ['National Geographic', 'Smithsonian Institution', 
+        'Verlag', 'College', 'University',  
+        'Harper & Brothers', 'Harper &amp; Brothers', 'Williams & Wilkins', 'Williams &amp; Wilkins',
```
leonardr: Rather than listing these twice it might be better to perform entity substitution on the incoming display name.

util/personal_names.py
```diff
@@ -21,3 +43,40 @@ def display_name_to_sort_name(display_name):
...
+    TODO: Consider: Further remove periods, commas, dashes, and all non-word characters.
+    TODO: consider what to do for multiple authors, like an et al or brothers grimm
+    """
+
+    # HumanName has a bug where two joiner words together cause an error.
+    # This is a quickie hack to fix that.
+    joiner_fix = re.compile('of the', re.IGNORECASE)
+    name = joiner_fix.sub('of', name)
+
+    name = HumanName(name)
```
leonardr: I don't think HumanName is imported in this file.

util/titles.py
```diff
@@ -0,0 +1,62 @@
+from nameparser import HumanName
+from nose.tools import set_trace
+import re
+
+from permanent_work_id import WorkIDCalculator;
+
+
+"""Fallback algorithms for dealing with personal names when VIAF fails us."""
+
+def is_corporate_title(display_name):
```
leonardr: How is this different from is_corporate_name in the other file?

metadata_layer.py
daryanypl: I patterned it on the contributor merge method.  I don't think it would be a problem to change the code.

util/personal_names.py
daryanypl: Sure

util/personal_names.py
```diff
@@ -1,14 +1,36 @@
...
 def is_corporate_name(display_name):
     """Does this display name look like a corporate name?"""
+    corporations = ['National Geographic', 'Smithsonian Institution', 
+        'Verlag', 'College', 'University',  
+        'Harper & Brothers', 'Harper &amp; Brothers', 'Williams & Wilkins', 'Williams &amp; Wilkins', 
+        'Estampie', 'Paul Taylor Dance', 'Gallery']
+
+    for corporation in corporations:
+        if corporation in display_name:
+            # TODO: consider making case-insensitive
```
leonardr: Case-insensitive comparison is a good idea here.

metadata_layer.py
leonardr: In private conversation we determined that the contributor merge method is the way it is because there's a preexisting determination that the two contributors are the same person and their data is interchangeable, so all that matters is filling in any gaps. In this case, we're replacing old data with new data, and the implication is the incoming stuff is better, so the new should always take precedence over the old.

util/titles.py
```diff
@@ -0,0 +1,62 @@
+from nameparser import HumanName
```
courte: This can be removed.

util/titles.py
```diff
@@ -0,0 +1,62 @@
...
+from nose.tools import set_trace
+import re
+
+from permanent_work_id import WorkIDCalculator;
+
+
+"""Fallback algorithms for dealing with personal names when VIAF fails us."""
+
+def is_corporate_title(display_name):
+    """Does this display name look like a corporate name?"""
```
courte: Would it be appropriate to change `display_name` to `title`?

metadata_layer.py

util/personal_names.py
daryanypl: done and done

util/personal_names.py
daryanypl: It's above, at line 2

util/titles.py
leonardr: I don't think so; `title` makes it sound like the title of a book.

util/titles.py
daryanypl: supported.

util/titles.py
daryanypl: Fair question.  I think, eventually, the pattern for finding a publisher's name present in the title will be different from finding the publisher as the author, but right now they're about same.

leonardr: What do you use `is_corporate_title` for? I don't see it mentioned elsewhere in the code. It doesn't seem to me like we want to do something special if the title of a book looks like the name of a company.

model.py
```diff
@@ -1885,8 +1885,15 @@ def author_contributor_tiers(cls):
 
     @classmethod
     def lookup(cls, _db, name=None, viaf=None, lc=None, aliases=None,
-               extra=None):
-        """Find or create a record for the given Contributor."""
+               extra=None, create_new=True):
```
leonardr: What's a scenario where we would pass in create_new=False? Given that we have heard about this contributor from somewhere, I think we would always want a database record to be created.

util/personal_names.py
leonardr: Got it, thanks.

leonardr:
This is pretty close to being ready. I do think you should save the new model class for another branch. A new model class is like a pet; we have to keep it around more or less forever and modifying it is a hassle. I'd prefer to hold off on the model class until we have some code that works with it and we know the schema is right.

model.py
daryanypl: When I search for contributor equivalencies, I do:
- get contributor from our db
- look them up in viaf
- for each viaf-returned cluster that is deemed the same person as looked-up contributor, I see if that cluster is in our db.  If it is, then I label it as related person.  So I don't want that lookup to create a new contributor here.

leonardr: Makes sense, thanks.

leonardr:
:+1:

PR #370, opened 2016-08-09
Convert entity-body and headers from Unicode to UTF-8 before going over the wire, to avoid triggering SSL bugs.

leonardr:
This branch fixes https://github.com/NYPL-Simplified/circulation/issues/338.

One thing this doesn't fix is the URL. It's still possible for a Unicode URL to get into the system and cause the error. I couldn't fix this easily because of the weird way the URL becomes part of `*args` within the HTTP wrapper method. I considered iterating over `*args` and converting all the Unicode to bytestrings, and I'm happy to make that change if you think it's smart.


aslagle:
:+1:

PR #372, opened 2016-08-10
Create a migration script

### LINKED ISSUE #342: Create migration script
It should use the system created to resolve #341.

courte:
Fixes #342 and NYPL-Simplified/operations#21.


scripts.py
```diff
@@ -707,6 +711,134 @@ def do_run(self):
...
+        previous = ''
+
+        for migration_file in sorted(migrations):
+            for d in directories:
+                if migration_file in migrations_by_dir[d]:
+                    # Create timestamp for previous migration, in case of
+                    # error while running this migration.
+                    if previous:
+                        set_trace()
+                        self.update_timestamp(timestamp, previous)
```
leonardr: Obvious comment: you have a set_trace() here that needs to be removed.

Less obvious comment: I think this will give the wrong behavior if there are two migration scripts from the same date. The timestamp will be updated after the first script runs; it should be updated after _all_ scripts from that date run.

leonardr:
Some ideas.

I think these are the things that are important to test but not currently tested:

* We look in the right directories for migration files.
* Files are picked up from those directories and run in the right order.
* The timestamp is updated appropriately as files are run.

It's a little trouble to test that the code/SQL in the files is actually run, but it's not too bad. You could test SQL by having it insert a timestamp, then checking for the timestamp. You could test module import by creating a mock class that stores modules as they're imported, then inspect the modules.

These changes will let you verify that the overall 'run' process works and that the timestamps are updated appropriately. You could also test this part by creating a mock class that just stores the lists of the files it "runs" in a list.

You can use Python's `tempfile` module to create a temporary directory, populate it with some files, and tell the script to look in file in that directory instead of the defaults. That will let you verify that the right files get picked up.

scripts.py
courte: The timestamp being set as the first time the date was run wouldn't be a huge problem right now because migrations are run if they are later than _or_ the same date as the provided timestamp. So if there was a failure between `20160810-2` and `20160810-3`, the whole date block (-1 to -n) would be run again.

It's not ideal, e.g. if there's a situation when someone creates a python migration that should only be run once for some reason (???), but I don't know where I could put the number otherwise without changing the Timestamp schema with an `extra` column or some such. Or maybe as hours in the datetime, lolomgno?

leonardr: I don't think any of our migration scripts up to this point would create havoc if they were run twice, but some of them could waste a lot of time.

There is a `counter` field in the `Timestamp` schema which you could use to keep track of the difference between `20160810-2` and `20160810-3`. How difficult would it be to use that?

courte: Wow. Yes. Not difficult at all.

scripts.py
```diff
@@ -72,7 +77,7 @@ def data_directory(self):
     @classmethod
     def parse_command_line(cls, _db=None, cmd_args=None):
         parser = cls.arg_parser()
-        return parser.parse_args(cmd_args)
```
courte: When nosetests was run, `parse_args` would return with an unexpected additional argument that caused tests to fail, so I replaced it with `parse_known_args`. I'm not aware of all of the impact this change might have, but there are a few number of tests for `parse_command_line`, still unbroken, and most subclasses overwrite it.

scripts.py
```diff
@@ -600,11 +612,6 @@ def arg_parser(cls):
         )
         return parser
 
-    @classmethod
```
courte: This was overwriting the parent class's method with the same code, so I nixed it.

aslagle:
What do you think someone should do to set up a brand new app? It might be good to have some kind of database initialization script that would set the timestamp to the latest migration without running any of the migrations.

scripts.py
leonardr: Nice, I think this is the right move.

scripts.py
```diff
@@ -707,6 +709,256 @@ def do_run(self):
...
+        last_run_counter = args.last_run_counter
+
+        existing_timestamp = get_one(self._db, Timestamp, service=self.name)
+        if last_run_date:
+            last_run_datetime = self.parse_time(last_run_date)
+            if existing_timestamp:
+                existing_timestamp.timestamp = last_run_datetime
+                if last_run_counter:
+                    existing_timestamp.counter = last_run_counter
+                self._db.commit()
```
leonardr: What is there to commit here?

scripts.py
```diff
@@ -707,6 +709,256 @@ def do_run(self):
...
+        WITHOUT_COUNTER = "-(\D+).(py|sql)"
+        migrations = sorted(migrations)
+
+        # The built-in sorted() method places migrations that start with
+        # just a date (e.g. '20160825-do-this.sql') AFTER migrations
+        # with a counter digit (e.g. '20160825-1-and-that.py').
+        #
+        # Therefore, if there are counter migrations, the list needs
+        # to be resorted to the more likely case: that a migration with
+        # a digit was created after a migration without one.
```
aslagle: I think it would be simpler to write a comparison function to pass into sorted.

scripts.py
```diff
@@ -707,6 +709,256 @@ def do_run(self):
...
+        """Ensures that migrations with a counter digit are sorted after
+        migrations without one.
+        """
+        WITHOUT_COUNTER = "-(\D+).(py|sql)"
+        migrations = sorted(migrations)
+
+        # The built-in sorted() method places migrations that start with
+        # just a date (e.g. '20160825-do-this.sql') AFTER migrations
+        # with a counter digit (e.g. '20160825-1-and-that.py').
+        #
```
leonardr: I think you're missing an explanatory step before the 'therefore'. I think this is it:

If we run migrations in this order, `Timestamp.counter` will be unusable. It won't be clear whether `Timestamp.counter` is None because all the migration scripts ran, or because none of them ran.

Is that right?

scripts.py
```diff
@@ -707,6 +709,256 @@ def do_run(self):
         print "Vacuumed in %.2f sec." % (b-a)
 
 
+class DatabaseMigrationScript(Script):
+    """Runs new migrations"""
+
+    name = "Database Migration"
+    MIGRATION_WITH_COUNTER = re.compile("\d{8}-(\d+)-(\w|-)+\.(py|sql)")
```
aslagle: I would make the (\w|-)+ less restrictive - I wouldn't want a migration to silently run out of order because someone put in a weird character that doesn't match this. Or, the script could throw an error up front if there's a migration file name that doesn't match the expected format.

tests/test_scripts.py
```diff
@@ -132,3 +136,309 @@ def test_make_query(self):
...
+
+            return [os.path.split(p)[1] for p in pathnames]
+
+        # Ensure that all the expected migrations from CORE are included in
+        # the 'core' directory array in migrations_by_directory.
+        core_migration_files = extract_filenames()
+
+        # At some point, August 10th, 2026 will come and go and a migration
+        # may be added and this unit test will start to fail right below.
+        # At that point, move all dates to the 2030s.
```
leonardr: It seems like `directories_by_priority` is the problem here. Can you create a subclass of `DatabaseMigrationScript` which overrides `directories_by_priority` to return the prepopulated temp directories?

scripts.py
leonardr: Good idea, seconded.

scripts.py
courte: Not quite. Timestamps for database migration are a direct reflection of the first 8 or 10 characters of the filename of the last database migration run. They don't hold any information about whether all of the migrations for a particular date have been run at all.

I'm making the assumption here that it's more likely that one dev writes a script with a timestamp ("20160825-migration.sql") and another one creates a script for the same day. The second dev may notice the overlap error when merging and choose to simply add a digit to their migration's filename instead of changing both. (This would put "20160825-migration.sql" and "20160825-1-other.sql" in the same directory.) I want to make sure that the migration created first gets run first.

It's a lot of complication around a particular edge case, but feels possible since there are two 20160602 in core/migration right now. (I'll be changing that in the next commit.)

The timestamp for the first migration wouldn't have a counter at all. The timestamp for the latter migration would, and it should be run second. As long as they're ordered correctly, neither timestamp would represent the "end" of what's been run for a particular date. If they're not ordered correctly, the second migration will be seen as a soft match for the counter-less timestamp and the first migration will be run again. I just want to make sure that "20160825-migration.sql" isn't sorted _after_ "20160825-1-" and run again because alphabet.

Hahaugh, this is terrible to try to explain via text. I'll add some kind of clarifying comment where you're suggesting while I'm restructuring this method to be passed into `sorted()`.

courte: Actually, I think switching to the compare method made it simpler to understand the order migrations are sorted in (if not the reasoning behind it).... 😕

scripts.py
```diff
@@ -739,6 +739,39 @@ def migratable_files(cls, filelist):
...
+            first_count = self.MIGRATION_WITH_COUNTER.search(first) or 0
+            second_count = self.MIGRATION_WITH_COUNTER.search(second) or 0
+            if not isinstance(first_count, int):
+                first_count = int(first_count.groups()[0])
+            if not isinstance(second_count, int):
+                second_count = int(second_count.groups()[0])
+            return first_count - second_count
+
+        return sorted(migrations, cmp=compare_migrations)
+
```
leonardr: Definitely an improvement.

aslagle:
:+1:, looks great to me

courte:
😭 😭 😭

PR #375, opened 2016-08-16
Commit the database session after processing every batch.

leonardr:
This branch creates a default implementation of `CoverageProvider.finalize_batch` that commits the database session. Without this, a script that needs to create a lot of coverage records will tie up big chunks of the database, potentially causing deadlocks, and if it crashes it will have to start from scratch.

Only one CoverageProvider in our code subclasses `finalize_batch`: `MetadataWranglerCollectionReaper` in circulation. I'll be putting up a circulation branch that makes it call the superclass implementation.

I wasn't able to add a test for this because the difference between the uncommitted session and the actual state of the database are only visible from a second database session.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -165,7 +165,6 @@ class BaseCoverageProvider(object):
             transient_failures += t
             persistent_failures += p
             records += r
-            self._db.commit()
             index += self.batch_size
         return (successes, transient_failures, persistent_failures), records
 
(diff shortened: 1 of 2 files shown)


coverage.py
```diff
@@ -165,7 +165,6 @@ def run_on_specific_identifiers(self, identifiers):
             transient_failures += t
             persistent_failures += p
             records += r
-            self._db.commit()
```
leonardr: No need to do this because it just happened in the call to `process_batch_and_handle_results`.

aslagle:
:+1:

Issue #371, opened 2016-08-10
UnfulfillableWork exception makes OPDS feed that's not a valid acquisition feed


aslagle:
When there's an UnfulfillableWork exception we create an entry with an "I know about this work but can offer no way of fulfilling it" message. Since the entry has no acquisition link, the feed isn't a valid OPDS acquisition feed. This is a problem for the web client because it thinks it's a navigation feed.

leonardr:
We discussed this and decided that it was an abuse of OPDS to use the `<entry>` tag to convey custom messages. We're going to switch to use a custom `<simplified:message>` tag instead. This will require changing opds_import.py as well as the code that generates OPDS feeds.

PR #376, opened 2016-08-19
Dont put opds message in entry

leonardr:
This branch fixes https://github.com/NYPL-Simplified/server_core/issues/371

This branch stops overloading the OPDS `<entry>` tag to convey error statuses. Instead we use a new tag, `<simplified:message>`. The only current purpose of these messages is in OPDS import processes to convey error conditions; I changed opds_import.py to manage this.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -33,6 +33,15 @@ class CoverageFailure(object):
         self.exception = exception
         self.transient = transient
 
+    def __repr__(self):
+        if self.data_source:
+            data_source = self.data_source.name
+        else:
+            data_source = None
+        return "<CoverageFailure: obj=%r data_source=%r transient=%r exception=%r>" % (
+            self.obj, data_source, self.transient, self.exception
+        )
+
     def to_coverage_record(self, operation=None):
         """Convert this failure into a CoverageRecord."""
         if not self.data_source:
--- a/model.py
+++ b/model.py
@@ -1264,6 +1264,8 @@ class Identifier(Base):
                        autocreate=True):
         """Turn a foreign ID into an Identifier."""
 
+        if not foreign_identifier_type or not foreign_id:
+            return None
         if foreign_identifier_type in (
                 Identifier.OVERDRIVE_ID, Identifier.THREEM_ID):
             foreign_id = foreign_id.lower()
@@ -1301,7 +1303,7 @@ class Identifier(Base):
     @classmethod
     def type_and_identifier_for_urn(cls, identifier_string):
         if not identifier_string:
-            return None
+            return None, None
         m = cls.GUTENBERG_URN_SCHEME_RE.match(identifier_string)
         if m:
             type = Identifier.GUTENBERG_ID
--- a/opds.py
+++ b/opds.py
@@ -633,7 +633,7 @@ class AcquisitionFeed(OPDSFeed):
 
 
     def __init__(self, _db, title, url, works, annotator=None,
-                 messages=[], precomposed_entries=[]):
+                 precomposed_entries=[]):
         """Turn a list of works, messages, and precomposed <opds> entries
         into a feed.
         """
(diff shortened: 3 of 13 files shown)


coverage.py
```diff
@@ -33,6 +33,15 @@ def __init__(self, obj, exception, data_source=None, transient=True):
...
 
+    def __repr__(self):
+        if self.data_source:
+            data_source = self.data_source.name
+        else:
+            data_source = None
+        return "<CoverageFailure: obj=%r data_source=%r transient=%r exception=%r>" % (
+            self.obj, data_source, self.transient, self.exception
+        )
+
```
leonardr: This is unrelated but it was useful while debugging.

model.py
```diff
@@ -1301,7 +1303,7 @@ class UnresolvableIdentifierException(Exception):
     @classmethod
     def type_and_identifier_for_urn(cls, identifier_string):
         if not identifier_string:
-            return None
+            return None, None
```
leonardr: An unrelated bug I found while writing tests.

opds.py
```diff
@@ -611,27 +612,9 @@ def single_entry(cls, _db, work, annotator, force_create=False):
                                  force_create=force_create)
 
     @classmethod
-    def render_messages(cls, messages_by_urn):
```
leonardr: I've checked that `render_messages` is not used anywhere except in code that's removed below.

opds.py
```diff
@@ -650,7 +633,7 @@ def facet_links(self, annotator, facets):
 
 
     def __init__(self, _db, title, url, works, annotator=None,
-                 messages_by_urn={}, precomposed_entries=[]):
+                 precomposed_entries=[]):
```
leonardr: Here's the core of the change. Instead of a separate map of message data structures that get rendered into OPDS entries, `precomposed_entries` can now contain a mixed list of <entry> tags and OPDSMessage objects.

opds_import.py
```diff
@@ -532,27 +540,18 @@ def data_detail_for_feedparser_entry(cls, entry, data_source):
...
             kwargs_meta = cls._data_detail_for_feedparser_entry(entry, data_source)
-            return identifier, kwargs_meta, failure
+            return identifier, kwargs_meta, None
         except Exception, e:
             _db = Session.object_session(data_source)
             identifier_obj, ignore = Identifier.parse_urn(_db, identifier)
-            failure = CoverageFailure(identifier_obj, traceback.format_exc(), data_source, transient=True)
+            failure = CoverageFailure(
+                identifier_obj, traceback.format_exc(), data_source,
+                transient=True
```
leonardr: Now the only time this method can return a CoverageFailure is here, when there's an exception turning the OPDS entry into a Metadata object.

opds_import.py
```diff
@@ -663,6 +662,90 @@ def summary_to_linkdata(detail):
...
+            status_code = None
+        else:
+            status_code = status_code_tag.text
+
+        return cls.coveragefailure_from_message_data(
+            data_source, urn, status_code, description
+        )
+
+    @classmethod
+    def coveragefailure_from_message_data(
```
leonardr: This is the method that gets heavily tested; the methods above are only tested via higher-level methods that call them.

tests/test_opds.py
```diff
@@ -744,23 +747,31 @@ def test_acquisition_feed_image_links_respect_cdn(self):
...
-        eq_("500", bar['simplified_status_code'])
-
+                               [], precomposed_entries=messages)
+        feed = unicode(feed)
+        for m in messages:
+            assert m.urn in feed
+            assert str(m.status_code) in feed
+            assert str(m.message) in feed
+
+    def test_precomposed_entries(self):
```
leonardr: An unrelated test I felt was missing.

tests/test_opds_import.py
```diff
@@ -131,24 +131,25 @@ def test_extract_next_links(self):
 
     def test_extract_last_update_dates(self):
         importer = OPDSImporter(self._db, DataSource.NYT)
+
```
leonardr: A bunch of these tests were changed because they do something to every <entry> tag and messages are no longer <entry> tags.

opds.py
```diff
@@ -611,27 +612,9 @@ def single_entry(cls, _db, work, annotator, force_create=False):
...
-
-            message_tag = AtomFeed.makeelement("{%s}message" % AtomFeed.SIMPLIFIED_NS)
-            message_tag.text = unicode(message)
-
-            entry.append(message_tag)
-            yield entry
-
-    @classmethod
     def error_message(cls, identifier, error_status, error_message):
         """Create a minimal OPDS entry for an error message."""
```
aslagle: This docstring shouldn't say entry anymore.

aslagle:
:+1:

aslagle:
:+1:

Issue #353, opened 2016-07-22
Search tests sometimes fail on travis.


aslagle:
Restarting the travis build will fix it. We probably just need to increase the sleep time for the search index to update.

leonardr:
I haven't noticed this in a long time so I'm assuming this very old issue has been fixed.

PR #377, opened 2016-08-23
Increased working memory for search document query

aslagle:


model.py
```diff
@@ -3923,6 +3923,11 @@ def to_search_documents(cls, works):
 
         _db = Session.object_session(works[0])
 
+        # If this is a batch of search documents, postgres needs extra working
+        # memory to process the query quickly.
+        if len(works) > 50:
+            _db.execute("set work_mem='200MB'")
```
aslagle: This is the command I got from Ashok. It only applies to the current database session.

tests/test_external_search.py
```diff
@@ -189,7 +189,7 @@ def setup(self):
 
             self.search.bulk_update(works)
 
-            time.sleep(1)
+            time.sleep(2)
```
aslagle: I want to see if this fixes https://github.com/NYPL-Simplified/server_core/issues/353

leonardr:
:joy_cat: :+1:

PR #378, opened 2016-08-23
Handle connection timeout errors during bulk search index update.

aslagle:
I was doing some indexing on the elasticsearch instance Brett set up on AWS, and it would sometimes start timing out. I made it retry the batch when that happens and added code to handle that type of error if it still fails.

It took 26 minutes to index the entire qa db on that instance, but there were a few batches that timed out twice and didn't get indexed. Using the elasticsearch on the same instance as the qa circ manager took 38 minutes (this was via jenkins), but it never timed out. My local elasticsearch took 10 minutes.


leonardr:
:+1:

PR #379, opened 2016-08-30
Handle missing data when gathering the availability link list, so long as the missing data is something relatively unimportant like the book's title.

leonardr:
This branch fixes part of https://github.com/NYPL-Simplified/circulation/issues/350

### DIFF
--- a/overdrive.py
+++ b/overdrive.py
@@ -391,7 +391,7 @@ class OverdriveRepresentationExtractor(object):
     log = logging.getLogger("Overdrive representation extractor")
 
     @classmethod
-    def availability_link_list(self, book_list):
+    def availability_link_list(cls, book_list):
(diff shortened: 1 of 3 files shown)


overdrive.py
```diff
@@ -401,8 +401,12 @@ def availability_link_list(self, book_list):
 
         products = book_list['products']
         for product in products:
-            data = dict(id=product['id'],
-                        title=product['title'],
+            if not 'id' in product:
+                cls.log.warn("No ID found in %r", product)
+                continue
+            book_id = product['id']
```
leonardr: Another bit of code further down uses `book_id`, which wasn't previously defined, so I defined it here.

aslagle:
:+1:

PR #381, opened 2016-08-30
Tweak the logs for the DatabaseMigrationScript

courte:
The logs were doing very annoying things when I was testing thing on another branch. Now they do fewer annoying things.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -11,6 +11,7 @@ from requests.exceptions import (
     HTTPError,
 )
 import sys
+import traceback
 
 from collections import defaultdict
 from nose.tools import set_trace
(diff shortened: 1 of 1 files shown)


scripts.py
```diff
@@ -930,7 +940,7 @@ def _run_migration(self, migration_path, timestamp):
                 sql = clause.read()
                 self._db.execute(sql)
         if migration_path.endswith('.py'):
-            module_name = migration_filename
+            module_name = migration_filename[:-3]
```
courte: This was causing an error for every global "import" in the migration file. Gross.

scripts.py
```diff
@@ -916,9 +916,19 @@ def run_migrations(self, migrations, migrations_by_dir, timestamp):
...
+                        self._run_migration(full_migration_path, timestamp)
+                        self._db.commit()
+                        previous = migration_file
+                    except Exception:
+                        print
+                        print "ERROR: Migration has been halted."
+                        print "%s must be migrated manually." % full_migration_path
+                        print "=" * 50
+                        print traceback.print_exc(file=sys.stdout)
+                        sys.exit(1)
```
courte: Later migrations were running when earlier migrations had caused an error. Tragedy.

aslagle:
:+1:

PR #382, opened 2016-08-30
Support identifier resolution coverage provider

courte:
This PR contains some work to support NYPL-Simplified/metadata_wrangler#84 in its bid to rid the source of all things UnresolvedIdentifier.

Beyond moving the IdentifierResolutionMonitor back into metadata, the biggest work that this branch does is create a DataSource for internal processes and a CoverageRecord operation for resolving identifiers. It also gets rid of several unused imports, including those UnresolvedIdentifier imports that were no longer in use.


model.py
```diff
@@ -647,6 +647,7 @@ class DataSource(Base):
     PLYMPTON = "Plympton"
     OA_CONTENT_SERVER = "Library Simplified Open Access Content Server"
     PRESENTATION_EDITION = "Presentation edition generator"
+    INTERNAL_PROCESSING = "Library Simplified Internal Process"
```
leonardr: My original idea was to leave DataSource blank for internal processes, but making it explicit is probably better.

leonardr:
:+1: Nice to give some code the :scissors:

PR #383, opened 2016-09-02
Support linkeddata provider bugfix

courte:
The big crux of the work here is allowing ContributorData to be applied to other ContributorData. But ultimately, I was just fixing errors as they popped up during NYPL-Simplified/metadata_wrangler#88.


metadata_layer.py
```diff
@@ -220,23 +231,27 @@ def apply(self, destination, replace=None):
             if not k in destination.extra:
                 destination.extra[k] = v
 
-        if self.lc != destination.lc:
```
courte: This and the changes below are actually ~kinda a big deal~. Instead of replacing information wholesale, we only replace contributor information that exists and leave the rest. I'd like to wait and discuss this with @daryanypl, who probably has a better sense of where contributors are being overwritten in the system.

metadata_layer.py
```diff
@@ -159,9 +159,9 @@ def __init__(self, sort_name=None, display_name=None,
         self.lc = lc
         self.viaf = viaf
         self.biography = biography
-        self.aliases = aliases
+        self.aliases = aliases or []
```
courte: Attribute changes allow `ContributorData` to feign `Contributor` persona more readily.

metadata_layer.py
```diff
@@ -313,22 +328,23 @@ def _display_name_to_sort_name(
...
-            and response.headers['Content-Type'].startswith('text/plain')):
-            sort_name = response.content.decode("utf8")
-            log.info(
-                "Canonicalizer found sort name for %r: %s => %s",
-                identifier_obj,
-                self.display_name,
-                sort_name
-            )
+
+        if isinstance(response, basestring):
```
courte: In the MetadataWrangler, the response is a string direct from the Canonicalizer -- instead of a response object.

metadata_layer.py
```diff
@@ -202,14 +203,24 @@ def apply(self, destination, replace=None):
...
 
-        if self.sort_name != destination.name:
-            destination.name = self.sort_name
+        if hasattr(destination, 'name'):
+            # For Contributor objects.
+            name_attr = 'name'
+        elif hasattr(destination, 'sort_name'):
+            # For ContributorData objects.
+            name_attr = 'sort_name'
+
```
leonardr: Now (or the near future) might be the time to fix this by renaming the database field from contributors.name to contributors.sort_name.

leonardr:
Looks good overall. The general rule we follow elsewhere in metadata_layer.py is that new data replaces old data. Applying that to this situation would mean `foo.apply(bar)` would overwrite `bar` with data from `foo`, except when _`foo`_ had missing data. I believe what you have now is the opposite: `bar`'s data is immutable once set. I don't know which we want in a real situation or even if it matters, but I'd need to be convinced that making `bar` immutable is the right call.

metadata_layer.py
courte: I vote now 👍 . Will add.

model.py
```diff
@@ -1957,6 +1957,13 @@ def lookup(cls, _db, name=None, viaf=None, lc=None, aliases=None,
 
         return contributors, new
 
+    @property
```
leonardr: These changes are good for the final version, but comment them out when you update the other components to make sure you find as much as possible of the code that needs to be changed.

model.py
```diff
@@ -1889,8 +1889,8 @@ def author_contributor_tiers(cls):
         yield cls.PERFORMER_ROLES
 
     @classmethod
-    def lookup(cls, _db, name=None, viaf=None, lc=None, aliases=None,
-               extra=None, create_new=True):
+    def lookup(cls, _db, sort_name=None, viaf=None, lc=None, aliases=None,
+               extra=None, create_new=True, name=None):
```
leonardr: Same here -- remove `name` altogether while getting the other components to work.

daryanypl:
👍

PR #384, opened 2016-09-07
Rename 3M to Bibliotheca in the database.

leonardr:
This is a minimal branch to fix https://github.com/NYPL-Simplified/circulation/issues/304

I change the database to rename 3M to its new owner, Bibliotheca. I change the DataSource and Identifier lookups so that "3M" and "3M ID" can still be used. This means old URLs should continue working. 

It's also important for me because I always misspell "Bibliotheca" (see the branch name), and because "bibliotheca" is Latin for "library", which I find very confusing.

Leonard


aslagle:
:+1:

leonardr:
Just to be 100% sure that the name is right, here's the site:

http://www.bibliotheca.com/3/index.php/en-us/

PR #388, opened 2016-09-13
Added Annotation model.

aslagle:


model.py
```diff
@@ -404,6 +404,8 @@ class Patron(Base):
     loans = relationship('Loan', backref='patron')
     holds = relationship('Hold', backref='patron')
 
+    annotations = relationship('Annotation', backref='patron')
```
leonardr: Does it make sense to add an `order_by` argument here to make sure we get most recent annotation first? Or should that go into the circ manager side where the list is actually botained?

model.py
```diff
@@ -615,6 +617,32 @@ def update(self, start, end, position):
...
+        IDLING
+    ]
+
+    __tablename__ = 'annotations'
+    id = Column(Integer, primary_key=True)
+    patron_id = Column(Integer, ForeignKey('patrons.id'), index=True)
+    identifier_id = Column(Integer, ForeignKey('identifiers.id'), index=True)
+    motivation = Column(
+        Enum(*MOTIVATIONS, name="motivation"),
+        index=True
```
aslagle: I'll change it. Even if we only ever allow a small number, we won't need migrations as we add the others.

model.py
```diff
@@ -615,6 +617,32 @@ def update(self, start, end, position):
...
+    )
+    timestamp = Column(DateTime, index=True)
+    active = Column(Boolean, default=True)
+    content = Column(Unicode)
+    target = Column(Unicode)
```
aslagle: Is it better to use the content than the timestamps? Would it work for a page's ETag to be a hash of the most recent timestamp on that page?

leonardr: Oh, yeah, it doesn't even need to be a hash. An ETag can be a timestamp.

However we should preface it with `W/` (e.g. `ETag: W/"2016-08-01:01:04:11"`) to make it a weak validator since we can't guarantee a new version is byte-for-byte identical with an old version. (Not that the content is likely to change without a timestamp update, but the way we generate the JSON-LD might change.)

aslagle: Hmm, what would the ETag be if the container is empty?

leonardr: Probably the empty string. e.g. `Etag: W/""`

PR #389, opened 2016-09-13
Allow internal process DataSource to offer licenses

courte:
A teensy branch to support NYPL-Simplified/metadata_wrangler#90.


leonardr:
:+1:

PR #390, opened 2016-09-28
"Add classification" script

leonardr:
This branch introduces a new command-line script that manually classifies an identifier under a certain subject. The main intended purpose is marking as children's books books that came in under a generic "juvenile" heading.


daryanypl:
Code looks good to me.  I would like to see classification code not auto-create new subjects without a force flag.

daryanypl:
👍

Issue #391, opened 2016-09-28
Change default identifier ID for Bibliotheca to "Bibliotheca ID" from "3M ID"


leonardr:
"3M ID" is a deprecated name for the Bibliotheca ID, but the `datasources` table still says "3M ID" is the default identifier type. This breaks any code that uses the default identifier type and doesn't check the Identifier.DEPRECATED_NAMES dictionary.

I've fixed the most egregious problems but we need a migration script that changes the default identifier type, and we need to make sure changing the default identifier type doesn't cause other problems.

PR #392, opened 2016-09-28
Added migration script.

leonardr:
This branch adds a migration script that updates `datasources` so that the primary identifier type of the Bibliotheca data source is 'Bibliotheca ID', not '3M ID'.

I've run this on the SimplyE QA database and I believe I've verified that it doesn't cause any new problems when borrowing books. I'm going to wait until tomorrow to run it on the production database, so that I have plenty of time to deal with any fallout.


leonardr:
Oh, this fixes https://github.com/NYPL-Simplified/server_core/issues/391.

On second thought, there's still a problem (to do with coverage records) that might happen on production, so I'm going to run this on production today and stick around in case there are problems.

PR #393, opened 2016-10-03
Add Lane.depth and Lane.includes_language

leonardr:
This branch defines some utility methods for determining how deep a Lane is in a hierarchy and whether or not books in a given language might be found in that lane.

`Lane.depth` already existed but it was attached to the wrong class.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -101,19 +101,6 @@ class Facets(FacetConstants):
             yield (self.COLLECTION_FACET_GROUP_NAME, self.collection)
 
     @property
-    def depth(self):
-        """How deep is this lane in this site's hierarchy?
-
-        i.e. how many times do we have to follow .parent before we get None?
-        """
-        depth = 0
-        tmp = self
-        while tmp.parent:
-            depth += 1
-            tmp = tmp.parent
-        return depth
-
-    @property
     def query_string(self):
         return "&".join("=".join(x) for x in sorted(self.items()))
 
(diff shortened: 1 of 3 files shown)


util/__init__.py
```diff
@@ -580,9 +580,9 @@ def iso_639_2_for_locale(cls, locale):
             language, place = locale.lower().split("-",1)
         else:
             language = locale
-        if language in cls.two_to_three:
+        if cls.two_to_three[language]:
```
leonardr: I was encountering a problem where `'eng' in cls.two_to_three` was True even though `cls.two_to_three['eng']` was None. I haven't been able to reproduce this problem but this change is a more robust way to express the same underlying idea as the old code.

lane.py
```diff
@@ -667,6 +667,22 @@ def set_list(field_name, value, default=None):
...
+    def includes_language(self, language):
+        """Would you expect to find books in the given language in
+        this lane?
+        """
+        if self.exclude_languages:
+            # We include all language except the ones on the exclude list.
+            if language in self.exclude_languages:
+                return False
+            else:
+                return True
```
daryanypl: So, by including an exclude_languages, I could nullify any include_language specifications?

leonardr: Yes, you can use _either_ languages or exclude_languages, not both.

daryanypl: OK, was hoping that was intentional.

daryanypl:
👍

PR #394, opened 2016-10-07
Support audienced recommendation lanes

courte:
A small refactoring to support some tricky business in NYPL-Simplified/circulation#376. Now sublanes / subgenre lanes are set via a new method, instead of being set directly inside of `Lane#__init__`.

I also stopped using implicit argument placement when calling `Lane#apply_filters` and removed a parameter that is no longer in use.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -683,7 +683,7 @@ class Lane(object):
         # We include all languages.
         return True        
 
-    def audience_list_for_age_range(self, audiences, age_range, default=[]):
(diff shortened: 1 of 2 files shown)


courte:
Updated to include a fix to a recurring error on Travis CI. We'll allow the database to decide when to delete WorkGenres instead of going our own way because it seems -- in certain situations, the database is not giving us a choice.

PR #395, opened 2016-10-11
Oneclick integration

daryanypl:
First methods to call OneClick endpoints.


oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+        "book by isbn" request is less comprehensive than the data returned 
+        by the "search titles/genres/etc." endpoint.
+
+        This method hits the "by isbn" endpoint and updates the bibliographic 
+        metadata returned by it. 
+        """
+        try:
+            response_dictionary = self.api.get_metadata_by_isbn(identifier)
+        except ValueError as error:
+            return CoverageFailure(identifier, error.message, data_source=self.output_source, transient=False)
```
leonardr: At the moment, a non-transient error from a bibliographic coverage provider almost never happens. Even if a book isn't in the library's collection (to pick an example of one way you use ValueError), it might become part of the collection later on. I would err on the side of making these errors transient.

In the future we'll resolve this by replacing the transient/permanent field with a 'retry after' field. If a book isn't in the collection the 'retry after' value would be, like, a week.

overdrive.py
```diff
@@ -854,6 +854,8 @@ def book_info_to_metadata(cls, book, include_bibliographic=True, include_formats
                         format_id, overdrive_id
                     )
 
+                # TODO: while reviewing the code, saw the medium variable
+                # is not used after declaration.  Consult with original coder and fix.
```
leonardr: Good catch. I think you can take this out. In book_info_to_metadata we use mediaType to get a value for medium.

tests/test_oneclick.py
```diff
@@ -0,0 +1,295 @@
...
+        response = api.search(mediatype='ebook', author="Alexander Mccall Smith", title="Tea Time for the Traditionally Built")
+        response_dictionary = response.json()
+        eq_(1, response_dictionary['pageCount'])
+        eq_(u'Tea Time for the Traditionally Built', response_dictionary['items'][0]['item']['title'])
+
+    def test_get_all_available_through_search(self):
+        api = MockOneClickAPI(self._db)
+        datastr, datadict = self.get_data("response_search_five_items_1.json")
+        api.queue_response(status_code=200, content=datastr)
+
```
leonardr: We discussed in person the conundrum of why this test works given your get_all_available_through_search implementation.

tests/test_oneclick.py
```diff
@@ -0,0 +1,295 @@
...
+        response_list = api.get_ebook_availability_info()
+        eq_(u'9781420128567', response_list[0]['isbn'])
+        eq_(False, response_list[0]['availability'])
+
+
+
+class TestOneClickRepresentationExtractor(OneClickTest):
+    def test_availability_info(self):
+        # TODO: the download links for items become available on checkout request
+        # See if need to go around this.
```
leonardr: I don't think this test corresponds to any code you wrote.

tests/test_oneclick.py
```diff
@@ -0,0 +1,295 @@
...
+            [
+                (Identifier.ISBN, "9780307378101"),
+            ],
+            sorted(ids)
+        )
+
+        # Available formats.      
+        [epub] = sorted(metadata.circulation.formats, key=lambda x: x.content_type)        
+        eq_(Representation.EPUB_MEDIA_TYPE, epub.content_type)       
+        eq_(DeliveryMechanism.ONECLICK_DRM, epub.drm_scheme)
```
leonardr: Once you change your format code we expect this would become ADOBE_DRM.

tests/test_oneclick.py
```diff
@@ -0,0 +1,295 @@
...
+            OneClickBibliographicCoverageProvider, self._db, [],
+            oneclick_api=self.api
+        )
+        assert isinstance(script.provider, 
+                          OneClickBibliographicCoverageProvider)
+        eq_(script.provider.api, self.api)
+
+
+    def test_invalid_or_unrecognized_guid(self):
+        # A bad or malformed GUID can't get coverage.
```
leonardr: There's some leftover code from Overdrive here. Can you change these error messages to the messages you see from One Click when you send an invalid ISBN?

Also change "guid" to "ISBN".

tests/test_oneclick.py
```diff
@@ -0,0 +1,295 @@
...
+
+        failure = self.provider.process_item(identifier)
+        assert isinstance(failure, CoverageFailure)
+        eq_(False, failure.transient)
+        eq_("ISBN [bad guid] is not in library [library_id_123]'s catalog.", failure.exception)
+
+
+    def test_process_item_creates_presentation_ready_work(self):
+        # Test the normal workflow where we ask Axis for data,
+        # Axis provides it, and we create a presentation-ready work.
```
leonardr: One Click -- comment needs updating.

tests/test_oneclick.py
```diff
@@ -0,0 +1,295 @@
...
+        self.api.queue_response(200, content=datastr)
+        
+        # Here's the book mentioned in response_isbn_found_1.
+        identifier = self._identifier(identifier_type=Identifier.ONECLICK_ID)
+        identifier.identifier = '9780307378101'
+
+        # This book has no LicensePool.
+        eq_(None, identifier.licensed_through)
+
+        # Run it through the Axis360BibliographicCoverageProvider
```
leonardr: Another comment that needs to be updated.

tests/test_oneclick.py
```diff
@@ -0,0 +1,295 @@
...
+        [lpdm] = pool.delivery_mechanisms
+        eq_('application/epub+zip (OneClick DRM)', lpdm.delivery_mechanism.name)
+
+        # A Work was created and made presentation ready.
+        eq_('Tea Time for the Traditionally Built', pool.work.title)
+
+        # TODO: determine what endpoints we want to call to get the licensing info 
+        # and make the work presentation-ready.
+        #eq_(True, pool.work.presentation_ready)
+
```
leonardr: The work should be made presentation ready once process_item() finishes. A work is made presentation ready once we have enough bibliographic information. There has to be _a_ license pool, but that pool doesn't have to actually have any copies. See test_overdrive.py for an example.

leonardr (changes requested):
This is good stuff! I have a lot of comments but they're pretty minor.

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+    # a complete response returns the json structure with more data fields than a basic response does
+    RESPONSE_VERBOSITY = {0:'basic', 1:'compact', 2:'complete', 3:'extended', 4:'hypermedia'}
+
+    log = logging.getLogger("OneClick API")
+
+    def __init__(self, _db, library_id=None, username=None, password=None, 
+        remote_stage=None, base_url=None, basic_token=None):
+        self._db = _db
+        (env_library_id, env_username, env_password, 
+         env_remote_stage, env_base_url, env_basic_token) = self.environment_values()
```
daryanypl: Checked millenium_patron.py, the functionality appears very similar, although the method name is environment_values.  Updated a bit to clean up, and changed method name to reflect its configuration-using nature.

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+    def request(self, url, method='get', extra_headers={}, data=None,
+                params=None, verbosity='complete'):
+        """Make an HTTP request, acquiring/refreshing a bearer token
+        if necessary.
+        """
+        if verbosity not in self.RESPONSE_VERBOSITY.values():
+            verbosity = self.RESPONSE_VERBOSITY[2]
+
+        headers = dict(extra_headers)
+        headers['Authorization'] = "Basic " + self.token
```
daryanypl: done

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+        """
+        page = 0;
+        response = self.search(availability='available', verbosity=self.RESPONSE_VERBOSITY[0])
+
+        respdict = response.json()
+        if not respdict:
+            raise IOError("OneClick availability response not parseable - has no respdict.")
+
+        if not ('pageIndex' in respdict and 'pageCount' in respdict):
+            raise IOError("OneClick availability response not parseable - has no page counts.")
```
daryanypl: Looked into it, the RemoteInitiatedServerError is defined in circulation_exceptions.  I'll keep it IOError in core, and then call the better error from the circulation oneclick code.

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+        page_count = respdict['pageCount']
+
+        while (page_count > (page_index+1)):
+            page_index += 1
+            response = self.search(availability='available', verbosity=self.RESPONSE_VERBOSITY[0], page_index=page_index)
+            tempdict = response.json()
+            if not ('items' in tempdict):
+                raise IOError("OneClick availability response not parseable - has no next dict.")
+            item_interest_pairs = tempdict['items']
+            respdict['items'].append(item_interest_pairs)
```
daryanypl: fixed

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+        url = "%s/libraries/%s/media/%s" % (self.base_url, str(self.library_id), identifier_string) 
+
+        response = self.request(url)
+
+        respdict = response.json()
+        if not respdict:
+            # should never happen
+            raise IOError("OneClick isbn search response not parseable - has no respdict.")
+
+        if "message" in respdict:
```
daryanypl: Good catch.  I'm not sure.  Most other errors would carry a non-200 http code.  But yes, there's no guarantee that's the only non-fatal type of error.

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+class OneClickRepresentationExtractor(object):
+    """ Extract useful information from OneClick's JSON representations. """
+    DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ" #ex: 2013-12-27T00:00:00Z
+    DATE_FORMAT = "%Y-%m-%d" #ex: 2013-12-27T00:00:00Z
+
+    log = logging.getLogger("OneClick representation extractor")
+
+    # TODO: the formats/encoding/drm information needs confirmation from OneClick
+    oneclick_formats = {
+        "ebook-epub-oneclick" : (
```
daryanypl: Would we ever use the information from more than one method?

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+        metadata = Metadata(
+            data_source=DataSource.ONECLICK,
+            primary_identifier=primary_identifier,
+        )
+
+        if include_bibliographic:
+            title = book.get('title', None)
+            series_name = book.get('seriesName', None)
+            # ignored for now
+            series_position = book.get('seriesPosition', None)
```
daryanypl: done

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+                language = LanguageCodes.string_to_alpha_3(book['language'])
+            else:
+                language = 'eng'
+
+            contributors = []
+            if 'authors' in book:
+                authors = book['authors']
+                for author in authors.split(";"):
+                    sort_name = author.strip()
+                    roles = [Contributor.AUTHOR_ROLE]
```
daryanypl: narrators are different, and are represented in their own field.  authors are lumped all together into one field, no roles.  I tried to look for mentions of translators or introduction writers or editors in the authors fields, but no go.  I'll confirm with them that there are no specific author roles.

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+                    contributor = ContributorData(sort_name=sort_name, roles=roles)
+                    contributors.append(contributor)
+
+            subjects = []
+            if 'genres' in book:
+                # example: "FICTION / Humorous / General"
+                genres = book['genres']
+            for genre in genres.split("/"):
+                subject = SubjectData(
+                    type=Subject.ONECLICK, identifier=genre.strip(),
```
daryanypl: done

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+            # A cover and its thumbnail become a single LinkData.
+            # images come in small (ex: 71x108px), medium (ex: 95x140px), 
+            # and large (ex: 128x192px) sizes
+            if 'images' in book:
+                images = book['images']
+                for image in images:
+                    if image['name'] == "large":
+                        image_data = cls.image_link_to_linkdata(image['url'], Hyperlink.IMAGE)
+                    if image['name'] == "medium":
+                        thumbnail_data = cls.image_link_to_linkdata(image['url'], Hyperlink.THUMBNAIL_IMAGE)
```
daryanypl: done

overdrive.py
daryanypl: done

tests/test_oneclick.py
daryanypl: fixed

oneclick.py
leonardr: I think when I gave you this advice I was looking at my refactored millenium_patron.py, which hasn't landed yet. Leave this alone for now; it's not vital and we can change it later.

oneclick.py
leonardr: That makes sense.

oneclick.py
```diff
@@ -0,0 +1,600 @@
...
+        response = self.request(url)
+
+        respdict = response.json()
+        if not respdict:
+            # should never happen
+            raise IOError("OneClick isbn search response not parseable - has no respdict.")
+
+        if "message" in respdict:
+            # can happen if searched for item that's not in library's catalog
+            raise ValueError("ISBN [%s] is not in library [%s]'s catalog." % (identifier_string, str(self.library_id)))
```
leonardr: What does the message say when you search  for an item not in the catalog? If there's a distinctive "not in the catalog" message, then check for that and either return None or raise a distinctive exception (which the calling code can check for). You can raise a more generic exception for other types of message.

oneclick.py
leonardr: OK, great. I know Bibliotheca treats all roles as author roles. One Click might have made a similar mistake but it's less likely since they handle narrators specially.

oneclick.py
```diff
@@ -0,0 +1,696 @@
...
+        return HTTP.request_with_timeout(
+            method, url, headers=headers, data=data,
+            params=params, **kwargs
+        )
+
+
+    def request(self, url, method='get', extra_headers={}, data=None,
+                params=None, verbosity='complete'):
+        """Make an HTTP request, acquiring/refreshing a bearer token
+        if necessary.
```
leonardr: Take out this part of the comment. It's left over from Overdrive code.

leonardr:
:+1: as a first step. Transactional API support can come next.

tests/test_oneclick.py
daryanypl: done

leonardr (approved):

PR #396, opened 2016-10-18
Circulation event types

aslagle:
This branch adds constants for the new circulation event types we're tracking, and organizes them by the source of the event. We can add additional client event types as needed.

Once these are getting sent to google analytics on production, we won't be able to migrate old event labels if we change any of the constants, so we should think about what they should be. I think the distinction between circ manager events and distributor events could be confusing.

### DIFF
--- a/model.py
+++ b/model.py
@@ -6409,6 +6409,18 @@ class CirculationEvent(Base):
     TYPE = u"event"
 
     # The names of the circulation events we recognize.
+    # They may be sent to third-party analytics services
+    # as well as used locally.
+
+    # Events that happen in a circulation manager.
+    NEW_PATRON = u"circulation_manager_new_patron"
+    CM_CHECKOUT = u"circulation_manager_check_out"
+    CM_CHECKIN = u"circulation_manager_check_in"
+    CM_HOLD_PLACE = u"circulation_manager_hold_place"
+    CM_HOLD_RELEASE = u"circulation_manager_hold_release"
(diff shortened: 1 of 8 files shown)


model.py
```diff
@@ -6409,6 +6409,18 @@ class CirculationEvent(Base):
     TYPE = u"event"
 
     # The names of the circulation events we recognize.
+    # They may be sent to third-party analytics services
+    # as well as used locally.
+
+    # Events that happen in a circulation manager.
+    NEW_PATRON = u"circulation_manager_new_patron"
```
aslagle: I'll use this one once the authenticator refactoring is done.

model.py
```diff
@@ -6409,6 +6409,18 @@ class CirculationEvent(Base):
...
+    CM_CHECKOUT = u"circulation_manager_check_out"
+    CM_CHECKIN = u"circulation_manager_check_in"
+    CM_HOLD_PLACE = u"circulation_manager_hold_place"
+    CM_HOLD_RELEASE = u"circulation_manager_hold_release"
+    CM_FULFILL = u"circulation_manager_fulfill"
+
+    # Events that we hear about from a distributor.
     CHECKOUT = u"check_out"
```
leonardr: As per Slack conversation suggest you prepend "distributor_" to these to reduce the possibility of confusion between the two different types of events.

leonardr:
:+1:

PR #397, opened 2016-10-19
Support for authenticator refactor

leonardr:
This branch makes some minor changes to core in support of a large authentication refactoring project in the circulation manager.

Notably, some logic about what it means for patrons to have borrowing privileges has been consolidated underneath Patron.

### DIFF
--- a/model.py
+++ b/model.py
@@ -472,7 +472,7 @@ class Patron(Base):
             # This patron has never been synced.
             return True
         
-        now = datetime.datetime.now()
+        now = datetime.datetime.utcnow()
         if self.has_borrowing_privileges:
             # A patron who has borrowing privileges gets synced every twelve
             # hours. Their account is unlikely to change rapidly.
--- a/model.py
+++ b/model.py
@@ -399,7 +399,7 @@ class Patron(Base):
     authorization_expires = Column(Date, index=True)
 
     # Outstanding fines the user has, if any.
-    fines = Column(Unicode)
+    fines = Column(Numeric)
 
     loans = relationship('Loan', backref='patron')
     holds = relationship('Hold', backref='patron')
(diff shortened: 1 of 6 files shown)


model.py
```diff
@@ -6520,13 +6567,15 @@ def lookup_by_temporary_token(cls, _db, data_source, type, token):
         return credential
 
     @classmethod
-    def temporary_token_create(self, _db, data_source, type, patron, duration):
+    def temporary_token_create(
+            self, _db, data_source, type, patron, duration, value=None
+    ):
```
leonardr: The token we get from an OAuth provider is now stored as a temporary token. This means we need to make it possible to create a temporary token with a given value (which we got from the OAuth provider), rather than with a randomly generated value.

testing.py
```diff
@@ -70,13 +70,6 @@ def package_setup():
     # Initialize basic database data needed by the application.
     _db = Session(connection)
     SessionManager.initialize_data(_db)
-
-    # Create the patron used by the dummy authentication mechanism.
-    # TODO: This can be probably be moved to circulation.
-    get_one_or_create(
```
leonardr: This has been moved to circulation.

model.py
```diff
@@ -458,15 +458,62 @@ def can_borrow(self, work, policy):
...
+        hours. Patrons who lack borrowing privileges can always stand
+        to be synced, since their privileges may have just been
+        restored.
+        """
+        if not self.last_external_sync:
+            # This patron has never been synced.
+            return True
+        
+        now = datetime.datetime.now()
+        expired_at = self.last_external_sync + datetime.timedelta(hours=12)
```
aslagle: We might want a way to configure the 12 hours.

model.py
aslagle: I think I originally didn't use this because it seemed like the temporary tokens were single-use - lookup_by_temporary_token expires them. Maybe we could rename that method.

leonardr: I'll address this in my next branch.

PR #398, opened 2016-10-20
Support isbns with works through oclc

courte:
Allows `LicensePool.set_presentation_edition` to take an external list of editions.

Support NYPL-Simplified/metadata_wrangler#100 to fix NYPL-Simplified/metadata_wrangler#66.

### DIFF
--- a/model.py
+++ b/model.py
@@ -889,8 +889,8 @@ class DataSource(Base):
     # LicensePool, editions are processed based on their data source,
     # in the following order:
     #
-    # [all other sources] < [source of the license pool] < [metadata
-    # wrangler] < [library staff] < [manual intervention]
+    # [all other sources] < [metadata wrangler] < [source of the license
+    # pool] < [library staff] < [manual intervention]
     #
     # This list keeps track of the high-priority portion of that
     # ordering.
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -5698,13 +5698,13 @@ def sort_key(edition):
...
+            editions = self.identifier.primarily_identifies
+        return sorted(editions, key=sort_key)
 
-        return sorted(self.identifier.primarily_identifies, key=sort_key)
-
-
-    # TODO:  policy is not used in this method.  Removing argument
-    # breaks many-many tests, and needs own branch.
-    def set_presentation_edition(self, policy=None):
+    def set_presentation_edition(self, external_editions=None):
```
leonardr: This needs a test, and a docstring explaining what's up with `external_editions`. Right now I have to guess what `external_editions` is doing. I believe it's some other list of Editions that we've decided could be components of a presentation edition, even though they're not directly connected to `self.identifier`.

courte:
This branch may not be necessary at all with the changes in-progress in NYPL-Simplified/metadata_wrangler#100. If they are, I'll def make the recommended changes.

courte:
@leonardr finally, finally ready for a fresh look 👀

PR #399, opened 2016-10-24
Patron fines are numeric

leonardr:
This branch changes `Patron.fines` from a string field to a numeric field. Patron fines are always sent to us as a string, but parsing that string is the job of the authentication plugin. By the time the data is stored in the database it needs to be a number that can be easily compared against other numbers.

### DIFF
--- a/model.py
+++ b/model.py
@@ -6580,7 +6580,7 @@ class Credential(Base):
             return None
 
     @classmethod
-    def lookup_by_temporary_token(cls, _db, data_source, type, token):
+    def lookup_and_expire_temporary_token(cls, _db, data_source, type, token):
         """Look up a temporary token and expire it immediately."""
         credential = cls.lookup_by_token(_db, data_source, type, token)
         if not credential:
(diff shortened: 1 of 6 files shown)


model.py
```diff
@@ -6580,7 +6580,7 @@ def lookup_by_token(self, _db, data_source, type, token,
             return None
 
     @classmethod
-    def lookup_by_temporary_token(cls, _db, data_source, type, token):
+    def lookup_and_expire_temporary_token(cls, _db, data_source, type, token):
```
leonardr: Made this change as per Amy's suggestion on an earlier branch. I don't know if this method is used at all outside of its test.

util/__init__.py
```diff
@@ -1091,3 +1092,19 @@ def process_data(cls, data, bigrams):
 english_bigrams = Bigrams(Counter())
 english_bigrams.proportional = Counter(english_bigram_frequencies)
 
+
+class MoneyUtility(object):
```
leonardr: A simple class that will use whatever heuristics we end up coming up with to turn a string into a Money object.

model.py
aslagle: :+1:

PR #400, opened 2016-10-25
Support for Basque PD eLiburutegia collection

aslagle:
Includes: 
- New data source
- New identifier type
- BIC classifier

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -46,6 +46,7 @@ class Classifier(object):
     OVERDRIVE = "Overdrive"
     THREEM = "3M"
     BISAC = "BISAC"
+    BIC = "BIC"
     TAG = "tag"   # Folksonomic tags.
 
     # Appeal controlled vocabulary developed by NYPL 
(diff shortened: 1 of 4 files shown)


classifier.py
```diff
@@ -86,7 +87,6 @@ class Classifier(object):
     SIMPLIFIED_GENRE = "http://librarysimplified.org/terms/genres/Simplified/"
     SIMPLIFIED_FICTION_STATUS = "http://librarysimplified.org/terms/fiction/"
 
-    # TODO: This is currently set in model.py in the Subject class.
```
aslagle: This was no longer true, it's set at the bottom of this file.

model.py
```diff
@@ -3211,9 +3215,9 @@ def target_age_string(self):
         upper = self.target_age.upper
         if not upper and not lower:
             return ""
-        if lower and not upper:
+        if lower and upper is None:
             return str(lower)
-        if upper and not lower:
+        if upper and lower is None:
```
aslagle: We have 0-8 as the age range for early grades in the metadata spreadsheet for the basque titles, and it was showing up as "8".

PR #405, opened 2016-11-01
Empty the exception of a previously-unsuccessful CoverageRecord if it becomes successful

courte:
I thought we had 23k+ unresolved identifiers being ignored in the Metadata Wrangler (exception: "No work done yet"). It turns out they're all successfully covered; it's just that their exceptions hadn't been removed.


leonardr:
nice :sunglasses:

PR #406, opened 2016-11-01
Made it possible to initialize an ExternalSearchIndex with arguments and no configuration.

aslagle:
This is a small branch to support the open access search lambda. It can't have environment variables, so I'm initializing the search index by passing in a url and index name instead of having it read the config file. The arguments were already there, but url wasn't being used. 

I might come up with a better way to handle configuration for the lambda, but this seems like a good change anyway.


aslagle:
I'm also using this in the script that creates the index, so the url and index can be arguments to the script.

PR #407, opened 2016-11-02
Delegated patron identifiers

leonardr:
This branch introduces the DelegatedPatronIdentifier model class: a place for library A to map a patron identifier from library B onto some other identifier (for which library A is the naming authority). Currently the only use case for this is Adobe IDs.

### DIFF
--- a/model.py
+++ b/model.py
@@ -4,7 +4,6 @@ from collections import (
     Counter,
     defaultdict,
 )
-import contextlib
 from lxml import etree
 from nose.tools import set_trace
 import cairosvg
@@ -329,7 +328,6 @@ def get_one(db, model, on_multiple='error', **kwargs):
 
 def get_one_or_create(db, model, create_method='',
                       create_method_kwargs=None,
-                      create_method_kwargs_methods=None,
                       **kwargs):
     one = get_one(db, model, **kwargs)
     if one:
--- a/model.py
+++ b/model.py
@@ -4,6 +4,7 @@ from collections import (
     Counter,
     defaultdict,
 )
+import contextlib
 from lxml import etree
 from nose.tools import set_trace
 import cairosvg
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -6633,6 +6633,69 @@ def persistent_token_create(self, _db, data_source, type, patron):
...
+    type = Column(String(255), index=True)
+    library_uri = Column(String(255), index=True)
+
+    # This is the ID the foreign library gives us when referring to
+    # this patron.
+    patron_identifier = Column(String(255), index=True)
+
+    # This is the identifier we made up for the patron. This is what the
+    # foreign library is trying to look up.
+    delegated_identifier = Column(String)
```
aslagle: Why isn't this column indexed?

model.py
```diff
@@ -6633,6 +6633,69 @@ def persistent_token_create(self, _db, data_source, type, patron):
...
+
+    __table_args__ = (
+        UniqueConstraint('type', 'library_uri', 'patron_identifier'),
```
aslagle: should there also be a constraint on type, library_uri, delegated_identifier? That seems more important than patron_identifier.

model.py
leonardr: We never have to look up a patron based on the delegated identifier. We only look up based on `foreign_identifier`.

To make the example concrete, BPL needs to know a patron's Adobe ID, so they send us the ID they used to register that ID. If they have the Adobe ID there's no need for them to look anything up.

aslagle: Doesn't Adobe give us the delegated_identifier to look up?

model.py
leonardr: `get_one_or_create` does a lookup based on ('type', 'library_uri', 'patron_identifier'), so that one really needs to be unique. (type, library_uri, delegated_identifier) should be unique for Adobe ID but I don't know if it has to be unique in general.

model.py
aslagle: ok, I think I get it

aslagle (approved):

aslagle (approved):

PR #408, opened 2016-11-02
Set explicit elasticsearch version.

aslagle:
The latest version of the elasticsearch python package doesn't work with our code, and is meant to be used with newer versions of elasticsearch.


courte:
Yeah, this will definitely need to be done in circulation, too.

PR #410, opened 2016-11-04
Patron input script

leonardr:
This branch creates a `PatronInputScript` abstract class that functions like the `IdentifierInputScript` in that its job is to take certain command-line arguments (patrons in this case), turn them into data model objects, and process them in some undefined way.

### DIFF
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -139,6 +139,26 @@ class TestPatronInputScript(DatabaseTest):
         )
         eq_([], parsed.patrons)
 
+
+    def test_do_run(self):
+        class MockPatronInputScript(PatronInputScript):
+            def process_patron(self, patron):
+                patron.processed = True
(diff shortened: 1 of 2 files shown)


tests/test_scripts.py
```diff
@@ -99,6 +100,68 @@ def test_parse_command_line_no_identifiers(self):
...
+            def process_patron(self, patron):
+                patron.processed = True
+        p1 = self._patron()
+        p2 = self._patron()
+        p3 = self._patron()
+        p3.processed = False
+        p1.authorization_identifier = self._str
+        p2.authorization_identifier = self._str
+        cmd_args = [p1.authorization_identifier, p2.authorization_identifier]
+        stdin = MockStdin(p2.authorization_identifier)
```
aslagle: Did you mean to have p2 in both cmd_args and stdin? Might be good to test one that's only in stdin.

PR #412, opened 2016-11-07
Add query generated lane

courte:
Moving QueryGeneratedLane from `circulation/api/lanes.py`, so I can use it in the content server. This supports NYPL-Simplified/content_server#89.

The only changes from circulation are noted in comments below.

Helps resolve NYPL-Simplified/content_server#87.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -539,7 +539,7 @@ class AcquisitionFeed(OPDSFeed):
         for args in cls.facet_links(annotator, facets):
             OPDSFeed.add_link_to_feed(feed=feed.feed, **args)
 
-        if len(works) > 0:
+        if len(works) > 0 and pagination.has_next_page:
             # There are works in this list. Add a 'next' link.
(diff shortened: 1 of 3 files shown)


opds.py
```diff
@@ -539,7 +539,7 @@ def page(cls, _db, title, url, lane, annotator=None,
         for args in cls.facet_links(annotator, facets):
             OPDSFeed.add_link_to_feed(feed=feed.feed, **args)
 
-        if len(works) > 0:
+        if len(works) > 0 and pagination.has_next_page:
```
courte: This change has the potential to break controller tests in circulation, the metadata wrangler, and may also impact the clients. It removes the "next" link if there aren't any more works in the query.

aslagle: Yeah, this should be tested on the clients before it's released just in case something crashes when the next link is missing. I do think it's a good change though.

PR #413, opened 2016-11-07
Removed dependency on postgres NumericRange from classifier.

aslagle:
I needed to run the search code on AWS lambda, and we can't install postgres there, so I changed the classifier to use tuples for target age instead of NumericRange. The tuples get converted to a NumericRange in model before being stored in the db.


leonardr:
As the author of the original code, I approve this message.

PR #414, opened 2016-11-09
Core work for Feedbooks import

leonardr:
This branch makes some changes in core in preparation for the ability to import books from FeedBooks. 

Up to this point we have handled a new datasource by defining a new DataSource for the new source of books. I've decided to put a stop to that, by changing opds_import.py so that it automatically creates DataSource objects when it encounters an unrecognized DataSource in an OPDS feed (or as `data_source_name` in the constructor).

So the first time you run a FeedBooks import (or an import of FeedBooks books from the content server), the FeedBooks DataSource will show up in your database.

I also made some improvements to the KeywordClassifier so that it could handle some of the tags specific to FeedBooks, such as "non-fiction" spelled with a dash.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -246,7 +246,9 @@ class OPDSImporter(object):
                 # to this item.
                 self.log.error("Error importing an OPDS item", exc_info=e)
                 identifier, ignore = Identifier.parse_urn(self._db, key)
-                data_source = DataSource.lookup(self._db, self.data_source_name)
+                data_source = DataSource.lookup(
+                    self._db, self.data_source_name, autocreate=True
+                )
                 failure = CoverageFailure(identifier, traceback.format_exc(), data_source=data_source, transient=False)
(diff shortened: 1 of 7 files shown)


opds_import.py
```diff
@@ -347,7 +351,12 @@ def extract_feed_data(self, feed, feed_url=None):
         """Turn an OPDS feed into lists of Metadata and CirculationData objects, 
         with associated messages and next_links.
         """
-        data_source = DataSource.lookup(self._db, self.data_source_name)
+        # This is one of these cases where we want to create a
+        # DataSource if it doesn't already exist. This way you don't have
+        # to predefine a DataSource for every source of OPDS feeds.
+        data_source, ignore = get_one_or_create(
+            self._db, DataSource, name=self.data_source_name
+        )
```
leonardr: You're right, this is old code from before I added autocreate to lookup.

PR #415, opened 2016-11-10
Add the drm: namespace to cached entries.

leonardr:
This branch fixes https://github.com/NYPL-Simplified/circulation/issues/408

We could do without this branch by regenerating the OPDS cache for every entry in our collections. I think it's less work to make sure the drm: namespace is present whenever an <entry> tag is about to be used as the root tag of an XML document.

Also, this probably isn't the only time this will happen, and it's good to have a precedent in the code base.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -831,10 +831,8 @@ class OPDSImporter(object):
                 ratings.append(v)
         data['measurements'] = ratings
 
-        entry_rights = parser._xpath(entry_tag, 'rights')
-        
         data['links'] = cls.consolidate_links([
-            cls.extract_link(link_tag, feed_url, entry_rights)
+            cls.extract_link(link_tag, feed_url)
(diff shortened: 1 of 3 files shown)


opds_import.py
```diff
@@ -831,8 +831,10 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None):
                 ratings.append(v)
         data['measurements'] = ratings
 
+        entry_rights = parser._xpath(entry_tag, 'rights')
```
aslagle: This looks unrelated, did you add this file by accident?

leonardr: :blush:

aslagle:
This looks fine to me. It would be nice to update the cached entry when you find one without the NS, but I don't know how you'd know where it was.

leonardr:
That would be nice but I don't think it's worth the effort. We only have to do this when someone gets a single-entry feed.

PR #417, opened 2016-11-15
Feedbooks import

leonardr:
Framework changes necessary to support imports from Feedbooks's idiosyncratic OPDS feed.

* It's possible for an open-access link to have a rights status of "in copyright". (In Feedbooks's case this is a book that is public domain in Europe but not in the US). Links like these do not get mirrored to S3.

* A LicensePool can have a DeliveryMechanism that contains an open-access link with a status of "in copyright" is not treated as an open-access LicensePool. I decided to do things this way (as opposed to changing the link relation or removing the DeliveryMechanism altogether) because it focuses the problem in the rights status. If the rights status ever changes it will be easy to reevaluate and make the LicensePool open-access.

* When creating a new DataSource at runtime it's possible to set its `.offers_licenses`.

* It's possible for the feedparser code to decline to contribute a value it's expected to contribute, with the expectation that the elementtree code will fill it in later. For Feedbooks we do this with the entry's default rights URI.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -698,13 +698,22 @@ class CirculationData(MetaToModelUtility):
                 # An open-access link or open-access rights implies a FormatData object.
                 open_access_link = (link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD and link.href)
                 # try to deduce if the link is open-access, even if it doesn't explicitly say it is
-                open_access_rights_link = (link.media_type in Representation.BOOK_MEDIA_TYPES 
-                                           and link.href
-                                           and (link.rights_uri or self.default_rights_uri) in RightsStatus.OPEN_ACCESS)
+                rights_uri =  link.rights_uri or self.default_rights_uri
+                open_access_rights_link = (
+                    link.media_type in Representation.BOOK_MEDIA_TYPES 
+                    and link.href
+                    and rights_uri in RightsStatus.OPEN_ACCESS
+                )
                 
                 if open_access_link or open_access_rights_link:
                     rights_uri = link.rights_uri or self.default_rights_uri
-                    if open_access_link and not rights_uri in RightsStatus.OPEN_ACCESS:
+                    if (open_access_link
+                        and rights_uri != RightsStatus.IN_COPYRIGHT
+                        and not rights_uri in RightsStatus.OPEN_ACCESS):
+                        # We don't know exactly what's going on here but
+                        # the link said it was an open-access book
+                        # and the rights URI doesn't contradict it,
+                        # so treat it as a generic open-access book.
                         rights_uri = RightsStatus.GENERIC_OPEN_ACCESS
                     format_found = False
                     for format in self.formats:
@@ -804,7 +813,10 @@ class CirculationData(MetaToModelUtility):
         """Does this Circulation object have an associated open-access link?"""
         return any(
             [x for x in self.links 
-             if x.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD and x.href]
+             if x.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD
+             and x.href
+             and x.rights_uri != RightsStatus.IN_COPYRIGHT
+            ]
         )
 
 
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -706,7 +706,6 @@ class CirculationData(MetaToModelUtility):
                 )
                 
                 if open_access_link or open_access_rights_link:
-                    rights_uri = link.rights_uri or self.default_rights_uri
                     if (open_access_link
                         and rights_uri != RightsStatus.IN_COPYRIGHT
                         and not rights_uri in RightsStatus.OPEN_ACCESS):
@@ -932,7 +931,19 @@ class CirculationData(MetaToModelUtility):
                         as_of=self.last_checked
                     )
 
-        made_changes = made_changes or changed_licenses
+        # Changes to the delivery mechanisms may have changed the work's
+        # open-access status.
+        old_open_access = pool.open_access                    
+        for lpdm in pool.delivery_mechanisms:
+            if lpdm.rights_status.uri in RightsStatus.OPEN_ACCESS:
+                pool.open_access = True
+                break
+        else:
+            pool.open_access = False
+        open_access_status_changed = (old_open_access != pool.open_access)
+                    
+        made_changes = (made_changes or changed_licenses
+                        or open_access_status_changed)
 
         return pool, made_changes
 
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -1307,7 +1307,6 @@ class Metadata(MetaToModelUtility):
         self.log.info(
             "APPLYING METADATA TO EDITION: %s",  self.title
         )
-
         fields = self.BASIC_EDITION_FIELDS+['permanent_work_id']
         for field in fields:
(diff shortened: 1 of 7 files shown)


metadata_layer.py
```diff
@@ -690,13 +698,21 @@ def links(self, arg_links):
...
+                    link.media_type in Representation.BOOK_MEDIA_TYPES 
+                    and link.href
+                    and rights_uri in RightsStatus.OPEN_ACCESS
+                )
                 
                 if open_access_link or open_access_rights_link:
-                    rights_uri = link.rights_uri or self.default_rights_uri
-                    if open_access_link and not rights_uri in RightsStatus.OPEN_ACCESS:
+                    if (open_access_link
+                        and rights_uri != RightsStatus.IN_COPYRIGHT
```
leonardr: This change is tested in the content server but I don't have a test of it in core.

opds_import.py
```diff
@@ -596,7 +607,8 @@ def _data_detail_for_feedparser_entry(cls, entry, metadata_data_source):
             if circulation_data_source_name:
                 _db = Session.object_session(metadata_data_source)
                 circulation_data_source = DataSource.lookup(
-                    _db, circulation_data_source_name, autocreate=True
+                    _db, circulation_data_source_name, autocreate=True,
+                    offers_licenses=True
```
leonardr: If we're importing from an OPDS feed and it's a data source that's not in the list of data sources (e.g. not the metadata wrangler), it's a safe bet that it offers licenses.

opds_import.py
```diff
@@ -671,15 +683,32 @@ def summary_to_linkdata(detail):
...
     @classmethod
-    def rights_uri_for_entry(cls, entry):
+    def rights_uri(cls, rights_string):
         """Determine the URI that best encapsulates the rights status of
         the downloads associated with this book.
+        """
+        return RightsStatus.rights_uri_from_string(rights_string)
 
-        :return: A URI
+    @classmethod
```
leonardr: This is an annoying mess -- both the feedparser entry-creation code and the elementtree entry-creation code need to retrieve the rights URI for an entry, if there is one. In the Feedbooks case, the feedparser code can't determine the rights URI (it depends on information unavailable to feedparser) so the elementtree code has to come in later and fix it.

opds_import.py
```diff
@@ -199,12 +199,14 @@ class OPDSImporter(object):
...
-                 identifier_mapping=None, mirror=None, http_get=None):
+                 identifier_mapping=None, mirror=None, http_get=None,
+                 metadata_client=None
+    ):
         self._db = _db
         self.log = logging.getLogger("OPDS Importer")
         self.data_source_name = data_source_name
         self.identifier_mapping = identifier_mapping
-        self.metadata_client = SimplifiedOPDSLookup.from_config()
+        self.metadata_client = metadata_client or SimplifiedOPDSLookup.from_config()
```
leonardr: We don't currently have an end-to-end test that would have caused this code to call out to metadata.librarysimplified.org, but in my content branch I added code that did cause network requests during tests.

model.py
```diff
@@ -6441,6 +6445,9 @@ def rights_uri_from_string(cls, rights):
             return RightsStatus.CC_BY_NC_SA
         elif rights == 'cc by-nc-nd':
             return RightsStatus.CC_BY_NC_ND
+        elif (rights in RightsStatus.OPEN_ACCESS
+              or rights == RightsStatus.IN_COPYRIGHT):
+            return rights
```
leonardr: If the rights string is one of our rights URIs, we should recognize it.

opds_import.py
```diff
@@ -938,13 +995,26 @@ def extract_link(cls, link_tag, feed_url=None):
...
+            rights_uri = entry_rights_uri
         if feed_url and not urlparse(href).netloc:
             # This link is relative, so we need to get the absolute url
             href = urljoin(feed_url, href)
-        return LinkData(rel=rel, href=href, media_type=media_type, rights_uri=rights_uri)
+        return cls.make_link_data(rel, href, media_type, rights_uri)
+
+    @classmethod
+    def make_link_data(cls, rel, href=None, media_type=None, rights_uri=None,
+                       content=None):
```
leonardr: This code moved out into its own method so it can be overridden in a subclass.

opds_import.py
```diff
@@ -404,7 +407,9 @@ def extract_feed_data(self, feed, feed_url=None):
             metadata[internal_identifier.urn] = Metadata(**combined_meta)
 
             # form the CirculationData that would correspond to this Metadata
-            c_data_dict = m_data_dict.get('circulation')
+            c_circulation_dict = m_data_dict.get('circulation')
+            xml_circulation_dict = xml_data_dict.get('circulation', {})
+            c_data_dict = self.combine(c_circulation_dict, xml_circulation_dict)
```
leonardr: This is where we allow elementtree to define its own arguments to the CirculationData constructor. It's not tested in core but I have a test in my content branch.

opds_import.py
```diff
@@ -446,8 +448,13 @@ def combine(self, d1, d2):
...
         for k, v in d2.items():
-            if k in new_dict and isinstance(v, list):
-                new_dict[k].extend(v)
+            if k in new_dict:
+                if isinstance(v, list):
+                    new_dict[k].extend(v)
+                elif isinstance(v, dict):
+                    new_dict[k] = self.combine(new_dict[k], v)
+                elif new_dict[k] is None:
+                    new_dict[k] = v
```
leonardr: This is where we allow the feedparser code to abstain from providing a value it would normally provide (by returning None). If elementtree provides a value it will override the None value.

opds_import.py
```diff
@@ -354,12 +356,13 @@ def extract_feed_data(self, feed, feed_url=None):
         # DataSource if it doesn't already exist. This way you don't have
         # to predefine a DataSource for every source of OPDS feeds.
         data_source = DataSource.lookup(
-            self._db, name=self.data_source_name, autocreate=True
+            self._db, name=self.data_source_name, autocreate=True,
+            offers_licenses=True
```
aslagle: This code can be used with data sources that don't offer licenses. If it's only the MetadataWranglerCoverageProvider, that's ok since its data source has already been created. But I think it could be confusing.

opds_import.py
aslagle: In this case we know it offers licenses because the OPDS feed said so. It's the data_source_name that's passed into the class that could be the metadata wrangler. 

You could move up the comment right below this, and take out the below line that sets offers_licenses to true. It's probably safer to not change the licensing status of an existing data source, even if an OPDS feed told us it offers licenses.

opds_import.py
```diff
@@ -671,15 +681,32 @@ def summary_to_linkdata(detail):
...
+        return cls.rights_uri(rights)
+        
+    @classmethod
+    def rights_uri_from_entry_tag(cls, entry):
+        """Extract a rights string from an lxml <entry> tag.
 
+        :return: A rights URI.
+        """
+        for name in ('atom:rights', 'rights'):
+            rights = OPDSXMLParser._xpath1(entry, 'rights')
```
aslagle: did you mean to use name on this line?

leonardr: Yes, and the fact that everything still works makes me think line 705 isn't necessary.

opds_import.py
leonardr: I've changed this in a way I think you'll like.

PR #418, opened 2016-11-15
Support lanes in static feeds

courte:
**Active ingredient:** Basic change to `AcquisitionFeed.groups` to support NYPL-Simplified/content_server#91.

Inactive ingredients:
- Removes unused `lane_link` argument from `AcquisitionFeed.create_entry` and its submethods.
- Uses unicode strings in `CachedFeed` creation to avoid database warnings.


leonardr:
:+1:

Issue #419, opened 2016-11-21
Feedbooks books take precedent over Gutenberg books


leonardr:
Currently when we import a Feedbooks book, if there's a Gutenberg edition of the same work, the Gutenberg edition takes precedence. The Feedbooks edition needs to take precedence.

The simplest way to solve this is to introduce a DataSource.FEEDBOOKS constant and make it higher priority than Gutenberg.

PR #420, opened 2016-11-21
Feedbooks books are better than Gutenberg books

leonardr:
I was hoping to avoid the creation of a DataSource.FEEDBOOKS constant, but it seems like the quickest way to get Feedbooks books properly set up in the content server.

A more flexible solution would be to create a `default_quality` field on `DataSource`. When you create a `DataSource` you would be allowed to specify its `default_quality`. We'd be able to get rid of OPEN_ACCESS_SOURCE_PRIORITY and instead calculate priority based on comparisons of `default_quality`.


leonardr:
This fixes https://github.com/NYPL-Simplified/server_core/issues/419

PR #423, opened 2016-11-23
Use limit for fast_query_count

courte:
Queries with a limit attached were returning a count for their non-limited length before. This branch fixes that.


util/__init__.py
```diff
@@ -22,6 +22,9 @@ def batch(iterable, size=1):
 
 def fast_query_count(query):
     """Counts the results of a query without using super-slow subquery"""
+    if query._limit:
+        return query._limit
+
```
leonardr: Shouldn't it be the smaller of `query._limit` and the actual number of results?

leonardr:
:+1:

PR #424, opened 2016-11-28
oneclick catalog populate and delta now working

daryanypl:


oneclick.py
```diff
@@ -367,6 +369,96 @@ def get_metadata_by_isbn(self, identifier):
...
+                e = "Could not extract metadata from OneClick data: %r" % catalog_item
+                make_note = CoverageFailure(identifier, e, data_source=self.output_source, transient=True)
+
+            # convert IdentifierData into Identifier, if can
+            identifier, made_new = metadata.primary_identifier.load(_db=self._db)
+            if identifier and not made_new:
+                [work] = Work.from_identifiers(self._db, [identifier])
+                if work:
+                    for cr in work.coverage_records:
+                        self._db.delete(cr)
```
leonardr: This is okay but it would be better to change core, either to create a `Work.delete()` method that takes care of this stuff or to create a cascade-on-delete so that this stuff is deleted automatically (the way WorkGenres are).

oneclick.py
```diff
@@ -367,6 +369,96 @@ def get_metadata_by_isbn(self, identifier):
...
+                    for pool in work.license_pools:
+                        if pool.presentation_edition:
+                            for contribution in pool.presentation_edition.contributions:
+                                self._db.delete(contribution)
+                            self._db.delete(pool.presentation_edition.contributions[0])
```
leonardr: Once you delete all the presentation edition's contributions, why is there still one left to be deleted?
Why is it necessary to delete the presentation edition's contributions, but not the presentation edition itself?
I don't think you should be deleting the presentation edition at all--it might be used by another license pool for the same book.

oneclick.py
```diff
@@ -367,6 +369,96 @@ def get_metadata_by_isbn(self, identifier):
...
+                make_note = CoverageFailure(identifier, e, data_source=self.output_source, transient=True)
+
+            # convert IdentifierData into Identifier, if can
+            identifier, made_new = metadata.primary_identifier.load(_db=self._db)
+            if identifier and not made_new:
+                [work] = Work.from_identifiers(self._db, [identifier])
+                if work:
+                    for cr in work.coverage_records:
+                        self._db.delete(cr)
+                    for pool in work.license_pools:
```
leonardr: I think you should check the data source and only delete a license pool that comes from OneClick. It's possible that another distributor could be providing the same book, and you wouldn't want to delete that LicensePool just because OneClick stopped providing the book.

oneclick.py
```diff
@@ -367,6 +369,96 @@ def get_metadata_by_isbn(self, identifier):
...
+
+    def update_metadata(self, catalog_item):
+        metadata = OneClickRepresentationExtractor.isbn_info_to_metadata(catalog_item)
+
+        if not metadata:
+            # generate a CoverageFailure to let the system know to revisit this book
+            # TODO:  if did not create a Work, but have a CoverageFailure for the isbn, 
+            # check that re-processing that coverage would generate the work.
+            e = "Could not extract metadata from OneClick data: %r" % catalog_item
+            make_note = CoverageFailure(identifier, e, data_source=self.output_source, transient=True)
```
leonardr: This is the second time you've written this code; it should be refactored.

oneclick.py
```diff
@@ -589,6 +682,7 @@ def isbn_info_to_metadata(cls, book, include_bibliographic=True, include_formats
                     subjects.append(subject)
 
             # audience options are: adult, beginning-reader, childrens, young-adult
+            # NOTE: audience can be set to "Adult" while publisher is "HarperTeen".
```
leonardr: I don't think this is a NOTE situation. Either it's a bug in core/classifier.py (in which case you should file the bug) or there's some other aspect of the metadata that outweighs the "HarperTeen" -> "Young Adult" mapping in classifier.py.

My guess is that OneClick is sending HarperTeen as the publisher, but classifier.py is expecting Harper Collins as the publisher and HarperTeen as the imprint. If this is true, the best solution would be to change `weigh_metadata` in classifier.py to collapse the distinction between 'publisher' and 'imprint' -- always check both fields, since we can't be sure which one might say "HarperTeen".

scripts.py
```diff
@@ -766,6 +767,66 @@ def run(self):
...
+    @classmethod
+    def arg_parser(cls):
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            '--library-id', 
+            help='OneClick library id.  Ex: 1931 for Wethersfield Public Library.'
+        )
+        parser.add_argument(
+            '--mock', 
+            help='OneClick library id.  Ex: 1931 for Wethersfield Public Library.'
```
leonardr: This is the wrong description for this argument.

scripts.py
```diff
@@ -766,6 +767,66 @@ def run(self):
...
+    OneClick-subscribed library catalog.
+    """
+
+    def __init__(self, _db=None, cmd_args=None):
+        super(OneClickDeltaScript, self).__init__(_db=_db)
+
+
+    def do_run(self):
+        print "OneClickDeltaScript.do_run"
+        logging.info("OneClickDeltaScript.do_run().")
```
leonardr: Either take this out or use `self.log.info` instead of `logging.info`.

tests/test_scripts.py
```diff
@@ -596,3 +608,142 @@ def test_autocreate(self):
...
+        return data, json.loads(data)
+
+
+    def test_parse_command_line(self):
+        cmd_args = ["--library-id", "1931", "--mock", "y"]
+        parsed = OneClickImportScript.parse_command_line(
+            _db=self._db, cmd_args=cmd_args
+        )
+        eq_('1931', parsed.library_id)
+        eq_('y', parsed.mock)
```
leonardr: Most of these tests aren't necessary if you make the changes I suggest above.

oneclick.py
daryanypl: done

scripts.py
```diff
@@ -766,6 +767,66 @@ def run(self):
...
 
+class OneClickImportScript(Script):
+    """Import all books from a OneClick-subscribed library catalog."""
+
+    @classmethod
+    def arg_parser(cls):
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            '--library-id', 
+            help='OneClick library id.  Ex: 1931 for Wethersfield Public Library.'
```
daryanypl: Good catch, I do get it from config.

oneclick.py
daryanypl: Sorry, I don't mean this is set in our systems, I mean this is conflicting data coming in from OneClick.  So, I wanted to reflect this finding, so that if we're surprised by how things end up classified on our end, we'll be one step further in analyzing the cause.  I'll edit the note to clarify.

scripts.py
daryanypl: done

scripts.py
daryanypl: ok

oneclick.py
```diff
@@ -682,7 +667,7 @@ def isbn_info_to_metadata(cls, book, include_bibliographic=True, include_formats
                     subjects.append(subject)
 
             # audience options are: adult, beginning-reader, childrens, young-adult
-            # NOTE: audience can be set to "Adult" while publisher is "HarperTeen".
+            # NOTE: In OneClick metadata, audience can be set to "Adult" while publisher is "HarperTeen".
```
leonardr: I filed https://github.com/NYPL-Simplified/server_core/issues/426 to address this.

leonardr:
:+1:

PR #427, opened 2016-12-01
Moved circulation-specific patron code into the circulation manager.

leonardr:
The code removed from this branch is added to circulation in https://github.com/NYPL-Simplified/circulation/pull/427


courte (approved):
👍

PR #429, opened 2016-12-01
Support static feeds with featured works

courte:
This branch is mainly to further support the creation of static feeds in the content server. Relevant changes include:
  - Adding an S3 bucket variable to Configuration for static feeds.
  - Helping `S3Uploader` understand what to do with such a bucket (creating URLs and what not).

It also includes some cleaning up in `model.py` because why not:
  - A few unused or duplicate imports have been removed.
  - Database constants in each class are now unicode, which should avoid some SQLAlchemy warnings about inserting non-unicode into Unicode columns of the database.
  - `LicensePool.set_presentation_edition` no longer accepts a `policy` parameter (which it didn't use), and the tests and test helper methods are adjusted accordingly.

Supports NYPL-Simplified/content_server#97.


config.py
```diff
@@ -147,6 +147,7 @@ class Configuration(object):
     S3_SECRET_KEY = "secret_key"
     S3_OPEN_ACCESS_CONTENT_BUCKET = "open_access_content_bucket"
     S3_BOOK_COVERS_BUCKET = "book_covers_bucket"
+    S3_STATIC_FEED_BUCKET = "static_feed_bucket"
```
leonardr: I feel like all the static feed stuff should go into your content server branch, but I could be wrong.

s3.py
```diff
@@ -111,9 +127,9 @@ def cover_image_url(cls, data_source, identifier, filename=None,
         return root + "%s/%s/%s" % tuple(args)
 
     @classmethod
-    def feed_url(cls, filename, extension='.opds', open_access=True):
+    def feed_url(cls, filename, extension='.xml', open_access=True):
         """The path to the hosted file for an OPDS feed with the given filename"""
-        root = cls.content_root(open_access)
+        root = cls.static_feed_root(open_access)
```
leonardr: Was it wrong to use content_root for this before?

leonardr:
Looks like a good cleanup branch, there's just one thing I'm confused on.

s3.py
courte: It was a temporary measure while the dev S3 bucket was being created, though I never actually intended to put anything in the content bucket.

leonardr:
:+1:

Issue #430, opened 2016-12-06
Bibliotheca IDs (rarely) contain slash characters, which crashes the API


leonardr:
At some point a Bibliotheca ID "hhot389/open_book" got inserted into the NYPL production circulation manager. I don't know whether it came from Bibliotheca or from some source internal to NYPL.  "hhot389" is a real book that is in NYPL's collection. 

The presence of the slash in "hhot389/open_book" breaks any attempt to use the identifier in the Bibliotheca API. (We get a 404 error.) Since we bundle 25 IDs together when doing lookups, this means we can't look up whatever other Bibliotheca IDs we bundled it with.

I've deleted "hhot389/open_book" from the database, which should address the problem, but since we don't know how the ID got into the database in the first place, it could come back. A slash is an invalid character in a Bibliotheca ID and we should write code that prevents such Identifiers from being created.

aslagle:
It sounds like this happened accidentally when we were figuring out how to create analytics events. The apps create the analytics url based on the work's url, and it looks like we tried appending the "open_book" event type without adding in the "analytics" part of the route, so it got interpreted as part of the identifier.

leonardr:
Thanks, that makes sense.

PR #431, opened 2016-12-06
Reject a Bibliotheca ID that contains a slash.

leonardr:
This fixes https://github.com/NYPL-Simplified/server_core/issues/430.

I also started using Identifier.BIBLIOTHECA_ID instead of (the identical) Identifier.THREEM_ID constant within model.py.


leonardr:
It depends on the API. A quick look at the APIs tells me we should also prohibit commas in Bibliotheca and Axis 360 IDs, because we separate them with commas when looking up a group of IDs. With Overdrive I don't think there are any special requirements because there's no way to deal with multiple Overdrive IDs.

PR #432, opened 2016-12-13
Editions issue402

daryanypl:


scripts.py
```diff
@@ -686,6 +686,7 @@ def process_work(self, work):
 
     def do_run(self):
         super(WorkConsolidationScript, self).do_run()
+        set_trace()
```
leonardr: :skull_and_crossbones:

leonardr:
This looks solid, just get rid of the set_trace call in the script.

leonardr:
:+1:

PR #435, opened 2016-12-15
Support sorted static feed lanes

courte:
This branch fixes a couple of issues around static feeds (and probably feeds in general).

- It fixes some bad math in Pagination.has_next_page which may be the reason why the clients can only scroll down 100 books. It also fixes some bad math in the tests of this method. 😢 
- It makes using CachedFeeds optional in AcquisitionFeed.groups and AcquisitionFeed.pages, to avoid overwriting of similarly named feeds when creating Static feed options.

Supports NYPL-Simplified/content_server#104.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -367,12 +367,16 @@ class Pagination(object):
 
     @property
     def has_next_page(self):
-        """Returns boolean reporting whether pagination is done for a query"""
+        """Returns boolean reporting whether pagination is done for a query
+
+        This method only returns valid information _after_ self.apply
+        has been run on a query.
+        """
         if self.query_size is None:
             return True
         if self.query_size==0:
             return False
-        return (self.offset+1) * self.size < self.query_size
+        return self.offset + self.size < self.query_size
 
     def apply(self, q):
         """Modify the given query with OFFSET and LIMIT."""
--- a/opds.py
+++ b/opds.py
@@ -302,6 +302,9 @@ class Annotator(object):
                 active_license_pool = p
         return active_license_pool
 
+    def sort_works_for_groups_feed(self, works, **kwargs):
+        return works
+
 
 class VerboseAnnotator(Annotator):
     """The default Annotator for machine-to-machine integration.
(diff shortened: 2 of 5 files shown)


lane.py
```diff
@@ -367,12 +367,16 @@ def previous_page(self):
...
+
+        This method only returns valid information _after_ self.apply
+        has been run on a query.
+        """
         if self.query_size is None:
             return True
         if self.query_size==0:
             return False
-        return (self.offset+1) * self.size < self.query_size
+        return self.offset + self.size < self.query_size
```
courte: I don't know what I was trying to do before, but it was very wrong. :(

opds.py
```diff
@@ -302,6 +302,9 @@ def active_licensepool_for(cls, work):
                 active_license_pool = p
         return active_license_pool
 
+    def sort_works_for_groups_feed(self, works, **kwargs):
```
courte: This is a hook that, by default, doesn't impact the works at all. It is overwritten in `StaticFeedAnnotator`

opds.py
```diff
@@ -751,13 +766,11 @@ def create_entry(self, work, even_if_no_license_pool=False,
     def _create_entry(self, work, license_pool, edition, identifier,
                       force_create=False, use_cache=True):
         xml = None
-        cache_hit = False
```
courte: This variable isn't being used.

opds.py
```diff
@@ -407,24 +410,27 @@ class AcquisitionFeed(OPDSFeed):
     FEED_CACHE_TIME = int(Configuration.get('default_feed_cache_time', 600))
 
     @classmethod
-    def groups(cls, _db, title, url, lane, annotator, cache_type=None,
-               force_refresh=False, use_materialized_works=True):
+    def groups(cls, _db, title, url, lane, annotator, use_cache=True,
+               cache_type=None, force_refresh=False, use_materialized_works=True):
```
leonardr: I'm not totally against this, but rather than adding a new argument to this method I think it would be better to add to the behavior of cache_type.

Either make the default value of cache_type be the same as CachedFeed.GROUPS_TYPE, or (more reliable) define a constant `NO_CACHE = object()` which you can pass in to turn off caching.

lane.py
leonardr: I probably reviewed this change and I'm bad at math. :(

leonardr:
:+1: Let's get this in as soon as possible, because that code could indeed be causing major problems.

leonardr:
Now we're :+1:ing

PR #438, opened 2016-12-25
Load Lane.works with all licensepools

courte:
This change fixes a problem not in `Lane` itself, but in its subclasses that apply additional filtering.

Because `contains_eager` applies filtering to joined collections, later filtering on `Identifier` in classes like `StaticFeedBaseLane` (in NYPL-Simplified/content_server:lanes.py) meant that other `LicensePools` on the work were filtered out of the session because their Identifiers weren't specifically requested. Later, when an active `Work.license_pools` was requested, it wasn't available in the session and an error was raised. By using `joinedload` instead, the entire collection is pulled into the db session regardless of later filtering.

See the SQLAlchemy docs for more information: http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#using-contains-eager-to-load-a-custom-filtered-collection-result.

Fixes NYPL-Simplified/content_server#98.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -23,6 +23,7 @@ from sqlalchemy import (
 from sqlalchemy.orm import (
     contains_eager,
     defer,
+    joinedload,
     lazyload,
 )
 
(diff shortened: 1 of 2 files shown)


leonardr:
:+1: If it's easy to add a test (a self-contained case that gives unexpected results with the old code but works now), then do it, but don't spend all day working on a test.

courte:
Test added!

leonardr:
Nice. :heart_decoration:

PR #439, opened 2017-01-04
Oneclick prod testing

daryanypl:
Added setting of display_name and quality, and delivery mechanisms, and changed OneClickAPI to read configuration.

### DIFF
--- a/oneclick.py
+++ b/oneclick.py
@@ -40,6 +40,7 @@ from metadata_layer import (
     IdentifierData,
     LinkData,
     Metadata,
+    ReplacementPolicy,
     SubjectData,
 )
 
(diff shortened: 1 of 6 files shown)


oneclick.py
```diff
@@ -856,6 +867,9 @@ def update_metadata(self, catalog_item, identifier=None, metadata_replacement_po
             identifier, metadata, metadata_replacement_policy=metadata_replacement_policy
         )
 
+        if not isinstance(result, CoverageFailure):
+            self.handle_success(identifier)
+
```
leonardr: I would expect `handle_success` to be called by the code that calls `process_item`, like `BibliographicCoverageProvider.process_batch`. I get why you're doing this -- you don't call `process_batch` because OneClick data isn't batched. You get the whole thing at once so you might as well process it all at once, in `populate_all_catalog`.

But in `populate_all_catalog` you already have an `if not isinstance(result, CoverageFailure):`. How about calling `handle_success` there? Then you could get rid of this, and if you ever do call `process_batch` then `handle_success` won't be called twice.

oneclick.py
```diff
@@ -658,17 +670,19 @@ def isbn_info_to_metadata(cls, book, include_bibliographic=True, include_formats
...
             if 'narrators' in book:
                 narrators = book['narrators']
                 for narrator in narrators.split(";"):
                     sort_name = narrator.strip()
-                    roles = [Contributor.NARRATOR_ROLE]
-                    contributor = ContributorData(sort_name=sort_name, roles=roles)
-                    contributors.append(contributor)
+                    if sort_name:
+                        roles = [Contributor.NARRATOR_ROLE]
+                        contributor = ContributorData(sort_name=sort_name, display_name=sort_name, roles=roles)
```
leonardr: You shouldn't use a sort name as a display name. It's better to leave it blank and let the metadata wrangler fill it in. Is the absence of a display name causing problems?

daryanypl: Yes, I had to start filling it in as part of getting my catalog to show up in the web app.

oneclick.py
daryanypl: I was trying to keep the code that modifies the db together with the code that reports the db modified.  Under what circumstances would handle_success be called twice?

daryanypl: Got it, moved the code outside.

leonardr:
This looks good. I'm not giving :+1: yet because your circulation branch indicates there's a commit in this branch you haven't pushed yet (40e081541507ae34346ebd34fea4c95edb26e514).

leonardr:
Oh, there it is. :+1:

PR #440, opened 2017-01-04
DRM Device ID

leonardr:
This branch creates a new model class, DRMDeviceID, which will be used to store the information used in the [DRM Device Management Protocol](https://github.com/NYPL-Simplified/Simplified/wiki/DRM-Device-Management).

### DIFF
--- a/model.py
+++ b/model.py
@@ -6674,6 +6674,13 @@ class DelegatedPatronIdentifier(Base):
     # foreign library is trying to look up.
     delegated_identifier = Column(String)
 
+    # One DelegatedPatronIdentifier can keep track of many
+    # DRMDeviceIdentifiers.
+    relationship(
+        "DRMDeviceIdentifier", backref=backref("delegated_patron_identifier",
+                                               lazy='joined')
+    )
+    
     __table_args__ = (
         UniqueConstraint('type', 'library_uri', 'patron_identifier'),
     )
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -6558,6 +6558,11 @@ class Credential(Base):
     credential = Column(String)
     expires = Column(DateTime)
 
+    # One Credential can have many associated DRMDeviceIdentifiers.
+    drm_device_identifiers = relationship(
+        "DRMDeviceIdentifier", backref=backref("credential", lazy='joined')
+    )
+
```
leonardr: I associate DRMDeviceIdentifier with Credential rather than Patron because your Adobe device IDs are associated with your Adobe ID, not your library account. I want to support tracking DRM device identifiers for (let's say) a patron's accounts with different URMS stores. This requires that one Patron keep multiple lists of DRMDeviceIdentifiers, one for each DRM-related Credential.

Although the Adobe ID itself is in DelegatedPatronIdentifier, I associate DRMDeviceIdentifier with Credential rather than with DelegatedPatronIdentifier. That's because I want each library to keep track of the DRM device IDs of its own patrons, and most libraries won't have access to DelegatedPatronIdentifiers.

problem_details.py
```diff
@@ -2,32 +2,42 @@
 from util.http import INTEGRATION_ERROR
 from flask.ext.babel import lazy_gettext as _
 
+# Generic problem detail documents that recapitulate HTTP errors.
+# call detailed() to add more specific information.
+
```
leonardr: I needed these for a number of cases in circ manager. Since they are generic versions of HTTP errors I put them in core.

testing.py
```diff
@@ -422,6 +424,37 @@ def _complaint(self, license_pool, type, source, detail, resolved=None):
...
+                    type=None, patron=None):
+        data_source = DataSource.lookup(self._db, data_source_name)
+        type = type or self._str
+        patron = patron or self._patron()
+        credential, is_new = Credential.persistent_token_create(
+            self._db, data_source, type, patron
+        )
+        return credential
+    
+    def _delegated_patron_identifier(
```
leonardr: In an earlier version of this branch, I had DRMDeviceIdentifier associated with DelegatedPatronIdentifier, and I needed this helper method to get the tests to work. I moved away from that design later, but this helper method seems generally useful, so I kept it.

PR #442, opened 2017-01-05
Added option to ReplacementPolicy to modify Representation content before mirroring.

aslagle:
This supports a forthcoming content server pr that will replace css in imported epub files.


leonardr:
Lovely. :+1:

PR #445, opened 2017-01-06
Support suppressed covers

courte:
This branch allows us to hide covers that we don't want in the feeds for whatever reason. It will hide the cover on a list of Works and any Works or Editions that use the same cover.

I'm not sold on the design yet--particularly the fact that it's done either by batch or individually by `Work`--but it's an initial fix for NYPL-Simplified/content_server#109.

Right now, it suppresses the cover entirely. In the future, we may want to search for a different cover image if more are available on the same or related Editions.

### DIFF
--- a/model.py
+++ b/model.py
@@ -1681,8 +1681,7 @@ class Identifier(Base):
             _db, identifier_ids, Hyperlink.IMAGE)
         images = images.join(Resource.representation)
         images = images.filter(Representation.mirrored_at != None).\
-            filter(Representation.mirror_url != None).\
-            filter(Resource.suppressed==False)
+            filter(Representation.mirror_url != None)
         images = images.all()
 
         champions = Resource.best_covers_among(images)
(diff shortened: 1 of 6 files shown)


model.py
```diff
@@ -4796,22 +4795,17 @@ class Resource(Base):
 
     # The average of human-entered values for the quality of this
     # resource.
-    voted_quality = Column(Float)
+    voted_quality = Column(Float, default=float(0))
```
leonardr: I think these require a migration script. 

```
ALTER TABLE ALTER COLUMN voted_quality SET DEFAULT 0.0;
ALTER TABLE ALTER COLUMN votes_for_quality SET DEFAULT 0;
```

Something like that.

leonardr:
Looks good with the migration script.

When we have an admin interface that acts like a CMS for the collection we'll probably need to keep track of all the activity on the Resources, but there's no need to change it now.

leonardr:
:+1:

PR #447, opened 2017-01-17
Support custom list static feeds

courte:
A lot of this code aims to minimize duplicates in the CustomList (and duplicates _of_ the CustomList). There's also a `CustomList.remove_entry` method to allow other types of list updates from the upload script in the content_server (NYPL-Simplified/content_server#114).

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -581,8 +581,15 @@ class Lane(object):
 
         # The lane may be restricted to books that are on a list
         # from a given data source.
-        self.list_data_source_id, self.list_ids = self.custom_lists_for_identifier(
+        custom_list_details = self.custom_lists_for_identifier(
             list_data_source, list_identifier)
+        (self.list_data_source_id,
+         self.list_ids,
+         self.list_featured_works_query) = custom_list_details
+        if not self.list_featured_works_query:
+            # The parent may have featured works from the list that
+            # could be included here.
+            set_from_parent('list_featured_works_query', None)
         set_from_parent(
             'list_seen_in_previous_days', list_seen_in_previous_days)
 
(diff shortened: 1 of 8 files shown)


model.py
```diff
@@ -7953,8 +7953,8 @@ class CustomList(Base):
     id = Column(Integer, primary_key=True)
     primary_language = Column(Unicode, index=True)
     data_source_id = Column(Integer, ForeignKey('datasources.id'), index=True)
-    foreign_identifier = Column(Unicode, index=True)
-    name = Column(Unicode)
+    foreign_identifier = Column(Unicode, index=True, unique=True)
+    name = Column(Unicode, index=True, unique=True)
```
courte: I'm not sure about this choice, but in the script I'm creating `foreign_identifier` based on name. I don't want there to be any confusion around this in the future...

leonardr: Maybe (data_source_id, foreign_identifier) and (data_source_id, name) could each have a unique constraint? I can see several different data sources providing a list called 'Best Sellers' or some other common phrase.

leonardr: Can (data_source_id, foreign_identifier) and (data_source_id, name) each have a unique constraint? I can see different data sources each providing a generic list name like 'Best Sellers'.

leonardr:
This looks good apart from my concern about the uniqueness constraint.

lane.py
```diff
@@ -800,11 +807,34 @@ def custom_lists_for_identifier(self, list_data_source, list_identifier):
...
+        if not lists:
+            return None
+
+        work_ids = list()
+        for custom_list in lists:
+            work_ids += [work.id for work in custom_list.featured_works]
+
+        works_query = self._db.query(Work).with_labels().\
+            filter(Work.id.in_(work_ids))
+        return works_query
```
leonardr: I don't know if this will be a problem anytime soon but there are a couple things that would be a lot ore efficient here.

First, you could do this in one query joining CustomListEntry against LicensePool against Work.

Or, you could do this in a couple queries by gathering all the custom list IDs, then defining a class method on CustomList that takes a list of list IDs and looks for the featured works in any of those lists, all at once.

I think this is okay for now, but if we start having really big lists this is an easy place to save some time.

lane.py
```diff
@@ -1315,27 +1360,63 @@ def featured_works(self, use_materialized_works=True):
...
+    def randomized_sample_works(self, query, target_size=None, use_min_size=False):
+        """Find a random sample of works for a feed"""
         offset = 0
-        target_size = Configuration.featured_lane_size()
         smallest_sample_size = target_size
 
         if use_min_size:
             smallest_sample_size = self.MINIMUM_SAMPLE_SIZE or (target_size-5)
-        total_size = query.count()
+        total_size = fast_query_count(query)
```
leonardr: :)

model.py
```diff
@@ -8029,6 +8040,8 @@ def add_entry(self, edition, annotation=None, first_appearance=None):
             entry.annotation = unicode(annotation)
         if edition.license_pool and not entry.license_pool:
             entry.license_pool = edition.license_pool
+        if featured:
+            entry.featured = True
```
leonardr: I think you should set entry.featured to False if featured is False, since this can be used to edit an existing list entry.

leonardr:
:+1: with a minor change. There's a performance improvement I'd like you to think about, but I think it's fine for now without it.

lane.py
courte: This is a helpful reminder, thanks. I'll take care of it on the next pass with `CustomList` --> static feeds.

PR #450, opened 2017-01-31
Fix cached feed content #449

### LINKED ISSUE #449: Cached feeds with no timestamp or content are making it into the database
Here are a number of entries in the `cachedfeeds` table:

```
  id   |                                      lane_name                                      | languages | timestamp |  type  |                   facets                   |   pagination    | content | license_pool_id 
-------+-------------------------------------------------------------------------------------+-----------+-----------+--------+--------------------------------------------+-----------------+---------+-----------------
 20314 | Horror                                                                              | eng       |           | groups |                                            |                 |         |                
 45191 | Fantasy                                                                             | eng       |           | page   |                                            |                 |         |                
 45336 | Horror                                                                              | eng       |           | page   |                                            |                 |         |                
 57650 | Books Related to The Secrets of Blueberries, Brothers, Moose & Me by Sara Nickerson | eng       |           | groups |                                            |                 |         |          217398
 57590 | Books Related to The Busy Book Bundle by Trish Kuffner                              | eng       |           | groups |                                            |                 |         |          230687
 46394 | Books Related to Drama Queens in the House by Julie Williams                        | eng       |           | groups |                                            |                 |         |          189035
 49987 | Books Related to A Young Wife by Pam Lewis                                          | eng       |           | groups |                                            |                 |         |          144811
 56711 | Books Related to The Immortalist by Scott Britz                                     | eng       |           | groups |                                            |                 |         |          240760
 52304 | Books Related to Blood and Salt by Kim Liggett                                      | eng       |           | groups |                                            |                 |         |          219834
 56117 | Books Related to 64: Una novela by Sigal Ratner-Arias                               | spa       |           | groups |                                            |                 |         |          148418
 55015 | Books Related to Who Was Woodrow Wilson? by Margaret Frith                          | eng       |           | groups |                                            |                 |         |          220273
 57017 | Fantasy                                                                             | eng       |           | groups |                                            |                 |         |                
 57544 | Books Related to The Pack by Tom Pow                                                | eng       |           | groups |                                            |                 |         |          188955
 37174 | Adult Nonfiction                                                                    | chi       |           | groups |                                            |                 |         |                
 37175 | Adult Nonfiction                                                                    | chi       |           | groups | available=all&collection=main&order=author | after=0&size=50 |         |                
 40493 | Poetry                                                                              | eng       |           | groups |                                            |                 |         |                
 40495 | Poetry                                                                              | eng       |           | groups | available=all&collection=main&order=author | after=0&size=50 |         |                
 44269 | Books Related to Hijacking the Runway by Teri Agins                                 | eng       |           | groups |                                            |                 |         |          159959
```

All of these entries have no `timestamp` and no `content`. They are feeds of all types created both by scripts and the web app. They should not be in this table. I believe the problem is that the feed is retrieved using `fetch()` but the `update()` method (which sets timestamp and content) is never called.

I'm not sure what's causing the underlying issue but we need to stop generating these rows, and stop treating these rows as real cachedfeeds if we do find them.

Suggestions:

1. Look at all the places in core and circulation that call `CachedFeed.fetch()` and see if there's one place where `update()` might not get called.

2. In the get_one_or_create call in CachedFeed.fetch(), try to ignore feeds that don't have a content or timestamp. This might be difficult because I don't think get_one_or_create allows you to put additional restrictions on the initial "does this exist?" query.

3. get_one_or_create might not be a good match for this query in general, because `on_multiple='interchangeable'` leads to useless, duplicate feeds sticking around in the database once they're no longer needed. 

Basically, you might have to improve get_one_or_create, or copy its logic for a CachedFeed-specific lookup method.

courte:
There are a few stray feeds without content or timestamp in the database, which makes them completely unusable to the controllers.

There wasn't a place in the code base where `CachedFeed.fetch` was being called without `CachedFeed#update`, but `CachedFeed#update` wasn't committing its changes to the database immediately. This may have caused problems if scripts or startups erred or stopped unexpectedly.

In accordance with the issue, I also added a CachedFeed-specific "get_one_or_create"-style method that deletes any found instances of CachedFeed that lack required information.

Fixes #449.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5603,6 +5603,7 @@ class CachedFeed(Base):
             facets=facets_key,
             pagination=pagination_key,
             )
+
         if force_refresh is True:
             # No matter what, we've been directed to treat this
             # cached feed as stale.
@@ -5641,9 +5642,10 @@ class CachedFeed(Base):
         # Either there is no cached feed or it's time to update it.
         return feed, False
 
-    def update(self, content):
+    def update(self, _db, content):
         self.content = content
         self.timestamp = datetime.datetime.utcnow()
+        _db.commit()
 
     def __repr__(self):
         if self.content:
--- a/model.py
+++ b/model.py
@@ -5571,7 +5571,7 @@ class CachedFeed(Base):
 
         license_pool = None
         if lane:
-            lane_name = lane.name
+            lane_name = unicode(lane.name)
             if hasattr(lane, 'license_pool'):
                 license_pool = lane.license_pool
         else:
--- a/model.py
+++ b/model.py
@@ -346,9 +346,6 @@ def get_one_or_create(db, model, create_method='',
             for key in get_one_keys:
                 if key in kwargs:
                     del kwargs[key]
-            # if 'on_multiple' in kwargs:
-            #     del kwargs['on_multiple']
-            # if 'filter_constraints'
             obj = create(db, model, create_method, create_method_kwargs, **kwargs)
             __transaction.commit()
             return obj
--- a/model.py
+++ b/model.py
@@ -5662,7 +5662,7 @@ class CachedFeed(Base):
     def update(self, _db, content):
         self.content = content
         self.timestamp = datetime.datetime.utcnow()
-        _db.commit()
+        _db.flush()
 
     def __repr__(self):
         if self.content:
--- a/model.py
+++ b/model.py
@@ -5557,6 +5557,24 @@ class CachedFeed(Base):
     log = logging.getLogger("CachedFeed")
 
     @classmethod
+    def get_feed_or_create(cls, _db, **kwargs):
+        cached_feed = get_one(_db, cls, **kwargs)
+        is_new = False
+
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -328,9 +341,14 @@ def get_one_or_create(db, model, create_method='',
...
         try:
-            if 'on_multiple' in kwargs:
-                # This kwarg is supported by get_one() but not by create().
-                del kwargs['on_multiple']
+            # These kwargs are supported by get_one() but not by create().
+            get_one_keys = ['on_multiple', 'constraint']
+            for key in get_one_keys:
+                if key in kwargs:
+                    del kwargs[key]
+            # if 'on_multiple' in kwargs:
```
leonardr: You can get rid of this code.

leonardr:
:+1: Looks great, just get rid of the commented-out code.

model.py
courte: Yikes, how embarrassing. 😳 👻

PR #451, opened 2017-02-02
Support FeedBooks short stories genre

courte:
Supports the work done in NYPL-Simplified/content_server#116, including:
 - allowing URI-type identifiers to be differentiated by DataSource,
 - creating the FeedBooks DataSource at initialization,
 - passing bypass database sessions, and
 - trimming some long lines and unused variables.

### DIFF
--- a/model.py
+++ b/model.py
@@ -4021,7 +4021,7 @@ class Work(Base):
             self, operation=WorkCoverageRecord.QUALITY_OPERATION
         )
 
-    def assign_genres(self, identifier_ids, cutoff=0.15):
+    def assign_genres(self, identifier_ids):
         """Set classification information for this work based on the
         subquery to get equivalent identifiers.
 
--- a/model.py
+++ b/model.py
@@ -893,6 +893,7 @@ class DataSource(Base):
                 (cls.NOVELIST, False, True, Identifier.NOVELIST_ID, None),
                 (cls.PRESENTATION_EDITION, False, False, None, None),
                 (cls.INTERNAL_PROCESSING, True, False, None, None),
+                (cls.FEEDBOOKS, True, False, Identifier.URI, None)
         ):
 
             extra = dict()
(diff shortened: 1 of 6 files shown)


model.py
```diff
@@ -4020,7 +4021,7 @@ def calculate_quality(self, identifier_ids, default_quality=0):
             self, operation=WorkCoverageRecord.QUALITY_OPERATION
         )
 
-    def assign_genres(self, identifier_ids, cutoff=0.15):
```
courte: `cutoff` wasn't being used.

scripts.py
```diff
@@ -232,6 +235,10 @@ def arg_parser(cls):
             help='Process identifiers of this type. If IDENTIFIER is not specified, all identifiers of this type will be processed. If IDENTIFIER is specified, this argument is required.'
         )
         parser.add_argument(
+            '--identifier-data-source',
```
courte: This supports cases like Identifier.URI.

scripts.py
```diff
@@ -232,6 +235,10 @@ def arg_parser(cls):
...
+            help='Process identifiers with this LicensePool.data_source'
```
leonardr: This is a little confusing to me. I would say "Process only identifiers which have a license pool associated with this data source."

leonardr:
:+1: with one minor clarification.

leonardr:
:+1:

courte:
Since we're going to keep working with FeedBooks, we may want to create a FeedBooksClassifier at some point. Their identifiers have subject information that can be more detailed or direct than their subject names.

Some examples:
 - `FBFIC027040 - Gothic` and `FBFIC024010N - Gothic` are both gothic fiction, but "Gothic" alone is too general a keyword to make the assumption.
 - `FBDRA011000N - Greek` is a Greek drama, while `FBJUV022020 - Greek & Roman` is a juvenile work with Greek / Roman aspects.
 - `FBFIC004030N - Modern (<1799)` and `FBFIC014040N - Modern (<1799)` are fiction. `FBPOE009000N - Modern (<1799)` is poetry, and `FBDRA050000N - Modern (<1799)` is drama.
 - So on and so forth.

PR #453, opened 2017-02-06
Remove 'biographical' from nonfiction (i.e. biographical fiction)

courte:
When I ran this against QA, it turns out this subject is for fiction and the keyword is generally used to describe fiction.


leonardr:
:+1:

PR #454, opened 2017-02-06
Sort name cleaning

daryanypl:


scripts.py
```diff
@@ -1225,11 +1219,141 @@ def do_run(self):
...
+                if edition.contributions:
+                    for contribution in edition.contributions:
+                        self.process_contribution(self._db, contribution, edition)
+            offset += batch_size
+            self._db.commit()
+        self._db.commit()
+
+
+    @classmethod
+    def process_contribution(cls, _db, contribution, edition):
```
leonardr: If you have a contribution you don't also need the edition; you can get it from `contribution.edition`.

util/personal_names.py
```diff
@@ -32,27 +32,54 @@ def is_corporate_name(display_name):
     return False
 
 
-def display_name_to_sort_name(display_name):
+def display_name_to_sort_name(display_name, advanced=False):
```
leonardr: When would you use advanced=False? Every time you call this method in this branch or the metadata branch you pass in advanced=True. Looking at the code, it looks like you've improved the algorithm without adding latency. If you're confident in the new algorithm I would just use it to replace the old one.

scripts.py
```diff
@@ -1225,11 +1219,141 @@ def do_run(self):
...
+    @classmethod
+    def check_sort_name(cls, computed_sort_name, contributor, edition, data_source, operation):
+        success = True
+        if computed_sort_name.strip().lower() != contributor.sort_name.strip().lower():
+            record, is_new = CoverageRecord.add_for(
+                edition=edition, data_source=DataSource.VIAF, operation=CoverageRecord.REPAIR_SORT_NAME_OPERATION,
+                status=CoverageRecord.TRANSIENT_FAILURE
+            )
+            # NOTE:  not the best way -- storing id info in the message field, but 
+            # perhaps better than making CoverageRecords cover Contributor-type objects.
```
leonardr: I'm uneasy about this for two reasons. First, as you point out, CoverageRecords aren't designed to cover Contributor-type objects. Second, this code is in core but the code for dealing with it (doing a VIAF lookup) is in metadata.

I would take this code out and let the core script do the best it can with your improved display_name_to_sort_name algorithm. In metadata, I would have the child of this script immediately do a VIAF lookup if it feels like it needs to, rather than saving the work until later.

If you really want to do a VIAF lookup from outside metadata, you can call `SimplifiedOPDSLookup.from_config()` to get an object that will let you ask metadata wrangler for the sort name of an author, using `canonicalize_author_name`. 

I think this makes sense in a repair script. Just have the metadata subclass of the script omit this step, since it _is_ the metadata wrangler and can do the VIAF lookup itself.

To summarize:

1. If a VIAF lookup needs to happen, do it immediately, don't save it for later. It'll slow down the script, but this is a repair script, we're not in a hurry here.
2. If you're not in metadata, get a `SimplifiedOPDSLookup` object and ask metadata to do the VIAF lookup.
3. If you're in metadata, do the VIAF lookup yourself.

util/personal_names.py
daryanypl: I was thinking that there might be reasons I don't know about for using the older code, and I wanted to keep it a possibility.  Especially in case there's legacy code I haven't found yet that expects a certain specific response?
Will add unit test.

scripts.py
daryanypl: I split the code that deals with it into two portions.  The VIAF lookup code is in metadata, but the display_name --> sort_name code is in core.  So it is possible to resolve the transient failure from within core.  That code will follow in future push.

Not sure I understand the second part of the comment.

util/personal_names.py
leonardr: A unit test can verify that the old behavior is preserved, which should eliminate any concerns about code that expects the old behavior.

model.py
```diff
@@ -2088,6 +2116,11 @@ def merge_into(self, destination):
             destination,
             destination.viaf
         )
+        
+
+        # TODO: put in logic to keep sort_name is one of the contributor objects has it.
```
daryanypl: done

scripts.py
daryanypl: changed

util/personal_names.py
daryanypl: removed old code and added unit test (in new test_personal_names.py file).

scripts.py
```diff
@@ -1221,8 +1220,16 @@ def do_run(self):
...
-    """ TODO """
+class CheckContributorNamesInDB(IdentifierInputScript):
+    """ Checks that contributor sort_names are display_names in 
+    "last name, comma, other names" format.  
+
+    NOTE:  There's also CheckContributorNamesOnWeb in metadata, 
+    it's a child of this script.  Use it to check our knowledge against 
+    viaf, with the newer better sort_name selection and formatting.
+
+    TODO: make sure don't start at beginning again when interrupt while batch job is running.
```
leonardr: An `IdentifierSweepMonitor` can do this, but right now you can't combine an `IdentifierInputScript` with a `RunMonitorScript`.

Is it worth making the change? I'm not sure how often we plan on running this script.

bin/repair/check_contributor_names
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+"""Explain what the system knows about a book."""
```
leonardr: Fix docstring.

scripts.py
```diff
@@ -1269,77 +1267,114 @@ def make_query(self, _db, identifier_type, identifiers, log=None):
...
-        FROM editions, identifiers \n
-        WHERE editions.primary_identifier_id IN (:primary_identifier_id_1) AND identifiers.type = :type_1'
-        '''
 
         if log:
             log.info(
                 "Processing %d editions.", query.count()
             )
+            print "Processing %d editions.", query.count()
+
```
leonardr: This should go back to being a log statement.

scripts.py
```diff
@@ -1269,77 +1267,114 @@ def make_query(self, _db, identifier_type, identifiers, log=None):
...
 
 
     def run(self, batch_size=10):
         param_args = self.parse_command_line(self._db)
         
-        #if param_args.identifiers:
+        if param_args.identifiers:
             # we're asked about a specific set of work contributors
-            #identifier_ids = [x.id for x in param_args.identifiers]
+            identifier_ids = [x.id for x in param_args.identifiers]
```
leonardr: Is this used? Seems like param_args.identifiers is passed into make_query(), which does the work.

scripts.py
```diff
@@ -1269,77 +1267,114 @@ def make_query(self, _db, identifier_type, identifiers, log=None):
...
 
         self.query = self.make_query(
             self._db, param_args.identifier_type, param_args.identifiers, self.log
         )
 
         editions = True
         offset = 0
+        output = "ContributorID|\tSortName|\tDisplayName|\tComputedSortName|\tMatchRatio_Etc|\tResolution"
+        print output.encode("utf8")
```
leonardr: Not a big deal, but the right way to do this is with a csv writer. For an example see metadata/scripts.py, `PermanentWorkIDStressTestGenerationScript`. You'll need to make sure that all the output is encoded as UTF-8.

scripts.py
```diff
@@ -1269,77 +1267,114 @@ def make_query(self, _db, identifier_type, identifiers, log=None):
...
         while editions:
-            editions = self.query.offset(offset).limit(batch_size).all()
+            my_query = self.query.offset(offset).limit(batch_size)
+            editions = my_query.all()
+
             for edition in editions:
                 if edition.contributions:
                     for contribution in edition.contributions:
-                        self.process_contribution(self._db, contribution, edition)
+                        self.process_contribution(self._db, contribution, self.log)
```
leonardr: You don't need to pass in self.log to another instance method--it'll still be available inside process_contribution().

scripts.py
```diff
@@ -1269,77 +1267,114 @@ def make_query(self, _db, identifier_type, identifiers, log=None):
...
+                # suffix, we can fix without asking for human intervention.  if the names are very different, 
+                # there's a chance the sort and display names are different on purpose, s.a. when foreign names 
+                # are passed as translated into only one of the fields, or when the author has a popular pseudonym. 
+                # best ask a human.
+
+                # if the relative lengths are off than by a stray space or comma, ask a human
+                # it probably means that a human metadata professional had added an explanation/expansion to the 
+                # sort_name, s.a. "Bob A. Jones" --> "Bob A. (Allan) Jones", and we'd rather not replace this data 
+                # with the "Jones, Bob A." that the auto-algorigthm would generate.
+                length_difference = len(contributor.sort_name.strip()) - len(computed_sort_name_local_new.strip())
```
leonardr: Would it make sense to calculate Levenshtein distance here instead of the length of the strings? I don't like what seems to happen here if the names are totally different but the same length.

scripts.py
```diff
@@ -1269,77 +1267,114 @@ def make_query(self, _db, identifier_type, identifiers, log=None):
...
 
-        computed_sort_name_local = display_name_to_sort_name(contributor.display_name, advanced=True)
-        success = cls.check_sort_name(computed_sort_name_local, contributor, edition, DataSource.INTERNAL_PROCESSING, CoverageRecord.REPAIR_SORT_NAME_OPERATION)
-        if not success:
-            output = "Contributor[%s]: contributor_sort_name=%s, contributor_display_name=%s: INTERNAL MISMATCH" % (contributor.id, contributor.sort_name, contributor.display_name)
-            print output.encode("utf8")
+                    # also change edition.sort_author, if the author was primary
+                    edition_contributors = contribution.edition.author_contributors
+                    if (len(edition_contributors) > 0 and (edition_contributors[0].id == contributor.id)):
+                        contribution.edition.sort_author = computed_sort_name_local_new
```
leonardr: I would expect calculate_author() to manage Edition.sort_author. Is there a reason why it wouldn't work to set the contributor sort_name and then call Edition.calculate_author()?

scripts.py
```diff
@@ -1269,77 +1267,114 @@ def make_query(self, _db, identifier_type, identifiers, log=None):
...
+
+        detail = "Contributor[id=%s].sort_name is oddly different from computed_sort_name, human intervention required." % contributor.id
+        pool = contribution.edition.is_presentation_for
+        complaint_type = "http://librarysimplified.org/terms/problem/wrong-author";
+        try:
+            complaint, is_new = Complaint.register(pool, complaint_type, source, detail)
+        except ValueError, e:
+            # log and move on, don't stop run
+            log.error("Error registering complaint: %r", contributor, exc_info=e)
+            print("Error registering complaint: %r", contributor)
```
leonardr: Yeah, this should be a log statement but you don't need the log from the instance, you can just do a `logging.getlogger` call.

scripts.py
daryanypl: I don't know.  We'll only run it once, hopefully, but the metadata database is very large, and things can crash.  Won't hurt to try and see.

bin/repair/check_contributor_names
daryanypl: fixed

scripts.py
daryanypl: removed

scripts.py
daryanypl: Good to know.  I'll write the class reference into my personal notes, and let me know if you want me to change the code.

scripts.py
daryanypl: I moved process_contribution to be a classmethod.

scripts.py
daryanypl: they're now set in the same method, set_contributor_sort_name

scripts.py
daryanypl: do you still want to remove class instance-level logging?

util/personal_names.py
```diff
@@ -33,6 +39,18 @@ def replacePhD(match):
...
+    If there was found to be improper punctuation at the end of the name string, 
+    clean it off.
+    :param match: a regular expression matched to a string
+    """
+    if not match or len(match.groups()) < 3:
+        return match
+
+    return match.groups()[0] + match.groups()[1]
+
+
```
leonardr: Sorry to nitpick but there are a couple problems with this function. First, it should be called replace_end_punctuation for PEP 8 compliance. Second, it's designed for use in one specific place -- inside `name_tidy` -- so the best thing to do is to define it within that function, just before it's used. Otherwise it looks like a general name cleanup utility function that can be called by whoever.

PR #457, opened 2017-02-13
Support static feeds from lists

courte:
This branch expands the list of attributes that a sublane will inherit from its parent to include exclude_genres and any `CustomList` information. It supports Lanes that have defined `exclude_genres` without `include_genres` (like "General Fiction"). It also moves `make_lanes` from circulation into core.

Supports NYPL-Simplified/content_server#118.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -688,7 +688,7 @@ class Lane(object):
 
         # Find all the genres that will go into this lane.
         genres, self.fiction = self.gather_matching_genres(
-            genres, fiction, full_exclude_genres
+            _db, genres, fiction, full_exclude_genres
         )
         self.genre_ids = [x.id for x in genres]
         self.genre_names = [x.name for x in genres]
@@ -906,16 +906,32 @@ class Lane(object):
         return genre_obj, genre_data
 
     @classmethod
-    def all_matching_genres(cls, genres, exclude_genres=None):
+    def all_matching_genres(cls, _db, genres, exclude_genres=None, fiction=None):
         matches = set()
         exclude_genres = exclude_genres or []
+        if exclude_genres:
+            if not isinstance(exclude_genres, list):
+                exclude_genres = list(exclude_genres)
+
+            for excluded_genre in exclude_genres[:]:
+                exclude_genres += excluded_genre.self_and_subgenres
+
+            if not genres:
+                # This is intended for lanes like "General Fiction"
+                # when we want everything EXCEPT for some given lanes.
+                genres = _db.query(Genre).all()
+                if fiction is not None:
+                    # Filter the genres according to their expected
+                    # fiction status.
+                    genres = filter(lambda g: g.default_fiction==fiction, genres)
+
         if genres:
             for genre in genres:
                 matches = matches.union(genre.self_and_subgenres)
         return [x for x in matches if x not in exclude_genres]
 
     @classmethod
-    def gather_matching_genres(cls, genres, fiction, exclude_genres=[]):
+    def gather_matching_genres(cls, _db, genres, fiction, exclude_genres=[]):
         """Find all subgenres of the given genres which match the given fiction
         status.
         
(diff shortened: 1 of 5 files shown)


lane.py
```diff
@@ -583,31 +571,20 @@ def set_list(field_name, value, default=None):
...
-            # could be included here.
-            set_from_parent('list_featured_works_query', None)
-        set_from_parent(
-            'list_seen_in_previous_days', list_seen_in_previous_days)
-
-        set_from_parent(
+        self.set_from_parent(
             'subgenre_behavior', subgenre_behavior, self.IN_SUBLANES)
 
-        if self.searchable and (self.list_data_source_id or self.list_ids):
```
courte: This is no longer true. I couldn't figure out a check that would make sense to maintain this error.

lane.py
```diff
@@ -618,54 +595,127 @@ def set_list(field_name, value, default=None):
             formats=formats
         )
         if include_staff_picks:
-            full_name = "%s - Staff Picks" % self.name
```
courte: Extracted to a method.

lane.py
```diff
@@ -618,54 +595,127 @@ def set_list(field_name, value, default=None):
...
-                )
-            except UndefinedLane, e:
-                # Not a big deal, just don't add the lane.
-                staff_picks_lane = None
-            if staff_picks_lane:
-                self.sublanes.lanes.insert(0, staff_picks_lane)
-
+            self.include_staff_picks(**base_args)
         if include_best_sellers:
-            full_name = "%s - Best Sellers" % self.name
```
courte: Extracted to a method.

lane.py
```diff
@@ -548,32 +548,20 @@ def __init__(self,
...
             age_range = [age_range]
         if age_range is not None:
             age_range = sorted(age_range)
-        set_from_parent('age_range', age_range)
+        self.set_from_parent('age_range', age_range)
 
         self.audiences = self.audience_list_for_age_range(audiences, age_range)
 
+
+        def set_list(field_name, value, default=None):
```
courte: I made a change to this in removing the repeat code from `set_from_parent()`, since its called in actual `set_from_parent()` call in the end.

lane.py
```diff
@@ -548,32 +548,20 @@ def __init__(self,
         # e.g. "All Science Fiction"
         self.include_all_feed = include_all
 
-        def set_from_parent(field_name, value, default=None):
```
courte: I moved this method out of `Lane.__init__()` so that other supporting initialization methods could use it.

lane.py
```diff
@@ -618,54 +595,127 @@ def set_list(field_name, value, default=None):
...
+
+        for sublane in self.sublanes:
+            # If the sublanes were set beforehand, they need to
+            # inherit this information now.
+            sublane.set_customlist_information(None, None, None)
+
+        self.set_from_parent(
+            'list_seen_in_previous_days', list_seen_in_previous_days)
+
+    def custom_lists_for_identifier(self, list_data_source, list_identifier):
```
courte: This method and the one below haven't been changed -- just moved up in the file.

lane.py
leonardr: I think this existed because I hadn't written the feature, and now you have.

model.py
```diff
@@ -8069,6 +8069,10 @@ def add_entry(self, edition, annotation=None, first_appearance=None,
             entry.license_pool = edition.license_pool
         if featured is not None:
             entry.featured = featured
+
+        if was_new:
+            self.updated = datetime.datetime.utcnow()
+            _db.commit()
```
leonardr: I'm not convinced this commit() is necessary. If you want it for symmetry with the commit() in remove_entry(), it should always happen, not just when the entry is new.

leonardr:
:+1: with a minor nitpick.

PR #460, opened 2017-02-17
Abstract implementation of the User Profile Management Protocol

leonardr:
This branch implements and tests a simple implementation of https://github.com/NYPL-Simplified/Simplified/wiki/User-Profile-Management-Protocol that uses a swappable data store.

The data store used for testing keeps the read-only items in one dictionary and the writable items in another dictionary.

If there are any discrepancies between this and the spec, it's possible the spec needs to change, but more likely it's a bug in the implementation.

### DIFF
--- a/tests/test_user_profile.py
+++ b/tests/test_user_profile.py
@@ -8,7 +8,7 @@ from user_profile import (
     DictionaryBasedProfileStore,
 )
 
-class TestProfileManager(object):
+class TestProfileController(object):
 
     def setup(self):
         self.read_only = dict(key="value")
--- a/tests/test_user_profile.py
+++ b/tests/test_user_profile.py
@@ -5,32 +5,32 @@ from nose.tools import (
 import json
 from user_profile import (
     ProfileController,
-    DictionaryBasedProfileStore,
+    MockProfileStorage,
 )
 
 class TestProfileController(object):
 
     def setup(self):
-        self.read_only = dict(key="value")
-        self.writable = dict(writable_key="old_value")
-        self.store = DictionaryBasedProfileStore(self.read_only, self.writable)
-        self.controller = ProfileController(self.store)
+        self.read_only_settings = dict(key="value")
+        self.writable_settings = dict(writable_key="old_value")
+        self.storage = MockProfileStorage(self.read_only_settings, self.writable_settings)
+        self.controller = ProfileController(self.storage)
 
-    def test_representation(self):
+    def test_profile_document(self):
         """Test that the default setup becomes a dictionary ready for
         conversion to JSON.
         """
         eq_({'key': 'value', 'settings': {'writable_key': 'old_value'}},
-            self.store.representation)
+            self.storage.profile_document)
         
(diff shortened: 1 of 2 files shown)


user_profile.py
```diff
@@ -0,0 +1,172 @@
...
+    MEDIA_TYPE = "vnd.librarysimplified/user-profile+json"
+    LINK_RELATION = "http://librarysimplified.org/terms/rel/user-profile"
+    SETTINGS_KEY = 'settings'
+    
+    def __init__(self, store):
+        """Constructor.
+
+        :param store: An instance of ProfileStore.
+        """
+        self.store = store
```
daryanypl: is the store pointing to the profile server, or does it represent the library the user is signed up to?

leonardr: It represents the user themselves. In my follow-up branch it's a wrapper around the `Patron` object.

user_profile.py
```diff
@@ -0,0 +1,172 @@
...
+        
+    def set(self, settable, full):
+        """(Try to) make the local store look like the provided Profile
+        document.
+
+        :param settable: The portion of the Profile document containing
+            settings that the client wants to change.
+
+        :param full: The full Profile document as provided by the client.
+            Should not be necessary but provided in case it's userful.
```
daryanypl: useful

user_profile.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        raise NotImplementedError()
+    
+
+class DictionaryBasedProfileStore(object):
```
daryanypl: inherit from ProfileStore

tests/test_user_profile.py
```diff
@@ -0,0 +1,151 @@
...
+        
+    def test_attempt_to_set_read_only_setting(self):
+        """You can't change the value of a setting that's not
+        writable.
+        """
+        headers = {"Content-Type" : ProfileController.MEDIA_TYPE}
+        body = json.dumps(dict(settings=dict(key="new value")))
+        problem = self.controller.put(headers, body)
+        eq_(400, problem.status_code)
+        eq_('"key" is not a writable setting.', problem.detail)
```
daryanypl: specific key?

leonardr: I'm not sure what this comment means. I'm testing that the name of the key ('key', defined on line 136) is used in the error message. If 'key' is a confusing name for a key I can change it.

user_profile.py
```diff
@@ -0,0 +1,182 @@
...
+                if k not in writable:
+                    return INVALID_INPUT.detailed(
+                        _('"%s" is not a writable setting.' % k)
+                    )
+            try:
+                # Update the profile storage with the new settings.
+                self.storage.update(new_settings, profile_document)
+            except Exception, e:
+                # There was a problem updating the profile storage.
+                if hasattr(e, 'as_problem_detail_document'):
```
daryanypl: I really like this as_problem_detail_document idea.

daryanypl:
👍

PR #461, opened 2017-02-17
Implement the User Profile Management Protocol for Patron objects

leonardr:
This branch introduces `PatronProfileStorage`, an implementation of `ProfileStorage` which uses a `Patron` object as its backing store. I create a new database field to store a patron's annotation synchronization settings, and make it a writable setting that can be manipulated through the UPMP.

The date on which the patron's card expires is also available through UPMP, though it's a profile item, not a writable setting.

I'm not bothering with fines at this point--it's more work and the client won't use it.

The `Patron.synchronize_annotations` field has two implemented effects. First, whenever it is set to false, all of the patron's annotations are deleted. Second, when it is set to false, an attempt to create a new annotation for the patron will raise a ValueError.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -55,7 +55,7 @@ from model import (
     LicensePool,
     Measurement,
     Patron,
-    PatronProfileStore,
+    PatronProfileStorage,
     Representation,
     Resource,
     RightsStatus,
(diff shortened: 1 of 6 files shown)


tests/test_model.py
```diff
@@ -4574,7 +4575,101 @@ def test_external_type_regular_expression(self):
...
+        self.patron = self._patron()
+        self.store = PatronProfileStorage(self.patron)
+        
+    def test_writable_setting_names(self):
+        """Only one setting is currently writable."""
+        eq_(set([self.store.SYNCHRONIZE_ANNOTATIONS]),
+            self.store.writable_setting_names)
+
+    def test_profile_document(self):
+        # synchronize_annotations always shows up as settable, even if
```
daryanypl: but not if it's false, right?

leonardr: If it's false, you can set it to true by flipping the opt-in switch in settings.

daryanypl:
👍

PR #462, opened 2017-02-17
Initialize database with date

### LINKED ISSUE #441: Change DatabaseMigrationInitializationScript to warn instead of raising an exception, accept a date.
Right now, if there's already a timestamp in the database, running `bin/initialize_database` raises an exception.

Instead, it would be nice if it shared a warning and went on to create a new timestamp anyway.

It would also be nice to pass in a timestamp (for folks who know what their last run datetime was).

courte:
Previously, the database initialization script was basically a toggle that would only create a **new** `"Database Migration"` timestamp for the **most recent** migration file. With this branch, an existing timestamp can be overwritten as needed, and the timestamp (new or existing) can be set to a date passed in via the command line. This provides some flexibility that should make working with the `DatabaseMigrationInitializationScript` more bearable in development without impacting production.

There was also a dumb bug in the `DatabaseMigrationScript` where the timestamp counter wasn't being reset to `None`. That's fixed here.

Fixes #441.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1256,12 +1256,29 @@ class DatabaseMigrationInitializationScript(DatabaseMigrationScript):
     DatabaseMigrationScript to manage migrations.
     """
 
-    def do_run(self):
+    @classmethod
+    def arg_parser(cls):
+        parser = super(DatabaseMigrationInitializationScript, cls).arg_parser()
+        parser.add_argument(
+            '-f', '--force', action='store_true',
+            help="Force reset the initialization, ignoring any existing timestamps."
+        )
+        return parser
+
+    def do_run(self, cmd_args=None):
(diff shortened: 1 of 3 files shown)


scripts.py
```diff
@@ -1256,12 +1257,42 @@ class DatabaseMigrationInitializationScript(DatabaseMigrationScript):
...
-            raise Exception(
-                "Timestamp for Database Migration script already exists"
-            )
+            if parsed.force:
+                self.log.warn(
+                    "Overwriting existing %s timestamp: %r",
+                    self.name, existing_timestamp)
+            else:
+                raise RuntimeError(
+                    "%s timestamp already exists: %r" %
```
aslagle: I'd like this message to remind me that I can run with -f to force an update.

PR #463, opened 2017-02-22
Support instance initialization

courte:
This branch uses an alias to access the configured Elasticsearch works_index. It then searching against that alias to improve the process of creating new indices.

It also adds a small change to allow the programmatic creation of a new index without an alias.

Fixes NYPL-Simplified/operations#49.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -94,10 +94,23 @@ class ExternalSearchIndex(object):
         else:
             if current_alias.endswith(self.CURRENT_ALIAS_SUFFIX):
                 # The alias culled from configuration is intended to be
-                # a current alias, but we couldn't find an index with that
-                # alias.
-                raise ValueError(
-                    "External search alias '%s' not found." % current_alias)
+                # a current alias, but an index with that alias wasn't
+                # found. Find or create an appropriate index.
+                base_index_name = self.base_index_name(current_alias)
+                new_index = base_index_name+'-'+ExternalSearchIndexVersions.latest()
+                exists = self.indices.exists(index=new_index)
+                created = None
+
+                if not exists:
+                    created = ExternalSearchIndexVersions.create_new_version(
+                        self, base_index_name)
+                if exists or created:
+                    _set_works_index(new_index)
+                else:
+                    raise ValueError(
+                        ("Requested external search alias '%s' not found"
+                         " and new index '%s' could not be found or created.")
+                        % (current_alias, new_index))
             else:
                 # Without the CURRENT_ALIAS_SUFFIX, assume the index string
                 # from config is the index itself and needs to be swapped.
--- a/external_search.py
+++ b/external_search.py
@@ -15,6 +15,7 @@ import time
 class ExternalSearchIndex(object):
     
     work_document_type = 'work-type'
+    CURRENT_ALIAS_SUFFIX = '-current'
     __client = None
 
     def __init__(self, url=None, works_index=None):
@@ -23,7 +24,7 @@ class ExternalSearchIndex(object):
 
         # By default, assume that there is no search index.
         self.works_index = None
-
+        use_alias = False
         if not ExternalSearchIndex.__client:
             if not url or not works_index:
                 integration = Configuration.integration(
@@ -34,6 +35,11 @@ class ExternalSearchIndex(object):
                     works_index = integration.get(
                         Configuration.ELASTICSEARCH_INDEX_KEY
                     ) or None
+
+                    # An alias should only be created if we're using the
+                    # configured works_index. Otherwise, the included
+                    # index might be temporary.
+                    use_alias = True
 
                 if not url:
                     if not integration:
@@ -64,6 +70,8 @@ class ExternalSearchIndex(object):
             
         if not self.indices.exists(self.works_index):
             self.setup_index()
+        if use_alias:
+            self.setup_current_alias()
 
     def setup_index(self):
         """
@@ -135,8 +143,35 @@ class ExternalSearchIndex(object):
                 body=mapping,
                 index=self.works_index,
             )
-            
-                
+
+    def setup_current_alias(self):
+        """Put an alias ending with '-current' on the existing works_index."""
+        if self.works_index.endswith(self.CURRENT_ALIAS_SUFFIX):
+            # The alias has already been created.
+            return
+
+        alias_details = self.indices.get_alias(index=self.works_index)
+        existing = alias_details[self.works_index]['aliases']
+        current_alias = [k for k in existing.keys()
+                         if k.endswith(self.CURRENT_ALIAS_SUFFIX)]
+
+        if current_alias:
+            # Assume there's only one '-current' alias on this index.
+            current_alias = current_alias[0]
+        else:
+            # Build an appropriate alias name.
+            base_works_index = re.sub(r'-v[0-9]+$', '', self.works_index)
+            alias_name = base_works_index + self.CURRENT_ALIAS_SUFFIX
+            current_alias = alias_name
+
+            # Create the alias.
+            response = self.indices.put_alias(index=self.works_index, name=alias_name)
+            if not response.get('acknowledged'):
+                self.log.error("Alias '%s' could not be created", alias_name)
+                return
+
+        # Make sure we're pointing to the alias instead of the index.
+        self.works_index = self.__client.works_index = current_alias
 
     def query_works(self, query_string, media, languages, exclude_languages, fiction, audience,
                     age_range, in_any_of_these_genres=[], fields=None, size=30, offset=0):
--- a/external_search.py
+++ b/external_search.py
@@ -117,7 +117,7 @@ class ExternalSearchIndex(object):
         the works_index directly for search queries.
         """
 
-        base_works_index = self._base_works_index(self.works_index)
+        base_works_index = self.base_index_name(self.works_index)
         alias_name = base_works_index+self.CURRENT_ALIAS_SUFFIX
         exists = self.indices.exists_alias(name=alias_name)
 
@@ -170,8 +170,17 @@ class ExternalSearchIndex(object):
             raise ValueError(
                 "Index '%s' does not exist on this client." % new_index)
 
+        current_base_name = self.base_index_name(self.works_index)
+        new_base_name = self.base_index_name(new_index)
+
+        if new_base_name != current_base_name:
+            raise ValueError(
+                ("Index '%s' is not in series with current index '%s'. "
+                 "Confirm the base name (without version number) of both indices"
+                 "is the same.") % (new_index, self.works_index))
+
         self.works_index = self.__client.works_index = new_index
-        alias_name = self._base_works_index(new_index)+self.CURRENT_ALIAS_SUFFIX
+        alias_name = self.base_index_name(new_index)+self.CURRENT_ALIAS_SUFFIX
 
         exists = self.indices.exists_alias(name=alias_name)
         if not exists:
@@ -191,7 +200,7 @@ class ExternalSearchIndex(object):
 
         self.works_alias = self.__client.works_alias = alias_name
 
-    def _base_works_index(self, index_or_alias):
+    def base_index_name(self, index_or_alias):
         """Removes version or current suffix from base index name"""
 
         current_re = re.compile(self.CURRENT_ALIAS_SUFFIX+'$')
--- a/external_search.py
+++ b/external_search.py
@@ -15,9 +15,19 @@ import time
 class ExternalSearchIndex(object):
     
     work_document_type = 'work-type'
+    __client = None
+
     CURRENT_ALIAS_SUFFIX = '-current'
     VERSION_RE = re.compile('-v([0-9]+)$')
-    __client = None
+
+    @classmethod
+    def reset(cls):
+        """Resets the __client object to None so a new configuration
+        can be applied during object initialization.
+
+        This method is only intended for use in testing.
+        """
+        cls.__client = None
 
     def __init__(self, url=None, works_index=None):
     
--- a/external_search.py
+++ b/external_search.py
@@ -78,7 +78,9 @@ class ExternalSearchIndex(object):
         self.bulk = bulk
 
     def set_works_index_and_alias(self, current_alias):
-        """Finds or creates the index based on provided configuration"""
+        """Finds or creates the works_index and works_alias based on
+        provided configuration.
+        """
         index_details = self.indices.get_alias(name=current_alias, ignore=[404])
         found = not (index_details.get('status')==404 or 'error' in index_details)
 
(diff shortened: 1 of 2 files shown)


external_search.py
```diff
@@ -63,84 +71,112 @@ def bulk(docs, **kwargs):
...
+    def setup_index(self, new_index=None, current_alias=False):
+        """Create the search index with appropriate mapping.
 
         This will destroy the search index, and all works will need
         to be indexed again. In production, don't use this on an
         existing index. Use it to create a new index, then change the 
         alias to point to the new index.
         """
-
-        if self.works_index:
```
courte: Despite the diff, this method has barely changed. Only the top three lines and the last three lines are any different. Everything else has just been tabbed left to adjust for the removal of this `if` clause.

external_search.py
```diff
@@ -52,7 +59,8 @@ def __init__(self, url=None, works_index=None):
             if not url:
                 raise Exception("Cannot connect to Elasticsearch cluster.")
 
-        self.works_index = self.__client.works_index
+        self.works_alias = self.works_index = self.__client.works_index
```
courte: The alias is initialized as the index name, so it can always be used for search - even if there isn't actually an alias.

courte:
There is more work that could be done here to make switching from one index to another easier, but it's not directly related to initialization so I'm leaving it for the future (e.g. a method to automatically create a new index that has the appropriate version, a method to change the "current" alias from one index to another). This is at a starting place.

aslagle:
I'm a little confused about what should be in the config file. If I understand this correctly, you would first set up your config file with the _index_ name, and it creates the index and an alias that's based on that index name. 

Later, when everything's set up and you ready to run the app, you would change your config file to have the _alias_ name? I'd rather not have to do that. 

Maybe it should be the _alias_ name that goes in the config file, or a base index name, and the code could contain an internal version number that it uses to create the index.

It might help to talk through the process of switching to a new version of the index?

courte:
In this setup, you wouldn't need to change the config file at all -- the alias will always be found based on the given index name, even if that index name is something like with "circulation-works-v3" or some other version number.

But yeah, let's talk about how you switch things over.

external_search.py
```diff
@@ -527,6 +585,43 @@ def bulk_update(self, works, retry_on_batch_failure=True):
...
+    def latest(cls):
+        version_re = re.compile('v(\d+)')
+        versions = [int(re.match(version_re, v).groups()[0]) for v in cls.VERSIONS]
+        latest = sorted(versions)[-1]
+        return 'v%d' % latest
+
+    @classmethod
+    def latest_mapping(cls):
+        version_method = cls.latest() + '_mapping'
+        return getattr(cls, version_method)()
```
courte: I hate this. Suggestions from outside the abyss are very welcome.

tests/test_external_search.py
```diff
@@ -38,13 +39,10 @@ def setup(self):
         with temp_config() as config:
             config[Configuration.INTEGRATIONS][Configuration.ELASTICSEARCH_INTEGRATION] = {}
             config[Configuration.INTEGRATIONS][Configuration.ELASTICSEARCH_INTEGRATION][Configuration.URL] = "http://localhost:9200"
-            config[Configuration.INTEGRATIONS][Configuration.ELASTICSEARCH_INTEGRATION][Configuration.ELASTICSEARCH_INDEX_KEY] = "test_index"
+            config[Configuration.INTEGRATIONS][Configuration.ELASTICSEARCH_INTEGRATION][Configuration.ELASTICSEARCH_INDEX_KEY] = "test_index-current"
 
             try:
-                ExternalSearchIndex.__client = None
```
courte: `__client` can only be set in the class, so I moved it into a method that gets run in `teardown`

external_search.py
courte: Well, now the way the mapping value is generated has been extracted.

external_search.py
```diff
@@ -17,12 +17,25 @@ class ExternalSearchIndex(object):
...
+        """
+        cls.__client = None
+
     def __init__(self, url=None, works_index=None):
     
         self.log = logging.getLogger("External search index")
 
         # By default, assume that there is no search index.
         self.works_index = None
+        from_configuration = False
```
aslagle: I'm not sure whether or not a config file was used is the right way to determine if we should create the alias.

The constructor could always create the alias if it doesn't exist yet, but I think we could separate out the code that moves the alias to a new index version, so it never happens as part of the constructor. I'd like to have that happen in a separate script anyway, not when you're running the app or creating a new index.

external_search.py
```diff
@@ -52,95 +66,139 @@ def __init__(self, url=None, works_index=None):
...
                 raise Exception("Cannot connect to Elasticsearch cluster.")
 
-        self.works_index = self.__client.works_index
         self.indices = self.__client.indices
         self.search = self.__client.search
         self.index = self.__client.index
         self.delete = self.__client.delete
         self.exists = self.__client.exists
+
+        self.works_index = self.works_alias = self.__client.works_index
```
aslagle: I think it might make sense for the alias to be determined once, and set on the __client the same way the index is.

external_search.py
courte: Yeah, you're right. I'm going to remove the `from_configuration` check here. It doesn't make a difference anyway, anymore.

Right now, there's no versioning implemented in this branch. The only time the version mapping would be accessed in the current arrangement is when the index doesn't exist.

If an `ELASTICSEARCH_INDEX_KEY` string is provided that doesn't exist as an index or an alias, there's no way for the initialization code to know which index to look for and upload documents to. This is the only case when it might create an index at a higher version than on that exists (but is unknown to the config).

Is there another case you see that I'm missing? Or could it be helpful for me to add a method to swap over the alias? That might make it clearer.

aslagle: I'd say if the provided string doesn't exist, the initialization code should create both the index and the alias. Otherwise, the code shouldn't change the alias, but it can create an index if that doesn't exist.

So if you provided works-v2 and nothing exists yet, it would create index works-v2 and alias works-current pointing to it.
If you provided works-v2 and works-current exists but works-v2 doesn't, it would create works-v2 but leave works-current the way it was.

Does that make sense?

I don't think the method to swap over the alias needs to be part of this branch, but you can add it now if you want to.

aslagle: There's also the case where you provide works-v2 and it exists already but has no alias. That shouldn't happen for any circ managers set up once we have this branch merged, but if it did I would think it should create the alias.

courte: Makes total sense, and that's almost what's happening right now: we create the alias or the index as needed. (I think the code flow is confusing. I'll add some more docstrings.)

The _only_ difference is that if you provide `works-v2` in configuration, and the `works-current` alias is set on `some-other-index` then the alias is moved from `some-other-index` to `works-v2`. Sounds like this isn't what we want. My assumption was that if you've changed to a different index name in config, that's the index you want to be using in production and thus should be "current".

So in that case, should the expectation be that you'll be point to `works-v2` without the alias or should the expectation be that you'll be pointing to `some-other-index` and the `works-current` alias despite what's in your configuration?

aslagle: I think if you point to `works-v2` which doesn't exist yet and the alias is set to something else, that means you are setting up a new index. The code you're running with this `works-v2` config should use the index instead of the alias, and any other code that's still running somewhere else with your old config can keep using the alias and continue to work.

When you're ready, you should do a separate step to move the alias, and the other code that's running will automatically start using the new index without any config change or downtime.

courte: Okay, I see, I see! Thanks!

courte:
@aslagle Would you take another look when you have a moment? I pulled out any alias overwriting from initialization and added some helper methods that you might be able to use in the scripts that you have in mind.

external_search.py
```diff
@@ -48,99 +60,189 @@ def __init__(self, url=None, works_index=None):
...
+        if current_alias.endswith(self.CURRENT_ALIAS_SUFFIX):
+            # The alias culled from configuration is intended to be
+            # a current alias. Get the index name from here.
+            base_works_index = self._base_works_index(current_alias)
+            version = ExternalSearchIndexVersions.latest()
+            _set_works_index(base_works_index+'-'+version)
+        else:
+            # Without the CURRENT_ALIAS_SUFFIX, assume the index string
+            # from config is the index itself and needs to be swapped.
+            _set_works_index(current_alias)
```
aslagle: This looks a lot better :)

Sorry I keep changing my mind about how this should work. Looking at it now I think the new index creation process is a little unclear.

I think we have two options, that maybe both work right now, and we should pick one.
1. Run a script with a config that has the index name instead of the alias to create the new index.
2. Update the latest version number in the code, and run a script with the regular config file.

I think I like 2, but I'm trying to think through what happens. When you switch the alias to the new index, would the app be unhappy if it happened to restart before the new code was deployed, and its internal version number didn't match the index pointed to by the alias?

external_search.py
```diff
@@ -48,99 +60,189 @@ def __init__(self, url=None, works_index=None):
...
-                                },
+        index = new_index or self.works_index
+
+        if self.indices.exists(index):
+            self.indices.delete(index)
+
+        self.log.info("Creating index %s", index)
+        self.indices.create(
+            index=index,
+            body={
```
aslagle: The body should be tied to the version number as well as the mapping - the analyzers are one of the more likely things to change between versions.

aslagle: Although, it wouldn't be a breaking change if it did.

external_search.py
courte: It's all good -- we're getting there!

Both work, but #1 is really just a desire to keep apps that are currently running with index-based configuration going without incident. In practice, I would prefer we use a script with `setup_index(new_index="")` to create something new instead of using configuration for it at all. I've been trying to cover all bases, but we've been through some iterations now so might have cleared up what I was worried about. Is there anything obvious you see that I can remove? I'll do a more thorough walkthrough, too.

If we depend on scripts instead of configuration in that way, for #2 the workflow would look like:
  1. Write some code to represent the newest version, add it to `ExternalSearchIndexVersions.VERSIONS`, deploy wherever
  2. Run the (currently nonexistent) create-new-search-version script, which would create an index, appropriately named, with the latest version structure.
  3. Run the existing `UpdateSearchIndexScript` to fill up the new index and do whatever testing or evaluation needs doing.
  4. When satisfied, run the (currently nonexistent) transfer-alias-to-new-index script to alter where the alias is pointing.

Does that seem workable to you?

external_search.py
courte: On it!

external_search.py
aslagle: That sounds good for 2. I'd like to have a migration script to create the alias for any apps that are still running with an index in their config. They'll have to make a config change, but I don't think there are any in production, unless brooklyn. Hmm...

courte:
@aslagle: Okay, the versions include the whole body now, and we're not every making indices anymore. I put the migration on the circulation branch, which hasn't changed much otherwise.

external_search.py
```diff
@@ -33,114 +42,158 @@ def __init__(self, url=None, works_index=None):
...
+        self.set_works_index_and_alias(works_index)
+
         def bulk(docs, **kwargs):
             return elasticsearch_bulk(self.__client, docs, **kwargs)
         self.bulk = bulk
-            
+
+    def set_works_index_and_alias(self, current_alias):
+        """Finds or creates the index based on provided configuration"""
+        index_details = self.indices.get_alias(name=current_alias, ignore=[404])
```
aslagle: I'm not sure what this does. Does get_alias return the index, not the alias?

external_search.py
```diff
@@ -33,114 +42,158 @@ def __init__(self, url=None, works_index=None):
...
+
         if not self.indices.exists(self.works_index):
             self.setup_index()
+        self.setup_current_alias()
 
-    def setup_index(self):
-        """
-        Create the search index with appropriate mapping.
+    def setup_current_alias(self):
+        """Put an alias ending with '-current' on the existing works_index."""
```
aslagle: This docstring could use an update.

external_search.py
courte: It returns a dict of the index and it's alias information, like`{"circulation-works-v2" : { "aliases" : { "circulation-works-current" : {}, "other-alias" : {}}}}`. If you use the `ignore` keyword, it will return something like `{ "status" : 404, "error" : "Not found" }` if the alias doesn't exist.

external_search.py
```diff
@@ -33,114 +42,158 @@ def __init__(self, url=None, works_index=None):
...
+        self.indices.create(index=index, body=body)
+
+    def transfer_current_alias(self, new_index):
+        """Force -current alias onto a new index"""
+        if not self.indices.exists(index=new_index):
+            raise ValueError(
+                "Index '%s' does not exist on this client." % new_index)
+
+        self.works_index = self.__client.works_index = new_index
+        alias_name = self._base_works_index(new_index)+self.CURRENT_ALIAS_SUFFIX
```
aslagle: It shouldn't be possible for the name of the alias to change as part of the transfer. Maybe that means this method should raise an error if the base_works_index differs between the new index and the old index.

courte:
@aslagle: Changes made! I rebased so it might be a bit of a pain depending on your process - my bad.

external_search.py
```diff
@@ -33,114 +42,176 @@ def __init__(self, url=None, works_index=None):
...
+            # We found an index for the alias in configuration. Assume
+            # there is only one.
+            _set_works_index(index_details.keys()[0])
+        else:
+            if current_alias.endswith(self.CURRENT_ALIAS_SUFFIX):
+                # The alias culled from configuration is intended to be
+                # a current alias, but we couldn't find an index with that
+                # alias.
+                raise ValueError(
+                    "External search alias '%s' not found." % current_alias)
```
aslagle: Is this where you end up if nothing exists yet and you put the alias name in your config file? In that case, I think it should set the index based on the latest version so a new one gets created. It looks like that would work if you took out this error. Is there another case this is supposed to handle?

courte: Yes! I'm changing this, but if the index that would be created (e.g. "test_index-v2") already exists (This won't happen if the migration has been run.), do we want to just point the alias at that index? That's the behavior the error was aiming to avoid way back when.

aslagle: I think that's fine, now that we're not moving the alias if it's already pointing to something else.

courte: Just pushed an option.

external_search.py
```diff
@@ -33,114 +42,189 @@ def __init__(self, url=None, works_index=None):
...
+                if not exists:
+                    created = ExternalSearchIndexVersions.create_new_version(
+                        self, base_index_name)
+                if exists or created:
+                    _set_works_index(new_index)
+                else:
+                    raise ValueError(
+                        ("Requested external search alias '%s' not found"
+                         " and new index '%s' could not be found or created.")
+                        % (current_alias, new_index))
```
aslagle: This is good, but could you set the new index and let the call to self.setup_index below handle it? That would be simpler if it works.

(sorry if you get this comment twice, I thought I wrote it on the commit but it disappeared.)

courte:
🎊 😄 Thanks for all the review support!

PR #465, opened 2017-02-23
Rename Collection to Catalog, and create a new Collection data model class

leonardr:
This finishes the implementation of https://github.com/NYPL-Simplified/circulation/issues/451, and implements https://github.com/NYPL-Simplified/circulation/issues/452.

This branch creates a `Collection` data model class, which represents the set of `LicensePool`s which were obtained from a distributor with a certain set of credentials. The `Collection` is not tied to `DataSource`; rather, the `Collection` is tied to a protocol (a piece of code you run to get `LicensePool`s) and a set of credentials (username, password, library ID, source URL). Most protocols are closely tied to a `DataSource` (the "Overdrive" protocol only gives you Overdrive books), but it's not necessarily so.

The goal of `Collection` is to replace the `"integrations"` JSON configuration for Overdrive, Bibliotheca, Axis 360, OneClickdigital, the Library Simplified content server, and any future ODL or OPDS-based integrations. Instead of being stored in JSON, this information will be stored in the database, where it can be edited.

The schema of `Collection` is designed to hold the basic information associated with collection credentials. The `CollectionSetting` data model class keeps any extra configuration information which doesn't fit the schema. Right now I think the only example is Overdrive's "library ID".

This is a good time to critique the design of the `Collection` data model. It's closely based on https://github.com/NYPL-Simplified/Simplified/wiki/MultipleCollections, except I took out the tie between `Collection` and `DataSource` because I don't think it's necessary.

There already is a `Collection` class, which is a collection of `Identifier`s. That class is intended to be used in the metadata wrangler to keep track of which books a given library wants to track metadata updates for. There is controller code for this class but it's not used in production yet, so I've simply renamed `Collection` to `Catalog` and written a migration script which drops the old tables (they'll be automatically recreated under the new names). The difference between `Collection` and `Catalog` is the difference between a bookshelf and a card catalog.

This branch also follows up on PR #458 by adding two fields to `Library`: `library_registry_short_name` and `library_registry_shared_secret`. This lets us move a couple fields out of JSON configuration and into the database. (The actual work of doing this happens in https://github.com/NYPL-Simplified/circulation/pull/455.)

### DIFF
--- a/model.py
+++ b/model.py
@@ -8345,13 +8345,7 @@ class Library(Base):
     
     # A URN that uniquely identifies the library, used to serve the
     # library's Authentication for OPDS document.
-    urn = Column(Unicode, index=True)
-    
-    # A short name and secret for this library, shared with a library
-    # registry, used to create short client tokens that a patron can
-    # use to get an Adobe Account ID.
-    adobe_short_name = Column(Unicode, index=True)
-    adobe_shared_secret = Column(Unicode)
+    urn = Column(Unicode)
 
     __table_args__ = (
         UniqueConstraint('urn'),
@@ -8367,6 +8361,14 @@ class Library(Base):
             )
         return value.upper()    
 
+    @classmethod
+    def instance(cls, _db):
+        """Find the one and only library."""
+        return get_one(
+            _db, Library, create_method_kwargs=dict(
+                uuid=uuid.uuid4()
+            )
+        )
 
 class Admin(Base):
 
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -8434,6 +8432,68 @@ def resolve(self):
...
+        library, is_new = get_one_or_create(
+            _db, Library, create_method_kwargs=dict(
+                uuid=str(uuid.uuid4())
+            )
+        )
+        return library
+
+    @hybrid_property
+    def library_registry_short_name(self):
+        """Gets library_registry_shared_secret from database"""
```
aslagle: This comment doesn't match the method.

aslagle (approved):
Looks good except the one comment

leonardr:
I've merged this into the `library-table` branch as a basis for future work.

PR #467, opened 2017-02-27
Pull API configuration from Collection objects

leonardr:
This branch changes the constructors for API objects (ThreeMAPI, OverdriveAPI, etc) so that rather than taking individual bits of configuration information, they take a `Collection` object with the configuration information. The idea is that you instantiate an API object for a specific API account. There's no longer an implication that every piece of the circulation manager always authenticates against the Overdrive API with the exact same credentials.

However, that assumption is still present in class methods called `from_environment` (which originally pulled API configuration from environment variables, and then from JSON files), which assume there is a single `Library` object with (at most) a single associated `Collection` of the appropriate type. Once these assumptions go away, the `from_environment` methods will also have to go away.

### DIFF
--- a/axis.py
+++ b/axis.py
@@ -49,6 +49,12 @@ class Axis360API(object):
 
     PRODUCTION_BASE_URL = "https://axis360api.baker-taylor.com/Services/VendorAPI/"
     QA_BASE_URL = "http://axis360apiqa.baker-taylor.com/Services/VendorAPI/"
+
+    # Map simple nicknames to server URLs.
+    SERVER_NICKNAMES = {
+        "production" : PRODUCTION_BASE_URL,
+        "qa" : QA_BASE_URL,
+    }
     
     DATE_FORMAT = "%m-%d-%Y %H:%M:%S"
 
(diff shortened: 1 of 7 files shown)


overdrive.py
```diff
@@ -247,15 +251,18 @@ def _get_book_list_page(self, link, rel_to_follow='next'):
         """
         # We don't cache this because it changes constantly.
         status_code, headers, content = self.get(link, {})
-        data = json.loads(content)
+        if isinstance(content, basestring):
+            content = json.loads(content)
```
leonardr: See https://github.com/NYPL-Simplified/circulation/pull/455/files#r103262011. I made this change, and similar changes in core, to accommodate the way Overdrive is mocked in the circulation tests.

axis.py
```diff
@@ -54,50 +63,50 @@ class Axis360API(object):
...
+        self.password = collection.password
+
+        # Convert the nickname for a server into an actual URL.
+        base_url = collection.url
+        if base_url in self.SERVER_NICKNAMES:
+            base_url = self.SERVER_NICKNAMES[base_url]
+        self.base_url = base_url
+
+        if (not self.library_id or not self.username
+            or not self.password):
```
aslagle: is it ok if there's no base url?

overdrive.py
```diff
@@ -98,49 +101,50 @@ class OverdriveAPI(object):
...
    
-    def __init__(self, _db, testing=False):
+    def __init__(self, _db, collection):
         self._db = _db
+        
+        if collection.protocol != collection.OVERDRIVE:
+            raise ValueError(
+                "Collection protocol is %s, but passed into OverdriveAPI!" %
+                collection.protocol
+            )
```
aslagle: Is there a reason Overdrive is the only one that needs to check this?

leonardr: No, they should all do it.

axis.py
leonardr: I think it should default to production.

PR #468, opened 2017-02-28
Add library and collection scripts

leonardr:
This branch adds four new scripts. One of them lists the libraries in the database, and one of them creates a new library or reconfigures an existing library.

The other two scripts do the same for collections.

The goal is to provide a simple way of doing initial configuration from the command line, and to provide a stopgap until the administrative interface allows the configuration of library and collection.

These scripts are in core rather than in circulation because I'm pretty sure the content server, at least, is going to use the concept of collection.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -812,7 +812,7 @@ class ConfigureCollectionScript(Script):
                     )
                 key, value = setting.split('=', 1)
                 collection.setting(key).value = value
-        if args.library:
+        if hasattr(args, 'library'):
             for name in args.library:
                 library = get_one(_db, Library, short_name=name)
                 if not library:
--- a/scripts.py
+++ b/scripts.py
@@ -621,7 +621,6 @@ class ConfigureLibraryScript(Script):
     def do_run(self, _db=None, cmd_args=None, output=sys.stdout):
         _db = _db or self._db
         args = self.parse_command_line(_db, cmd_args=cmd_args)
-        self.args = args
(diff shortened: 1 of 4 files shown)


aslagle:
Did you forget to add script files in bin or will those be in circulation?

leonardr:
They'll be in circulation.

scripts.py
```diff
@@ -541,6 +546,289 @@ def refresh_metadata(self, identifier):
...
+        if library_names:
+            return '"' + '", "'.join(library_names) + '"'
+        return ""
+    
+    def do_run(self, _db=None, cmd_args=None, output=sys.stdout):
+        _db = _db or self._db
+        args = self.parse_command_line(_db, cmd_args=cmd_args)
+
+        # Find or create the collection
+        create_kwargs = dict()
```
aslagle: you didn't end up using this?

tests/test_model.py
```diff
@@ -5292,6 +5316,40 @@ def test_set_key_value_pair(self):
         eq_(setting2, collection.setting("website_id"))
 
 
+    def test_explain(self):
+        """Test that Library.explain gives all relevant information
+        about a Library.
```
aslagle: This is Collection.explain

aslagle (approved):
Looks good minus an extraneous variable and an incorrect comment

leonardr:
Travis integration is broken because of the S3 outage. I've run the test suite locally and verified it passes.

PR #469, opened 2017-03-01
Begin support for Overdrive Advantage

leonardr:
This branch makes it possible to represent an Overdrive Advantage account as a `Collection`. 

A library consortium may have an Overdrive account. The consortial collection is shared among all member libraries. On top of this, each member library may have an Overdrive Advantage account associated with the library consortium. Books in a library's OA account are only accessible to patrons of that library.

An OA account is associated with an Overdrive account, and uses that account's credentials to make requests. But it has its own ID which is used along with the library ID of the consortial Overdrive account.

Overdrive makes it easy for a library to see a combined collection that incorporates the consortial collection and the OA collection. In a world where each library has its own circulation manager, this is good enough. But under multi-tenancy, it's extremely inefficient--it means twenty or thirty `Collection`s on a single circulation manager that are almost entirely duplicates. It's more efficient to have one big `Collection` representing the consortial collection, and twenty or thirty tiny `Collection`s that only contain a few books.

This branch makes it possible for one `Collection` to have another `Collection` as its parent. In the Overdrive context, this means the child is an Overdrive Advantage account that uses the same credentials as its parent.

This branch also creates code that parses the list of Overdrive Advantage collections available for a given Overdrive account (documentation: https://developer.overdrive.com/docs/advantage-link), potentially turning those collections into `Collection` objects.

### DIFF
--- a/model.py
+++ b/model.py
@@ -8556,6 +8556,20 @@ class Collection(Base):
     settings = relationship(
         "CollectionSetting", backref="collection"
     )
+
+    # A Collection may reuse the credentials from some other
+    # Collection. For instance, an Overdrive Advantage collection is a
+    # specialization of an ordinary Overdrive collection. It uses the
+    # same access key and secret as the Overdrive collection, but it
+    # has a distinct external_account_id.
+    parent_id = Column(Integer, ForeignKey('collections.id'), index=True)
+
+    # A collection may have many child collections. For example,
+    # An Overdrive collection may have many children corresponding
+    # to Overdrive Advantage collections.
+    children = relationship(
+        "Collection", backref="parent", remote_side = [id]
+    )
     
     # A Collection can provide books to many Libraries.
     libraries = relationship(
--- a/model.py
+++ b/model.py
@@ -8603,7 +8603,8 @@ class Collection(Base):
     # An Overdrive collection may have many children corresponding
     # to Overdrive Advantage collections.
     children = relationship(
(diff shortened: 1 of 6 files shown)


model.py
```diff
@@ -8645,7 +8662,7 @@ def explain(self, include_password=False):
         if self.url:
             lines.append('URL: "%s"' % self.url)
         if self.username:
-            lines.append('Username: "%s"' % self.url)
+            lines.append('Username: "%s"' % self.username)
```
leonardr: This was a stupid error not caught earlier.

leonardr:
Still having problems with Travis but I've run the test suite locally.

overdrive.py
```diff
@@ -880,8 +954,75 @@ def book_info_to_metadata(cls, book, include_bibliographic=True, include_formats
...
+    """
+    
+    def __init__(self, parent_library_id, library_id, name, type):
+        """Constructor.
+        
+        :param parent_library_id: The library ID of the parent Overdrive 
+            account.
+        :param library_id: The library ID of the Overdrive Advantage account.
+        :param name: The name of the library whose Advantage account this is.
+        :param type: Should be the literal string 'Library Advantage Account'
```
aslagle: Why is this an argument if there's only one allowed value?

leonardr: I don't know for sure that it's the only allowed value. I think I could take that out--it's not important right now.

PR #470, opened 2017-03-02
CustomListEntry links to Work, not LicensePool

leonardr:
This branch gets rid of the CustomListEntry.license_pool_id field and institutes CustomListEntry.work_id instead. When a Lane is built off of a CustomList, we join against the work ID instead of the license pool ID.

We don't want a book to show up in a Lane twice just because its Work has two LicensePools, but there's already a DISTINCT clause in the relevant part of Lane.apply_filters() (lane.py line 1169), to take care of the case where a book shows up twice because it's on multiple list. That should also take care of the case where a Work has multiple LicensePools.

Getting rid of this connection is a prerequisite for allowing multiple LicensePools for the same identifier (which is a prerequisite for supporting multiple collections that might offer the same book).

I've run the circulation tests and the content tests against this branch and didn't encounter any problems. I'm pretty sure those components only interact with CustomListEntries via Lanes.

### DIFF
--- a/external_list.py
+++ b/external_list.py
@@ -193,7 +193,7 @@ class TitleFromExternalList(object):
             
         list_entry.annotation = self.annotation
 
-        list_entry.set_license_pool(self.metadata, metadata_client)
+        list_entry.set_work(self.metadata, metadata_client)
         return list_entry, is_new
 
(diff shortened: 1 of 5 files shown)


courte:
Hey there - I missed that the migration wasn't added. 😿 (At least it's easier than adding Collections to everything!)

PR #472, opened 2017-03-02
Add customlistentries migration

courte:
This adds a migration to support #470 and shifts the migration naming so they all have numbers. (Turns out the name changing wasn't necessary, but it's done now.) It also removes an extraneous semicolon on a previous migration.


leonardr:
:+1: Sorry for putting you through this. I wrote a very similar migration but forgot to add it to the branch!

PR #473, opened 2017-03-03
Add cascade to collections table drop

courte:
An error came up for foreign_key constraints without this cascade.


leonardr:
:+1: thanks!

PR #475, opened 2017-03-10
Fixed problem that caused app to use test database

aslagle:
A recent change to core introduced an import of testing.py during app initialization, and testing.py was setting os.environ['TESTING'] to true, which caused the app to use the test database. 

I tried removing the line that set os.environ['TESTING'], and it didn't seem to cause any problems with the tests, but log.py was checking to determine the log level. I moved the variable to get set in `./test`, but maybe there's a better place for it. It won't be set when running a single test with nose, for example. And it would also need to be added to `./test` in the apps.

### DIFF
--- a/model.py
+++ b/model.py
@@ -8484,7 +8484,7 @@ class Library(Base):
     # consumption by the library registry.
     library_registry_shared_secret = Column(Unicode, unique=True)
 
-    def __repr__(cls):
+    def __repr__(self):
         return '<Library: name="%s", short name="%s", uuid="%s", library registry short name="%s">' % (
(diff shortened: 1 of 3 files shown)


model.py
```diff
@@ -8484,7 +8484,7 @@ class Library(Base):
     # consumption by the library registry.
     library_registry_shared_secret = Column(Unicode, unique=True)
 
-    def __repr__(cls):
+    def __repr__(self):
```
aslagle: This is an unrelated problem I found.

leonardr:
I think this is fine.

PR #478, opened 2017-03-15
Switch OneClickAPI over to the Collection-based system.

leonardr:
This branch corrects an earlier oversight, where I converted the API classes for other content sources over to the Collection-based system, but I didn't touch the OneClickAPI.

Some of the items found in the JSON OneClickAPI configuration are being removed in the conversion. I don't think we actually use them and I'm changing the way we treat them to reflect what I think actually happens within OneClickAPI. For example, the Collection.password field contains the Basic Auth token we get from OneCliek, not the 'password' we are given by OneClick, because  we use the Basic Auth token to authenticate requests (the purpose of a password) and we don't use the 'password' for anything.

There are also some other changes in here to support https://github.com/NYPL-Simplified/circulation/pull/473/files

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -32,6 +32,8 @@ class ExternalSearchIndex(object):
     def __init__(self, url=None, works_index=None):
     
         self.log = logging.getLogger("External search index")
+        self.works_index = None
+        self.works_alias = None
 
         if not ExternalSearchIndex.__client:
             if not url or not works_index:
@@ -59,7 +61,6 @@ class ExternalSearchIndex(object):
             ExternalSearchIndex.__client = Elasticsearch(
                 url, use_ssl=use_ssl, timeout=20, maxsize=25
             )
-
         self.indices = self.__client.indices
         self.search = self.__client.search
         self.index = self.__client.index
@@ -69,8 +70,6 @@ class ExternalSearchIndex(object):
         # Sets self.works_index and self.works_alias values.
         # Document upload runs against the works_index.
         # Search queries run against works_alias.
-        self.works_index = None
-        self.works_alias = None
         self.set_works_index_and_alias(works_index)
 
         def bulk(docs, **kwargs):
--- a/oneclick.py
+++ b/oneclick.py
@@ -81,7 +81,6 @@ class OneClickAPI(object):
             )
         
         self.library_id = collection.external_account_id.encode("utf8")
-        self.username = collection.username.encode("utf8")
         self.token = collection.password.encode("utf8")
 
         # Convert the nickname for a server into an actual URL.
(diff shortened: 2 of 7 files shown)


external_search.py
```diff
@@ -32,6 +32,8 @@ def reset(cls):
     def __init__(self, url=None, works_index=None):
     
         self.log = logging.getLogger("External search index")
+        self.works_index = None
+        self.works_alias = None
```
leonardr: This stops a crash in a test scenario where the app server is started up without any search configuration.

overdrive.py
```diff
@@ -407,6 +406,7 @@ def _do_post(self, url, payload, headers, **kwargs):
 class MockOverdriveAPI(OverdriveAPI):
 
     def __init__(self, _db, collection=None, *args, **kwargs):
+        self.access_token_requests = []
```
leonardr: In https://github.com/NYPL-Simplified/circulation/pull/473 we become very interested in which requests were made to `token_post`, because we need to see whether we included a PIN or told Overdrive not to check the PIN.

util/http.py
```diff
@@ -27,7 +27,8 @@ def __init__(self, url_or_service, message, debug_message=None):
            (e.g. "Overdrive"), or the specific URL that had the problem.
         """
         super(RemoteIntegrationException, self).__init__(message)
-        if any(url_or_service.startswith(x) for x in ('http:', 'https:')):
+        if (url_or_service and
```
leonardr: Sometimes `url_or_service` isn't set. Usually during testing, but it could happen for real (if we try to connect to a server but the hostname is specified as None). There's no reason why that should cause an error here.

oneclick.py
```diff
@@ -52,28 +55,44 @@
...
-        self.token = basic_token
+
+        if collection.protocol != collection.ONE_CLICK:
+            raise ValueError(
+                "Collection protocol is %s, but passed into OneClickAPI!" %
+                collection.protocol
+            )
+        
+        self.library_id = collection.external_account_id.encode("utf8")
+        self.username = collection.username.encode("utf8")
```
daryanypl: Fyi: I'm looking at the OneClick code, and it looks like we're not actually using the library username to talk to them.

oneclick.py
```diff
@@ -91,38 +110,26 @@ def create_identifier_strings(cls, identifiers):
...
 
+        if len(collections) > 1:
+            raise ValueError(
+                "Multiple OneClick collections found for one library. This is not yet supported."
+            )
+        [collection] = collections 
+
+        try:
+            return cls(_db, collection)
+        except CannotLoadConfiguration, e:
```
daryanypl: maybe output an error to log

daryanypl:
👍  but there's one extra log.error that I'd like

oneclick.py
leonardr: Thanks, this is exactly the kind of thing I wanted your opinion on.

oneclick.py
leonardr: After looking at the code I don't think it's necessary to muffle these errors. If a Collection is present but improperly configured, an exception _should_ be raised here.

PR #481, opened 2017-03-22
Support cataloguing collections

courte:
This branch gets rid of the separate concept of the `Catalog` and uses `Collection` to hold its own Catalog on the metadata wrangler. It also creates a `ClientServer` class for authenticated server-to-server interactions. A couple additional areas for feedback:

- I don't want to hold the unique identifier (`Collection` username, password, url, etc.) all willy-nilly in the database, so on the metadata wangler a collection will have its "metadata_identifier" as its name. Thoughts?
- Identifying a ClientServer by its `url` -- any changes to that url will need to be handled in the admin interface by hitting an endpoint in the metadata wrangler, as will registering the server in the first place

This supports an upcoming metadata wrangler branch in fending off NYPL-Simplified/circulation#467.

### DIFF
--- a/model.py
+++ b/model.py
@@ -8789,7 +8789,7 @@ collections_licensepools = Table(
 )
 
 collections_identifiers = Table(
-    'collectionsidentifiers', Base.metadata,
+    'collections_identifiers', Base.metadata,
     Column(
         'collection_id', Integer, ForeignKey('collections.id'),
         index=True, nullable=False
--- a/model.py
+++ b/model.py
@@ -8724,145 +8724,92 @@ collections_licensepools = Table(
      UniqueConstraint('collection_id', 'licensepool_id'),
  )
 
-    
-class Catalog(Base):
 
-    """A Catalog is like a Collection, but it doesn't hold any actual
-    LicensePools, it just records Identifiers.
 
-    The Collections associated with a Library in its circulation
-    manager will show up as Catalogs associated with the corresponding
-    Library in the metadata wrangler.
+class ClientServer(Base):
+    """A server that has authenticated access to this application.
+
+    Currently used to represent circulation managers that have access
+    to the metadata wrangler.
     """
-
-    __tablename__ = 'catalogs'
+    __tablename__ = 'servers'
 
     id = Column(Integer, primary_key=True)
-    name = Column(Unicode, unique=True, nullable=False)
-    client_id = Column(Unicode, unique=True, index=True)
-    _client_secret = Column(Unicode, nullable=False)
 
-    # A catalog can have one DataSource
-    data_source_id = Column(
-        Integer, ForeignKey('datasources.id'), index=True
-    )
+    # URL or human readable name to represent the server.
+    name = Column(Unicode)
 
-    # A catalog can include many Identifiers
-    catalog = relationship(
-        "Identifier", secondary=lambda: catalogs_identifiers,
-        backref="catalogs"
-    )
+    # Unique identifier
+    key = Column(Unicode, unique=True, index=True)
 
+    # Encrypted secret
+    _secret = Column(Unicode, nullable=False)
+
+    created = Column(DateTime)
+    last_accessed = Column(DateTime)
 
     def __repr__(self):
-        return "%s ID=%s DATASOURCE_ID=%d" % (
-            self.name, self.id, self.data_source.id
-        )
+        return (u"%s ID=%s" % (self.name, self.id)).encode('utf8')
 
     @hybrid_property
-    def client_secret(self):
-        """Gets encrypted client_secret from database"""
-        return self._client_secret
+    def secret(self):
+        """Gets encrypted client secret from database"""
+        return self._secret
 
-    @client_secret.setter
-    def _set_client_secret(self, plaintext_secret):
+    @secret.setter
+    def _set_secret(self, plaintext_secret):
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -8630,6 +8648,48 @@ class Collection(Base):
...
+    def metadata_identifier(self):
+        """Identifier based on collection details that uniquely represents
+        this Collection on the metadata wrangler. This identifier is
+        composed of the Collection protocol and account identifier.
+
+        In the metadata wrangler, this identifier is used as the unique
+        name of the collection.
+        """
+        account_id_column = self.UNIQUE_IDENTIFIER_BY_PROTOCOL.get(self.protocol)
+        account_id = getattr(self, account_id_column.name)
```
leonardr: If the collection has a parent ID, its metadata identifier needs to include the metadata identifier for its parent. So an Overdrive Advantage account might have a metadata Id like `MY LIBRARY / 27`.

model.py
```diff
@@ -8630,6 +8648,48 @@ class Collection(Base):
...
+
+        if ':' in self.protocol:
+            # This will break the base64 decoding on the metadata wrangler.
+            raise ValueError(
+                "Identifiers for protocols with ':' cannot be generated."
+            )
```
leonardr: If necessary we can loosen this requirement by encoding the colon before doing the base64 encoding, but this is good enough for now.

tests/test_model.py
```diff
@@ -5390,64 +5387,26 @@ def test_explain(self):
...
-        eq_(None, result)
-
     def test_catalog_identifier(self):
         """#catalog_identifier associates an identifier with the catalog"""
-
         identifier = self._identifier()
-        self.catalog.catalog_identifier(self._db, identifier)
-        eq_(1, len(self.catalog.catalog))
-        eq_(identifier, self.catalog.catalog[0])
+        self.collection.catalog_identifier(self._db, identifier)
```
leonardr: You need to verify that the identifier was actually inserted.

model.py
```diff
@@ -8722,147 +8804,117 @@ class CollectionSetting(Base):
...
 
-    """A Catalog is like a Collection, but it doesn't hold any actual
-    LicensePools, it just records Identifiers.
 
-    The Collections associated with a Library in its circulation
-    manager will show up as Catalogs associated with the corresponding
-    Library in the metadata wrangler.
-    """
+class ClientServer(Base):
+    """A server that has authenticated access to this application.
```
leonardr: `ClientServer` is a very generic name. I would call it something like `IntegrationAccount` or `IntegrationClient`. It doesn't matter to us whether the thing making these requests is a 'server' or some other piece of software.

model.py
```diff
@@ -8610,6 +8609,25 @@ class Collection(Base):
     # external_account_id.
     parent_id = Column(Integer, ForeignKey('collections.id'), index=True)
 
+    # A dict capturing the column location of the unique collection
+    # account identifier in the collections table.
+    UNIQUE_IDENTIFIER_BY_PROTOCOL = {
+        # Axis 360 'username'
+        AXIS_360 : username,
```
leonardr: I think `external_account_id` (Axis "library ID") is a more reliable identifier of the collection. One user account in Axis 360 could conceivably have access to multiple libraries.

model.py
```diff
@@ -8610,6 +8609,25 @@ class Collection(Base):
...
+
+        # Bibliotheca 'account_id'
+        BIBLIOTHECA : username,
```
leonardr: external_account_id here as well.

model.py
```diff
@@ -8610,6 +8609,25 @@ class Collection(Base):
...
+
+        # OPDS_IMPORT 'url'
+        OPDS_IMPORT : url,
+
+        # One Click 'basic_token'
+        ONE_CLICK : password,
```
leonardr: `external_account_id` here as well.

model.py
```diff
@@ -8610,6 +8609,25 @@ class Collection(Base):
...
+
+        # Overdrive 'client_key'
+        OVERDRIVE : username,
```
leonardr: `external_account_id` here as well.

leonardr:
This is looking good. I was skeptical you'd be able to reuse `Collection` but you pulled it off. My main concern here is the name of the `ClientServer` class.

model.py
courte: "*Orange* you glad I didn't call it `Server`?" said in the tone of that knock-knock joke. (I'm going to go with `IntegrationClient`.)

model.py
```diff
@@ -8662,16 +8662,27 @@ def metadata_identifier(self):
...
-        account_id_column = self.UNIQUE_IDENTIFIER_BY_PROTOCOL.get(self.protocol)
-        account_id = getattr(self, account_id_column.name)
+        def get_account_id(collection):
+            account_id_column = self.UNIQUE_IDENTIFIER_BY_PROTOCOL.get(collection.protocol)
+            return getattr(collection, account_id_column.name)
+
+        collection = self
+        account_ids = list()
+        while collection:
+            account_id = get_account_id(collection)
```
leonardr: I think you could simplify this quite a bit by defining `account_ids` as a recursive method that got `parent.account_ids` as a list and stuck the final account ID on the end.

leonardr:
:+1: but take a stab at turning the loop inside `metadata_identifier` into a recursive call.

Issue #212, opened 2016-05-04
Multiple license pools per identifier+data source


leonardr:
Currently an Identifier can have at most one LicensePool. This makes it impossible to properly keep track of the Project Gutenberg and Project GITenberg editions of a given text, since they use the same numbering scheme.

If distributors ever stop coming up with their own numbering schemes and start identifying their books with ISBNs, it will be possible to buy two versions of the same book from different distributors, but not possible to represent this in a collection.

Expanding the requirement to "one LicensePool per Identifier+DataSource" will solve this problem, but it won't solve the consortium problem, when a library may have access to the same book _from the same distributor_ in multiple ways. (Say, one LicensePool from the state library and one LicensePool obtained locally.)

So we need to make LicensePool unique across identifier + data source + collection.

PR #483, opened 2017-03-24
Allow multiple licence pools for the same Identifier so long as Identifier+DataSource+Collection is unique

leonardr:
This branch addresses https://github.com/NYPL-Simplified/server_core/issues/212. It allows the database to contain multiple LicensePools for the same book, so long as Identifier+DataSource+Collection is unique.

All LicensePools with a given Identifier will have the same presentation Edition and be assigned the same Work.

Anything that might create or manipulate a `LicensePool` now needs to have a `Collection` handy so that we can distinguish between otherwise identical `LicensePool`s. 

I make the various API classes keep track of a `Collection` object. Previously their constructors took a `Collection` but it was discarded once the API credentials were extracted from it.

I've limited the number of things that require a `Collection` by making it possible to go directly from an `Identifier` to a `Work`. This is possible because I've added a rule that all `LicensePool`s for a given `Identifier` must have the same `Work`.

We are not yet at the point where we can usefully have different Collections that contain the same book, but this is a big step in that direction.

## Timestamp

I extended the `Timestamp` class to have a reference to a Collection. This lets us keep track of when a given collection-specific monitor ran for a certain Collection. I also had to give `Timestamp` a numeric primary key, since the name of the monitor no longer works as a unique identifier.

## Monitors

I introduced `CollectionMonitor`, which does... something... for all Collections that implement a certain protocol, one after the other. The idea is that monitors like OverdriveCirculationMonitor will start subclassing this class. This removes the need to specify which Collection you mean when updating Overdrive circulation information. All Collections will be processed, one after the other.

## Coverage Providers

Changing the `CoverageProvider` constructor to take a `Collection` meant I had to touch every single `CoverageProvider` subclass and jiggle its constructor. I was really worried about this causing hard-to-detect problems, and then I noticed why we have this problem. 

The coverage providers in core are abstract classes that must be specialized by subclassing. But a lot of the specializations (which data source to use, which identifier types to process) are passed into the constructor when the object is created. This makes tests a little easier to write, since you don't need to make a subclass just for a test, but it makes the constructor very complicated.

So I changed `BaseCoverageProvider` and its subclasses to take most of their configuration in class constants like SERVICE_NAME, OPERATION, and INPUT_IDENTIFIER_TYPES. This removes most of the complexity from the constructor signature. And the tests are actually better now--there are more one-off subclasses but there's less code overall.

I renamed `CoverageProvider` to `IdentifierCoverageProvider` so that we don't have both `CoverageProvider` and `BaseCoverageProvider`.

I went through `test_coverage.py` and verified that we have good test coverage for all our code. In places where specific methods were not tested in isolation I added stubs to test just those methods.

### BibliographicCoverageProvider

This branch exposes, but does not resolve, a problem with coverage providers that is most obvious with `BibliographicCoverageProvider`.  A coverage provider does some work for an Identifier, and then checks the identifier off of a list and never does any more work. In some situations, this can leave work done for one Collection and undone for others.

It often makes sense to use one Collection or another when doing this work. Some third parties (Bibliotheca, Axis 360, and OneClick) won't allow work to be done on a title except in the context of a Collection that has that title available. But the _result_ of running a CoverageProvider is supposed to be Collection-independent, because you're only doing the work once. Anything that might differ between collections is supposed to be handled in a `Monitor` or in some other way.

The problem is that Overdrive, Bibliotheca, and OneClick all provide data about available formats along with bibliographic information. It's possible for two different Collections to have the same book available--that's the whole reason we're doing this. The only time we learn about formats is when the BibliographicCoverageProvider runs. Since BCP now runs in the context of a specific Collection, the only formats we're going to learn about are the one in the Collection that happens to be passed into the BCP. The other LicensePool will never have any DeliveryMechanisms associated with it.

(Axis 360 always sends bibliographic and circulation data at the same time, so the BibliographicCoverageProvider isn't really necessary. We need to make sure it doesn't have this problem, but at least there's an obvious solution.)

## Resources

Another issue I don't tackle in this branch is the relationship between `Resource` and `LicensePool`. The simplest thing to do is probably remove this relationship--I think it's redundant.

### DIFF
--- a/axis.py
+++ b/axis.py
@@ -70,8 +70,7 @@ class Axis360API(object):
                 "Collection protocol is %s, but passed into Axis360API!" %
                 collection.protocol
             )
-        self.collection = collection
-        self._db = Session.object_session(self.collection)
+        self._db = Session.object_session(collection)
         self.library_id = collection.external_account_id.encode("utf8")
         self.username = collection.username.encode("utf8")
         self.password = collection.password.encode("utf8")
--- a/axis.py
+++ b/axis.py
@@ -262,7 +262,7 @@ class Axis360BibliographicCoverageProvider(BibliographicCoverageProvider):
                     self._db, Identifier.AXIS_360_ID, identifier_string
                 )
                 result = CoverageFailure(
-                    identifier, "Book not in collection", data_source=self.output_source, transient=False
+                    identifier, "Book not in collection", data_source=self.data_source, transient=False
                 )
                 batch_results.append(result)
         return batch_results
--- a/axis.py
+++ b/axis.py
@@ -282,8 +282,8 @@ class Axis360BibliographicCoverageProvider(BibliographicCoverageProvider):
                 identifier, ignore = Identifier.for_foreign_id(
                     self._db, Identifier.AXIS_360_ID, identifier_string
                 )
-                result = CoverageFailure(
-                    identifier, "Book not in collection", data_source=self.data_source, transient=False
+                result = self.failure(
+                    identifier, "Book not in collection", transient=False
                 )
                 batch_results.append(result)
         return batch_results
--- a/axis.py
+++ b/axis.py
@@ -6,6 +6,7 @@ import os
 import json
 import logging
 import re
+from sqlalchemy.orm.session import Session
 
 from config import (
     Configuration, 
@@ -63,15 +64,14 @@ class Axis360API(object):
 
     log = logging.getLogger("Axis 360 API")
 
-    def __init__(self, _db, collection):
+    def __init__(self, collection):
         if collection.protocol != collection.AXIS_360:
             raise ValueError(
                 "Collection protocol is %s, but passed into Axis360API!" %
                 collection.protocol
             )
-
-        self._db = _db
         self.collection = collection
+        self._db = Session.object_session(self.collection)
         self.library_id = collection.external_account_id.encode("utf8")
         self.username = collection.username.encode("utf8")
         self.password = collection.password.encode("utf8")
@@ -183,7 +183,9 @@ class Axis360API(object):
 
 class MockAxis360API(Axis360API):
 
-    def __init__(self, _db, with_token=True, *args, **kwargs):
+    @classmethod
+    def mock_collection(self, _db):
+        """Create a mock Axis 360 collection for use in tests."""
         library = Library.instance(_db)
         collection, ignore = get_one_or_create(
             _db, Collection,
@@ -194,7 +196,19 @@ class MockAxis360API(Axis360API):
             )
         )
         library.collections.append(collection)
-        super(MockAxis360API, self).__init__(_db, collection, *args, **kwargs)
+        return collection
+
+    def __init__(self, collection, with_token=True, **kwargs):
+        """Constructor.
+
+        :param collection: Get Axis 360 credentials from this
+            Collection.
+
+        :param with_token: If True, this class will assume that
+            it already has a valid token, and will not go through
+            the motions of negotiating one with the mock server.
+        """
+        super(MockAxis360API, self).__init__(collection, **kwargs)
         if with_token:
             self.token = "mock token"
         self.responses = []
@@ -222,19 +236,26 @@ class Axis360BibliographicCoverageProvider(BibliographicCoverageProvider):
     not normally necessary because the Axis 360 API combines
     bibliographic and availability data.
     """
-    def __init__(self, _db, metadata_replacement_policy=None, axis_360_api=None,
-                 input_identifier_types=None, input_identifiers=None, **kwargs):
+
+    SERVICE_NAME = "Axis 360 Bibliographic Coverage Provider"
+    DATA_SOURCE_NAME = DataSource.AXIS_360
+    PROTOCOL = Collection.AXIS_360
+    INPUT_IDENTIFIER_TYPES = Identifier.AXIS_360_ID
+    DEFAULT_BATCH_SIZE = 25
+    
+    def __init__(self, collection, api_class=Axis360API, **kwargs):
+        """Constructor.
+
+        :param collection: Provide bibliographic coverage to all
+            Axis 360 books in the given Collection.
+        :param api_class: Instantiate this class with the given Collection,
+            rather than instantiating Axis360API.
         """
-        :param input_identifier_types: Passed in by RunCoverageProviderScript, data sources to get coverage for.
-        :param input_identifiers: Passed in by RunCoverageProviderScript, specific identifiers to get coverage for.
-        """
-        self.parser = BibliographicParser()
         super(Axis360BibliographicCoverageProvider, self).__init__(
-            _db, axis_360_api, DataSource.AXIS_360,
-            batch_size=25, 
-            metadata_replacement_policy=metadata_replacement_policy,
-            **kwargs
+            collection, **kwargs
         )
+        self.api = api_class(collection)
+        self.parser = BibliographicParser()
 
     def process_batch(self, identifiers):
         identifier_strings = self.api.create_identifier_strings(identifiers)
--- a/bibliotheca.py
+++ b/bibliotheca.py
@@ -211,9 +211,7 @@ class MockBibliothecaAPI(BibliothecaAPI):
     def __init__(self, collection, *args, **kwargs):
         self.responses = []
         self.requests = []
-        super(MockBibliothecaAPI, self).__init__(
-            collection, *args, **kwargs
-        )
+        super(MockBibliothecaAPI, self).__init__(collection, *args, **kwargs)
 
     def now(self):
         """Return an unvarying time in the format Bibliotheca expects."""
@@ -428,12 +426,11 @@ class BibliothecaBibliographicCoverageProvider(BibliographicCoverageProvider):
             rather than instantiating BibliothecaAPI.
         :param input_identifiers: Passed in by RunCoverageProviderScript. 
             A list of specific identifiers to get coverage for.
-
         """
         super(BibliothecaBibliographicCoverageProvider, self).__init__(
             collection, **kwargs
         )
-        self.api = (api_class or BibliothecaAPI)(collection)
+        self.api = api_class(collection)
         
     def process_item(self, identifier):
         # We don't accept a representation from the cache because
--- a/bibliotheca.py
+++ b/bibliotheca.py
@@ -193,10 +193,9 @@ class BibliothecaAPI(object):
 
 class MockBibliothecaAPI(BibliothecaAPI):
 
-    def __init__(self, _db, *args, **kwargs):
-        self.responses = []
-        self.requests = []
-
+    @classmethod
+    def mock_collection(self, _db):
+        """Create a mock Bibliotheca collection for use in tests."""
         library = Library.instance(_db)
         collection, ignore = get_one_or_create(
             _db, Collection,
@@ -207,12 +206,17 @@ class MockBibliothecaAPI(BibliothecaAPI):
             )
         )
         library.collections.append(collection)
+        return collection
+        
+    def __init__(self, collection, *args, **kwargs):
+        self.responses = []
+        self.requests = []
         super(MockBibliothecaAPI, self).__init__(
             collection, *args, **kwargs
         )
 
     def now(self):
-        """Return an unvarying time in the format 3M expects."""
+        """Return an unvarying time in the format Bibliotheca expects."""
         return datetime.strftime(
             datetime(2016, 1, 1), self.AUTH_TIME_FORMAT
         )
@@ -407,6 +411,7 @@ class BibliothecaBibliographicCoverageProvider(BibliographicCoverageProvider):
 
     Then mark the works as presentation-ready.
     """
+    SERVICE_NAME = "Bibliotheca Bibliographic Coverage Provider"
     DATA_SOURCE_NAME = DataSource.BIBLIOTHECA
     PROTOCOL = Collection.BIBLIOTHECA
     INPUT_IDENTIFIER_TYPES = Identifier.BIBLIOTHECA_ID
--- a/bibliotheca.py
+++ b/bibliotheca.py
@@ -8,6 +8,7 @@ import os
 import re
 import logging
 from datetime import datetime, timedelta
+from sqlalchemy.orm.session import Session
 
 from config import (
     Configuration,
@@ -15,8 +16,7 @@ from config import (
     temp_config,
 )
 from coverage import (
-    BibliographicCoverageProvider,
-    CoverageFailure,
+    BibliographicCoverageProvider
 )
 from model import (
     get_one_or_create,
@@ -47,7 +47,7 @@ from metadata_layer import (
 from util.http import HTTP
 from util.xmlparser import XMLParser
 
-class ThreeMAPI(object):
+class BibliothecaAPI(object):
 
     # TODO: %a and %b are localized per system, but 3M requires
     # English.
@@ -93,7 +93,7 @@ class ThreeMAPI(object):
 
     @property
     def source(self):
-        return DataSource.lookup(self._db, DataSource.THREEM)
+        return DataSource.lookup(self._db, DataSource.BIBLIOTHECA)
 
     def now(self):
         """Return the current GMT time in the format 3M expects."""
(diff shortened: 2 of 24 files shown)


coverage.py
```diff
@@ -395,26 +396,47 @@ class CoverageProvider(BaseCoverageProvider):
     Turns errors in processing into coverage records with failure flags.
     """
 
-    # Does this CoverageProvider get its data from a source that also
-    # provides licenses for books?
-    CAN_CREATE_LICENSE_POOLS = False
```
leonardr: This is no longer relevant. A CoverageProvider that is instantiated with a Collection can create LicensePools; a CoverageProvider that has no Collection cannot.

coverage.py
```diff
@@ -395,26 +396,47 @@ class CoverageProvider(BaseCoverageProvider):
...
+    def __init__(self, service_name, output_source, collection=None,
+                 input_identifier_types=None, batch_size=100,
+                 cutoff_time=None, operation=None, input_identifiers=None
+    ):
```
leonardr: There's a lot of cleanup here and adding of docstrings because I changed the signature of these constructors. I had to do a lot of work to make sure arguments were still being properly passed in.

coverage.py
```diff
@@ -463,101 +485,18 @@ def output_source(self):
...
-                # pools, so it's okay to automatically create
-                # a license pool for this book.
-                license_pool, ignore = LicensePool.for_foreign_id(
-                    self._db, self.output_source, identifier.type, 
-                    identifier.identifier
-                )
-            else:
-                return None
-        return license_pool
-
```
leonardr: This got moved to CollectionCoverageProvider. Normal CoverageProviders aren't supposed to call it.

(Maybe normal CoverageProviders shouldn't store a Collection at all?)

axis.py
```diff
@@ -63,15 +64,14 @@ class Axis360API(object):
 
     log = logging.getLogger("Axis 360 API")
 
-    def __init__(self, _db, collection):
+    def __init__(self, collection):
```
leonardr: I changed all the API constructors to just take a Collection and acquire the database connection from it using Session.object_session().

axis.py
```diff
@@ -89,26 +89,6 @@ def __init__(self, _db, collection):
             )
             
         self.token = None
-
-    @classmethod
-    def from_environment(cls, _db):
```
leonardr: I removed every `from_environment` call. You can no longer get "the default API" because there is no longer a "default Collection". This is going to break a lot of scripts.

axis.py
```diff
@@ -203,7 +183,9 @@ def _make_request(self, url, method, headers, data=None, params=None,
 
 class MockAxis360API(Axis360API):
 
-    def __init__(self, _db, with_token=True, *args, **kwargs):
+    @classmethod
+    def mock_collection(self, _db):
+        """Create a mock Axis 360 collection for use in tests."""
```
leonardr: I changed all the mock APIs to split out the code that creates a mock collection, so that it can be called separately.

bibliotheca.py
```diff
@@ -47,7 +47,7 @@
 from util.http import HTTP
 from util.xmlparser import XMLParser
 
-class ThreeMAPI(object):
+class BibliothecaAPI(object):
```
leonardr: I decided since I was already changing all the signatures, the time was right to rename all remaining "3M" stuff to "Bibliotheca".

coverage.py
```diff
@@ -350,83 +351,208 @@ def items_that_need_coverage(self, identifiers, **kwargs):
...
+
+    @property
+    def data_source(self):
+        """Look up the DataSource object corresponding to the
+        service we're running this data through.
+
+        Out of an excess of caution, we look up the DataSource every
+        time, rather than storing it, in case a CoverageProvider is
+        ever used in an environment where the database session is
+        scoped (e.g. the circulation manager).
```
leonardr: I'm not sure if this is helpful anymore, since we're also storing the `Collection` object with the `CoverageProvider`.

coverage.py
```diff
@@ -450,192 +576,51 @@ def ensure_coverage(self, item, force=False):
...
+        view of a given Identifier.
+        """
         edition, ignore = Edition.for_foreign_id(
-            self._db, license_pool.data_source, identifier.type,
+            self._db, self.data_source, identifier.type,
             identifier.identifier
         )
         return edition
 
-    def work(self, identifier):
```
leonardr: Also moved to CollectionCoverageProvider.

coverage.py
```diff
@@ -450,192 +576,51 @@ def ensure_coverage(self, item, force=False):
...
-        return work
-
-
-    def set_metadata(self, identifier, metadata, 
-                     metadata_replacement_policy=None):
-        return self.set_metadata_and_circulation_data(
-            identifier, metadata, None, metadata_replacement_policy,
-        )
-
-    def set_metadata_and_circulation_data(self, identifier, metadata, circulationdata,
```
leonardr: Also moved to CollectionCoverageProvider.

coverage.py
```diff
@@ -450,192 +576,51 @@ def ensure_coverage(self, item, force=False):
...
-            e = "No license pool available"
-            return CoverageFailure(identifier, e, data_source=self.output_source, transient=True)
-        work, created = license_pool.calculate_work(even_if_no_author=True)
-        if not work:
-            e = "Work could not be calculated"
-            return CoverageFailure(identifier, e, data_source=self.output_source, transient=True)
-        return work
-
-
-    def set_metadata(self, identifier, metadata,
```
leonardr: This was _not_ moved to CollectionCoverageProvider, it was just moved below this part of the diff.

coverage.py
```diff
@@ -450,192 +576,51 @@ def ensure_coverage(self, item, force=False):
...
-            self.log.warn(
-                "Error applying circulationdata to pool %d: %s",
-                pool.id, e, exc_info=e
-            )
-            return CoverageFailure(identifier, repr(e), data_source=self.output_source, transient=True)
 
         return identifier
 
-
-    def set_presentation_ready(self, identifier):
```
leonardr: Moved to CollectionCoverageProvider.

coverage.py
```diff
@@ -683,12 +669,281 @@ def failure_for_ignored_item(self, item):
...
+        """
+        qu = super(CollectionCoverageProvider, self).items_that_need_coverage(
+            identifiers, **kwargs
         )
+        qu = qu.join(Identifier.licensed_through).filter(
+            LicencePool.collection_id==self.collection_id
+        )
+        return qu
+        
+    def license_pool(self, identifier):
```
leonardr: Lots of stuff was moved here from IdentifierCoverageProvider.

model.py
```diff
@@ -6413,6 +6458,20 @@ def calculate_work(self, even_if_no_author=False, known_edition=None):
...
+                "LicensePools for %r have more than one Work between them. Removing them all and starting over."
+            )
+            for lp in self.identifier.licensed_through:
+                lp.work = None
+                if lp.presentation_edition:
+                    lp.presentation_edition.work = None
+        else:
+            # There is a consensus Work for this Identifier.
+            [self.work] = existing_works
+
```
leonardr: This is an important invariant that I'm introducing.

opds.py
```diff
@@ -1155,11 +1155,15 @@ def create_entry(self, work):
...
         # specific LicensePool, and it's possible that the client is
-        # asking for a lookup centered around a different LicensePool.
+        # asking for a lookup centered around a different edition of the
+        # same book.
         default_licensepool = self.annotator.active_licensepool_for(work)
-        active_licensepool = identifier.licensed_through
+        if identifier.licensed_through:
+            active_licensepool = identifier.licensed_through[0]
+        else:
+            active_licensepool = default_licensepool
```
leonardr: The OPDS side of things is going to need some more work as well, once we actually have multiple LicensePools per Work.

scripts.py
```diff
@@ -904,9 +945,9 @@ def run(self):
                     self.subject.identifier, self.subject.name,
                     self.weight
                 )
-                pool = identifier.licensed_through
-                if pool and pool.work:
-                    pool.work.calculate_presentation(policy=policy)
+                for pool in identifier.licensed_through:
+                    if pool.work:
+                        pool.work.calculate_presentation(policy=policy)
```
leonardr: TODO: I think this would be simpler if I used `identifier.work`.

axis.py
```diff
@@ -63,15 +64,14 @@ class Axis360API(object):
...
         if collection.protocol != collection.AXIS_360:
             raise ValueError(
                 "Collection protocol is %s, but passed into Axis360API!" %
                 collection.protocol
             )
-
-        self._db = _db
-
+        self.collection = collection
+        self._db = Session.object_session(self.collection)
```
aslagle: Check if we keep the api object around in the app server

coverage.py
```diff
@@ -79,18 +82,40 @@ class BaseCoverageProvider(object):
...
+    def __init__(self, _db, batch_size=None, cutoff_time=None):
         """Constructor.
 
-        :param service_name: The name of the coverage provider. Used in
-        log messages and Timestamp objects.
+        :param service_name: The name of the service that is providing
+        coverage. Used in log messages and Timestamp objects.
+
+        :param operation: An optional operation being performed by the
+        service. This lets one service perform multiple operations.
```
aslagle: Take these out

coverage.py
```diff
@@ -450,192 +576,51 @@ def ensure_coverage(self, item, force=False):
...
-        )
-
         edition = self.edition(identifier)
         if isinstance(edition, CoverageFailure):
             return edition
 
         if not metadata:
             e = "Did not receive metadata from input source"
-            return CoverageFailure(identifier, e, data_source=self.output_source, transient=True)
+            return CoverageFailure(
```
aslagle: use self.failure

coverage.py
```diff
@@ -450,192 +576,51 @@ def ensure_coverage(self, item, force=False):
...
-            return pool
-
-        if not circulationdata:
-            e = "Did not receive circulationdata from input source"
-            return CoverageFailure(identifier, e, data_source=self.output_source, transient=True)
-
-        try:
-            circulationdata.apply(
-                pool, replace=circulationdata_replacement_policy,
+            return CoverageFailure(
```
aslagle: use self.failure

coverage.py
```diff
@@ -683,12 +669,281 @@ def failure_for_ignored_item(self, item):
         """
         return CoverageFailure(
             item, "Was ignored by CoverageProvider.", 
-            data_source=self.output_source, transient=True
+            data_source=self.data_source, transient=True
```
aslagle: use self.failure

coverage.py
```diff
@@ -683,12 +669,281 @@ def failure_for_ignored_item(self, item):
...
+        
+    def items_that_need_coverage(self, identifiers=None, **kwargs):
+        """Find all Identifiers associated with this Collection but lacking
+        coverage through this CoverageProvider.
+        """
+        qu = super(CollectionCoverageProvider, self).items_that_need_coverage(
+            identifiers, **kwargs
         )
+        qu = qu.join(Identifier.licensed_through).filter(
+            LicencePool.collection_id==self.collection_id
```
aslagle: This should be LicensePool

coverage.py
```diff
@@ -683,12 +669,281 @@ def failure_for_ignored_item(self, item):
...
+            
+        if pool:
+            work, created = pool.calculate_work(even_if_no_author=True)
+            if not work:
+                error = "Work could not be calculated"
+        else:
+            error = "Cannot locate LicensePool"
+                
+        if error:
+            return CoverageFailure(
```
aslagle: use self.failure

coverage.py
```diff
@@ -683,12 +669,281 @@ def failure_for_ignored_item(self, item):
...
+        context of this Collection), and LicensePool (ditto), and that
+        all the information is up to date.
+
+        :return: The Identifier (if successful) or an appropriate
+        CoverageFailure (if not).
+        """
+
+        if not metadata and not circulationdata:
+            e = "Received neither metadata nor circulation data from input source"
+            return CoverageFailure(
```
aslagle: use self.failure

coverage.py
```diff
@@ -683,12 +669,281 @@ def failure_for_ignored_item(self, item):
...
+        :return: The Identifier (if successful) or an appropriate
+        CoverageFailure (if not).
+        """
+        pool = self.license_pool(identifier)
+        if isinstance(pool, CoverageFailure):
+            return pool
+
+        if not circulationdata:
+            e = "Did not receive circulationdata from input source"
+            return CoverageFailure(identifier, e, data_source=self.data_source, transient=True)
```
aslagle: self.failure

monitor.py
```diff
@@ -93,6 +116,49 @@ def cleanup(self):
...
+
+    This class is designed to be subclassed rather than instantiated
+    directly. Subclasses should define SERVICE_NAME and PROTOCOL.
+    """
+
+    # In your subclass, set this to the name of the service,
+    # e.g. "Overdrive Circulation Monitor". All instances of your
+    # subclass will give this as their service name and track their
+    # Timestamps under this name.
+    SERVICE_NAME = None
```
aslagle: move to monitor

opds_import.py
```diff
@@ -694,16 +706,17 @@ def summary_to_linkdata(detail):
...
         )
             
-        # Only add circulation data if both the book's distributor *and*
-        # the source of the OPDS feed are lendable data sources.
-        if (circulation_data_source and circulation_data_source.offers_licenses
+        # Only add circulation data if there is a collection, and if
+        # both the book's distributor *and* the source of the OPDS
+        # feed are lendable data sources.
+        if (circulation_data_source
+            and circulation_data_source.offers_licenses
```
aslagle: should check if there's a collection

scripts.py
```diff
@@ -1573,7 +1579,9 @@ def do_run(self, cmd_args=None):
                     "%s timestamp already exists: %r. Use --force to update." %
                     (self.name, existing_timestamp))
 
-        timestamp = existing_timestamp or Timestamp.stamp(self._db, self.name)
+        timestamp = existing_timestamp or Timestamp.stamp(
+            self._db, service=self.name, collection=None
```
aslagle: OPDS Import should be using a collection

tests/test_circulation_data.py
```diff
@@ -270,7 +276,9 @@ def test_rights_status_default_rights_passed_in(self):
             formats=True,
         )
 
-        pool, ignore = circulation_data.license_pool(self._db)
+        pool, ignore = circulation_data.license_pool(
+            self._db, self._collection()
```
aslagle: use _default_collection throughout

axis.py
leonardr: Yes, we store it as `CirculationManager.circulation`.

monitor.py
```diff
@@ -26,93 +28,150 @@
...
     @property
     def log(self):
         if not hasattr(self, '_log'):
             self._log = logging.getLogger(self.service_name)
         return self._log        
 
+    @property
+    def collection(self):
+        """Retrieve the Collection object associated with this
+        CoverageProvider.
```
aslagle: this comment should say Monitor

monitor.py
```diff
@@ -156,103 +228,159 @@ def all(cls, _db, **kwargs):
...
+    DEFAULT_BATCH_SIZE = 500
+    
+    def __init__(self, _db, collection=None, subject_type=None,
+                 filter_string=None):
+        """Constructor.
+
+        :param collection: This value is ignored. There's no way
+            to run a SubjectSweepMonitor on a specific collection,
+            because there's no obvious connection between a Subject
+            and a Collection.
```
aslagle: Why have this param at all?

oneclick.py
```diff
@@ -829,10 +840,10 @@ def __init__(self, collection, api_class=OneClickAPI, api_class_kwargs={},
             # generally the done thing, but it is necessary when a
             # OneClickAPI object itself wants a
             # OneClickBibliographicCoverageProvider.
-            if api_class.collection != collection:
+            if api_class.collection_id != collection.id:
                 raise ValueError(
                     "Coverage provider and its API are scoped to different collections! (%s vs. %s)" % (
-                        api_class.collection.name, collection.name
+                        api_class.collection_id, collection._id
```
aslagle: should be collection.id, no _

monitor.py
leonardr: I was thinking it would be better if all the Monitors shared a baseline set of constructor arguments so that they could take the same set of command-line arguments. Looking at the code we have now, I don't  think that's a strong argument. My other reason is that I can't think of a connection between Subject and Collection _now_, but I don't want to foreclose on the possibility,

aslagle: You could always put the argument back when you find the connection - you'd have to modify this class anyway. Unless there are already places that pass in a collection.

leonardr: Ok, it's gone.

aslagle:
:+1: The new changes look good.

PR #484, opened 2017-03-27
Added a database table for storing admin authentication service config.

aslagle:
This is a proposal for storing AuthenticationService configuration in the database. 

I'm not sure what if anything should be shared between admin authentication and patron authentication. For the apps we have now, admin and patrons use different auth providers. But for example, we could authenticate patrons with Google OAuth if someone wanted that, and in that case the auth configuration could be the same except the allowed domains.

Here I did something in the middle and created a db table specifically for admin auth config, but made a base class so it will have the same columns as the patron auth config table.

I'd like some feedback on this before I get too much farther with the admin interface changes.

### DIFF
--- a/model.py
+++ b/model.py
@@ -8642,15 +8642,10 @@ class AdminAuthenticationService(Base):
     @property
     def external_integration(self):
         _db = Session.object_session(self)
-        if self.external_integration_id:
-            external_integration = get_one(
-                _db, ExternalIntegration, id=self.external_integration_id,
-            )
-        else:
-            external_integration, ignore = create(
-              _db, ExternalIntegration
-            )
-            self.external_integration_id = external_integration.id
+        external_integration, ignore = get_one_or_create(
+            _db, ExternalIntegration, id=self.external_integration_id,
+        )
+        self.external_integration_id = external_integration.id
         return external_integration
 
 
@@ -8723,15 +8718,10 @@ class Collection(Base):
     @property
     def external_integration(self):
         _db = Session.object_session(self)
-        if self.external_integration_id:
-            external_integration = get_one(
-                _db, ExternalIntegration, id=self.external_integration_id,
-            )
-        else:
-            external_integration, ignore = create(
-              _db, ExternalIntegration
-            )
-            self.external_integration_id = external_integration.id
+        external_integration, ignore = get_one_or_create(
+            _db, ExternalIntegration, id=self.external_integration_id,
+        )
+        self.external_integration_id = external_integration.id
         return external_integration
 
     @property
--- a/model.py
+++ b/model.py
@@ -8557,12 +8557,73 @@ class Admin(Base):
         _db.commit()
 
 
-class AuthenticationService(object):
-    """An AuthenticationService contains configuration for a third-party
-    service that can authenticate people. There will be a subclass of
-    AuthenticationService for each type of person that needs to be
-    authenticated.
+class ExternalIntegration(Base):
+
+    """An external integration contains configuration for connecting
+    to a third-party API.
     """
+
+    __tablename__ = 'externalintegrations'
+    id = Column(Integer, primary_key=True)
+
+    # If there is a special URL to use for access to this API,
+    # put it here. This is most common for OPDS and ODL integrations.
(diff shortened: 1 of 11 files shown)


model.py
```diff
@@ -8557,6 +8557,36 @@ def update_credentials(self, _db, access_token, credential):
...
+    provider = Column(Unicode, nullable=False, index=True)
+
+    # Supported values for the 'provider' field
+    GOOGLE_OAUTH = 'Google OAuth'
+
+    PROVIDERS = [GOOGLE_OAUTH]
+
+    # Additional provider-specific settings are stored as
+    # stringified JSON in the 'settings' field
+    settings = Column(Unicode)
```
aslagle: We could use a Postgres JSON column for this, but I think we would have to update sqlalchemy, and the string's probably fine since we don't have to load it that often.

model.py
```diff
@@ -8557,6 +8557,36 @@ def update_credentials(self, _db, access_token, credential):
         _db.commit()
 
 
+class AuthenticationService(object):
```
aslagle: I wasn't sure what to name this class.

model.py
leonardr: I want Collection configuration to be consistent with this. It's OK with me either to switch to this system and move  Collection settings to a JSONified string, or to keep authentication mechanism configuration in a special table.

aslagle: That's fine. I did this because google provides its client credentials as JSON but I can put that JSON in a settings table like Collection has.

leonardr:
Apropos putting different types of authentication in the same table, I think this makes sense for OAuth, but probably not beyond that. I'm not against the idea.

When I looked at doing this for Collection, I took inventory of all the existing collection types and what I would be removing from the JSON file. So here's something similar for our existing patron authentication techniques.

SIP2:
```
server host
server port
login_user_id
login_password
location_code
field_separator
```

Millenium:
```
url
```

FirstBook:
```
url
key
```

Clever:
```
client_id
client_secret
```

It's kind of all over the place, but there are enough similarities to the Collection configuration that I wonder if we can normalize the class even further. Create an object called `ExternalIntegration` which keeps track of the credentials necessary to connect to some external service and use its API. `Collection` and `AdminAuthenticationService` would both have an `ExternalIntegration` object that told you how to actually connect to the server and get the data. If it turned out that `AdminAuthenticationService` and `PatronAuthenticationService` both used the same OAuth server, then they could point to the same `ExternalIntegration`.

What do you think?

aslagle:
I like that idea, I'll try it out and see how it goes.

model.py
```diff
@@ -8557,6 +8557,103 @@ def update_credentials(self, _db, access_token, credential):
...
+
+    """An external integration contains configuration for connecting
+    to a third-party API.
+    """
+
+    __tablename__ = 'externalintegrations'
+    id = Column(Integer, primary_key=True)
+
+    # If there is a special URL to use for access to this API,
+    # put it here. This is most common for OPDS and ODL integrations.
```
leonardr: I think the OPDS/ODL bit isn't true anymore, since that has become the external_account_id for OPDS/ODL integrations.

model.py
```diff
@@ -8557,6 +8557,103 @@ def update_credentials(self, _db, access_token, credential):
...
+        if self.external_integration_id:
+            external_integration = get_one(
+                _db, ExternalIntegration, id=self.external_integration_id,
+            )
+        else:
+            external_integration, ignore = create(
+              _db, ExternalIntegration
+            )
+            self.external_integration_id = external_integration.id
+        return external_integration
```
leonardr: This seems like a job for `get_one_or_create`?

model.py
```diff
@@ -8654,10 +8721,23 @@ class Collection(Base):
...
```
leonardr: Ditto here -- I get that this is a little different from a normal scenario where the foreign key is managed with `relationship` but `get_one_or_create` seems like it would be safer. Is there a reason why it won't work here?

leonardr:
This all makes sense to me except for the external_integration_id bit.

leonardr:
:+1:

PR #487, opened 2017-04-05
Licensepooldeliverymechanism uses data source and identifier

leonardr:
This branch changes the LicensePoolDeliveryMechanism class to identify a book in terms of its data source and identifier, rather than a specific LicensePool. This reflects the fact that distributors make books available in the same formats to everyone. Only the number of copies is different from one LicensePool to another.

I still want to get from LicensePool to LicensePoolDeliveryMechanism, but there's no longer a proper foreign key so I couldn't figure out how to implement a normal SQLAlchemy `relationship`. I ended up writing properties that return queries, which I think is fine. We'll see if it causes performance problems later on.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -17,6 +17,8 @@ from classifier import (
 )
 
 from sqlalchemy import (
+    and_,
+    exists,
     or_,
     not_,
 )
@@ -35,6 +37,7 @@ from model import (
     Edition,
     Genre,
     LicensePool,
+    LicensePoolDeliveryMechanism,
     Work,
     WorkGenre,
 )
@@ -1215,10 +1218,14 @@ class Lane(object):
             )
 
         # Only find books the default client can fulfill.
-        query = query.filter(LicensePool.delivery_mechanisms.any(
-            DeliveryMechanism.default_client_can_fulfill==True)
-        )
-
+        LPDM = LicensePoolDeliveryMechanism
+        exists_clause = exists().where(
+            and_(LicensePool.data_source_id==LPDM.data_source_id,
+                and_(LicensePool.identifier_id==LPDM.identifier_id,
+                      DeliveryMechanism.default_client_can_fulfill==True
+        )))
+        query = query.filter(exists_clause)
+        
         # Only find books with unsuppressed LicensePools.
         if not show_suppressed:
             query = query.filter(LicensePool.suppressed==False)
@@ -1234,7 +1241,6 @@ class Lane(object):
             query = query.filter(
                 or_(LicensePool.licenses_available > 0, LicensePool.open_access)
             )
-
         return query
 
     @property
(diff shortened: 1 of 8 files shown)


lane.py
```diff
@@ -1214,11 +1217,14 @@ def only_show_ready_deliverable_works(
                 work_model.presentation_ready == True,
             )
 
-        # Only find books the default client can fulfill.
```
leonardr: This wasn't working properly and we didn't catch it because the code wasn't tested. Now that the SimplyE client hides books it can't fulfill, I decided to relax this restriction and only find books that have _some_ delivery mechanism.

tests/test_lane.py
```diff
@@ -1092,57 +1092,77 @@ def test_only_show_ready_deliverable_works(self):
...
+        w8.presentation_edition.title = "I have a weird delivery mechanism"
+        [pool] = w8.license_pools
+        for dm in pool.delivery_mechanisms:
+            self._db.delete(dm)
+        self._db.commit()
+        pool.set_delivery_mechanism(
+            "weird content type", "weird DRM scheme", "weird rights URI",
+            None
+        )
+
```
leonardr: This is the code that would have shown our "only delivery mechanisms the default client can render" code was broken.

lane.py
```diff
@@ -1214,11 +1217,14 @@ def only_show_ready_deliverable_works(
...
-        query = query.filter(LicensePool.delivery_mechanisms.any(
-            DeliveryMechanism.default_client_can_fulfill==True)
-        )
-
+        # Only find books that have some kind of DeliveryMechanism.
+        LPDM = LicensePoolDeliveryMechanism
+        exists_clause = exists().where(
+            and_(LicensePool.data_source_id==LPDM.data_source_id,
+                and_(LicensePool.identifier_id==LPDM.identifier_id)
```
aslagle: you don't need the second and_ here

migration/20170403-modify-licensepooldeliveries.sql
```diff
@@ -0,0 +1,30 @@
...
+
+-- Copy in appropriate values using license_pool_id.
+update licensepooldeliveries set
+ data_source_id=subquery.data_source_id,
+ identifier_id=subquery.identifier_id from (
+    select lpdm.id as delivery_id,
+    	   lp.identifier_id as identifier_id,
+	   lp.data_source_id as data_source_id
+    from licensepooldeliveries lpdm
+    	 join licensepools lp on lpdm.license_pool_id=lp.id limit 10
```
aslagle: what does the limit do?

leonardr: Oops, I added that while testing to avoid selecting 300,000 records on alpha.

PR #488, opened 2017-04-05
Circulationdata without collection

leonardr:
This branch changes `CirculationData.apply()` to take an optional Collection instead of a LicensePool. If you provide a Collection, it automatically finds or creates the appropriate LicensePool. If you don't provide a Collection, that's fine, but you can't `apply()` any information that goes into a LicensePool (such as number of available copies). You can only `apply()` formatting information (which as of my previous branch is associated with DataSource and Identifier, not LicensePool).

This is a much less intrusive change than moving link and format information from `CollectionData` to `Metadata`.

It turns out every time we currently call `CirculationData.apply()` we will have a Collection in mind, because we either call it from `CollectionMonitor` or `BibliographicCoverageProvider`. The real work was done in the previous branch, which made sure that the work done by `BibliographicCoverageProvider` in creating `LicensePoolDeliveryMechanism`s applies to all LicensePools for a given book. Now that that's in place we don't actually have to change `BibliographicCoverageProvider`.

I thought about changing `BibliographicCoverageProvider` so it wasn't a `CollectionCoverageProvider`, and we could do that for the Overdrive coverage provider, but not for the others. The problem is we need a Collection to get access to the third-party data in the first place. Collection is where we keep the API credentials. With the Overdrive API you can ask about a book that's not in your collection, but you can't do that with the Bibliotheca, Axis 360, or (I think) OneClick API.

So although this branch makes a pretty big change to `CirculationData`, we don't actually make use of it yet and I'm not sure if we ever will. Maybe in the open-access context it will make sense.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -892,12 +892,13 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
         CoverageFailure (if not).
         """
         error = None
-        if not circulationdata:
+        if circulationdata:
+            primary_identifier = circulationdata.primary_identifier(self._db)
+            if identifier != primary_identifier:
+                error = "Identifier did not match CirculationData's primary identifier."
+        else:
             error = "Did not receive circulationdata from input source"
 
-        primary_identifier = circulationdata.primary_identifier(self._db)
-        if identifier != primary_identifier:
-            error = "Identifier did not match CirculationData's primary identifier."
         if error:
             return self.failure(identifier, error, transient=True)
                                 
@@ -908,7 +909,7 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
         except Exception as e:
             self.log.warn(
                 "Error applying circulationdata to collection %s: %s",
-                collection.name, e, exc_info=e
+                self.collection.name, e, exc_info=e
             )
             return self.failure(identifier, repr(e), transient=True)
 
--- a/coverage.py
+++ b/coverage.py
@@ -691,9 +691,11 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
     left alone.
 
     For this reason it's important that subclasses of this
-    CoverageProvider only deal with bibliographic information, and
-    never circulation information. (Circulation information includes
-    formatting information and links to open-access downloads.)
+    CoverageProvider only deal with bibliographic information and
+    format availability information (such as links to open-access
+    downloads). You'll have problems if you try to use
+    CollectionCoverageProvider to keep track of information like the
+    number of licenses available for a book.
 
     In addition to defining the class variables defined by
     CoverageProvider, you must define the class variable PROTOCOL when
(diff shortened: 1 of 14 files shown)


coverage.py
```diff
@@ -889,20 +891,24 @@ def _set_circulationdata(self, identifier, circulationdata):
...
         """
-        pool = self.license_pool(identifier)
-        if isinstance(pool, CoverageFailure):
-            return pool
-
+        error = None
         if not circulationdata:
-            e = "Did not receive circulationdata from input source"
-            return self.failure(identifier, e, transient=True)
+            error = "Did not receive circulationdata from input source"
```
aslagle: Removing the return here means the next line will raise an exception.

coverage.py
```diff
@@ -889,20 +891,24 @@ def _set_circulationdata(self, identifier, circulationdata):
...
-            circulationdata.apply(pool, replace=self.replacement_policy)
+            circulationdata.apply(
+                self._db, self.collection, replace=self.replacement_policy
+            )
         except Exception as e:
             self.log.warn(
-                "Error applying circulationdata to pool %d: %s",
-                pool.id, e, exc_info=e
+                "Error applying circulationdata to collection %s: %s",
+                collection.name, e, exc_info=e
```
aslagle: should this be self.collection.name?

metadata_layer.py
```diff
@@ -898,67 +941,73 @@ def apply(self, pool, replace=None):
...
 
         return pool, made_changes
 
-
-
+    def _availability_needs_update(self, pool):
+        """Does this CirculationData represent information more recent than 
+        what we have for the given LicensePool?
+        """
+        if not self.last_checked:
```
aslagle: You changed self.last_checked so that it's always defined. Does that break this check?

leonardr: No, that check is just in case it got un-defined somehow.

metadata_layer.py
```diff
@@ -898,67 +941,73 @@ def apply(self, pool, replace=None):
...
-        for lpdm in pool.delivery_mechanisms:
+        # Changes to the delivery mechanisms may have changed the
+        # open-access status of all LicensePools for this work.
+        pools = identifier.licensed_through
+        old_open_access = any(pool.open_access for pool in pools)
+        for lpdm in identifier.delivery_mechanisms:
             if (lpdm.rights_status
                 and lpdm.rights_status.uri in RightsStatus.OPEN_ACCESS):
-                pool.open_access = True
+                new_open_access = True
```
aslagle: LPDM.set does the open_access update so this may not be necessary

leonardr: We need it in case a book _stops_ being open access because a LPDM was _removed_. I'll move that logic into model.py though.

metadata_layer.py
```diff
@@ -950,29 +957,19 @@ def apply(self, _db, collection, replace=None):
...
                     for loan in lpdm.fulfills:
                         self.log.info("Loan %i is associated with a format that is no longer available. Deleting its delivery mechanism." % loan.id)
                         loan.fulfillment = None
-                    _db.delete(lpdm)
+                    # This can cause an open-access LicensePool to go
+                    # non-open-access.
+                    lpdm.delete()
 
-        # Changes to the delivery mechanisms may have changed the
-        # open-access status of all LicensePools for this work.
```
leonardr: This is now handled by `LicensePoolDeliveryMechanism.set()` and `LicensePoolDeliveryMechanism.delete()`

PR #491, opened 2017-04-12
Support bibblio integration

courte:
This branch supports the Bibblio integration currently baked into the content_server (though it will have to move into circulation if we decide we want to catalogue nonfiction books outside of the public domain). Changes include:

- Creating a Bibblio DataSource and identifier type for works catalogued as Bibblio Content Items.
- Extracting an `EpubAccessor` utility class from Amy's work from the content_server. It cracks open an EPUB and get at the manifest and other goodies within and uses contextlib to make sure the EPUB is closed up again afterwards.
- Removing all references to the deprecated `flask.ext.babel`, referencing the `flask_babel` module instead.
- Adding an `IF EXISTS` clause to an otherwise occasionally broken migration.
- Using the preferred `[[` function in bash conditionals.

### DIFF
--- a/util/epub.py
+++ b/util/epub.py
@@ -1,0 +1,77 @@
+import contextlib
+import logging
+import os, sys
+from lxml import etree
+from nose.tools import set_trace
+from StringIO import StringIO
+from zipfile import ZipFile
+
+from http import HTTP
+
+
+class EpubAccessor(object):
+
+    CONTAINER_FILE = "META-INF/container.xml"
+    IDPF_NAMESPACE = "http://www.idpf.org/2007/opf"
+
+    @classmethod
+    @contextlib.contextmanager
+    def open_epub(cls, url, content=None):
+        """Cracks open an EPUB to expose its contents
+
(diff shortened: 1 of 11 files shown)


util/epub.py
```diff
@@ -0,0 +1,77 @@
...
+
+                if rootfiles_element is None:
+                    raise ValueError("Invalid EPUB file, not modifying: %s" % url)
+
+                rootfile_element = rootfiles_element.find("{urn:oasis:names:tc:opendocument:xmlns:container}rootfile")
+                if rootfile_element is None:
+                    raise ValueError("Invalid EPUB file, not modifying: %s" % url)
+
+                package_document_path = rootfile_element.get('full-path')
+            yield zip_file, package_document_path
```
courte: I don't know much about EPUB structure. There may come a time where we need more than these two items, but this met the immediate needs of the FeedBooks import and Bibblio cataloguing tasks.

leonardr:
:+1:

leonardr:
:+1:

PR #493, opened 2017-04-25
Fix CustomListUploadScript

courte:
I ran into some errors while running scripts to upload to Bibblio, and this was a core fix on a problem with `CustomList#entries_for_work` where `Query` wasn't iterable and raised an error, but only in certain cases. I'm not sure why, but extracting the query into a list before filtering it removed the error in practice.


leonardr:
:+1:

PR #495, opened 2017-05-01
Fixes from 1.1.21 patch for master

courte:
There were some issues when deploying 1.1.21 to production. These are the fixes, laid on top of master. In the meantime, production is pointing to server_core at branch hot-fix-1.1.20, otherwise known as commit 3322f5cf2407a4b9d5c4f65abfa92b29eebdcd48.


leonardr:
:+1:

PR #496, opened 2017-05-01
Circulation fixes

courte:
Circulation and metadata wrangler fixes for multi-tenancy, including:
- Giving Bibliographic APIs a collection variable
- The CatalogCoverageProvider, used primarily in the metadata wrangler
- Accounting for CoverageProviders that cover multiple protocols
- `WorkBasedLane` instead of `LicensePoolBasedLane`, which really just makes a lot more sense
- Relatedly: CachedFeeds identified by `Work` instead of `LicensePool`
- Giving a `Work` a direct connection to all of its complaints
- Giving a `Complaint` knowledge about whether it's likely to be LicensePool-specific or not
- Test fixin'


migration/20170426-2-cached-feeds-from-works-not-pools.py
```diff
@@ -0,0 +1,40 @@
...
+    CachedFeed.license_pool_id.isnot(None),
+    CachedFeed.work_id.is_(None)).all()
+
+if remaining_feeds:
+    logging.error(
+        "ERROR: %d affected CachedFeeds remaining without associated Work",
+        len(remaining_feeds)
+    )
+else:
+    logging.info("All affected CachedFeeds given associated Work")
```
leonardr: This is pretty intense -- I would have deleted these rows. The code looks fine though.

Are we going to delete `cachedfeeds.license_pool_id` later?

leonardr:
:+1:

migration/20170426-2-cached-feeds-from-works-not-pools.py
courte: I like to live life on the edge (🤘), but deleting them is an easier idea. I forgot what cachedfeeds were in my effort to fix them. 💁

If we delete them, the whole schema change can be made at once.

leonardr: Yeah, delete them and do the whole change at once.

model.py
```diff
@@ -5785,10 +5794,16 @@ class CachedFeed(Base):
     # The content of the feed.
     content = Column(Unicode, nullable=True)
 
-    # A feed may be associated with a LicensePool.
+    # TODO: After LicensePool-associated feeds have been successfully
+    # migrated to Work-associated feeds, drop this column and remove
+    # its relationship in LicensePool.
```
aslagle: Shouldn't we do this TODO now? I see a migration that drops the column.

model.py
```diff
@@ -9066,19 +9094,6 @@ def works_updated_since(self, _db, timestamp):
...
-     Column(
-         'collection_id', Integer, ForeignKey('collections.id'),
-         index=True, nullable=False
-     ),
-     Column(
-         'licensepool_id', Integer, ForeignKey('licensepools.id'),
-         index=True, nullable=False
-     ),
-     UniqueConstraint('collection_id', 'licensepool_id'),
-)
```
aslagle: we don't need a migration to drop this table?

leonardr: This table isn't present in the master branch, so there are no real instances that use it. I created it thinking we could make collection-licensepool many-to-many, but that didn't work out.

model.py
courte: Ah, I can get rid of this note. We're just getting rid of the feeds entirely now. They can be regenerated as WorkBasedLanes without much hassle.

courte: This only impacts Recommendations (which aren't configured in QA or production, I think) and their parent Related Books lanes.

PR #499, opened 2017-05-09
Patron has library

leonardr:
This branch adds `library_id` to the `patrons` table, associating each patron with a library. Fields that were formerly unique across the `patrons` table are now unique only in combination with `library_id`. So two patrons can have the same username so long as they belong to different libraries.

This is for https://github.com/NYPL-Simplified/circulation/issues/487

### DIFF
--- a/migration/20170509-add-library-to-patron.sql
+++ b/migration/20170509-add-library-to-patron.sql
@@ -16,3 +16,5 @@ ALTER TABLE patrons ADD CONSTRAINT patrons_library_id_username_key UNIQUE (libra
 CREATE INDEX ix_patron_library_id_authorization_identifier ON patrons USING btree (library_id, authorization_identifier);
 CREATE INDEX ix_patron_library_id_external_identifier ON patrons USING btree (library_id, external_identifier);
 CREATE INDEX ix_patron_library_id_username ON patrons USING btree (library_id, username);
+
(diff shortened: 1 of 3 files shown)


migration/20170509-add-library-to-patron.sql
```diff
@@ -0,0 +1,22 @@
...
+DROP INDEX ix_patrons_external_identifier;
+DROP INDEX ix_patrons_username;
+
+ALTER TABLE patrons ADD COLUMN library_id integer;
+ALTER TABLE patrons ADD CONSTRAINT patrons_library_id_fkey FOREIGN KEY (library_id) REFERENCES libraries(id);
+
+
+ALTER TABLE patrons ADD CONSTRAINT patrons_library_id_authorization_identifier_key UNIQUE (library_id, authorization_identifier);
+ALTER TABLE patrons ADD CONSTRAINT patrons_library_id_external_identifier_key UNIQUE (library_id, external_identifier);
+ALTER TABLE patrons ADD CONSTRAINT patrons_library_id_username_key UNIQUE (library_id, username);
```
aslagle: Is it not necessary to drop the old unique constraints on the single columns?

leonardr: Those constraints were implemented through unique indexes, so dropping the indexes gets rid of the constraints.

PR #500, opened 2017-05-12
Fix Best Sellers & Staff Picks lanes

courte:
This branch fixes the problems that @leonardr spotted in the Staff Picks and Best Sellers lanes where every single book was being included in both. It also seeks to find matching books by both LicensePool and Edition, since in the QA database it appears that finding a CustomListEntry by its connected via LicensePool is highly unlikely.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -579,11 +579,16 @@ class Lane(object):
             exclude_genres=exclude_genres, fiction=fiction
         )
 
-        # CustomList information must be set after sublanes.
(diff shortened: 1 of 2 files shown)


leonardr:
:+1: on the understanding that once we fix the production server you'll write a test to detect the problem if it happens again.

courte:
I added the test ✔️

leonardr:
Looks great.

PR #501, opened 2017-05-15
Avoid nltk error

courte:
This fixes the bug that @datalogics-cgreen found last week by catching the MissingCorpusError and ignoring in-depth summary evaluation in its wake.

I'd like to add a test, but I'm not sure how to test it, since the error is raised from the library's `TextBlob` class.


leonardr:
This looks good. To test it, you would change add() to take a parser class as an argument. The default would be `parser=TextBlob`, but in the test you would pass in a class that raised a `MissingCorpusError` when you tried to access its `noun_phrases` instance variable.

PR #504, opened 2017-05-22
Extend external integrations

### LINKED ISSUE #498: Extend external integrations to incorporate third party vendors from Configuration
As discussed in [this series of comments](https://github.com/NYPL-Simplified/content_server/pull/122/files/02252173baae86b30ccd11e238dd1f9023aeec8e#diff-3b895af42a91e96b13119492cfce77b9R53), extend the `ExternalIntegration` model to incorporate Configuration from non-`Collection` integrations. This list of integrations should include:

- Metadata wrangler
- Google Analytics
- New York Times API
- Amazon S3
- NYPL Shadowcat
- Content Cafe
- The CDN
- Loggly
- Elasticsearch (???)
- NoveList Select
- Google OAuth

courte:
This branch adds a name and provider to `ExternalIntegration` so it can be used for more than just distributor-related configuration; then it slorps in everything worth using from the Configuration json.

It also allows `CoverageRecords` to be associated with a `Collection`, a requirement for the Metadata Wrangler Collection syncing.

It creates a MetadataWranglerOPDSLookup, mainly because the Metadata Wrangler endpoints are doing their own kind of special something and now prefer authentication.

It helps the OPDSImporter reverse-engineer an identifier mapping from the results, which is required to grab metadata updates from the Wrangler.

There are also some changes to `CollectionMonitor` and the scripts associated with `CollectionCoverageProvider` to avoid (and perhaps create!) some signature confusion and depend work with CoverageProviders that can work with Collections of multiple protocols.

This half-fixes #498, but doesn't actually use it in all the places.


config.py
```diff
@@ -38,7 +38,6 @@ class Configuration(object):
     instance = None
 
     # Logging stuff
-    LOGGING = "logging"
```
courte: This was a duplicate.

coverage.py
```diff
@@ -754,12 +757,14 @@ def all(cls, _db, **kwargs):
         CollectionCoverageProvider (or, more likely, one of its subclasses).
 
         """
-        if not cls.PROTOCOL:
-            raise ValueError("%s must define PROTOCOL." % cls.__name__)
-        collections = _db.query(Collection).filter(
-            Collection.protocol==cls.PROTOCOL).order_by(func.random())
+        collections = _db.query(Collection)
+
+        if cls.PROTOCOL:
```
courte: I sense we may eventually need to let PROTOCOL be a list of protocols or perhaps have an `ANY_PROTOCOL = object()` constant, but it's not required for the MW right now.

scripts.py
```diff
@@ -1522,7 +1513,10 @@ def _run_migration(self, migration_path, timestamp):
 
         if migration_path.endswith('.sql'):
             with open(migration_path) as clause:
-                sql = clause.read()
+                # By wrapping the action in a transation, we can avoid
+                # rolling over errors and losing data in files
+                # with multiple interrelated SQL actions.
+                sql = 'BEGIN;\n%s\nCOMMIT;' % clause.read()
```
courte: This helps avoid some of the SQL problems I've had in recent migrations. This way, if it errors it errors. No more partial migrations.

model.py
```diff
@@ -8795,9 +8812,51 @@ class ExternalIntegration(Base):
...
+        GOOGLE_OAUTH, METADATA_WRANGLER, NOVELIST, NYPL_SHADOWCAT,
+        PATRON_WEB_CLIENT, NYT, STAFF_PICKS,
+    ]
+
     __tablename__ = 'externalintegrations'
     id = Column(Integer, primary_key=True)
 
+    # If this integration isn't related to a Collection, it should
+    # have a provider and type.
+    provider = Column(Unicode, nullable=True)
```
leonardr: I think this should replace `Collection.protocol`, but that can wait until later.

migration/20170522-2-move-configuration-to-external-integrations.py
```diff
@@ -0,0 +1,144 @@
...
+    )
+
+    if content_server_conf:
+        url = content_server_conf.get(Configuration.URL)
+
+        log_import(ExternalIntegration.CONTENT_SERVER)
+        service, ignore = get_one_or_create(
+            _db, ExternalIntegration, provider=ExternalIntegration.CONTENT_SERVER
+        )
+        service.url = unicode(url)
```
leonardr: Is there a "content server integration" that's not a Collection with protocol=OPDS_IMPORT? Maybe there is for metadata.

model.py
```diff
@@ -8795,9 +8812,51 @@ class ExternalIntegration(Base):
...
+    GOOGLE_OAUTH = u'Google OAuth'
+    ADMIN_AUTH_PROVIDERS = [GOOGLE_OAUTH]
+
+    PROVIDERS = [
+        ADOBE_VENDOR_ID, AMAZON_S3, BIBBLIO, CDN, CIRCULATION_MANAGER,
+        CONTENT_CAFE, CONTENT_SERVER, ELASTICSEARCH, GOOGLE_ANALYTICS,
+        GOOGLE_OAUTH, METADATA_WRANGLER, NOVELIST, NYPL_SHADOWCAT,
+        PATRON_WEB_CLIENT, NYT, STAFF_PICKS,
+    ]
+
```
leonardr: I don't see this used anywhere (in this branch or metadata) and I don't think it's good to have a comprehensive list of providers. (We used to try to have a comprehensive list of DataSources and we've now moved away from that for the same reason.) I'd rather have providers like NYPL_SHADOWCAT known only within the code components that use them.

leonardr:
Adding `CoverageRecord.collection_id` is not the right move, and we should be able to avoid it. `OPDSImporter` can handle `collection=None`, and it shouldn't be too difficult to make a version of `OPDSImportCoverageProvider` that subclasses `CoverageProvider` instead of `CollectionCoverageProvider`. Something like this:

```
class MetadataOPDSImportCoverageProvider(CoverageProvider)

class OPDSImportCoverageProvider(CollectionCoverageProvider, MetadataOPDSImportCoverageProvider)
```

scripts.py
```diff
@@ -195,27 +195,18 @@ def __init__(self, providers):
             self.providers.append(i)
 
     def do_run(self):
-        offsets = dict()
```
leonardr: I wasn't sure about this but looking at the implementation of `provider.run_once_and_update_timestamp` I think it makes sense.

model.py
courte: Monitors and CoverageProviders would definitely need to take a list of accepted PROTOCOLS instead of a single covered PROTOCOL or `None` in that case.

migration/20170522-2-move-configuration-to-external-integrations.py
courte: Oh, that's true. There is for metadata, in that metadata importing doesn't create a licensepool. Maybe I'm misunderstanding how Collections work, though?

courte:
The `collection_id` isn't for the OPDSImportCoverageProviders, it's for the Sync and Reap providers. Syncing and reaping is collection-specific because identifiers can be included in multiple Collections, and they need to be added or removed in the Metadata Wrangler independently for each Collection.

Another way around this is to have a Monitor go through each identifier and add for each Collection, but it would create a lot of duplicate work because there's currently no way to know when an Identifier is added or removed from a Collection or its individual licenses are added or removed.

leonardr:
In conversation with Courteney I was convinced that `CoverageRecord.collection_id` is in fact the right move. The metadata sync and metadata reap coverage providers are operating on an Identifier _as it is in a collection_. An Identifier that was reaped from one collection may stil lbe present in another.

Furthermore, it's possible that other integrations may require the same type of collection-aware coverage, so a solution like putting a flag in the `collections_identifiers` table is not appropriate -- we'd have to change to Courteney's proposed system the next time that happened.

migration/20170522-2-move-configuration-to-external-integrations.py
leonardr: No, I think you have it right.

model.py
leonardr: I don't understand this part. Do you have a case in mind where multiple providers are using the same protocol?

courte: No, I don't think so. Right now CollectionMonitor and CollectionCoverageProvider each have a single class variable that can be set to a single protocol or `None` (which means any). But for something like the MetadataWrangler or NoveList, they can cover a collection of any protocol.

I was wrong, though. I'd forgot that protocol and provider are set in different classes, and it wouldn't be a problem as long as we're looking for an attached Collection (we are) and not just the ExternalIntegration.

leonardr:
:+1:

PR #507, opened 2017-05-25
Merge Collection.protocol with ExternalIntegration.provider

leonardr:
This branch combines two fields created for the multi-tenancy branch, Collection.protocol and ExternalIntegration.provider. The new field is called ExternalIntegration.protocol. "protocol" is a better name than "provider" because it covers things like "OPDS Import" and "CDN" which are closer to protocols than to companies or services. There's also a name conflict between "provider" and our "CoverageProvider" concept, which would have led to a CoverageProvider being associated with a provider.

Since these fields were created for a branch that hasn't been released, I haven't written migration scripts for it -- anyone who might benefit from these scripts should instead drop their database and let it get rebuilt, and anyone who doesn't need these scripts won't be able to run them.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -701,7 +701,7 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
     number of licenses available for a book.
 
     In addition to defining the class variables defined by
-    CoverageProvider, you must define the class variable PROVIDER when
+    CoverageProvider, you must define the class variable PROTOCOL when
     subclassing this class. This is the entity that provides the
     licenses for this Collection. It should be one of the
     collection-type provider constants defined in the
@@ -717,7 +717,7 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
     # Set this to the name of the provider managed by this type of
     # CoverageProvider. If this CoverageProvider can manage collections
     # for any provider, leave this as None.
-    PROVIDER = None
+    PROTOCOL = None
     
     def __init__(self, collection, **kwargs):
         """Constructor.
@@ -732,10 +732,10 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
                 )
             )
 
-        if self.PROVIDER and collection.provider != self.PROVIDER:
+        if self.PROTOCOL and collection.protocol != self.PROTOCOL:
             raise ValueError(
-                "Collection provider (%s) does not match CoverageProvider provider (%s)" % (
-                    collection.provider, self.PROVIDER
+                "Collection protocol (%s) does not match CoverageProvider protocol (%s)" % (
+                    collection.protocol, self.PROTOCOL
                 )
             )
         _db = Session.object_session(collection)
@@ -754,7 +754,7 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
     @classmethod
     def all(cls, _db, **kwargs):
         """Yield a sequence of CollectionCoverageProvider instances, one for
-        every Collection that gets its licenses from cls.PROVIDER.
+        every Collection that gets its licenses from cls.PROTOCOL.
 
         CollectionCoverageProviders will be yielded in a random order.
 
@@ -762,16 +762,11 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
         CollectionCoverageProvider (or, more likely, one of its subclasses).
 
         """
-        collections = _db.query(Collection)
+        if cls.PROTOCOL:
+            collections = Collection.by_protocol(_db, cls.PROTOCOL)
+        else:
+            collections = _db.query(Collection)
 
-        if cls.PROVIDER:
-            collections = collections.join(
-                ExternalIntegration,
-                ExternalIntegration.id==Collection.external_integration_id).filter(
-                    ExternalIntegration.type==ExternalIntegration.LICENSE_TYPE
-                ).filter(
-                    ExternalIntegration.provider==cls.PROVIDER
-                )
         collections = collections.order_by(func.random())
         for collection in collections:
             yield cls(collection, **kwargs)
--- a/coverage.py
+++ b/coverage.py
@@ -714,9 +714,9 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
     
     DEFAULT_BATCH_SIZE = 10
 
-    # Set this to the name of the provider managed by this type of
+    # Set this to the name of the protocol managed by this type of
     # CoverageProvider. If this CoverageProvider can manage collections
-    # for any provider, leave this as None.
+    # for any protocol, leave this as None.
     PROTOCOL = None
     
     def __init__(self, collection, **kwargs):
--- a/coverage.py
+++ b/coverage.py
@@ -14,6 +14,7 @@ from model import (
     CoverageRecord,
     DataSource,
     Edition,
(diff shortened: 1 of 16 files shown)


model.py
```diff
@@ -8812,42 +8812,95 @@ class ExternalIntegration(Base):
     to a third-party API.
     """
 
-    # Utilities
+    # Possible types of ExternalIntegrations.
+    #
+    # These integrations are associated with external services such as
+    # Google Enterprise which authenticate library administrators.
+    ADMIN_AUTH_TYPE = 'admin_auth'
```
leonardr: Since we don't use most of these constants yet, this is mainly planning out what future use of ExternalIntegration will look like. I did this in this branch to see whether it was wise to remove the uniqueness constraint below.

model.py
```diff
@@ -8867,17 +8920,10 @@ class ExternalIntegration(Base):
...
         lazy="joined", cascade="save-update, merge, delete, delete-orphan",
     )
-
-    __table_args__ = (
-        UniqueConstraint('provider', 'type'),
-        Index(
-            'ix_unique_provider_with_null_type', provider,
-            unique=True, postgresql_where=(type.is_(None))),
-    )
-
```
leonardr: There are lots of situations -- not just with collections -- where different libraries in a single circ manager might need their own individual ExternalIntegrations for the same external integration. For example, two libraries on a circ manager might  have separate Novelist accounts.

model.py
```diff
@@ -9021,13 +9047,71 @@ class Collection(Base):
 
     def __repr__(self):
         return (u'<Collection "%s"/"%s" ID=%d>' %
-                (self.name, self.protocol, self.id)).encode('utf8')
+                (self.name, self.protocol, self.id)).encode('utf8')        
+
+    @classmethod
+    def by_name_and_protocol(cls, _db, name, protocol):
```
leonardr: Looking up and creating Collections has become more complicated because what used to be a field of Collection is now in an associated object.

overdrive.py
```diff
@@ -987,21 +992,22 @@ def to_collection(self, _db):
...
-        parent = get_one(
-            _db, Collection, external_account_id=self.parent_library_id,
-            protocol=Collection.OVERDRIVE
-        )
-        if not parent:
+        try:
+            parent = Collection.by_protocol(_db, ExternalIntegration.OVERDRIVE).filter(
+                Collection.external_account_id==self.parent_library_id
+            ).one()
+        except NoResultFound, e:
```
leonardr: With this and the implementation of `Collection.protocol` I decided that a collection with a parent inherits `.protocol` from its parent, along with other features of its `ExternalIntegration`. It doesn't need to be stored separately.

coverage.py
```diff
@@ -701,17 +702,21 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
...
     # all Identifiers in the given Collection, regardless of their type.
     INPUT_IDENTIFIER_TYPES = None
     
     DEFAULT_BATCH_SIZE = 10
 
-    # Set this to the name of the protocol managed by this type of
+    # Set this to the name of the provider managed by this type of
     # CoverageProvider. If this CoverageProvider can manage collections
-    # for any protocols, leave this as None.
+    # for any provider, leave this as None.
```
aslagle: This change seems backwards. Shouldn't these still say protocol?

model.py
```diff
@@ -8812,42 +8812,95 @@ class ExternalIntegration(Base):
...
+
+    # These integrations are associated with external services such as
+    # Google Analytics, which receive analytics events.
+    ANALYTICS_TYPE = 'analytics'
+       
+    # These integrations are associated with the Library Simplified
+    # applications themselves, as well as external services such as
+    # CDNs and Elasticsearch which are directly used by the
+    # applications.
+    SYSTEM_TYPE = 'system'
```
aslagle: What are the advantages of grouping these types of integrations together?

aslagle: As opposed to having single protocols for CDN or S3, and different integration types for the different uses of those.

model.py
```diff
@@ -8812,42 +8812,95 @@ class ExternalIntegration(Base):
...
+    
+    # ANALYTICS_TYPE integrations
     GOOGLE_ANALYTICS = u'Google Analytics'
 
-    # Patron Authentication
-    ADOBE_VENDOR_ID = u'Adobe Vendor ID'
-
-    # Admin Authentication
+    # ADMIN_AUTHENTICATION_TYPE integrations
     ADMIN_AUTH_TYPE = u'admin_auth'
```
aslagle: This constant is a duplicate from above.

model.py
```diff
@@ -8812,42 +8812,95 @@ class ExternalIntegration(Base):
...
     id = Column(Integer, primary_key=True)
 
-    # If this integration isn't related to a Collection, it should
-    # have a provider and type.
-    provider = Column(Unicode, nullable=True)
+    # Each integration should have a protocol (explaining how we actually
+    # get information through it) and a type (explaining what the integration
+    # is for).
+    protocol = Column(Unicode, nullable=True)
     type = Column(Unicode, nullable=True)
```
aslagle: Does it make sense for these to be nullable? When would it be ok for them to be null?

model.py
```diff
@@ -9021,13 +9047,71 @@ class Collection(Base):
...
+    
+    @property
+    def protocol(self):
+        """What protocol do we need to use to get licenses for this 
+        collection?
+        """
+        if self.external_integration.protocol:
+            return self.external_integration.protocol
+        if self.parent:
+            return self.parent.protocol
```
aslagle: Inheriting protocol from the parent is inconsistent with by_protocol and by_name_and_protocol. A Collection with a protocol from its parent won't be able to be looked up by that protocol.

tests/test_model.py
```diff
@@ -5567,9 +5567,9 @@ def setup(self):
         self.external_integration, ignore = create(self._db, ExternalIntegration)
 
     def test_data_source(self):
-        # For most collections, the protocol determines the
+        # For most collections, the provider determines the
```
aslagle: Another backwards comment change.

model.py
leonardr: It took a while for me to understand what you're saying, but I think it makes sense. You're saying look up the image CDN with (type="image" protocol="CDN") rather than (type="system" protocol="Image CDN").

This is not how I imagined using `type`--I imagined it as being more general than `protocol` and in this case it specializes `protocol`--but I think it could work.

model.py
leonardr: Upon inspection, it doesn't make sense. `provider` was nullable but that was only to support collections, which had a non-nullable `protocol` instead. Now that `provider` and `protocol` are the same, the field should not be nullable.

aslagle: I do like `purpose`.

I was thinking external integrations would normally be looked up by type - if you stored a one-off integration without a type/purpose you would need to keep its id somewhere else so you could find it again.

model.py
```diff
@@ -8988,7 +9005,7 @@ class Collection(Base):
     # to Overdrive Advantage collections.
     children = relationship(
         "Collection", backref=backref("parent", remote_side = [id]),
-        uselist=False
+        uselist=True
```
leonardr: Unrelated bug which showed up once I actually looked at `Collection.children` in a test.

PR #510, opened 2017-05-30
Support incorporate related lanes into entry

courte:
This branch creates an author tags for each author, even in simple OPDS entries. This means that creating the  simple OPDS entry will take longer and require the database. It also reduces the need for `Edition.author`, which can potentially be removed at this point. We use it in a lot of places, though, I'm reluctant to make that change right now. (I'll make an issue for it if this PR is approved.)

The branch also allows Works and Editions created during testing to have a series set at the time of creation. This reduces the need to recalculate the presentation later on.

One change here that I'm not sure about is potentially adding authors by sort name if their display name is not available. I'm not sure if that possibility happens in practice, but it appears in tests quite often (due to how we create contributors on the Edition we use in tests).

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -201,14 +201,15 @@ class Annotator(object):
     def authors(cls, work, license_pool, edition, identifier):
         """Create one or more <author> tags for the given work."""
         authors = list()
-        listed = list()
+        listed = set()
         for author in edition.author_contributors:
             name = author.display_name or author.sort_name
-            if name.lower() in listed:
+            name_key = name.lower()
+            if name_key in listed:
                 continue
 
             authors.append(AtomFeed.author(AtomFeed.name(name)))
-            listed.append(name.lower())
+            listed.add(name_key)
 
         if authors:
             return authors
--- a/opds.py
+++ b/opds.py
@@ -201,10 +201,14 @@ class Annotator(object):
     def authors(cls, work, license_pool, edition, identifier):
         """Create one or more <author> tags for the given work."""
         authors = list()
+        listed = list()
         for author in edition.author_contributors:
-            authors.append(AtomFeed.author(AtomFeed.name(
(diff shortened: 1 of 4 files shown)


opds.py
```diff
@@ -200,7 +200,15 @@ def categories(cls, work):
...
-        return [AtomFeed.author(AtomFeed.name(edition.author or ""))]
+        authors = list()
+        for author in edition.author_contributors:
+            authors.append(AtomFeed.author(AtomFeed.name(
+                author.display_name or author.sort_name
+            )))
+
+        if authors:
+            return authors
+        return [AtomFeed.author(AtomFeed.name(""))]
```
leonardr: As long as we're doing this, it would be good to use `atom:author` tags for contributors whose role is in `Contributor.AUTHOR_ROLES`, and `atom:contributor` tags for everyone else.

I don't have a problem using `sort_name` as a fallback when we don't have `display_name`. In almost all real cases we will have `display_name`, even if it's just a guess.

opds.py
```diff
@@ -200,7 +200,15 @@ def categories(cls, work):
     @classmethod
     def authors(cls, work, license_pool, edition, identifier):
         """Create one or more <author> tags for the given work."""
-        return [AtomFeed.author(AtomFeed.name(edition.author or ""))]
+        authors = list()
+        for author in edition.author_contributors:
+            authors.append(AtomFeed.author(AtomFeed.name(
+                author.display_name or author.sort_name
+            )))
```
leonardr: We sometimes have a problem where the same contributor (or possibly a different contributor with the same name) is listed as a contributor twice. It won't fix the problem, but putting a dupe check in this code is a really easy way to stop the problem from affecting patrons. Keep track of the names we've used, and skip an author if we've used their name before.

PR #512, opened 2017-05-31
Library short name in urls

aslagle:
This supports https://github.com/NYPL-Simplified/circulation/pull/501

### DIFF
--- a/model.py
+++ b/model.py
@@ -8707,7 +8707,7 @@ class Library(Base):
 
     # A short name of this library, to use when identifying it in
     # scripts. e.g. "NYPL" for NYPL.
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -8707,7 +8707,7 @@ class Library(Base):
 
     # A short name of this library, to use when identifying it in
     # scripts. e.g. "NYPL" for NYPL.
-    short_name = Column(Unicode, unique=True)
+    short_name = Column(Unicode, unique=True, nullable=False)
```
leonardr: Since this will be used in URLs, should we impose a length limit or a restriction on which characters are valid?

leonardr:
:+1:

PR #513, opened 2017-06-01
Fix scripts

leonardr:
This branch creates the `CollectionInputScript`, which allows you to name specific collections on the command line, and parses them into Collection objects.

The `OPDSImportScript` is now a `CollectionInputScript`, and it has some tests. `RunCollectionMonitorScript` does not subclass this class, but it should in the future.

This branch handles a now-common problem where a `CollectionMonitor` is passed in to `RunMonitorScript` when it should be passed in to `RunCollectionMonitorScript`. `RunMonitorScript` now complains and runs the monitor through `RunCollectionMonitorScript` instead.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -58,7 +58,10 @@ from model import (
 from external_search import ExternalSearchIndex
 from monitor import SubjectAssignmentMonitor
 from nyt import NYTBestSellerAPI
-from opds_import import OPDSImportMonitor
+from opds_import import (
+    OPDSImportMonitor,
+    OPDSImporter,
+)
 from oneclick import OneClickAPI, MockOneClickAPI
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -146,14 +150,32 @@ def load_configuration(self):
...
+        elif self.collection_monitor:
+            logging.warn(
+                "Running a CollectionMonitor by delegating to RunCollectionMonitorScript. "
+                "It would be better if you used RunCollectionMonitorScript directly."
+            )
+            RunCollectionMonitorScript(
+                self.collection_monitor, self._db, **self.collection_monitor_kwargs
+            ).run()
+        else:
+            self.monitor.run()
```
aslagle: This is already called in the if self.monitor at the top.

PR #514, opened 2017-06-02
Flexible configuration

leonardr:
This branch replaces ExternalIntegrationSetting with the more general ConfigurationSetting, which can be used to configure the site as a whole, an ExternalIntegration, a Library, or the combination of a Library and an ExternalIntegration.

This will radically simplify the move of authentication mechanisms from JSON into the database, and give a place for other things (especially library-specific links like terms-of-use) to go into the database as well.

### DIFF
--- a/model.py
+++ b/model.py
@@ -8981,7 +8981,7 @@ class ConfigurationSetting(Base):
       is a patron of, is associated with both a Library and an
       ExternalIntegration.
     """
-    __tablename__ = 'externalintegrationsettings'
+    __tablename__ = 'configurationsettings'
     id = Column(Integer, primary_key=True)
     external_integration_id = Column(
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -8952,23 +8959,77 @@ def get(self, key):
...
+    * The "website ID" for an Overdrive collection is associated with
+      an ExternalIntegration (the Overdrive integration), but not with
+      any particular Library (since multiple libraries might share an
+      Overdrive collection).
+
+    * The "identifier prefix" used to determine which library a patron
+      is a patron of, is associated with both a Library and an
+      ExternalIntegration.
     """
     __tablename__ = 'externalintegrationsettings'
```
aslagle: you're not changing the table name?

leonardr: :blush:

PR #516, opened 2017-06-05
Removed link constants and methods from config since they've been mov…

aslagle:
…ed to the db in the circ manager.


leonardr:
Awesome :+1:

PR #519, opened 2017-06-06
Support branch for moving authentication out of JSON config

leonardr:
This branch makes miscellaneous core improvements to the way we treat configuration settings in support of the circulation branch that stores authentication information in the database.

### DIFF
--- a/model.py
+++ b/model.py
@@ -8798,8 +8798,7 @@ class Library(Base):
                 'Short name (for library registry): "%s"' %
                 self.library_registry_short_name
             )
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -8948,16 +8963,37 @@ def setting(self, key):
             _db, key, self
         )
 
-    def get(self, key):
+    def get(self, key, default=None):
```
leonardr: I'm not convinced this method is necessary, but it wasn't working before and this should fix it.

PR #521, opened 2017-06-08
Support library config move

leonardr:
This branch offers core support for moving several circulation-specific configuration items out of the JSON configuration and into ConfigurationSettings.

* The maximum cache times for different types of OPDS feed have been made into site-wide settings.
* The default loan period and default reservation period have been made into Collection settings. I removed OneClick's separate values for default loan period for ebook and audio content. We need a consistent way of storing this value for different media types, and it doesn't matter all that much because when we create a loan, we always find out pretty quickly how long the loan is actually for.
* The regular expression used to determine a patron's external type from their authentication identifier (e.g. the first character of a FirstBook barcode) has been moved into circulation as a specific library's customization of an authentication-type integration.

I added a convenience class method `ConfigurationSetting.sitewide_secret` which finds a sitewide setting, populating it with a random 192-bit value if it doesn't already exist.

I added a convenience method `ConfigurationSetting.bool_value` which interprets a setting as a boolean.
I tried to move HOLDS_POLICY (basically, whether unavailable books should be shown or hidden) out of the JSON configuration, but since it's tied up with the way we generate OPDS feeds, I think it's easier to wait until we move Lane configuration into the database.

Once I discovered this problem I abandoned my plans to convert similar settings like "featured lane size".

### DIFF
--- a/model.py
+++ b/model.py
@@ -9322,7 +9322,7 @@ class Collection(Base):
             or self.STANDARD_DEFAULT_LOAN_PERIOD
         )
 
-    @protocol.setter
+    @default_loan_period.setter
     def set_default_loan_period(self, new_value):
         new_value = int(new_value)
         self.external_integration.setting(
--- a/model.py
+++ b/model.py
@@ -9281,14 +9281,19 @@ class Collection(Base):
         self.external_integration.protocol = new_protocol
         for child in self.children:
             child.protocol = new_protocol
-            
+
+    # TODO: The default loan period needs to be expanded to handle
+    # different defaults for different media types (e.g. on Overdrive
+    # the default loan period for audiobooks is 14 days, and for video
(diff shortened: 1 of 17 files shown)


model.py
```diff
@@ -9262,6 +9278,53 @@ def set_protocol(self, new_protocol):
         for child in self.children:
             child.protocol = new_protocol
 
+    # TODO: The default loan period needs to be expanded to handle
+    # different defaults for different media types (e.g. on Overdrive
+    # the default loan period for audiobooks is 14 days, and for video
+    # it's 5 days). This isn't a high priority because the license
+    # source generally tells us when each loan will end.
```
aslagle: It's important for hold time estimates, not that we do a good job of them even with this.

leonardr: True on both counts.

PR #524, opened 2017-06-12
Move lane configuration out of Configuration and into Library ConfigurationSettings

leonardr:
This branch changes Lane to take a Library in its constructor. This Library is the source for various items of configuration that were previously stored in the JSON configuration:

* For each facet group, which facets are enabled.
* Of the enabled facets in a facet group, which is the default.
* Whether the library allows patrons to put items on hold.
* The size of the 'featured' lanes used for the home page.
* The minimum quality a book must have to show up in a 'featured' lane.

I moved the `FacetConfig` class in from circulation. This class lets a controller override the Library's facet configuration for controller-specific purposes. (e.g. by default, the 'books in this series' feed orders books by series position, which is not a normally available ordering).

I changed the `cachedfeeds` table so that each feed keeps track of the library for which it is a feed.

I changed `LicensePool.with_complaint` to get LicensePools available through a specific Library, not all LicensePools on the system that have a complaint.

### DIFF
--- a/config.py
+++ b/config.py
@@ -93,9 +93,6 @@ class Configuration(object):
     OVERDRIVE_INTEGRATION = "Overdrive"
     THREEM_INTEGRATION = "3M"
 
-    MINIMUM_FEATURED_QUALITY = "minimum_featured_quality"
-    FEATURED_LANE_SIZE = "featured_lane_size"
-
     S3_INTEGRATION = u"S3"
     S3_ACCESS_KEY = u"access_key"
     S3_SECRET_KEY = u"secret_key"
@@ -189,18 +186,6 @@ class Configuration(object):
     def logging_policy(cls):
         default_logging = {}
         return cls.get(cls.LOGGING, default_logging)
-
-    # TODO: Needs to be a per-library ConfigurationSetting once
-    # lanes are configured in database.
-    @classmethod
-    def minimum_featured_quality(cls):
-        return float(cls.policy(cls.MINIMUM_FEATURED_QUALITY, 0.65))
-
-    # TODO: Needs to be a per-library ConfigurationSetting once
-    # lanes are configured in database.
-    @classmethod
-    def featured_lane_size(cls):
-        return int(cls.policy(cls.FEATURED_LANE_SIZE, 15))
 
     @classmethod
     def localization_languages(cls):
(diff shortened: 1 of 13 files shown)


config.py
```diff
@@ -137,8 +102,7 @@ class Configuration(object):
     CDN_INTEGRATION = u"CDN"
 
     BASE_OPDS_AUTHENTICATION_DOCUMENT = "base_opds_authentication_document"
-    SHOW_STAFF_PICKS_ON_TOP_LEVEL = "show_staff_picks_on_top_level"
```
leonardr: This constant is not used anywhere.

migration/20170612-add-library-id-to-cachedfeed.sql
```diff
@@ -0,0 +1,13 @@
+-- Remove the existing index from cachedfeeds
+drop index if exists ix_cachedfeeds_lane_name_type_facets_pagination;
+
+-- Add library_id as a foreign key
+alter table cachedfeeds add column library_id integer;
+alter table cachedfeeds add constraint coveragerecords_collection_id_fkey
```
aslagle: the name of this constraint should be cachedfeeds_library_id_fkey, not coveragerecords, right?

aslagle:
Do you need a new migration for the config items you moved?

leonardr:
Yes, but only for Open Ebooks. I'll add it.

leonardr:
Here's the config I used to test my changes to the migration scripts.

```
{
   "integrations" : {
      "Postgres" : {
         "test_url" : "postgres://simplified_test:test@localhost:5432/simplified_circulation_test",
      }
   },
   "policies" : {
       "holds": "hide",
       "minimum_featured_quality" : 200,
	"featured_lane_size" : 100,
	"facets" : {
            "enabled" : {
                "order" : [ "title", "author" ],
		"collection" : ["full"],
		"available" : ["now"]
            },
            "default" : {
                "order" : "title",
		"collection" : "full",
		"available" : "now"
            }
 	}      
   }
}
```

leonardr:
I should mention that the migration script went into https://github.com/NYPL-Simplified/circulation/pull/512 since only the circulation manager uses any of the configuration items that were moved.

PR #526, opened 2017-06-13
Use external integrations

### LINKED ISSUE #518: Move per-library configuration settings into ExternalIntegration
This includes policies like `Configuration.HOLD_POLICY` and `default_notification_email_address`, etc etc.

courte:
This branch pulls CDN, S3, and Elasticsearch integrations into the database, as well as supporting the import of other third-party integrations in the Circulation Manager.

I'm not in love with the way that S3 and CDN integrations are imported, especially since more often than not they're related to each other. I contemplated having a 'cdn' setting on the S3 External Integration (and then possibly incorporating cdn-ify somewhere in S3Uploader). But I'm not sure the overlap is 100% so I'd rather just get something stable and separate in for now.

Fixes #518.


util/cdn.py
```diff
@@ -1,25 +1,42 @@
...
-    if not cdns:
-        # No CDNs configured
-        return url
+def cdnify(_db, url):
     scheme, netloc, path, query, fragment = urlparse.urlsplit(url)
 
+    # TODO: Find a better way to import from model in core.util when
+    # using the parent directory (e.g. circulation) or move cdnify into
+    # an OPDS- or controller-focused class or file.
+    if not core_dir in sys.path:
```
courte: I'm not sure why these import problems from circ are only just popping up, but when referencing this file from circulation modules, `core` is not listed in the path. If there's a better way to do this, I'm happy to hear about it.

util/s3.py
```diff
@@ -1,189 +0,0 @@
-from nose.tools import set_trace
```
courte: This file was an outdated duplicate for core/s3.py. Or, now that I think about it, maybe this file is why the import in cdn was working. 😱

courte:
One thing that's missing here are the imports of Collection-connected / bibliographic license source configurations. I'm going to do that in a separate branch to get the Adobe Vendor ID changes from circulation incorporated sooner.

model.py
```diff
@@ -9058,11 +9082,31 @@ class ExternalIntegration(Base):
...
-    def lookup(cls, _db, protocol, goal=None):
-        integration = get_one(_db, cls, protocol=protocol, goal=goal)
-        return integration
+    def lookup(cls, _db, protocol, goal, library=None):
+        integrations = _db.query(cls).filter(
+            cls.protocol==protocol, cls.goal==goal
+        ).all()
+
+        if library:
+            integrations = filter(lambda i: library in i.libraries, integrations)
```
leonardr: Can you join the query against cls.libraries and filter then, rather than filtering after the query has run?

nyt.py
```diff
@@ -10,9 +10,10 @@
 from sqlalchemy.orm.exc import (
```
leonardr: Yeeees.... I think I wanted to use the NYT integration in metadata, but it's like Novelist in that the licensing prohibits sharing of the best-seller list data. And the title-author type metadata that comes from this source is of very low quality, so we wouldn't want to use it internally.

s3.py
```diff
@@ -7,22 +7,41 @@
...
-            secret_key = secret_key or integration[Configuration.S3_SECRET_KEY]
+        elif access_key or secret_key:
+            if not (access_key and secret_key):
+                raise ValueError(
+                    "Cannot create S3Uploader without both"
+                    " access_key and secret_key"
+                )
             self.pool = tinys3.Pool(access_key, secret_key)
+        else:
+            if not _db and goal:
```
leonardr: I think this needs parentheses.

tests/test_opds.py
```diff
@@ -739,29 +740,31 @@ def test_acquisition_feed_includes_image_links(self):
...
-            links = sorted([x['href'] for x in feed['entries'][0]['links'] if 
-                            'image' in x['rel']])
-            eq_(['http://bar/a', 'http://foo/b'], links)
+        # Create some CDNS.
+        self._external_integration(
+            ExternalIntegration.CDN, goal=u'thumbnail.com', url=u'http://foo/'
+        )
+        self._external_integration(
+            ExternalIntegration.CDN, goal=u'full.com', url=u'http://bar/'
+        )
```
leonardr: I don't understand what the .com bits are for.

leonardr:
This looks good overall but the CDN stuff needs work. I would like to take any database access out of `cdnify`, which is called a huge number of times. Since the import problem has to do with importing stuff from model, removing database access from cdnify should also solve that problem.

Maybe we could create a method of `Configuration` that loads site-wide configuration items like CDN, S3 and Elasticsearch. All the configuration items themselves are strings and don't need to be associated with the database; it's just that we're loading them from the database now instead of from JSON.

So long as you had previously initialized the `Configuration` object against the database, you could look up things like CDN URLs in the `Configuration` object without bothering with a database connection.

Since you've changed all the `cdnify` calls to take a database connection, we could try having `cdnify` call `Configuration.cdn_config(_db netloc)`, and have that method initialize the site-wide configuration if it's not already initialized. In this case `cdnify` could still cause database access, but it would only ever happen once.

courte:
I don't think the import stuff has to do with model at all. It's because we're trying to import from outside of core.util in general. The s3 import was only working because we had a duplicate of s3 inside of core.util. When a parent app (like `circulation`) pulls in anything from core.util it only has circulation-level modules and `core.util`-level modules available to it -- not `core` modules.

I like your solution of pulled the CDN configuration into the app, but if I'm right we're going to see the same problem if I try to import `config` at this level. I will give it a try to confirm.

config.py
```diff
@@ -230,6 +216,30 @@ def data_directory(cls):
...
+        if not cdns:
+            return
+
+        cdn_integration = dict()
+        for cdn in cdns:
+            netloc = CDN_GOAL_KEY.get(cdn.goal)
+            if not netloc:
+                continue
+
+            cdn_integration[netloc] = cdn.url
```
leonardr: I don't get this part. `netloc` is a urllib2 term for a domain name, but here the "netloc" would be "book_covers", "open_access_books", etc.

opds_import.py
```diff
@@ -1487,10 +1499,10 @@ def run_once(self, start_ignore, cutoff_ignore):
 class OPDSImporterWithS3Mirror(OPDSImporter):
     """OPDS Importer that mirrors content to S3."""
 
-    def __init__(self, _db, default_data_source, **kwargs):
+    def __init__(self, _db, goal, default_data_source, **kwargs):
```
leonardr: What does `goal` do here? You don't use it whtin this method. Maybe `S3Uploader` used to take it as an argument?

leonardr:
I put my finger on the CDN problem when I saw the JSON file you've been using to test imports.

https://github.com/NYPL-Simplified/circulation/pull/514#issuecomment-309038258

```
    "CDN" : {
      "opds" : "https://opds.cloudfront.net/",
      "book_covers" : "https://book-covers.cloudfront.net/",
      "open_access_books" : "https://open-access-books.cloudfront.net/"
    },
```

CDN configuration used to look like this, but it doesn't anymore. It turned out to be a mistake, because a CDN doesn't mirror "all the book covers", it mirrors a specific domain name, like `book-covers.nypl.org`. Our production CDN configuration looks like this:

```
        "CDN" : {
            "" : "<%= @opds_feeds_cdn_host %>",
            "book-covers.nypl.org" : "<%= @book_covers_cdn_host %>"
        },
```

Where "" indicates the CDN for the circulation manager itself.

I think the CDN integrations should take this form:

* goal="CDN"
* protocol="CDN" (value here doesn't really matter)
* url="http://whatever.cloudfront.net/"
* mirror_domain="book-covers.nypl.org"

Inside `load_cdns` you would find all the Integrations with goal="CDN" and build a mapping of `mirror_domain` (this is the "netloc") to `url`.

Any CDN code (as opposed to S3 code) that deals with constants like "book_covers" or "opds" is out of date and should be removed.

config.py
```diff
@@ -83,7 +83,8 @@ class Configuration(object):
     OVERDRIVE_INTEGRATION = "Overdrive"
     THREEM_INTEGRATION = "3M"
 
-    BASE_OPDS_AUTHENTICATION_DOCUMENT = "base_opds_authentication_document"
+    # ConfigurationSEtting key for a CDN's mirror domain
+    CDN_MIRROR_DOMAIN_KEY = u'mirror_domain'
```
leonardr: Looking at this in context, I think "mirrored_domain" is better, because "mirror_domain" is ambiguous whether it's the domain that you're mirroring or the domain _of_ the mirror.

leonardr:
:+1:  with the variable change.

PR #528, opened 2017-06-14
Changed url, username, and password on ExternalIntegration to be properties using ConfigurationSettings.

aslagle:
This makes it easier to convert between settings and form values in the admin interface - everything can be accessed as a setting. I changed the explain methods to do this too. In other places, they can be used the same way as before.

### DIFF
--- a/model.py
+++ b/model.py
@@ -9030,6 +9030,19 @@ class ExternalIntegration(Base):
     # List of such ADMIN_AUTH_GOAL integrations
     ADMIN_AUTH_PROTOCOLS = [GOOGLE_OAUTH]
 
+    # Keys for common configuration settings
+
+    # If there is a special URL to use for access to this API,
+    # put it here.
+    URL = "url"
+
+    # If access requires authentication, these settings represent the
+    # username/password or key/secret combination necessary to
+    # authenticate. If there's a secret but no key, it's stored in
+    # 'password'.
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -9575,14 +9596,9 @@ def explain(self, include_password=False):
...
-        if integration.url:
-            lines.append('URL: "%s"' % integration.url)
-        if integration.username:
-            lines.append('Username: "%s"' % integration.username)
-        if integration.password and include_password:
-            lines.append('Password: "%s"' % integration.password)
         for setting in integration.settings:
-            lines.append('Setting "%s": "%s"' % (setting.key, setting.value))
+            if (setting.key != ExternalIntegration.PASSWORD) or include_password:
+                lines.append('Setting "%s": "%s"' % (setting.key, setting.value))
```
leonardr: Among other ideas I've been having for improving discoverability of these configuration settings, I had the idea of marking certain items as "secret". So ExternalIntegration.PASSWORD would be an object that looks like a normal string, except hasattr(ExternalIntegration.PASSWORD, 'secret') would be True.

This would eliminate the need to put special cases in this code for knowing whether a field value is secret or not.

Not a high priority IMO.

leonardr:
:+1:

model.py
aslagle: That would be nice.

PR #531, opened 2017-06-15
Define site-wide and library-wide settings in Configuration.

aslagle:


scripts.py
```diff
@@ -677,7 +690,14 @@ def do_run(self, _db=None, cmd_args=None, output=sys.stdout):
...
                 key, value = setting.split('=', 1)
-                ConfigurationSetting.sitewide(_db, key).value = value
+                if not args.force and not key in [s.get("key") for s in self.config.SITEWIDE_SETTINGS]:
+                    raise ValueError(
+                        "'%s' is not a known site-wide setting. Use --force to set it anyway."
+                        % key
+                    )
+                else:
+                    ConfigurationSetting.sitewide(_db, key).value = value
+                    _db.commit()
```
leonardr: Why commit immediately?

aslagle: The script didn't seem to be committing at all before...

leonardr:
:+1:

scripts.py
leonardr: I looked at the code and it looks like `do_run()` is responsible for calling commit. Can you just move this to the end? (and add commit() to any of the other configuration scripts that don't have it at the end of do_run()?)

aslagle: Done. The other scripts had it already.

PR #532, opened 2017-06-16
Moved constants for editing Collections and their ExternalIntegrations

aslagle:


opds_import.py
```diff
@@ -275,7 +276,15 @@ class OPDSImporter(object):
...
-   
+
+    NAME = ExternalIntegration.OPDS_IMPORT
+    FIELDS = [
+        {
+            "key": Collection.EXTERNAL_ACCOUNT_ID_KEY,
+            "label": _("URL"),
+        }
+    ]
+
```
leonardr: This isn't a problem with this branch, but to understand this change I went looking in opds_import.py and I think it's a little confusing that `SimplifiedOPDSLookup` looks for the URL of the OPDS feed in `integration.url` and `OPDSImporter` looks in `integration.external_account_id`. Do you think this is worth changing?

aslagle: You're right, that is confusing. I can't remember why we wanted it in external_account_id. Maybe we wanted it to be on the Collection instead of the integration.

aslagle: And changing that would let me get rid of the constant in Collection and put it in BaseCirculationAPI instead.

leonardr: The URL of an OPDS feed is analogous to the "library ID" of an Overdrive account, and not to the URL of the Overdrive web service root, so I understand why we did that, and on balance I'd like to keep it on the Collection rather than in the integration.

leonardr:
:+1:

opds_import.py
leonardr: I don't have a strong opinion either way though.

aslagle: I don't either. I'll leave it alone for now.

PR #533, opened 2017-06-16
Integration config script

leonardr:
This branch does some refactoring of scripts that set `ConfigurationSetting`s. It changes `ConfigureLibraryScript` so that you can set `ConfigurationSetting`s on a Library.

It also introduces two new scripts for managing `ExternalIntegration`s. `ShowIntegrationsScript` lists all the integrations and `ConfigureIntegrationScript`lets you create an `ExternalIntegration`, or set `ConfigurationSetting`s on an existing `ExternalIntegration`.

You can't use the command line to associate an `ExternalIntegration` with a Library, or to set a `ConfigurationSetting` on the combination of an integration and a library -- for that we decided the admin interface is good enough.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1058,22 +1058,15 @@ class ConfigureIntegrationScript(ConfigurationSettingScript):
             'Set a configuration value on the integration. Format: --setting="key=value"'
         )        
         return parser
-    
-    def do_run(self, _db=None, cmd_args=None, output=sys.stdout):
-        _db = _db or self._db
-        args = self.parse_command_line(_db, cmd_args=cmd_args)
 
-        # Find or create the integration
-        protocol = None
-        id = args.id
-        name = args.name
-        protocol = args.protocol
-        goal = args.goal
-
+    @classmethod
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -953,6 +1027,79 @@ def do_run(self, _db=None, cmd_args=None, output=sys.stdout):
...
+                _db, ExternalIntegration, ExternalIntegration.id==id
+            )
+        if not integration and name:
+            integration = get_one(
+                _db, ExternalIntegration, ExternalIntegration.name==name
+            )
+        if not integration and (protocol and goal):
+            integration, is_new = get_one_or_create(
+                _db, ExternalIntegration, protocol=protocol, goal=goal
+            )
```
aslagle: I think you're missing something to set the integration's name...

aslagle:
Looks good, but I think you should be able to set an integration's name when you create it, or the name isn't very useful.

leonardr:
Thanks for the review. I need to merge Courteney's recent changes into the big branch I'm working on but I'll come back to this on Monday.

PR #536, opened 2017-06-19
Consolidate s3 integrations

### LINKED ISSUE #534: Consolidate S3 ExternalIntegrations
Instead of having multiple S3 ExternalIntegrations for the same account, use one ExternalIntegration to access multiple S3 buckets.

courte:
This changes the setup for the `S3Uploader` to resemble CDNs more closely. S3 has a single ExternalIntegration with a goal of 'storage', and all of its buckets and their respective URLs are loaded from the database by `Configuration.load`. When `S3Uploader` is accessed, these buckets are available on the class itself.

Fixes #534.

### DIFF
--- a/config.py
+++ b/config.py
@@ -8,6 +8,8 @@ import copy
 from util import LanguageCodes
 from flask.ext.babel import lazy_gettext as _
 
+from s3 import S3Uploader
+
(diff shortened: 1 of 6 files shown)


leonardr:
:+1:

PR #538, opened 2017-06-20
Facet config constants

aslagle:
This adds default and enabled facets to the library-wide configuration settings.

### DIFF
--- a/config.py
+++ b/config.py
@@ -9,6 +9,7 @@ from util import LanguageCodes
 from flask.ext.babel import lazy_gettext as _
 
 from s3 import S3Uploader
+from facets import FacetConstants
 
 class CannotLoadConfiguration(Exception):
     pass
@@ -110,6 +111,12 @@ class Configuration(object):
     # 'featured' lanes.
     FEATURED_LANE_SIZE = "featured_lane_size"
 
+    # Each facet group has two associated per-library keys: one
+    # configuring which facets are enabled for that facet group, and
+    # one configuring which facet is the default.
+    ENABLED_FACETS_KEY_PREFIX = "facets_enabled_"
+    DEFAULT_FACET_KEY_PREFIX = "facets_default_"
+
     # The name of the per-library per-patron authentication integration
     # regular expression used to derive a patron's external_type from
     # their authorization_identifier.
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -145,6 +152,24 @@ class Configuration(object):
...
         },
+    ] + [
+        { "key": ENABLED_FACETS_KEY_PREFIX + group,
+          "label": _("Enabled %(group)s facets", group=display_name),
+          "type": "list",
+          "options": [
+              { "key": facet, "label": FacetConstants.FACET_DISPLAY_TITLES.get(facet) }
+              for facet in FacetConstants.FACETS_BY_GROUP.get(group)
+          ],
+        } for group, display_name in FacetConstants.GROUP_DISPLAY_TITLES.iteritems()
```
aslagle: In the admin interface, this shows up with a heading for each group and a checkbox for each facet in the group.

config.py
```diff
@@ -145,6 +152,24 @@ class Configuration(object):
...
+    ] + [
+        { "key": DEFAULT_FACET_KEY_PREFIX + group,
+          "label": _("Default %(group)s facet", group=display_name),
+          "type": "select",
+          "options": [
+              { "key": facet, "label": FacetConstants.FACET_DISPLAY_TITLES.get(facet) }
+              for facet in FacetConstants.FACETS_BY_GROUP.get(group)
+          ],
+        } for group, display_name in FacetConstants.GROUP_DISPLAY_TITLES.iteritems()
```
aslagle: This shows up as a dropdown for each group. There's currently no validation that the default facet is one of the enabled facets. I'm not sure if that's important.

leonardr: It's not important right now; almost nobody will configure these settings.

facets.py
```diff
@@ -21,10 +31,25 @@ class FacetConstants(object):
...
     ORDER_RANDOM = 'random'
+    ORDER_FACETS = [
+        ORDER_TITLE,
+        ORDER_AUTHOR,
+        ORDER_LAST_UPDATE,
+        ORDER_ADDED_TO_COLLECTION,
+        ORDER_SERIES_POSITION,
+        ORDER_WORK_ID,
+        ORDER_RANDOM,
+    ]
```
leonardr: Some of these facets only make sense in certain contexts and shouldn't be options in the user interface, which controls how lane feeds are presented to patrons. "random" might be fun, but I don't think there's any reason to include "last update" or "work id", and series position won't work at all.

aslagle: Yeah, I wasn't sure about those. I'll remove them from this list I guess.

leonardr:
:+1:

PR #539, opened 2017-06-21
Minor changes to support config

leonardr:
Starting from a config file that looks like this, I started a circ manager and tried to get to the point where I had two functioning libraries.

```
{
    "policies" : {
        "languages" : {
            "large_collections" : ["eng"],
            "small_collections" : ["spa", "chi", "rus"]
        }
    },
    "integrations" : {
        "Postgres" : {
            "production_url" : "postgres://simplified:[password]@localhost:5432/simplified_circulation_dev",
            "test_url" : "postgres://simplified_test:test@localhost:5432/simplified_circulation_test"
        }
    }
}
```

I found some bugs and needed improvements, but I also was able to make a few minor changes myself that either helped me get things working or make the experience nicer for people who do this in the future.  This branch contains those minor changes.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -59,9 +59,9 @@ class ExternalSearchIndex(object):
                 url = url or integration.url
                 if not works_index:
                     setting = integration.setting(self.WORKS_INDEX_KEY).value
-                    if not setting.value:
-                        setting.value = self.DEFAULT_WORKS_INDEX
-                    works_index = setting.value
+                    works_index = setting.value_or_default(
+                        self.DEFAULT_WORKS_INDEX
+                    )
             if not url:
                 raise CannotLoadConfiguration(
                     "No URL configured to Elasticsearch server."
@@ -89,7 +89,7 @@ class ExternalSearchIndex(object):
         def bulk(docs, **kwargs):
             return elasticsearch_bulk(self.__client, docs, **kwargs)
         self.bulk = bulk
-
+        
     def update_integration_settings(self, integration, force=False):
         """Updates the integration with an appropriate index and alias
         setting if the index and alias have been updated.
--- a/external_search.py
+++ b/external_search.py
@@ -1,7 +1,10 @@
 from nose.tools import set_trace
 from elasticsearch import Elasticsearch
 from elasticsearch.helpers import bulk as elasticsearch_bulk
-from config import Configuration
+from config import (
+    Configuration,
+    CannotLoadConfiguration,
+)
 from classifier import (
     KeywordBasedClassifier,
     GradeLevelClassifier,
@@ -17,6 +20,8 @@ class ExternalSearchIndex(object):
     WORKS_INDEX_KEY = u'works_index'
     WORKS_ALIAS_KEY = u'works_alias'
 
+    DEFAULT_WORKS_ALIAS = u'works'
+    
     work_document_type = 'work-type'
     __client = None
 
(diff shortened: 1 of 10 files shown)


s3.py
```diff
@@ -27,58 +27,77 @@ class S3Uploader(MirrorUploader):
...
     def from_config(cls, _db):
+        """Create an S3Uploader from site configuration.
+
+        :return: An S3Uploader if S3 is configured; otherwise None.
+        """
         integration = cls.integration(_db)
-        cls.initialize_buckets(_db)
+        if not integration:
+            return None
+        cls.initialize_buckets(_db)
```
leonardr: If there's no S3 integration (as is the case in the circ manager) then this should return None (indicating that there is no way to upload things to S3) rather than raising an error.

scripts.py
```diff
@@ -143,7 +143,7 @@ def run(self):
             raise e
 
     def load_configuration(self):
-        if not Configuration.instance:
+        if not Configuration.loaded_from_database():
```
leonardr: When I ran a script for real (as opposed to from a test), the Configuration was loaded once without access to the database. Then this was running, but since the configuration had "already been loaded", it was never loaded with access to the database.

scripts.py
```diff
@@ -1151,9 +1151,17 @@ def _integration(self, _db, id, name, protocol, goal):
...
+                _db, ExternalIntegration, ExternalIntegration.id==id
+            )
+            if not integration:
+                raise ValueError("No integration with ID %s." % id)
+        if name:
             integration = get_one(_db, ExternalIntegration, name=name)
+            if not integration and not (protocol and goal):
+                raise ValueError(
+                    'No integration with name "%s". To create it, you must also provide protocol and goal.' % name
+                )
```
leonardr: This is to give better error messages when I ran `configure_integration` with bad arguments.

tests/test_overdrive.py
```diff
@@ -105,6 +107,20 @@ def test_failure_to_get_library_is_fatal(self):
             self.api.get_library
         )
 
+    def test_error_getting_library(self):
+        class MisconfiguredOverdriveAPI(MockOverdriveAPI):
+            """This Overdrive client has valid credentials but the library
+            can't be found -- probably because the library ID is wrong."""
```
leonardr: I frequently type in the library ID for the website ID and vice versa.

external_search.py
```diff
@@ -47,12 +52,20 @@ def __init__(self, _db, url=None, works_index=None):
                     goal=ExternalIntegration.SEARCH_GOAL
                 )
 
+                if not integration:
+                    raise CannotLoadConfiguration(
+                        "No Elasticsearch integration configured."
+                    )
```
leonardr: Without the integration configured I was getting an unhelpful message shortly below this point.

external_search.py
```diff
@@ -47,12 +52,20 @@ def __init__(self, _db, url=None, works_index=None):
...
                 url = url or integration.url
                 if not works_index:
-                    works_index = integration.setting(self.WORKS_INDEX_KEY).value
-
+                    setting = integration.setting(self.WORKS_INDEX_KEY).value
+                    works_index = setting.value_or_default(
+                        self.DEFAULT_WORKS_INDEX
+                    )
```
leonardr: Most of the time there's no need to make the administrator explicitly define a DEFAULT_WORKS_INDEX.

external_search.py
```diff
@@ -47,12 +52,20 @@ def __init__(self, _db, url=None, works_index=None):
...
 
+                if not integration:
+                    raise CannotLoadConfiguration(
+                        "No Elasticsearch integration configured."
+                    )
                 url = url or integration.url
                 if not works_index:
-                    works_index = integration.setting(self.WORKS_INDEX_KEY).value
-
+                    setting = integration.setting(self.WORKS_INDEX_KEY).value
```
aslagle: This line shouldn't have .value

PR #541, opened 2017-06-22
Monitor for configuration changes

leonardr:
This branch gives you a method you can call to see if the site configuration has changed while you weren't looking: `Configuration.check_for_site_configuration_update()`.

It also gives you a method you can call if you yourself have changed the site configuration: `site_configuration_has_changed` in model.py.

I've added listeners for every database event I can think of that implies a change to what we generally consider "the site configuration". I've also changed the scripts that modify the configuration so that they explicitly call `site_configuration_has_changed` just before committing.

The goal is to use this within the circulation manager to detect when site configuration has changed and automatically reload the `CirculationManager` configuration.

### DIFF
--- a/config.py
+++ b/config.py
@@ -289,34 +289,34 @@ class Configuration(object):
         return [LanguageCodes.three_to_two[l] for l in languages]
 
     # The last time the database configuration is known to have changed.
-    DATABASE_CONFIGURATION_LAST_UPDATE = "database_configuration_last_update"
+    SITE_CONFIGURATION_LAST_UPDATE = "site_configuration_last_update"
 
     # The last time we *checked* whether the database configuration had
     # changed.
-    LAST_CHECKED_FOR_DATABASE_CONFIGURATION_UPDATE = "last_checked_for_database_configuration_update"
+    LAST_CHECKED_FOR_SITE_CONFIGURATION_UPDATE = "last_checked_for_site_configuration_update"
     
     @classmethod
-    def database_configuration_last_update(cls):
+    def site_configuration_last_update(cls):
         """As far as we know, when is the last time the database configuration
         was updated?
         """
-        return cls.instance.get(cls.DATABASE_CONFIGURATION_LAST_UPDATE, None)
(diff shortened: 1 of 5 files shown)


tests/test_model.py
```diff
@@ -5955,9 +6100,9 @@ def test_explain(self):
              'Protocol: "Overdrive"',
              'Used by library: "The only library"',
              'External account ID: "id"',
+             'Setting "setting": "value"',
              'Setting "url": "url"',
              'Setting "username": "username"',
-             'Setting "setting": "value"'
```
leonardr: This test started failing on my machine because items started coming out differently from a hash. I changed explain() to sort the keys.

config.py
```diff
@@ -283,8 +283,77 @@ def logging_policy(cls):
     def localization_languages(cls):
         languages = cls.policy(cls.LOCALIZATION_LANGUAGES, default=["eng"])
         return [LanguageCodes.three_to_two[l] for l in languages]
+
+    # The last time the database configuration is known to have changed.
+    SITE_CONFIGURATION_LAST_UPDATE = "site_configuration_last_update"
```
aslagle: Why does this information need to be stored in the Configuration instance as well as the database?

leonardr: My goal is to not have to go to the database to check this more than once a minute. I was thinking I would put that code in circulation but maybe I got confused and it needs to go into core. I'll investigate.

leonardr: You're right, that caching code needs to go in here. I've added it.

model.py
```diff
@@ -9928,4 +9939,84 @@ def numericrange_to_tuple(r):
...
+        # means it's time to reset the Timestamp that says when the
+        # configuration last changed.
+        needs_commit = False
+        # Convert something that might not be a Connection object into
+        # a Connection object.
+        if isinstance(_db, Base):
+            _db = Session.object_session(_db)
+        elif isinstance(_db, Connection):
+            _db = Session(_db)
+            needs_commit = True
```
leonardr: I'm not sure about this because it's one of the random things I tried when getting something to work, but unlike many changes of that sort I think it holds up.

PR #543, opened 2017-06-23
No new database sessions

leonardr:
This branch fixes a couple problems I found when using the code that checks for changes to ConfigurationSettings.

First, the code that passed in a database connection (instead of a session) to site_configuration_has_changed was a bad idea. In an environment with a scoped session (like our  applications) you can't go creating new, non-scoped sessions on the same connection. Fortunately, it was easy to change `configuration_relevant_lifecycle_event` to pass in the target object instead. I didn't do this initially because I didn't really notice that that function took `target` as an argument.

Once that's fixed, there's another problem. The very first time `configuration_relevant_lifecycle_event` is ever called, it calls `create` to create the `Timestamp` that stores the last update time. `create` calls `flush`, but `configuration_relevant_lifecycle_event` is called when we're already in the middle of a flush, which raises the exception.

Ignoring the "you're already in the middle of a flush" exception seems to work, but I also noticed that the SQLAlchemy docs (http://docs.sqlalchemy.org/en/latest/orm/session_events.html#session-persistence-mapper) put restrictions on the things you can do from within a mapper-level event such as `before_insert`. In particular, you're not supposed to call `Session.add()`, which is what `create` does immediately before calling the `flush()` that causes us the problem.

I briefly tried using validators instead (http://docs.sqlalchemy.org/en/latest/orm/mapped_attributes.html#simple-validators), but validators seem to fall down even earlier: when you try to make a database query inside them.

I'm not sure what the solution is -- maybe making sure the last update time is present in the database immediately upon startup, so that we can guarantee  `configuration_relevant_lifecycle_event` never needs to call create().

### DIFF
--- a/model.py
+++ b/model.py
@@ -79,7 +79,6 @@ from sqlalchemy.sql.expression import (
 )
 from sqlalchemy.exc import (
     IntegrityError,
-    InvalidRequestError,
 )
 from sqlalchemy import (
     create_engine,
@@ -332,7 +331,7 @@ class SessionManager(object):
                 session, content_type, drm_scheme
             )
             mechanism.default_client_can_fulfill = True
-
+        site_configuration_has_changed(session)
         session.commit()
 
 def get_one(db, model, on_multiple='error', constraint=None, **kwargs):
@@ -397,11 +396,7 @@ def create(db, model, create_method='',
     kwargs.update(create_method_kwargs or {})
     created = getattr(model, create_method, model)(**kwargs)
     db.add(created)
-    try:
-        db.flush()
-    except InvalidRequestError, e:
-        if e.message != 'Session is already flushing':
-            raise e
+    db.flush()
     return created, True
 
 Base = declarative_base()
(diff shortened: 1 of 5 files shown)


leonardr:
Initializing the Timestamp as part of the initial dataset seems to solve the problem.

leonardr:
Yeah, we're still in trouble. This works, and the tests pass, but when I use the admin interface the timestamp doesn't actually get updated. I think the reason is I'm modifying `Timestamp.value` (on an object that never gets flushed) instead of sending raw SQL.

leonardr:
OK, I've got everything working using raw SQL to update the timestamp. I used a live site to test that making a change to the site configuration updates the timestamp, but that random things like loading a page don't update the timestamp.

The raw SQL only works if the Timestamp already exists, so I create it immediately, along with the DataSources. This required that I change a few tests that used "there are no Timestamps" as a shorthand for "the Timestamp I'm interested in doesn't exist".

model.py
```diff
@@ -9967,26 +9967,23 @@ def site_configuration_has_changed(_db, timeout=1):
         # a Connection object.
         if isinstance(_db, Base):
             _db = Session.object_session(_db)
-        elif isinstance(_db, Connection):
-            _db = Session(_db)
-            needs_commit = True
```
aslagle: you should update the docstring for the db param.

aslagle: you can also remove the needs_commit = False above

PR #544, opened 2017-06-23
Support content in multi-tenant

courte:
A couple changes to support bringing the content server over to multi-tenancy, including:

- Loading Configuration from the database for both the current Configuration class and any of its parent Configuration classes (otherwise errors occur in core/opds.py)
- Adding Gutenberg as a LICENSE_GOAL protocol
- Removing the 'static_feed_bucket' from S3 (Since it's only used in a couple configurable scripts on the content server, this'll be fine.)
- Slightly altering the `OPDSImportScript` for a new subclass

### DIFF
--- a/model.py
+++ b/model.py
@@ -9599,6 +9599,7 @@ class Collection(Base):
 
     @classmethod
     def by_datasource(cls, _db, data_source):
+        """Query collections that are associated with the given DataSource."""
         if isinstance(data_source, DataSource):
             data_source = data_source.name
 
--- a/model.py
+++ b/model.py
@@ -9544,6 +9544,18 @@ class Collection(Base):
             ).filter(ExternalIntegration.protocol==protocol)
         return qu
 
+    @classmethod
+    def by_datasource(cls, _db, data_source):
+        if isinstance(data_source, DataSource):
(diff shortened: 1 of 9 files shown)


model.py
```diff
@@ -9032,8 +9032,12 @@ class ExternalIntegration(Base):
     AXIS_360 = DataSource.AXIS_360
     ONE_CLICK = DataSource.ONECLICK
 
+    # These protocols are only used on the Content Server.
```
leonardr: Spell out that this is importing texts from a Project Gutenberg mirror, not importing Gutenberg texts from the content server via OPDS.

model.py
```diff
@@ -9589,6 +9597,18 @@ def by_protocol(cls, _db, protocol):
             ).filter(ExternalIntegration.protocol==protocol)
         return qu
 
+    @classmethod
+    def by_datasource(cls, _db, data_source):
+        if isinstance(data_source, DataSource):
```
leonardr: A docstring on this method would be helpful -- I know what happens inside but it's not clear from looking at the implementation why you would call it.

leonardr:
:+1:

PR #545, opened 2017-06-26
Remove default license-offering from INTERNAL_PROCESSING DataSource

courte:
I noticed that the INTERNAL_PROCESSING DataSource was set to offer licenses by default when it shouldn't be. Only the Metadata Wrangler currently needs this DataSource to offer up (fake) licenses.


leonardr:
:+1:

Issue #537, opened 2017-06-19
Allow one library to be specified as the default


leonardr:
We have put a lot of work into making sure all libraries in the system are equal and there is not "default" library. However there is one case where we _want_ a default: when someone makes an HTTP request and doesn't include a library name. In that case we want to designate one library as the default. This should be stored in the database (either as a field or a setting) and be settable through the admin interface.

If there is no default set in the database, then the default library should be the one with the lowest database ID.

This has a circulation component and a core component.

PR #550, opened 2017-06-27
Default library, not one library

leonardr:
This branch removes `Library.instance` and replaces it with the concept of the default library. One and only one library is supposed to be the default. This is stored in the new `libraries.is_default` field. I use a setter to make sure that if there are any libraries, one of them is always the default. If there should ever be more than one default library (e.g. due to a race condition), a winner is chosen.

I think we are almost never going to use the default library concept in a multi-library setting, which is why I'm okay with automatically detecting and correcting situations where the number of default libraries != 1.

### DIFF
--- a/migration/20170627-add-default-to-library.sql
+++ b/migration/20170627-add-default-to-library.sql
@@ -1,0 +1,8 @@
+-- Add library_id as a foreign key
+alter table libraries add column is_default boolean default false;
+
+create index "ix_libraries_default" on libraries (is_default);
+
+-- The first library in the system is set as the default.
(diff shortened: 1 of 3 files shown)


leonardr:
This is the backend work for https://github.com/NYPL-Simplified/server_core/issues/537.

migration/20170627-add-default-to-library.sql
```diff
@@ -0,0 +1,8 @@
+-- Add library_id as a foreign key
```
aslagle: wrong comment

PR #551, opened 2017-06-27
Moved analytics configuration to the database.

aslagle:
This branch moves the analytics configuration into an ExternalIntegration, and changes Analytics so its no longer a global singleton. It also defines the constants for LocalAnalyticsProvider to be configured in the admin interface.

### DIFF
--- a/analytics.py
+++ b/analytics.py
@@ -1,71 +1,34 @@
+from nose.tools import set_trace
 import importlib
 import contextlib
 import datetime
-from config import Configuration
+from model import ExternalIntegration
+from config import CannotLoadConfiguration
 
 class Analytics(object):
 
-    __instance = None
+    def __init__(self, _db):
+        self.providers = []
+        self.initialization_exceptions = {}
 
-    if '.' in __module__:
-        # We are operating in an application that imports this product
-        # as a package (probably called 'core'). The module name of
-        # the analytics provider should be scoped to the name of the
-        # package, i.e.  'core.local_analytics_provider'.
-        package_name = __module__[:__module__.rfind('.')+1]
-    else:
-        # This application is not imported as a package, probably
-        # because we're running its unit tests.
-        package_name = ''
+        # Find a list of all the ExternalIntegrations set up with a
+        # goal of analytics.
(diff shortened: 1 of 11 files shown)


metadata_layer.py
```diff
@@ -978,11 +979,13 @@ def apply(self, _db, collection, replace=None):
         if pool and self._availability_needs_update(pool):
             # Update availabily information. This may result in
             # the issuance of additional circulation events.
+            analytics = Analytics(_db)
```
aslagle: I don't like this, but I'm not sure anything else would be better.

aslagle:
Also added in two sitewide settings that weren't in the admin interface yet.

metadata_layer.py
leonardr: Yeah, this is a performance problem. We're looking at database queries and module imports every time we want to change a LicensePool. I think `analytics` should be passed into `ReplacementPolicy` just like `mirror` is.

analytics.py
```diff
@@ -30,5 +37,5 @@ def __init__(self, _db):
     def collect_event(self, library, license_pool, event_type, time=None, **kwargs):
         if not time:
             time = datetime.datetime.utcnow()
-        for provider in self.providers:
+        for provider in (self.sitewide_providers + self.library_providers[library.id]):
```
leonardr: When you call collect_event() to register the same event for multiple libraries, will this cause the same event to be registered multiple times for the sitewide providers?

If so, it might work better for the sitewide providers to be handled as providers that, if enabled, are presumed to be enabled for every library in the system.

aslagle: That's a good point. If there's a collection with multiple libraries it will cause duplicate events in the sitewide provider. I was thinking that I'd have to add the library id to the CirculationEvents table and the Google Analytics setup in a future pr though.

leonardr: Adding library ID to CirculationEvents makes a lot of sense. File a bug for that and then :+1:

aslagle: https://github.com/NYPL-Simplified/server_core/issues/556

Issue #553, opened 2017-06-27
"Where are my books" script


leonardr:
So far, everyone who sets up a circ manager has the same experience:

* Run an import script
* Hit the app server
* "Where are my books?"

There are a number of reasons why you might not see your books, even if everything goes right. A "where are my books" script would try some common tricks like refreshing the materialized views, clearing the `cachedfeeds` table, seeing whether your books are presentation ready, etc. This would save a lot of time debugging peoples' remote setups.

PR #555, opened 2017-06-28
Where are my books?

leonardr:
This branch fixes https://github.com/NYPL-Simplified/server_core/issues/553. It creates a script that automatically solves these issues, all of which result in someone not seeing in the app server books that they imported:

* Your works are not marked presentation-ready even though they totally are. (This happened due to a misconfigured search index.)
* You added works to your database but didn't refresh the materialized views.
* You refreshed the materialized views but old `CachedFeed` objects are preventing you from seeing the new feeds.

It diagnoses, but does not solve, these issues:

* You have no presentation-ready works.
* Your works are presentation-ready but don't show up in the materialized view because there's something else wrong with them (like no LicensePools).

This might be a branch that could be merged into master instead of multi-tenant.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -18,6 +18,7 @@ import traceback
 import unicodedata
 
 from collections import defaultdict
+from external_search import ExternalSearchIndex
 import json
 from nose.tools import set_trace
 from sqlalchemy import create_engine
@@ -36,6 +37,7 @@ from model import (
     get_one,
     get_one_or_create,
     production_session,
+    CachedFeed,
     Collection,
     Complaint,
     ConfigurationSetting,
@@ -50,6 +52,7 @@ from model import (
     LicensePool,
     Patron,
     PresentationCalculationPolicy,
+    SessionManager,
     Subject,
     Timestamp,
     Work,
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -2242,6 +2245,86 @@ def explain_work(cls, work):
...
 
+class FixInvisibleWorksScript(Script):
+    """Try to figure out why Works aren't showing up.
+
+    This is a common problem on a new installation.
+    """
+    def __init__(self, _db=None, output=None, search=None):
+        super(FixInvisibleWorksScript, self).__init__(_db)
+        self.output = output or sys.stdout
+        self.search = search or None # ExternalSearchIndex(_db)
```
courte: You can remove this comment.

scripts.py
```diff
@@ -2242,6 +2245,86 @@ def explain_work(cls, work):
...
+        )
+        
+        # If the number of works in the materialized view is more than
+        # 5% off from the number of presentation-ready works, rebuild
+        # the materialized views.
+        #
+        # If the number of works is very small, such that it's not big
+        # deal to refresh the materialized views, just refresh them
+        # every time.
+        if mv_works_count < 1000 or float(mv_works_count) / ready_count < 0.995:
```
courte: You might want to check `ready_count < 1000` here instead of mv_works_count. The current number of MaterializedWorks might be small, but the number of Works made ready to insert could be much larger.

scripts.py
```diff
@@ -2242,6 +2245,86 @@ def explain_work(cls, work):
...
+        self.output.write("%d works not presentation-ready.\n" % unready_count)
+
+        if unready_count > 0:
+            self.output.write(
+                "Attempting to make %d works presentation-ready based on their metadata.\n" % (unready_count)
+            )
+            for work in unready:
+                work.set_presentation_ready_based_on_content(self.search)
+            self.output.write(
+                "%d works are now presentation-ready.\n" % ready.count()
```
courte: If you reset `ready_count` before this log method, you won't have to recalculate it again below.

scripts.py
leonardr: We talked this over and decided to always refresh the materialized views.

scripts.py
```diff
@@ -2242,6 +2245,82 @@ def explain_work(cls, work):
...
+                "Attempting to make %d works presentation-ready based on their metadata.\n" % (unready_count)
+            )
+            for work in unready:
+                work.set_presentation_ready_based_on_content(self.search)
+
+        ready_count = ready.count()
+
+        if unready_count > 0:
+            self.output.write(
+                "%d works are now presentation-ready.\n" % ready_count
```
courte: Is this to update that some formerly unready Works have become ready, but not all?

leonardr: Yeah, it's so you can see if `set_presentation_ready_based_on_content` did anything.

PR #558, opened 2017-06-29
Stop using library registry fields

aslagle:
We're not using Library.library_registry_short_name and Library.library_registry_shared_secret anymore, so the library configuration script shouldn't set them.


leonardr:
:+1:

PR #559, opened 2017-06-30
Added admin interface constants to search.

aslagle:
I had to switch around the way model and external_search import each other so I could use constants from ExternalIntegration.


leonardr:
:+1:

PR #560, opened 2017-07-03
Made Lane constructor take db argument.

aslagle:
Having Lane._db come from the Library was causing a scoped session error when creating CachedFeeds, so Leonard suggested passing the db to the Lane constructor.


leonardr:
:+1:

PR #565, opened 2017-07-07
Fixed problems with search integration

aslagle:
This branch fixes some problems I found with the search setup when testing a multi-library circ manager.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -55,27 +55,32 @@ class ExternalSearchIndex(object):
         self.works_alias = None
         integration = None
 
-        if not ExternalSearchIndex.__client:
-            if _db and (not url or not works_index):
-                integration = ExternalIntegration.lookup(
-                    _db, ExternalIntegration.ELASTICSEARCH,
-                    goal=ExternalIntegration.SEARCH_GOAL
-                )
+        if not _db:
+            raise CannotLoadConfiguration(
+                "Cannot load Elasticsearch configuration without a database.",
+            )
+        if not url or not works_index:
+            integration = ExternalIntegration.lookup(
+                _db, ExternalIntegration.ELASTICSEARCH,
+                goal=ExternalIntegration.SEARCH_GOAL
+            )
 
-                if not integration:
-                    raise CannotLoadConfiguration(
-                        "No Elasticsearch integration configured."
-                    )
-                url = url or integration.url
-                if not works_index:
-                    setting = integration.setting(self.WORKS_INDEX_KEY)
-                    works_index = setting.value_or_default(
-                        self.DEFAULT_WORKS_INDEX
-                    )
-            if not url:
+            if not integration:
                 raise CannotLoadConfiguration(
-                    "No URL configured to Elasticsearch server."
+                    "No Elasticsearch integration configured."
                 )
+            url = url or integration.url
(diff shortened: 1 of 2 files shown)


external_search.py
```diff
@@ -55,27 +55,32 @@ def __init__(self, _db, url=None, works_index=None):
...
+            if not integration:
                 raise CannotLoadConfiguration(
-                    "No URL configured to Elasticsearch server."
+                    "No Elasticsearch integration configured."
+                )
+            url = url or integration.url
+            if not works_index:
+                setting = integration.setting(self.WORKS_INDEX_KEY)
+                works_index = setting.value_or_default(
+                    self.DEFAULT_WORKS_INDEX
```
aslagle: Before, if __client already existed, the integration and settings weren't looked up again. But the works_index was still being used to set the index and alias. Now, the integration is looked up every time, but the client is still only initialized once.

external_search.py
```diff
@@ -132,8 +138,12 @@ def set_works_index_and_alias(self, current_alias):
         """Finds or creates the works_index and works_alias based on
         provided configuration.
         """
-        index_details = self.indices.get_alias(name=current_alias, ignore=[404])
-        found = bool(index_details) and not (index_details.get('status')==404 or 'error' in index_details)
```
aslagle: If current_alias is None, self.indices.get_alias returns all the indices that have no alias. That makes index_details some arbitrary index that you probably don't want to use. I have lots of those in my local elasticsearch, and it could happen in production if you have an old index that you haven't deleted yet.

tests/test_external_search.py
```diff
@@ -54,259 +62,225 @@ def teardown(self):
...
 
 class TestExternalSearch(ExternalSearchTest):
-    """
-    These tests require elasticsearch to be running locally. If it's not, or there's
-    an error creating the index, the tests will pass without doing anything.
 
-    Tests for elasticsearch are useful for ensuring that we haven't accidentally broken
-    a type of search by changing analyzers or queries, but search needs to be tested manually
-    to ensure that it works well overall, with a realistic index.
+    def test_setup_index_creates_new_index(self):
```
aslagle: I didn't change these tests, but I moved them to a different class than the test that needs to create and index lots of works so they don't need to run the slow setup method.

tests/test_external_search.py
```diff
@@ -54,259 +62,225 @@ def teardown(self):
...
         if self.search:
-            works = []
 
             self.moby_dick = self._work(title="Moby Dick", authors="Herman Melville", fiction=True)
             self.moby_dick.presentation_edition.subtitle = "Or, the Whale"
             self.moby_dick.presentation_edition.series = "Classics"
             self.moby_dick.summary_text = "Ishmael"
             self.moby_dick.presentation_edition.publisher = "Project Gutenberg"
             self.moby_dick.set_presentation_ready()
-            works.append(self.moby_dick)
```
aslagle: I realized set_presentation_ready already adds everything to the search index, so the extra bulk update at the end of this method wasn't doing anything. It might be a little faster to create all the works before setting up the search index and do the bulk update, but if I create the search index first I can skip creating the works entirely if it doesn't work.

leonardr:
:+1:

PR #566, opened 2017-07-10
Add content source enki

datalogics-tarar:
Added support for the Enki content provider


scripts.py
```diff
@@ -539,6 +539,8 @@ def refresh_metadata(self, identifier):
             provider = OverdriveBibliographicCoverageProvider
         elif identifier.type==Identifier.AXIS_360_ID:
             provider = Axis360BibliographicCoverageProvider
+        elif identifier.type==Identifier.ENKI_ID:
+            provider = EnkiBibliographicCoverageProvider
```
leonardr: You're not importing the right class here. At any rate, this script doesn't work very well so I'm not sure it's worth fixing. Maybe just take this whole thing out.

datalogics-tarar: Right. We just thought the Axis import was missing but I'm guessing it's not required. And I refactored the EnkiBibliographicCoverageProvider out of core so I'll delete that.

leonardr: You're right, the Axis import should be required here. This whole conversation sort of shows how this code isn't working and doesn't make sense anymore.

leonardr:
Do we still need this branch or has it effectively been merged? I don't see a lot of nontrivial changes in here.

datalogics-tarar:
It's obsolete and was replaced by https://github.com/NYPL-Simplified/server_core/pull/603

PR #567, opened 2017-07-10
multi tenancy fixes

aslagle:
This fixes a bunch of problems I found when setting up a circ manager with a couple libraries.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -7,6 +7,7 @@ import traceback
 from sqlalchemy.sql.functions import func
 from sqlalchemy.sql.expression import (
     or_,
+    and_,
 )
 
 import log # This sets the appropriate log format and level.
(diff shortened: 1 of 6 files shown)


monitor.py
```diff
@@ -227,7 +228,12 @@ def all(cls, _db, **constructor_kwargs):
         service_match = or_(Timestamp.service==cls.SERVICE_NAME,
                             Timestamp.service==None)
         collections = Collection.by_protocol(_db, cls.PROTOCOL).outerjoin(
-            Collection.timestamps).filter(service_match)
+            Timestamp,
+            and_(
+                Timestamp.collection_id==Collection.id,
+                service_match,
+            )
+        )
```
aslagle: This was filtering out collections that had a Timestamp with a different service name. My change moves the filter into the on clause instead of the where clause, so it won't join Timestamps for other services, but the collections with those Timestamps will still be there.

overdrive.py
```diff
@@ -131,8 +132,6 @@ def __init__(self, collection):
         else:
             self.parent_library_id = None
 
-        self.collection_id = collection.id
```
aslagle: Here collection is the parent collection instead of the child collection. This collection id is used later to set the collection on license pools, so it should be the child collection.

scripts.py
```diff
@@ -2276,6 +2299,39 @@ def run(self):
...
+
+        # Check if the pools have available licenses.
+        mv_works = mv_works.filter(
+            or_(LicensePool.licenses_owned > 0, LicensePool.open_access)
+        )
+        if mv_works.count() == 0:
+            self.output.write(
+                "Here's your problem: your works aren't open access and have no licenses owned.\n"
+            )
+            return
```
aslagle: None of these problems can be fixed automatically, but it's still helpful to know what the problem is.

leonardr:
This is :100:, exactly the sort of problems I knew you would find! :+1:

PR #569, opened 2017-07-12
Fix migrations for 1.1.23

courte:
This removal of a transitional migration used in development and the pgcrypto extension support smooth migration to the next release of master.

The changes to Admin support [basic authentication usage](http://explore-flask.readthedocs.io/en/latest/users.html). It looks like there are some [more complex solutions](http://variable-scope.com/posts/storing-and-verifying-passwords-with-sqlalchemy) that better incorporate SQLAlchemy and would allow the sort of straightforward comparison that was possible in the database, but I didn't think it was necessary in this pass.


migration/20170328-move-collection-info-to-external-integration.sql
```diff
@@ -1,61 +0,0 @@
--- For OPDS Import collections, move the url to external_account_id.
-UPDATE collections SET external_account_id = url WHERE protocol = 'OPDS Import';
-UPDATE collections SET url = NULL WHERE protocol = 'OPDS Import';
```
aslagle: I think this piece of the migration got lost...

model.py
```diff
@@ -8535,23 +8533,32 @@ def update_credentials(self, _db, credential=None):
...
+        return self.password_hashed == bcrypt.hashpw(password, self.password_hashed)
+
+    @classmethod
+    def authenticate(cls, _db, email, password):
+        """Finds an authenticated Admin by email and password
+
+        :return: Admin or None
+        """
+        match = None
+        matches = _db.query(Admin).filter(Admin.email==unicode(email)).all()
```
aslagle: I think you could use get_one for this.

PR #570, opened 2017-07-13
Changed lane to include the parent collection when a library is associated with a child collection.

aslagle:
When I tried setting up an Overdrive Advantage collection, I expected that I would be able to add the Overdrive Advantage collection to a library and have that library's feeds also contain the books from the parent consortium collection. Without this change, I'd have to separately add the consortium collection to the library.

The downside is that now there's no way to get a feed that has the Overdrive Advantage books without the consortium's books.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -550,6 +550,8 @@ class Lane(object):
         self.library_id = library.id
         self.collection_ids = [
             collection.id for collection in library.collections
+        ] + [
+            collection.parent_id for collection in library.collections if collection.parent_id
         ]
         self.default_for_language = False
         self.searchable = searchable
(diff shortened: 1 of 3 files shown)


lane.py
```diff
@@ -550,6 +550,8 @@ def __init__(self,
         self.library_id = library.id
         self.collection_ids = [
             collection.id for collection in library.collections
+        ] + [
+            collection.parent_id for collection in library.collections if collection.parent_id
```
leonardr: I think it would be better to defined a method on Library that returned all its collections, recursively, and did `[c.id for collection in method()]` here. This can't be the only place we need all the collections. (Although the Lane stuff is centralized well enough that it might be.)

PR #573, opened 2017-07-14
Fix migrations for 2.0.0

courte:
This includes the changes in `fix-migrations-for-1.1.23` and all other changes on master, while also resorting and fixing migrations to transition from master to multi-tenancy.

There's a good chunk of ugly, temporary code to address issues when a database object's structure has changed but the existing table hasn't because the migration script hasn't been run yet. (This is primarily an issue in the case of Timestamps, which are used by the DatabaseMigrationScript and also changed by the same script, wow-ee.)


migration/20170328-move-collection-info-to-external-integration.sql
```diff
@@ -1,61 +0,0 @@
--- For OPDS Import collections, move the url to external_account_id.
```
courte: This file removal is merged from master.

migration/20170523-add-admin-password.sql
```diff
@@ -1,3 +1,2 @@
-CREATE EXTENSION pgcrypto;
```
courte: File merged from master.

migration/20170615-add-name-to-externalintegration.sql
```diff
@@ -1,2 +0,0 @@
-alter table externalintegrations add column name character varying;
```
courte: File moved [below](https://github.com/NYPL-Simplified/server_core/pull/573/files#diff-f1c46e8f6321fa77f3269f9972c8a783) with additional changes.

migration/20170626-no-internal-processing-licenses-by-default.sql
```diff
@@ -0,0 +1,3 @@
+update datasources
```
courte: Merged from master.

migration/20170627-remove-unresolved-identifiers.sql
```diff
@@ -0,0 +1 @@
+drop table if exists unresolvedidentifiers;
```
courte: Merged from master.

model.py
```diff
@@ -332,13 +313,72 @@ def initialize_data(cls, session):
             )
             mechanism.default_client_can_fulfill = True
 
-        # Set the timestamp which site_configuration_has_changed keeps
-        # updated.
-        Timestamp.stamp(session, Configuration.SITE_CONFIGURATION_CHANGED,
-                        collection=None)
+        # TODO: Remove this exception catch after version 2.0.0. See
```
courte: This is where the ugliness begins.

model.py
```diff
@@ -1019,7 +1059,7 @@ def well_known_sources(cls, _db):
                 (cls.OA_CONTENT_SERVER, True, False, None, None),
                 (cls.NOVELIST, False, True, Identifier.NOVELIST_ID, None),
                 (cls.PRESENTATION_EDITION, False, False, None, None),
-                (cls.INTERNAL_PROCESSING, True, False, None, None),
+                (cls.INTERNAL_PROCESSING, False, False, None, None),
```
courte: This and the UnresolvedIdentifier changes are from master. No need to review.

model.py
```diff
@@ -9101,29 +9054,35 @@ def update_credentials(self, _db, credential=None):
             self.credential = credential
         _db.commit()
 
-    class HashedPasswordComparator(Comparator):
```
courte: Changes in the rest of this file are merged from master.

tests/test_model.py
```diff
@@ -6531,14 +6531,17 @@ def test_license_data_source_is_stored_in_views(self):
```
courte: Changes in were reviewed on master.

courte:
With ruby's ActiveRecord, there's a table in the database that exists **only** to track migrations. If this is a recurring problem, that could help avoid having to create core-level codebase workarounds. We'd have to create one for the configuration timestamp as well, though, and I don't see this being a recurring problem.

migration/20170713-17-move-externalintegration-columns-to-settings.sql
```diff
@@ -18,4 +18,22 @@ WHERE e.password is not null;
...
+SET protocol = c.protocol
+FROM collections as c
+WHERE e.id = c.external_integration_id;
+
+-- Create an externalintegration for the Open Access Content Server.
+INSERT INTO externalintegrations(protocol)
+SELECT c.protocol
+FROM collections c
+WHERE c.external_integration_id is null;
+
```
leonardr: I don't understand this one.

courte: With the changes to Collection on master, collections and their external integrations are created before ExternalIntegration.protocol exists. At that point, external integrations are also only created for a Collection if it has a url, username, or password.

Content Server doesn't have any of those attributes, so it's never associated with an external integration in the database. Because the oversight should only occur for the content server at this point, this migration creates an integration for any collection that doesn't have one (content server, if anything) and then associates that collection with the new integration below.

PR #575, opened 2017-07-17
Migrations without config file

courte:
This merges the `sys.exit()` changes from master. It also sweeps away some lingering columns and Collection data from 1.1.23.


migration/20170713-17-move-externalintegration-columns-to-settings.sql
```diff
@@ -37,3 +37,10 @@ UPDATE collections c
 SET external_integration_id = e.id
 FROM externalintegrations e
 WHERE e.protocol = 'OPDS Import' and c.external_integration_id is null;
+
+ALTER TABLE collections DROP COLUMN protocol;
```
leonardr: Does this need an IF EXISTS or will it always be there at this point?

courte: It should always be there, but it makes sense to add `if exists` in general. No sense creating unhelpful migration errors.

Issue #579, opened 2017-07-20
Crash when search index alias already exists but as an index, not an alias


leonardr:
Brooklyn's QA Elasticsearch server has an index called "circulation-works-current". When `ExternalSearchIndex.set_works_index_and_alias` runs, it doesn't find an alias called "circulation-works-current" (because it's an index, not an alias), so it tries to create one, which causes a crash.

This is a problem in 1.1 but I think it becomes an even bigger problem in 2.0 since you can edit the name of the alias in the admin interface.

This needs to be addressed in code (I have a workaround but don't know if it's the best solution) and there's also an operational aspect of this where Brooklyn's server(s) should probably be changed.

PR #580, opened 2017-07-20
Avoid a crash if the supposed search alias exists as an index

leonardr:
I'm not sure if this is the right fix for https://github.com/NYPL-Simplified/server_core/issues/579, but it works and I figure it's a good place to start.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -86,29 +86,28 @@ class ExternalSearchIndex(object):
         except Exception, e:
             found_index = False
 
-        if found_index:
-            # The alias exists as an index. This isn't supposed to happen,
-            # but we can avoid a meltdown now by doing nothing.
-            self.log.error(
-                'Supposed alias "%s" already exists, as an index.',
-                current_alias
-            )
-            return
-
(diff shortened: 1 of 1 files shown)


external_search.py
```diff
@@ -80,13 +80,30 @@ def set_works_index_and_alias(self, current_alias):
...
+            found_index = False
+
+        if found_index:
+            # The alias exists as an index. This isn't supposed to happen,
+            # but we can avoid a meltdown now by doing nothing.
+            self.log.error(
+                'Supposed alias "%s" already exists, as an index.',
+                current_alias
+            )
+            return
```
aslagle: So after this we'd end up with self.works_index = None?

leonardr: That's a problem. It works when I run it for real (not sure how) but it makes tests fail.

aslagle: It may be able to read from the index without that but not write to the index.

leonardr:
We've decided to fix the BPL index manually and address this problem in 2.0.

PR #584, opened 2017-07-26
Fix Axis configuration error import

courte:
Fixes a small import bug reported by datalogics


courte:
I'm going to pull this against both `master` and `1.x`, once it's been approved.

leonardr:
When you merge with 1.x, would you do me a favor and also merge https://github.com/NYPL-Simplified/server_core/pull/587/files ?

PR #599, opened 2017-08-04
Finalize authentication for opds format

leonardr:
This branch rewrites the Authentication For OPDS code to take into account the (semi-) finalized A4OPDS format, defined at https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions#changes-to-the-core-spec.

Rather than taking a skeletal OPDS document and filling it out (something that has not been used for a long time if we ever used it at all), the new system builds a data structure and outputs a dictionary that becomes an A4OPDS document when serialized as JSON.


config.py
```diff
@@ -417,9 +410,6 @@ def site_configuration_last_update(cls, _db, known_value=None,
         else:
             last_update = known_value
 
-        logging.debug(
-            "Updating Configuration last update time to %s", last_update
-        )
```
leonardr: I'm sick of this debug statement happening all the time in development.

Issue #598, opened 2017-08-04
Materialized views draw data_source_id and primary_identifier_id from Edition (metadata source) not LicensePool (license source)


leonardr:
This means that when books are imported from the open-access content server, the data source in the materialized view (Open Access Content Server) doesn't match the data source associated with the LicensePool or the LicensePoolDeliveryMechanism (Feedbooks, Gutenberg, etc.).

This didn't used to matter, but now we use data_source_id and primary_identifier_id when joinin the materialized view against LicensePoolDeliveryMechanism looking for a way to fulfill the books in the license pool. We need to start using LicensePool's identifier_id and data_source_id in the materialized view.

leonardr:
I'm a little confused about the mechanism here, since `Library.restrict_to_ready_deliverable_works` doesn't work the way I thought it did. It uses an existing join against LicensePool and joins _that_ table against LicensePoolDeliveryMechanism. using `LicensePool.data_source_id` and `LicensePool.identifier_id`. So in theory that shouldn't be the problem. But making this change to the materialized view makes sense and seems to solve the problem in real life.

PR #600, opened 2017-08-07
Fix materialized view ids

leonardr:
This branch fixes https://github.com/NYPL-Simplified/server_core/issues/598.

The data source ID seems to matter much more than the identifier ID. I changed the name of the identifier field (from primary_identifier_id to identifier_id) and it didn't break any tests or seem to matter at all. I don't see primary_identifier_id in opds.py or lane.py, the two places that do the most work with the materialized views.

I'm piggybacking off of the most recent "recreate materialized views" migration rather than creating a new one, since we don't have any production deployments that have already run that script.

### DIFF
--- a/files/materialized_view_works.sql
+++ b/files/materialized_view_works.sql
@@ -3,8 +3,8 @@ as
  SELECT 
     distinct works.id AS works_id,
     editions.id AS editions_id,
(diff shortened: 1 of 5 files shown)


courte (approved):
Approved, once tests are fixed.

PR #603, opened 2017-08-08
Added Enki datasource

datalogics-dans:
When trying to get the Enki content import working, this seemed to be the one core change that was necessary. It's possible there are others as well, but for now it's just this.


leonardr:
This looks good. Now you can set ENKI_ID in your circulation classes to refer to Identifier.ENKI_ID.

PR #606, opened 2017-08-09
OneClick monitors

leonardr:
This branch makes it possible to run the OneClick import and delta scripts in a multi-collection environment. This works by creating `CollectionMonitor` classes that can be invoked with the standard `RunCollectionMonitorScript`.

Prior to this branch, invoking one of these scripts would cause an error, because nothing in the script invocation specified the Collection to use. With `RunCollectionMonitorScript` each relevant Collection is processed in turn.

If necessary, we can improve `RunCollectionMonitorScript` later on to give more precise control over specifying a Collection from the command line, and all the `RunCollectionMonitorScript`s will benefit from the work.

### DIFF
--- a/oneclick.py
+++ b/oneclick.py
@@ -45,6 +45,8 @@ from metadata_layer import (
     SubjectData,
 )
 
+from monitor import CollectionMonitor
+
 from util import LanguageCodes
 
 from util.personal_names import name_tidy, sort_name_to_display_name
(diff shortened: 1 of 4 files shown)


leonardr:
This fixes https://github.com/NYPL-Simplified/circulation/issues/629.

courte (approved):

PR #608, opened 2017-08-15
Cache Genre objects in a big dictionary

leonardr:
This branch caches all the Genre objects in a dictionary ahead of time. The Genres are detached from the database session that created them, and then when they're looked up they're attached with the database session that wants to do the lookup. This gives the appearance of a Genre individually looked up and fresh from the database, even though all the genres are loaded from the database in a single query.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5368,7 +5368,7 @@ class Genre(Base):
                 if genre is None:
                     logging.getLogger().error('"%s" is not a recognized genre.', name)
                     return None, False
-            cls._cache_genre(cache, genre)
+            cls._cache_genre(cls._cache, genre)
             
         # Now we know the genre is in the cache. Retrieve it and associate
         # it with this database session.
--- a/model.py
+++ b/model.py
@@ -5333,7 +5333,7 @@ class Genre(Base):
             self.name, len(self.subjects), len(self.works), length)
 
     @classmethod
-    def _cache_genre(cls, genre):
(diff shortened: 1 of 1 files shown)


model.py
```diff
@@ -5323,18 +5333,48 @@ def __repr__(self):
...
+        """
+        make_transient(genre)
+        make_transient_to_detached(genre)
+        cls._cache[genre.name] = genre
+        return genre
+    
+    @classmethod
+    def load_all(cls, _db):
+        cls._cache = {}
+        cache = cls._cache
```
aslagle: unused variable

leonardr:
I changed the code to use the same pattern I'm using for a ConfigurationSetting cache. The whole cache is replaced at once. This isn't really important for Genres, where the cache is only ever loaded once, but I want to use this pattern consistently so maybe it can be refactored later.

aslagle:
What about when lookup calls _cache_genre?

leonardr:
By "the whole cache is replaced at once" I mean that I create a new dict, populate it, and then assign it to Genre._cache, rather than assigning an empty dict to Genre._cache and then populating it. If additional things are added to the cache later, that's okay. This change is to minimize the chance of two threads stepping on each other by populating the same cache object.

leonardr:
Hmm, I think you were actually pointing out that I didn't change the other call to _cache_genre.

PR #609, opened 2017-08-15
Cache ConfigurationSettings

leonardr:
This branch creates a cache that contains the entire `configurationsettings` table. The lookup methods are changed to check this cache, with the database as a fallback.

This is similar to my earlier cache of the `genres` table, but a bit more complicated. Unlike `Genre`, we expect new `ConfigurationSetting` objects to be created and modified after the site starts up, so we need a mechanism for invalidating the cache. I use the same event monitoring system we use to keep an eye on the site configuration.

Since the cache can be invalidated, there's also the possibility that it gets invalidated while you're looking  up a `ConfigurationSetting`, leaving you with nowhere to cache the object.

I created a mix-in class for this "keep the whole table cached" strategy and adapted Genre to use it as well. In a future branch I'll apply this strategy to DataSource, Library, and Collection, which should save at least a couple database queries on every incoming request.

This branch has some known problems which are addressed in the follow-up: https://github.com/NYPL-Simplified/server_core/pull/610

I don't know whether it'd be easier for you to review that branch and do all this code at once, or review the basic stuff and then the branch I think completes the work.

### DIFF
--- a/model.py
+++ b/model.py
@@ -306,7 +306,7 @@ class SessionManager(object):
         list(DataSource.well_known_sources(session))
 
         # Load all existing Genre objects.
-        Genre.load_all(session)
+        Genre.populate_cache(session)
         
         # Create any genres not in the database.
         for g in classifier.genres.values():
@@ -5321,8 +5321,14 @@ class Genre(Base):
     work_genres = relationship("WorkGenre", backref="genre", 
                                cascade="all, delete, delete-orphan")
 
-    # A dictionary of all known Genre objects by name.
-    _cache = {}
+    # The in-memory Genre cache starts in an invalid state and must be
+    # populated before it can be used.
+    RESET = object()
+    _cache = RESET
+
+    @classmethod
+    def reset_cache(cls):
+        cls._cache = cls.RESET
     
     def __repr__(self):
         if classifier.genres.get(self.name):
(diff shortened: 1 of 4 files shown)


opds.py
```diff
@@ -1108,7 +1108,7 @@ def license_tags(cls, license_pool, loan, hold):
             return
         default_loan_period = default_reservation_period = None
         collection = license_pool.collection
-        if (loan or hold) and not open_access:
+        if (loan or hold) and not license_pool.open_access:
```
leonardr: This is an unrelated fix to code I landed earlier.

model.py
```diff
@@ -9497,21 +9540,56 @@ def for_externalintegration(cls, key, externalintegration):
...
+    @classmethod
+    def _cache_insert(cls, setting, cache):
+        """Cache a ConfigurationSetting for later retrieval, probably by a
+        different database session.
+        """
+        library_id = setting.library_id
+        external_integration_id = setting.external_integration_id
+        key = setting.key
+        cache_key = (library_id, external_integration_id, key)
+        cache[cache_key] = setting
```
aslagle: Do you need to override _cache_insert here? Won't HasFullTableCache._cache_insert work since you've defined _cache_key?

leonardr: You're right, I just caught that in the third branch, which I'm working on now.

PR #610, opened 2017-08-16
More full table caches

leonardr:
This branch expands on my earlier branch by adding full-table caches for DataSource,  Collection, and Library. The first two can be used as-is because they're tied into existing lookup methods. For Library I defined a new `lookup` method and made a change in my circulation branch to have `library_for_request` use that new method instead of going direct to the database.

The DataSource change is the big winner here; I think it will save seven database queries on an average  request for an OPDS feed. The Library change will save one query on every single request, once my circulation change is made. I don't think the Collection change will save us any queries in current typical usage, but I have an idea for future improvements.

If we could do this for ExternalIntegration we could save another four queries on an average request, but ExternalIntegration.lookup is complicated enough that it might not be possible.

This branch also changes HasFullTableCache._check_cache to repopulate the entire cache if it notices the cache has been reset. Previously this method was proceeding with an individual lookup once the cache was reset, which meant that although all the tests passed the cache was never actually being populated. This branch includes tests that actually look inside the internal cache to make sure it's been populated.

### DIFF
--- a/testing.py
+++ b/testing.py
@@ -165,8 +165,9 @@ class DatabaseTest(object):
 
         # Remove any database objects cached in the model classes but
         # associated with the now-rolled-back session.
-        Genre.reset_cache()
         ConfigurationSetting.reset_cache()
+        DataSource.reset_cache()
+        Genre.reset_cache()
         Library.reset_cache()
         
         # Also roll back any record of those changes in the
--- a/testing.py
+++ b/testing.py
@@ -167,6 +167,7 @@ class DatabaseTest(object):
         # associated with the now-rolled-back session.
         Genre.reset_cache()
         ConfigurationSetting.reset_cache()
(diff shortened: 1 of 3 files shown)


testing.py
```diff
@@ -165,8 +165,10 @@ def teardown(self):
 
         # Remove any database objects cached in the model classes but
         # associated with the now-rolled-back session.
-        Genre.reset_cache()
         ConfigurationSetting.reset_cache()
+        DataSource.reset_cache()
+        Genre.reset_cache()
+        Library.reset_cache()
```
aslagle: no Collection here?

aslagle (approved):

testing.py
leonardr: Thanks for the catch, I've fixed in branch #3.

PR #611, opened 2017-08-16
Fix Metadata Wrangler integration

courte:
This branch has a few fixes to get cataloguing working between the Circulation Manager and the Metadata Wrangler.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -720,7 +720,7 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
 
     # By default, Works calculated by a CollectionCoverageProvider update
     # the ExternalSearchIndex. Set this value to True for applications that
-    # use external search, such as the Metadata Wrangler.
+    # don't use external search, such as the Metadata Wrangler.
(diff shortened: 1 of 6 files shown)


coverage.py
```diff
@@ -717,7 +717,12 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
     # CoverageProvider. If this CoverageProvider can manage collections
     # for any protocol, leave this as None.
     PROTOCOL = None
-    
+
+    # By default, Works calculated by a CollectionCoverageProvider update
+    # the ExternalSearchIndex. Set this value to True for applications that
+    # use external search, such as the Metadata Wrangler.
+    EXCLUDE_SEARCH_INDEX = False
```
leonardr: The layers of negatives here confuse me. 1) Why do we want to exclude applications that use external search, instead of applications that don't use it? 2) I thought the Metadata Wrangler _doesn't_ use external search.

courte: Ah, yes, this is confusing and inaccurate. Thanks.

PR #612, opened 2017-08-16
Cache by ID

leonardr:
This branch creates a second cache managed by HasFullTableCache. This one tracks the contents of a table by its database ID. I define a new method to pull things from this cache (`HasFullTableCache.by_id`) and rename `HasFullTableCache._check_cache` to `HasFullTableCache.by_cache_key`.

This isn't really used in core, but there are several places where I think it can save us database queries in circulation.

This branch is also a lot more careful about avoiding race conditions where a cache is invalidated while we're looking something up.

Here's the description for another branch which I just decided to put in with this one because it's more of the same:

This branch eliminates a lot of database hits by replacing many calls to `get_one()` with calls to HasFullTableCache.by_cache_key and (mainly) HasFullTableCache.by_id. The single biggest win here is the change in Lane.library(), which was causing hundreds of database hits when the circ manager was started up.

This branch adds the HasFullTableCache mixin to two more classes: DeliveryMechanism and ExternalIntegration. The ExternalIntegration implementation is a little ugly, but we only need the by_id implementation, so it works.

### DIFF
--- a/model.py
+++ b/model.py
@@ -8405,7 +8405,12 @@ class DeliveryMechanism(Base, HasFullTableCache):
 
     @classmethod
     def lookup(cls, _db, content_type, drm_scheme):
-        return cls.by_cache_key(_db, (content_type, drm_scheme), None)
+        def lookup_hook():
+            return get_one_or_create(
+                _db, DeliveryMechanism, content_type=content_type,
+                drm_scheme=drm_scheme
+            )
+        return cls.by_cache_key(_db, (content_type, drm_scheme), lookup_hook)
 
     @property
     def implicit_medium(self):
--- a/model.py
+++ b/model.py
@@ -10481,6 +10481,14 @@ def refresh_datasource_cache(mapper, connection, target):
     # the cache will be repopulated.
     DataSource.reset_cache()
 
+@event.listens_for(DeliveryMechanism, 'after_insert')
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -10415,6 +10475,14 @@ def refresh_datasource_cache(mapper, connection, target):
...
     # the cache will be repopulated.
     DataSource.reset_cache()
+
+@event.listens_for(ExternalIntegration, 'after_insert')
+@event.listens_for(ExternalIntegration, 'after_delete')
+@event.listens_for(ExternalIntegration, 'after_update')
+def refresh_datasource_cache(mapper, connection, target):
+    # The next time someone tries to access an ExternalIntegration,
+    # the cache will be repopulated.
+    ExternalIntegration.reset_cache()
```
aslagle: DeliveryMechanism doesn't have event listeners yet.

testing.py
```diff
@@ -165,8 +165,10 @@ def teardown(self):
 
         # Remove any database objects cached in the model classes but
         # associated with the now-rolled-back session.
+        Collection.reset_cache()
         ConfigurationSetting.reset_cache()
         DataSource.reset_cache()
+        ExternalIntegration.reset_cache()
```
aslagle: And DeliveryMechanism should be here too.

PR #621, opened 2017-08-25
Made it possible to import an OPDS feed without a metadata wrangler integration.

### LINKED ISSUE #619: OPDS Import script should be able to run without a Metadata Wrangler integration
Currently you get an error like `core.config.CannotLoadConfiguration: No ExternalIntegration found for the Metadata Wrangler.`

aslagle:
This fixes #619.


aslagle:
This reduces the configuration needed for someone setting up a basic circ manager for testing.

PR #623, opened 2017-08-25
Correct ACSM media type.

leonardr:
I'm sick of this problem so I'm just going to fix https://github.com/NYPL-Simplified/circulation/issues/483 and see what breaks.


leonardr:
I don't see a way to test them without getting this change in. We'll back it out if it won't work.

PR #629, opened 2017-09-01
Remove edition license pool

leonardr:
This branch replaces `Edition.license_pool` with `Edition.license_pools`. This code should have been changed in the 2.0 release, but everything appears to work, until you get two collections that provide LicensePools for the same Identifier and DataSource, a case we hadn't tested.

I'll be checking all three components and doing branches for them if they use `Edition.license_pool`.

Update:

content: https://github.com/NYPL-Simplified/content_server/pull/139
circulation: https://github.com/NYPL-Simplified/circulation/pull/659
metadata (just a submodule update): https://github.com/NYPL-Simplified/metadata_wrangler/pull/122

### DIFF
--- a/model.py
+++ b/model.py
@@ -2805,13 +2805,14 @@ class Edition(Base):
         return r
 
     @property
-    def license_pool(self):
-        """The Edition's corresponding LicensePool, if any.
+    def license_pools(self):
+        """The LicensePools that provide access to the book described
+        by this Edition.
         """
         _db = Session.object_session(self)
-        return get_one(_db, LicensePool,
-                       data_source=self.data_source,
-                       identifier=self.primary_identifier)
+        return _db.query(LicensePool).filter(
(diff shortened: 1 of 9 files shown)


model.py
```diff
@@ -2807,13 +2807,14 @@ def for_foreign_id(cls, _db, data_source,
...
+        """The LicensePools that provide access to the book described
+        by this Edition.
         """
         _db = Session.object_session(self)
-        return get_one(_db, LicensePool,
-                       data_source=self.data_source,
-                       identifier=self.primary_identifier)
+        return _db.query(LicensePool).filter(
+            LicensePool.data_source==self.data_source,
+            LicensePool.identifier==self.primary_identifier).all()
```
leonardr: Here's the actual change. Everything else in here is getting the tests to work.

aslagle (approved):

PR #633, opened 2017-09-05
Support OPDS authentication on the Metadata Wrangler

courte:
These changes to core support OPDS authentication on the Metadata Wrangler, including changing the IntegrationClient and how it authenticates.

A change to DummyHTTPClient requires some other, unrelated test code to change order (but not content).

### DIFF
--- a/testing.py
+++ b/testing.py
@@ -1022,9 +1022,9 @@ class DummyHTTPClient(object):
                 headers[k.lower()] = v
         self.responses.append((response_code, headers, content))
 
-    def do_get(self, url, headers, **kwargs):
+    def do_get(self, url, *args, **kwargs):
         self.requests.append(url)
-        return self.responses.pop()
+        return self.responses.pop(0)
 
 
 class MockRequestsResponse(object):
--- a/testing.py
+++ b/testing.py
@@ -829,8 +829,7 @@ class DatabaseTest(object):
     def _integration_client(self, url=None):
         url = url or self._url
         return get_one_or_create(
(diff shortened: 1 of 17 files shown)


testing.py
```diff
@@ -1024,9 +1023,9 @@ def queue_response(self, response_code, media_type="text/html",
                 headers[k.lower()] = v
         self.responses.append((response_code, headers, content))
 
-    def do_get(self, url, headers, **kwargs):
+    def do_get(self, url, *args, **kwargs):
         self.requests.append(url)
-        return self.responses.pop()
+        return self.responses.pop(0)
```
courte: Plucking from the front to make loading test response queuing a mite simpler.

tests/test_opds_import.py
```diff
@@ -1103,20 +1103,20 @@ def test_resources_are_mirrored_on_import(self):
 </svg>"""
 
         http = DummyHTTPClient()
-        # The request to http://root/full-cover-image.png
-        # will result in a 404 error, and the image will not be mirrored.
-        http.queue_response(404, media_type="text/plain")
         http.queue_response(
-            200, content='I am 10557.epub.images',
-            media_type=Representation.EPUB_MEDIA_TYPE,
+        200, content='I am 10441.epub.images',
```
courte: All changes in this file swap the order of queued mock responses.

testing.py
leonardr: This is the right thing to do but it's probably going to affect circulation and content as well.

PR #636, opened 2017-09-06
Dont use ORM when migrating database

leonardr:
This branch changes `DatabaseMigrationScript` so that it never loads the database model. Loading the database model for version N can cause a crash if the actual database schema still reflects version N-1. This is the situation `DatabaseMigrationScript` is supposed to address, and this script has been using `Timestamp` from the start, so I don't understand why this ever worked.

I have manually tested this using the `20170831-add-data-source-integration-client-id.sql` script, with and without a migration row in the `timestamps` table.

### DIFF
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -793,7 +793,7 @@ class TestDatabaseMigrationScript(DatabaseTest):
         eq_(self.timestamp.counter, 3)
 
     def test_all_migration_files_are_run(self):
-        self.script.run()
+        self.script.run(test=True)
 
         # There are two test timestamps in the database, confirming that
         # the test SQL files created by self._create_test_migration_files()
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 import os
+import shutil
 import tempfile
 from StringIO import StringIO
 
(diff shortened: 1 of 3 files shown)


leonardr:
I can't actually run these particular tests on my computer (it doesn't like the `os.rename` call in the setup) so diagnosing this test failure is really slow for me. I'll keep at it but if you have any ideas it might save me come time.

courte:
I can pick this up and keep running with it tomorrow, since the tests run fine for me. I'm curious about how you were planning to handle the python migrations, since without ORM, they won't work. Building a database session at python migration runtime maybe? I'm thinking about how harmful it would be to run SQL migrations before python ones.

It's also possible that this is cacheing related, as Amy mentioned in Slack. Is there a way to turn off cacheing temporarily to check?

leonardr:
The Python scripts call `production_session` and get a separate database session. At _that_ point the database schema has to be up to date with what the model expects. This means you can't do a migration that goes SQL -> Python -> SQL, you have to make a pit stop at the Python version, but I don't think 

I don't think this is caching related because the exception is raised when I call `Base.metadata.create_all(engine)` in `SessionManager.initialize`. But the timing works out -- this stopped working around the time I added the cache.

leonardr:
I tried it again on core master and the problem actually happens in `SessionManager.initialize_data`. I had it happening in `Base.metadata.create_all` but I must have made some change that caused it to happen there.

`initialize_data` touches `DataSource`, `Genre`, `DeliveryMechanism`, and `Timestamp`. If we have a SQL migration that touches any of _those_ classes we'll have this problem. I think `20170831-add-data-source-integration-client-id.sql` is the first time we've changed one of those classes _after_ adding it to `initialize_data`.

tests/test_scripts.py
```diff
@@ -532,7 +533,7 @@ def _create_test_migrations(self):
         for migration_dir in directories:
             if not os.path.isdir(migration_dir):
                 temp_migration_dir = tempfile.mkdtemp()
-                os.rename(temp_migration_dir, migration_dir)
+                shutil.move(temp_migration_dir, migration_dir)
```
leonardr: These tests were always failing on my machine because I have a separate `/home` partition, and `os.rename` can't move things across partitions.

Issue #595, opened 2017-08-03
Rename OneClickDigital to "RBdigital"


leonardr:
As with 3M -> Bibliotheca, OneClickDigital has rebranded itself as RBdigital.

The most important thing that needs to change is the URLs we use by default to connect to their API.

Production base URL: https://api.rbdigital.com/
QA base URL: http://api.rbdigitalstage.com/

But we should also change the data source and the identifier type and all the classes and all the modules etc. etc. etc. to use the new name. :(

leonardr:
I've changed enough to call this fixed -- end-users won't notice any "OneClick" anything.

PR #637, opened 2017-09-07
Rename oneclick

leonardr:
This branch does the minimum work necessary to fix https://github.com/NYPL-Simplified/server_core/issues/595.

I update the API URLs.

I change the strings that end-users are likely to notice -- the name of the data source and the name of the identifier type.

I add a migration script that updates `datasources`, `identifiers`, `collections`, and `configurationsettings` to change the old names/URLs to the new ones. I don't think there are any other tables that need to be changed.

In addition, I define new constants and change existing code to use the new constants instead of the old ones, which isn't strictly necessary.

I'm not going all out because the Bibliotheca change was a huge pain. So I haven't changed the name of the subject classifications or the DRM technique. This clearly isn't going to be the last time a license provider changes its name, so I'm trying to hone in on the minimal set of changes that have to be made.


migration/20170907-rename-one-click.sql
```diff
@@ -0,0 +1,5 @@
+update datasources set name='RBdigital' where name='OneClick';
+update externalintegrations set protocol='RBdigital' where protocol='OneClick';
+update configurationsettings set value='https://api.rbdigital.com/' where key='url' and value='https://api.rbdigital.com/';
```
aslagle: I think the second url in this line is meant to have oneclick in it?

PR #640, opened 2017-09-11
Change a LicensePool based on a CirculationEvent

leonardr:
This branch makes it possible to update a LicensePool's availability information based on the information in a CirculationEvent. Usually we update a LicensePool's availability information and then send out CirculationEvents to match, but for Bibliotheca we do things in the other order.

This is not 100% reliable because we a single event doesn't give us certainty about the complete state of a LicensePool, but it's better than nothing and it should do a good job of filling in the gaps between runs of the `BibliothecaCirculationSweep`.

### DIFF
--- a/model.py
+++ b/model.py
@@ -6547,6 +6547,10 @@ class LicensePool(Base):
         _db = Session.object_session(self)
         if not as_of:
             as_of = datetime.datetime.utcnow()
+        elif as_of == CirculationEvent.NO_DATE:
+            # The caller explicitly does not want
+            # LicensePool.last_checked to be updated.
+            as_of = None
 
         old_licenses_owned = self.licenses_owned
         old_licenses_available = self.licenses_available
@@ -6598,7 +6602,7 @@ class LicensePool(Base):
             self.patrons_in_hold_queue = new_patrons_in_hold_queue
             any_data = True
 
-        if any_data or changes_made:
+        if as_of and (any_data or changes_made):
             # Sometimes update_availability is called with no actual
             # numbers, but that's not the case this time. We got
             # numbers and they may have even changed our view of the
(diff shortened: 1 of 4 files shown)


leonardr:
In Slack conversation with Amy she convinced me to try a different approach.

leonardr:
I think this is ready for another look.

model.py
```diff
@@ -6616,6 +6620,98 @@ def update_availability(
...
+            # We have a history for this LicensePool and we don't know
+            # where this event fits into that history. Ignore the
+            # event.
+            return
+
+        (new_licenses_owned, new_licenses_available, 
+         new_licenses_reserved, 
+         new_patrons_in_hold_queue) = self._calculate_change_from_one_event(
+             event_type, delta
+         )
```
aslagle: `update_availability` uses these four values to guess the event type, basically reversing the work you just did. I'm not sure how to improve it though.

model.py
```diff
@@ -6616,6 +6620,98 @@ def update_availability(
...
+        CE = CirculationEvent
+        added = False
+        if type == CE.DISTRIBUTOR_HOLD_PLACE:
+            new_patrons_in_hold_queue += delta
+        elif type == CE.DISTRIBUTOR_HOLD_RELEASE:
+            new_patrons_in_hold_queue = deduct(new_patrons_in_hold_queue)
+        elif type == CE.DISTRIBUTOR_CHECKIN:
+            new_licenses_available += delta
+        elif type == CE.DISTRIBUTOR_CHECKOUT:
+            new_licenses_available = deduct(new_licenses_available)
```
aslagle: update_availability has a bug where it misses events when a book with a long holds queue is checked out to the next person and licenses_available doesn't change. In this case, you could fix it because you know there was a checkout event even if the delta is 0.

model.py
```diff
@@ -6616,6 +6620,98 @@ def update_availability(
...
+        elif type == CE.DISTRIBUTOR_LICENSE_ADD:
+            new_licenses_owned += delta
+        elif type == CE.DISTRIBUTOR_LICENSE_REMOVE:
+            new_licenses_owned = deduct(new_licenses_owned)
+        elif type == CE.DISTRIBUTOR_AVAILABILITY_NOTIFY:
+            new_patrons_in_hold_queue = deduct(new_patrons_in_hold_queue)
+            new_licenses_reserved += delta
```
aslagle: Would licenses_available also change in this case? Or would the previous checkin event have had delta 0?

leonardr:
Should I send out an analytics event even for an event that is ignored because it's out of order?

leonardr:
This new branch should address everything except the "send analytics events for ignored events" issue.

In a lot of these cases there isn't an answer that's always right, which may be why this code was (surprisingly) never written before, so I just try to make a best guess.

model.py
```diff
@@ -6616,6 +6618,126 @@ def update_availability(
...
+                # available.  Checking in more books than there are
+                # patrons in the hold queue _does_ make books
+                # available.  However, in neither case do patrons
+                # leave the hold queue. That will happen in the near
+                # future as DISTRIBUTOR_AVAILABILITY_NOTIFICATION events 
+                # are sent out.
+                if delta > new_patrons_in_hold_queue:
+                    new_licenses_available += (delta-new_patrons_in_hold_queue)
+        elif type == CE.DISTRIBUTOR_CHECKOUT:
+            new_licenses_available = deduct(new_licenses_available)
```
aslagle: Is there a case when this changes licenses_reserved instead of licenses_available?

aslagle:
If an event is out of order because we've already processed it and we got the same event again, we wouldn't want to log a duplicate. But if the event is new but somehow ended up after a later event, I think we would still want to log it.

jce1028:
Late to the conversation but Tom Mercer of Bibliotheca reached out and asked if there was anything we needed.  For this situation is there a small change on there side that you could contemplate that would assist?

leonardr:
OK, I've changed it to always send an analytics event even if the circulation event itself is ignored.

leonardr:
@jce1028 We're seeing that a log of circulation events is not an audit log. You can't take a prior state, play back the event log, and reliably get the current state.

If I could change the API, the simplest change I'd make would be to record the current state of the license after an event happens, and to provide that information as part of the event record. This would let us separate the two pieces of code -- the piece that knows what just happened in business terms and the piece that (currently) has to guess as to what effect it has on the collection.

Up to this point, the piece of code that knows what events have happened has been separated from the piece of code that understands the current state of the collection. This branch gives the "what events have happened" code some heuristics for guessing at the current state of the collection, but it's still necessary to just go through all the books periodically and make sure we got it right.

PR #643, opened 2017-09-12
Run Python migration scripts after SQL migration scripts.

leonardr:
I tested this by migrating a dump of the Open Ebooks database and it worked well.


leonardr:
In that hypothetical scenario we'll have to do the migration in two steps, but we would have had to do it in two steps anyway.

PR #644, opened 2017-09-13
Made the Explain script testable and added a basic test.

leonardr:
This branch fixes a bug in `Explain` which embedded the pre-2.0 assumption that an Edition only has one associated LicensePool.

Most of the code is devoted to making the `Explain` script testable and providing it with basic test coverage.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -2164,7 +2164,7 @@ class Explain(IdentifierInputScript):
         policy = None
         for edition in editions:
             self.explain(self._db, edition, policy)
-            self.stdout.write("-" * 80)
+            self.write("-" * 80)
 
     def write(self, s):
         """Write a string to self.stdout."""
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -2153,51 +2153,60 @@ def register_problem(cls, source, contribution, computed_sort_name, error_messag
...
+                self.explain_contribution(contribution)
 
         # Tell about the LicensePool.
-        lp = edition.license_pool
-        if lp:
-            cls.explain_license_pool(lp)
+        lps = edition.license_pools
+        if lps:
+            for lp in lps:
+                self.explain_license_pool(lp)
```
leonardr: Here's the actual bug fix.

scripts.py
```diff
@@ -2153,51 +2153,60 @@ def register_problem(cls, source, contribution, computed_sort_name, error_messag
...
         editions = self._db.query(Edition).filter(
             Edition.primary_identifier_id.in_(identifier_ids)
         )
+        self.stdout = stdout
 
         policy = None
         for edition in editions:
             self.explain(self._db, edition, policy)
-            print "-" * 80
+            self.stdout.write("-" * 80)
```
aslagle: This could be self.write.

PR #651, opened 2017-09-21
Record a transient metadata upload failure when applying metadata that should be sent to the metadata wrangler.

aslagle:
This branch is the core component of https://github.com/NYPL-Simplified/circulation/issues/686.

Instead of making metadata upload part of Metadata.apply(), Metadata.apply creates a transient failure CoverageRecord indicating that the upload needs to be done, so a CoverageProvider can take care of it later.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -1592,10 +1592,16 @@ class Metadata(MetaToModelUtility):
             # Create a transient failure CoverageRecord for this edition
             # so it will be processed by the MetadataUploadCoverageProvider.
             internal_processing = DataSource.lookup(_db, DataSource.INTERNAL_PROCESSING)
-            CoverageRecord.add_for(edition, internal_processing,
-                                   operation=CoverageRecord.METADATA_UPLOAD_OPERATION,
-                                   status=CoverageRecord.TRANSIENT_FAILURE,
-                                   collection=collection)
+
+            # If there's already a CoverageRecord, don't change it to transient failure.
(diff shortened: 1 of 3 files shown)


metadata_layer.py
```diff
@@ -1576,6 +1576,26 @@ def _key(classification):
...
+            DataSource.AXIS_360,
+        ]
+        if made_core_changes and (not data_source.integration_client) and (data_source.name not in METADATA_UPLOAD_BLACKLIST):
+            # Create a transient failure CoverageRecord for this edition
+            # so it will be processed by the MetadataUploadCoverageProvider.
+            internal_processing = DataSource.lookup(_db, DataSource.INTERNAL_PROCESSING)
+            CoverageRecord.add_for(edition, internal_processing,
+                                   operation=CoverageRecord.METADATA_UPLOAD_OPERATION,
+                                   status=CoverageRecord.TRANSIENT_FAILURE,
+                                   collection=collection)
```
leonardr: If there already is a successful CoverageRecord for this action, `add_for()` will switch it back from success to transient failure. I feel like this is a shortcoming of add_for() rather than a problem with the new code.

aslagle: Okay, I was thinking we wanted to re-sync for every change.

leonardr: In theory, we do, I'm just concerned about performance and I think it's better to take it slow since right now we really only _need_ this for cover images.

PR #655, opened 2017-09-25
Change log files

courte:
These changes allow `bin/run` to be used for any script in the `bin/` directory of a parent application (e.g. circulation, content_server) and not just those in the first level. It will log output and create a pid file to assist with tracking long-running `repair/` scripts, for example.

It also changes the directory where logs and pid files are held. I'm trying to keep the system consistent in using "simplified" to reference app-related variables and files.

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -9,9 +9,14 @@
 # /bin directory. It assumes your LIBSIMPLE_DIR is /var/www/circulation
 # unless you set the environment variable otherwise.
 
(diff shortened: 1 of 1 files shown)


courte:
Not ready for review yet. This change coincide with changes to circulation-docker, so it'll need that submodule update, too.

courte:
Actually, I was wrong. This is good for a look-see.

PR #659, opened 2017-09-26
Add data_source to MW mirrored collections

courte:
This is the other side of the endpoint changes made in NYPL-Simplified/metadata_wrangler#127,
sending a data_source name for OPDS_IMPORT collections, so their identifiers can be looked up.

### DIFF
--- a/tests/test_opds_import.py
+++ b/tests/test_opds_import.py
@@ -1,5 +1,6 @@
 import os
 import datetime
+import urllib
 from StringIO import StringIO
 from lxml import builder
 from nose.tools import (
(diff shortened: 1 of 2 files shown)


tests/test_opds_import.py
```diff
@@ -143,6 +166,11 @@ def test_lookup_endpoint(self):
         lookup.collection = self.collection
         eq_('lookup', lookup.lookup_endpoint)
 
+        # With authentication and a collection, a specific endpoint is returned.
```
courte: This is a 'success' case for this test.

PR #660, opened 2017-09-26
Move log configuration into the database

leonardr:
This branch replaces `Configuration.log_policy` with `LogConfiguration.initialize`, a method that configures (or reconfigures) the Python logging system based on the current database configuration.

If there is no database configuration to be found, we use one set of defaults when running unit tests, and another set of defaults everywhere else.

`LogConfiguration.initialize` is automatically called from the unit test runner and from `production_session`. This ensures that logging is properly configured as soon as possible. The applications may need to be changed, but hopefully it will only be to remove code (and to add configuration settings to the admin interface).

### DIFF
--- a/log.py
+++ b/log.py
@@ -115,11 +115,13 @@ class LogConfiguration(object):
         be ignored in favor of a known test-friendly policy. (It's
         okay to pass in False during a test *of this method*.)
 
-        :return: A 3-tuple (handlers, database_log_level). `handlers`
-        is a list of Handler objects that will be associated with the
-        top-level logger. database_log_level is the log level to be
-        applied to the loggers for the database connector and other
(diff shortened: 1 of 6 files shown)


log.py
```diff
@@ -48,50 +47,181 @@ def format(self, record):
...
+
+        :param _db: A database connection. If None, the default
+        logging policy will be used.
+
+        :param testing: A boolean indicating whether a unit test is
+        happening right now. If True, the database configuration will
+        be ignored in favor of a known test-friendly policy. (It's
+        okay to pass in False during a test *of this method*.)
+
+        :return: A 2-tuple (handlers, database_log_level). `handlers`
```
aslagle: This comment is wrong, it's a 3-tuple.

PR #661, opened 2017-09-26
An image can be its own thumbnail

leonardr:
This branch adds support for a case (as with Enki) where the 'full-sized' image is about the size of a thumbnail. By setting an image as its own thumbnail, you can start showing an image immediately without needing to get a properly scaled-down image from the metadata wrangler.

This required a minor change to the model (but not the database schema) to avoid a CircularDependencyError when Representation.thumbnail_of pointed to itself.

A minor addition that is nice: if you give `apply` a Hyperlink that has a thumbnail, you don't also need to include that `Hyperlink` in the `links` list. The `apply` method will pick it up and (if it really is a thumbnail-type link) create it automatically.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -410,6 +410,22 @@ class LinkData(object):
         # rather than each edition.
         self.rights_uri = rights_uri
 
+    @property
+    def guessed_media_type(self):
+        """If the media type of a link is unknown, take a guess."""
+        if self.media_type:
+            # We know.
+            return self.media_type
+
+        if self.href:
+            # Take a guess.
+            return Representation.guess_media_type(self.href)
+
+        # No idea.
+        # TODO: We might be able to take a further guess based on the
+        # content and the link relation.
+        return None
+
     def __repr__(self):
         if self.content:
             content = ", %d bytes content" % len(self.content)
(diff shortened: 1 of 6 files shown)


metadata_layer.py
```diff
@@ -1508,6 +1508,7 @@ def _key(classification):
                 identifier.links = surviving_hyperlinks
         
         link_objects = {}
+        thumbnails = {}
```
aslagle: I don't think you need this.

metadata_layer.py
```diff
@@ -1517,6 +1518,25 @@ def _key(classification):
...
+                    thumbnail = link.thumbnail
+                    thumbnail_obj, ignore = identifier.add_link(
+                        rel=thumbnail.rel, href=thumbnail.href, 
+                        data_source=data_source, 
+                        media_type=thumbnail.media_type,
+                        content=thumbnail.content
+                    )
+                    thumbnail_obj.resource.representation.thumbnail_of = (
+                        link_obj.resource.representation
+                    )
```
aslagle: Looks like this breaking tests....

leonardr:
OK, I think I have the tests working. There's one place where we still have a problem. If `Metadata.apply` creates an image Resource A whose Representation B has a thumbnail representation C.

...and then we try to mirror all the resources, and it turns out that Resource A is of a different media type than we thought...

...then we create a new Resource, D, and a new Representation, E, but we don't update C to point to E instead of to B.

If we could, I'd prefer to update C to point to E instead of to B, but `Representation.get` doesn't have any knowledge of A or B, so it never knows that there are thumbnails that need to be updated.

I don't think this will be a problem in real life because the components that mirror resources also create their own thumbnails, but it's kind of annoying.

PR #663, opened 2017-09-28
Suppress the logging of the sending of logs to loggly, to avoid infin…

leonardr:
…ite loops.


leonardr:
See https://github.com/NYPL-Simplified/library_registry/pull/53

PR #665, opened 2017-09-28
Debuggable response handling

leonardr:
This branch creates an alternate way to make HTTP requests where errors on the other side are passed through in detail rather than hidden under a generic "server integration error" message. This is for use in administrative interfaces and other places where we'd rather know more about a problem than risk freaking someone out with an unfriendly error message.

* If a response yields a problem detail document, it's wrapped in another problem detail document (so that the response code is 502 instead of whatever the original response code was).
* If a response returns a status code outside the 2xx/3xx range, but no problem detail document is returned, the content from the server is reported inside a new problem detail document.
* Otherwise, the response is returned as-is.

### DIFF
--- a/util/http.py
+++ b/util/http.py
@@ -260,20 +260,23 @@ class HTTP(object):
             )
         return response
 
-    def debuggable_get(self, url, **kwargs):
+    @classmethod
+    def debuggable_get(cls, url, **kwargs):
         """Make a GET request that returns a detailed problem
         detail document on error.
         """
-        return self._debuggable_request("GET", url, **kwargs)
+        return cls._debuggable_request("GET", url, **kwargs)
 
-    def debuggable_post(self, url, payload, **kwargs):
+    @classmethod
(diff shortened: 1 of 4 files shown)


util/http.py
```diff
@@ -257,3 +260,56 @@ def _process_response(cls, url, response, allowed_response_codes=None,
...
+        :param response: A Response object from the requests library.
+        """
+        content_type = response.headers.get('Content-Type')
+        if content_type == PROBLEM_DETAIL_JSON_MEDIA_TYPE:
+            return INTEGRATION_ERROR.detailed(
+                _('Remote service returned a problem detail document: %r') % (
+                    response.content
+                )
+            )
+        elif response.status_code / 100 not in (2,3):
```
aslagle: It might be useful to keep the argument for allowed response codes - I could imagine a case where you'd expect a 401 but want to use this in case you got something else.

PR #669, opened 2017-10-04
Add debugging and recovery for rare problems when objects can't be merged in with a database session

leonardr:
The main purpose of this branch is to give better visibility into a rare instance where an object loaded from the cache can't be merged in with the active database session.

As I tried to solve the problem I did a lot of research around the `cascade` option and what it means for merge (and other actions) to cascade. I still don't understand exactly what I'm seeing w/r/t this error, but I learned enough from http://docs.sqlalchemy.org/en/latest/orm/cascades.html to believe we really only need three values for `cascade`:

* The default value, mainly used for many-to-many relationships.
* `all`, for when we want the deletion of a parent to result in the deletion of all its children. 
* `all, delete-orphan`, for when we _also_ want a child object to be deleted rather than removed from its parent.

I went through `model.py` and changed a number of `cascade` values to one of these three, generally `all, delete-orphan`. I have not audited all our `relationship`s to make sure that the default value is appropriate, but I'm pretty sure that whenever we do specify a `relationship` it's now doing what we want.

However, I don't think this has any effect on the problem we're seeing -- it's just a minor bit of cleanup.

The name of this branch is misleading; this code ensures that merge operations _are_ cascaded in more cases than previously.

### DIFF
--- a/model.py
+++ b/model.py
@@ -276,7 +276,7 @@ class HasFullTableCache(object):
                 )
                 # Try to look up a fresh copy of the object.
                 obj, new = lookup_hook()
-                if obj and obj in _db and False:
+                if obj and obj in _db:
                     logging.error("Was able to look up a fresh copy of %r", obj)
                     return obj, new
 
--- a/model.py
+++ b/model.py
@@ -274,6 +274,15 @@ class HasFullTableCache(object):
                     "Unable to merge cached object %r into database session",
                     obj, exc_info=e
                 )
+                # Try to look up a fresh copy of the object.
+                obj, new = lookup_hook()
+                if obj and obj in _db and False:
+                    logging.error("Was able to look up a fresh copy of %r", obj)
+                    return obj, new
+
+                # That didn't work. Re-raise the original exception.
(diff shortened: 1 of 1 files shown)


model.py
```diff
@@ -10122,7 +10133,7 @@ class Collection(Base, HasFullTableCache):
     # A Collection can include many LicensePools.
     licensepools = relationship(
         "LicensePool", backref="collection",
-        cascade="save-update, merge, delete"
+        cascade="all, delete-orphan"
```
leonardr: A LicensePool must be associated with a Collection; it's not optional.

model.py
```diff
@@ -10138,7 +10149,7 @@ class Collection(Base, HasFullTableCache):
     # for Identifiers in its catalog.
     coverage_records = relationship(
         "CoverageRecord", backref="collection",
-        cascade="save-update, merge, delete"
+        cascade="all"
```
leonardr: It's okay for a CoverageRecord not to be associated with any Collection.

PR #671, opened 2017-10-05
Put collection ids in search document

leonardr:
This branch incorporates into a Work's search document the IDs of the Collections in which the Work is currently available. By passing a Library into `query_works` you can restrict results to Works that are in Collections that are in that library.

This branch does not actually pass in that Library argument, so searches work as they did before. This  allows for an interim period where this code is deployed and the search index is rebuilt. If we start passing in Library immediately, there will be a period where search doesn't work at all, while the index is being rebuilt. This period will only be about an hour, but it's not that much more work to avoid the outage.

The other thing that's missing from this branch is some kind of hook that re-indexes a Work whenever necessary. It's now possible for a Work to need re-indexing even if nothing about the Work itself has changed. I'm not sure where this hook would go (an event listener might work, but what would it be listening for?) and I'd like feedback.

### DIFF
--- a/model.py
+++ b/model.py
@@ -4599,6 +4599,23 @@ class Work(Base):
             )
         ).alias('works_alias')
 
+        # This subquery gets Collection IDs.
+        collections = select(
+            [LicensePool.collection_id]
+        ).where(
+            LicensePool.work_id==literal_column(works_alias.name + '.' + works_alias.c.work_id.name
+        ).select_from(
+            LicensePool
+        ).alias("collections_subquery")
+
+        # Create a json array from the set of Collections.
+        collections_json = select(
+            [func.array_to_json(
+                    func.array_agg(
+                        func.row_to_json(
+                            literal_column(collections.id)
+                        )))]
+        ).select_from(collections)
 
         # This subquery gets Contributors, filtered on edition_id.
         contributors = select(
@@ -4745,6 +4762,7 @@ class Work(Base):
              works_alias.c.popularity,
 
              # Here are all the subqueries.
+             collections_json.label("collections"),
              contributors_json.label("contributors"),
              subjects_json.label("classifications"),
              genres_json.label('genres'),
--- a/model.py
+++ b/model.py
@@ -4603,9 +4603,7 @@ class Work(Base):
         collections = select(
             [LicensePool.collection_id]
         ).where(
-            LicensePool.work_id==literal_column(works_alias.name + '.' + works_alias.c.work_id.name
-        ).select_from(
-            LicensePool
+            LicensePool.work_id==literal_column(works_alias.name + '.' + works_alias.c.work_id.name)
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -4599,6 +4599,21 @@ def to_search_documents(cls, works):
...
+        ).where(
+            LicensePool.work_id==literal_column(works_alias.name + '.' + works_alias.c.work_id.name)
+        ).alias("collections_subquery")
+
+        # Create a json array from the set of Collections.
+        collections_json = select(
+            [func.array_to_json(
+                    func.array_agg(
+                        func.row_to_json(
+                            literal_column(collections.name)
```
aslagle: This should be collections.id, not collections.name.

aslagle: Or not, I think this is the name of the subquery and it works because it only has one column.

tests/test_external_search.py
```diff
@@ -863,12 +920,13 @@ def test_query_works_from_lane_definition_handles_languages(self):
...
         )
         filter = search.make_filter(
+            [self._default_collection.id],
             lane.media, lane.languages, lane.exclude_languages,
             lane.fiction, list(lane.audiences), lane.age_range,
             lane.genre_ids,
         )
         
-        languages_filter, medium_filter = filter['and']
+        collection_filter, languages_filter, medium_filter = filter['and']
```
aslagle: You could check the collection_filter's contents as well.

aslagle:
I guess the hook would have to listen for creation or deletion of LicensePools?

model.py
leonardr: I originally thought the same, but this isn't pulling a field out of the 'collections_subquery' alias, it's turning the alias itself (with _all_ of its fields) into a JSON dictionary. `collections.name` here is 'collections_subquery', the name of the subquery.

You'll see below this code that `contributors_json` does something similar. It generates SQL that looks kind of like this toy example:

```
select row_to_json(subquery) from (select sort_name, family_name from contributors limit 1) as subquery;
```

The results:

```
{"sort_name":"Sen, Chaitali","family_name":null}
```

In the toy example, row_to_json() is called on "subquery", not "subquery.sort_name" or "subquery.family_name"

leonardr:
We've decided I'll do a follow-up branch that does the following:

1. Store a collection ID in the search document only if the LicensePool for that Work in that collection is actually available through that collection (open-access or has licenses_owned).
2. Track events that would make a collection ID show up or not show up in a Work's search document, and when they happen, use a CoverageRecord to indicate that the Work needs to be reindexed.
3. Write a new script that runs every few minutes and reindexes all the works that have outstanding CoverageRecords.

Eventually we can make a change where all Works are reindexed in the background in batches. I think this can save a lot of time overall, since it's much more efficient to index 10 works simultaneously than to index 10 works one at a time.

PR #673, opened 2017-10-06
Create a CoverageProvider that makes sure Works stay indexed

leonardr:
This branch introduces a CoverageProvider that makes sure all Works have a WorkCoverageRecord associated with the current name of the search index (which means they have an up-to-date search document in the index).

Works will go into the 'registered' state (indicating that work needs to be done) when they enter or drop out of a Collection. 

A future branch will change `update_external_index` to add the coverage record in the 'registered' state, instead of doing the work and then adding the coverage record in the 'success' state. This should improve performance quite a bit.

When writing tests I noticed some incorrect edge cases in the ExternalSearchIndex error handling, which I think I've fixed. They had to do with what happens when every single document you try to index comes back with an error.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -855,12 +855,8 @@ class SearchIndexCoverageProvider(WorkCoverageProvider):
         return ExternalSearchIndex.search_index_update_operation(self._db)
 
     def process_batch(self, works):
-        """:return: A 2-tuple (counts, records). 
-
-        `counts` is a 3-tuple (successes, transient failures,
-        persistent_failures).
-
-        `records` is a mixed list of Works and CoverageFailure objects.
+        """
+        :return: a mixed list of Works and CoverageFailure objects.
         """
         successes, failures = self.search_index_client.bulk_update(works)
 
--- a/external_search.py
+++ b/external_search.py
@@ -558,7 +558,16 @@ class ExternalSearchIndex(object):
 
         clauses = []
         if collection_ids is not None:
-            clauses.append(dict(terms=dict(collection_id=list(collection_ids))))
+            # Either the collection ID field must be completely
+            # missing (as it will be in older indexes) or it must
+            # include one of the collection IDs we're looking for.
+            collection_id_matches = dict(
+                term=dict(collection_id=list(collection_ids))
+            )
+            no_collection_id = dict(
+                bool=dict(must_not=dict(exists=dict(field="collection_id")))
+            )
+            clauses.append({'or': [collection_id_matches, no_collection_id]})
         if languages:
             clauses.append(dict(terms=dict(language=list(languages))))
         if exclude_languages:
--- a/external_search.py
+++ b/external_search.py
@@ -16,6 +16,10 @@ from model import (
     Work,
     WorkCoverageRecord,
 )
+from coverage import (
+    CoverageFailure,
+    WorkCoverageProvider,
+)
 import os
 import logging
 import re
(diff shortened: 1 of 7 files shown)


external_search.py
```diff
@@ -633,14 +666,15 @@ def bulk_update(self, works, retry_on_batch_failure=True):
...
-        ]
+        def get_error_id(error):
+            return error.get('data', {}).get('_id', None) or error.get('index', {}).get('_id', None)   
+        error_ids = [get_error_id(error) for error in errors]
 
+        missing_works = [
+            work for work in works 
+            if work.id not in doc_ids and work.id not in error_ids
+        ]
+
```
leonardr: Previously we counted a work both as 'error' and as 'missing' if it caused an error.

external_search.py
```diff
@@ -621,9 +651,12 @@ def bulk_update(self, works, retry_on_batch_failure=True):
...
         # If the entire update failed, try it one more time before giving up on the batch.
-        if retry_on_batch_failure and len(errors) == len(docs):
-            self.log.info("Elasticsearch bulk update timed out, trying again.")
-            return self.bulk_update(works, retry_on_batch_failure=False)
+        if len(errors) == len(docs):
+            if retry_on_batch_failure:
+                self.log.info("Elasticsearch bulk update timed out, trying again.")
+                return self.bulk_update(works, retry_on_batch_failure=False)
+            else:
+                docs = []
```
leonardr: Previously, if every document in the update failed for reasons other than a timeout (or if a timeout happened twice), each document was treated both as a success and an error.

leonardr:
When integrating this into the script runner I discovered that the parts that connect to the larger system are slightly wrong, so don't review this one yet.

leonardr:
OK, this is ready for review now.

external_search.py
```diff
@@ -48,6 +56,35 @@ def reset(cls):
...
+        )
+
+    @classmethod
+    def works_index_name(cls, _db):
+        """Look up the name of the search index."""
+        integration = cls.search_integration(_db)
+        if not integration:
+            return None
+        setting = integration.setting(cls.WORKS_INDEX_KEY)
+        return setting.value_or_default(cls.DEFAULT_WORKS_INDEX)
```
aslagle: I'm not sure about this. I think the index name in the database is usually the base index name, without a version number, and in some cases could be an alias instead. The code at the end of __init__ uses this setting to determine the actual versioned index name and the alias. We'd want to use the versioned index name in coverage records.

external_search.py
```diff
@@ -799,3 +833,39 @@ def bulk(self, docs, **kwargs):
...
+        )
+
+    def get_operation(self):
+        return ExternalSearchIndex.search_index_update_operation(self._db)
+
+    def process_batch(self, works):
+        """:return: A 2-tuple (counts, records). 
+
+        `counts` is a 3-tuple (successes, transient failures,
+        persistent_failures).
```
aslagle: Did you take out this part of the return value?

model.py
```diff
@@ -10730,6 +10748,50 @@ def site_configuration_has_changed(_db, timeout=1):
...
+
+    This shouldn't ever happen.
+    """
+    work = target.work
+    if not work:
+        return
+    if value == oldvalue:
+        return
+    work.external_index_needs_updating()
+
```
aslagle: one other thing that should never happen is a license pool changing its collection.

leonardr:
After an in-person discussion we decided to revisit the issue of index versioning later. For now, I've made the search index backwards-compatible with documents created using the earlier code.

Since we only use one index at a time, I've also changed the coverage provider to use a generic operation instead of a versioned operation. Whenever we make backwards-incompatible changes for any reason (whether in a migration script or in code), we'll need to delete all existing CoverageRecords with that operation so that the entire index will be rebuilt.

PR #677, opened 2017-10-11
Implemented and tested WorkCoverageRecord.bulk_add.

leonardr:
This branch implements a new method which takes a list of Works and makes sure every work has an up-to-date WorkCoverageRecord, whether that means inserting a new row into the database or updating an old row.

### DIFF
--- a/model.py
+++ b/model.py
@@ -1555,9 +1555,13 @@ class WorkCoverageRecord(Base, BaseCoverageRecord):
         # Make sure that works that previously had a
         # WorkCoverageRecord for this operation have their timestamp
         # and status updated.
+        #
+        # We always set exception to None because if there are
+        # individual differences between the records being created,
+        # they need to be created one at a time and not in a batch.
         update = WorkCoverageRecord.__table__.update().where(
             Work.id.in_(work_ids)
-        ).values(timestamp=timestamp, status=status)
+        ).values(dict(timestamp=timestamp, status=status, exception=None))
         _db.execute(update)
 
 
--- a/model.py
+++ b/model.py
@@ -73,9 +73,11 @@ from sqlalchemy.sql.expression import (
     or_,
     select,
     join,
+    literal,
     literal_column,
     case,
     table,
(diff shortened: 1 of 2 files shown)


tests/test_model.py
```diff
@@ -5727,6 +5727,64 @@ def test_add_for(self):
...
+            not_affected, irrelevant_operation, 
+            status=WorkCoverageRecord.SUCCESS
+        )
+
+        # Tell bulk_add to update or create WorkCoverageRecords for
+        # not_already_covered and already_covered, but not not_affected.
+        new_timestamp = datetime.datetime.utcnow()
+        new_status = WorkCoverageRecord.REGISTERED
+        WorkCoverageRecord.bulk_add(
+            [not_already_covered],
```
aslagle: how did already_covered get updated when it's not in here?

leonardr: Uh-oh.

leonardr:
I made the test a lot more thorough and fixed some bugs.

model.py
```diff
@@ -1497,6 +1498,70 @@ def add_for(self, work, operation, timestamp=None,
...
+        _db.execute(insert)
+
+        # Make sure that works that previously had a
+        # WorkCoverageRecord for this operation have their timestamp
+        # and status updated.
+        update = WorkCoverageRecord.__table__.update().where(
+            and_(WorkCoverageRecord.work_id.in_(work_ids),
+                 WorkCoverageRecord.operation==operation)
+        ).values(dict(timestamp=timestamp, status=status, exception=exception))
+        _db.execute(update)
```
aslagle: Does doing the insert before the update mean that the new records are updated unnecessarily? Maybe it doesn't make a difference, but doing the update first could be faster.

leonardr:
I agree it doesn't make a big difference; I was thinking about the case where two processes run this method simultaneously and the SQL calls are interleaved. But even then I guess it doesn't matter which timestamp the records end up with.

PR #678, opened 2017-10-11
Run ALTER TYPE migrations without a transaction

courte:
*\~quietly sobbing into elbow~*

I had to break out of SQLAlchemy to accurately update an enum. I also scraped out / reused some old code from those good ol' 2.0 migration days.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1901,18 +1901,52 @@ class DatabaseMigrationScript(Script):
                 transactionless = any(filter(
                     lambda c: c in sql.lower(), self.TRANSACTIONLESS_COMMANDS
                 ))
-                if not transactionless:
+                if transactionless:
+                    new_session = self._run_migration_without_transaction(sql)
+                else:
                     # By wrapping the action in a transation, we can avoid
                     # rolling over errors and losing data in files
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -1901,19 +1901,52 @@ def _run_migration(self, migration_path, timestamp):
...
+        # database: no transactions, no guardrails.
+        for command in sql_commands:
+            connection.execution_options(isolation_level='AUTOCOMMIT')\
+                .execute(text(command))
+
+        # Update the script's Session to a new one that has the changed schema
+        # and other important info.
+        self._session = Session(connection)
+        self.load_configuration()
+        DataSource.well_known_sources(self._db)
```
leonardr: Is this enough or should you go all out and call initialize_data?

courte: I'm not sure on a deep leve, but in testing `initialize_data` raised an error because the MaterializedViews already exist.

leonardr: OK, that makes sense. Let's see what happens.

courte: Oh, that was actually the initialization of the engine and the collection. I'll try this.

PR #680, opened 2017-10-12
Only put presentation-ready works into the index

leonardr:
All code that used to immediately update a work's index entry now registers the work to be done later by the SearchIndexCoverageProvider. This should improve performance of nearly all scripts that run on the script runner.

I copied the SearchIndexMonitor and its tests in from circulation; I figure it should be in the same file as the SearchIndexCoverageProvider. The code is the same, but I changed the SearchIndexMonitor to use `WorkCoverageRecord.bulk_add` rather than adding WorkCoverageRecords one at a time.

Works are removed from the index in `bulk_update` if they're not presentation-ready. The removal of a work from an index is now treated as a success, where the failure to index a non-presentation-ready work used to be an error. My rationale is that it's a success because the state of the work has been accurately mirrored to the index. I'm open to treating them separately, but we need `SearchIndexMonitor` to treat them as successes.

The `SearchIndexCoverageProvider` uses `bulk_update` to update all the works it successfully covered at once, rather than one at a time. This is where we'll see the big performance improvements. This required some minor refactoring of the `BaseCoverageProvider`. The existing tests validate the refactoring, but there's no new test that specifically checks that we're doing bulk updates.


aslagle:
It makes sense to me that deleting non-presentation-ready works counts as success. I think I made those fail so it would make another attempt to index them, in case they became presentation-ready later. The code that changes presentation-ready status should update search index CoverageRecords in order to handle that.

external_search.py
```diff
@@ -634,7 +635,21 @@ def bulk_update(self, works, retry_on_batch_failure=True):
         """Upload a batch of works to the search index at once."""
 
         time1 = time.time()
-        docs = Work.to_search_documents(works)
+        needs_add = []
+        needs_delete = []
```
aslagle: you didn't use needs_delete

external_search.py
```diff
@@ -634,7 +635,21 @@ def bulk_update(self, works, retry_on_batch_failure=True):
...
+        successes = []
+        for work in works:
+            if work.presentation_ready:
+                needs_add.append(work)
+            else:
+                # Works are removed one at a time, which shouldn't
+                # pose a performance problem because works almost never
+                # stop being presentation ready.
```
aslagle: It is possible to do a bulk delete if we ever need to, but this is fine.

PR #681, opened 2017-10-13
Take the site configuration timeout from a sitewide setting.

leonardr:
Our circulation managers are still having mysterious errors that seem to coincide with the time the site configuration is being reloaded, compounded by another mysterious problem that is causing it to appear as though the site configuration is constantly changing. Until we can diagnose these errors I'd like the site configuration to be reloaded as infrequently as possible, but I don't want to have to change core and circulation every time I change the timeout value.

This branch creates a sitewide setting which is used to determine how often the site configuration is reloaded.

Hopefully we won't need a user interface for this setting because it will become unnecessary once we fix the underlying problems. I'd like to go back to always checking once a minute.

### DIFF
--- a/config.py
+++ b/config.py
@@ -341,6 +341,10 @@ class Configuration(object):
     # changed.
     LAST_CHECKED_FOR_SITE_CONFIGURATION_UPDATE = "last_checked_for_site_configuration_update"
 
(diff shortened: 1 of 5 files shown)


leonardr:
The change exposed a problem and some duplicated code in the site-wide configuration settings script, which I've fixed.

PR #683, opened 2017-10-18
Store lane configuration in database

leonardr:
This branch replaces the old `Lane` class with a new `Lane` class that's an ORM object. The features of the old lanes are still available, with the following exceptions:

1. Individual lanes are no longer explicitly set as 'searchable'. All lanes are searchable so long as they are not limited to a genre or custom list. Lanes that are the root lane for one or more patron types are also searchable, no matter what. (This can cause problems if a lane based on a custom list is also the root lane for a patron type -- I don't think this will work, but I also don't think it wlil happen in real life.)
2. It's no longer possible to define a lane in terms of the languages _not_ covered by that lane.
3. It is now possible to define a Lane as recursively including some Genres (i.e. those Genres and their subgenres) but recursively excluding other Genres. This currently has no real effect beyond what was possible before, because our genre structure only goes two levels deep.
4. The purpose of lane visibility has changed. Previously, invisible but searchable lanes were used to get the desired search behavior. Now, a lane is made visible or invisible to hide it from or show it to library patrons. The behavior has also changed: the children of an invisible lane are now always invisible themselves.
5. Previously, a constructor setting allowed a lane to inherit its parent's behavior when it was created. Now, this behavior is controlled by a database field, `inherit_parent_restrictions`. If this is set, a Work will only show up in a Lane if it would have _also_ shown up in the parent lane. This works even when one or both Lanes are based on CustomList membership.

The `LaneList` class has been removed, as have the various subclasses of `Lane`. Both of these are replaced by the `WorkList` class, a generic implementation of "something that can obtain a bunch of Works on demand, and also might have children that are also `WorkList`s". The `Lane` class uses `WorkList` as a mixin.

Other notable changes:

1. The code that runs the core "find a bunch of Works" query no longer supports `Work` as its `work_model`. Now you have to use a materialized view. Although the `Work` feature was originally how this whole thing worked before we created materialized views, at this point it  was only used in tests and it made the code significantly more complicated.
2. When choosing which books to feature for a grouped feed representation of a lane based on a CustomList, we privilege books that are featured _on the CustomList_ over books that are of 'featurable' quality but not featured _on that particular list_. So a "bad" book might be featured over a "good" book if the librarian decides this is that kind of list.

This is the basis for a new feature branch, so I won't be merging it immediately after the PR is approved.


lane.py
```diff
@@ -189,78 +216,34 @@ def dy(new_value):
...
+            cls.ORDER_SERIES_POSITION : work_model.series_position,
             cls.ORDER_RANDOM : work_model.random,
         }
         return order_facet_to_database_field[order_facet]
 
-    @classmethod
-    def database_field_to_order_facet(cls, database_field):
-        """The inverse of order_facet_to_database_field.
-
-        TODO: This method may not be necessary.
```
leonardr: NARRATOR: It wasn't.

lane.py
```diff
@@ -387,897 +372,796 @@ def has_next_page(self):
...
-            q = facets.apply(self._db, q, work_model, edition_model,
-                             distinct=distinct)
-        if pagination:
-            q = pagination.apply(q)
+        :return: A list of genre IDs, or None if this Lane does not
+        consider genres at all.
+        """
+        if not hasattr(self, '_genre_ids'):
+            self._genre_ids = self._gather_genre_ids()
+        return self._genre_ids
```
leonardr: I'm concerned that a change to the WorkGenres for a lane doesn't force this to be recalculated. I'll probably need to add an event listener.

lane.py
```diff
@@ -1306,364 +1199,147 @@ def search(self, query, search_client, pagination=None):
...
-                    )
-
-        if not results:
-            logging.debug("No elasticsearch results, falling back to database query")
-            results = self._search_database(query).limit(pagination.size).offset(pagination.offset).all()
-        return results
+                    results = self.works_for_specific_ids(_db, doc_ids)
 
-    def _search_database(self, query):
-        """Do a really awful database search for a book using ILIKE.
```
leonardr: This whole thing is gone.

lane.py
```diff
@@ -387,897 +372,796 @@ def has_next_page(self):
...
-        Basically, forward slash is changed to "__". This is necessary
-        because Flask tries to route "feed/Suspense%2FThriller" to
-        feed/Suspense/Thriller.
+    @property
+    def visible_children(self):
+        """A WorkList's children can be used to create a grouped acquisition
+        feed for that WorkList.
         """
-        return self.name.replace("/", "__")
+        return self.children
```
aslagle: Is self.children only visible children? It's not all children?

lane.py
```diff
@@ -387,897 +372,796 @@ def has_next_page(self):
...
+        yield (Facets.COLLECTION_FEATURED, Facets.AVAILABLE_NOW, False)
+        yield (Facets.COLLECTION_FEATURED, Facets.AVAILABLE_ALL, False)
+
+        # The 'main' collection contains everything except low-quality
+        # open-access works, 
+        yield (Facets.COLLECTION_MAIN, Facets.AVAILABLE_NOW, False)
+        yield (Facets.COLLECTION_MAIN, Facets.AVAILABLE_ALL, False)
+
+        # The 'full' collection contains absolutely everything.
+        yield (Facets.COLLECTION_FULL, Facets.AVAILABLE_ALL, False)
```
aslagle: Why is featured always false in these? It seems like you could start with true and change to false if you get more desperate. Are you thinking that's something that will be configurable when you add a list to the lane?

lane.py
```diff
@@ -387,897 +372,796 @@ def has_next_page(self):
...
+    id = Column(Integer, primary_key=True)
+    lane_id = Column(Integer, ForeignKey('lanes.id'), index=True,
+                     nullable=False)
+    genre_id = Column(Integer, ForeignKey('genres.id'), index=True,
+                      nullable=False)
+
+    # An inclusive relationship means that books classified under the
+    # genre are included in the lane. An exclusive relationship means
+    # that books classified under the genre are excluded, even if they
+    # would otherwise be included.
```
aslagle: Does this mean that if a lane has a recursive exclusive relationship with a genre, any inclusive relationships with its subgenres will be ignored?

lane.py
```diff
@@ -387,897 +372,796 @@ def has_next_page(self):
...
+        nullable=True
+    )
 
-        Works will:
+    # Only books on one or more CustomLists obtained from this
+    # DataSource will be shown.
+    _list_datasource_id = Column(
+        Integer, ForeignKey('datasources.id'), index=True,
+        nullable=True
+    )
```
aslagle: What will this be used for?

aslagle: I'm guessing NYT, but that's one list so I'm wondering why it would be associated with the data source instead.

lane.py
```diff
@@ -387,897 +372,796 @@ def has_next_page(self):
...
 
-        * Be filed under of the genres listed in `self.genre_ids` (or, if
-          `self.include_subgenres` is True, any of those genres'
-          subgenres).
+    # This has no effect unless list_datasource_id or
+    # list_identifier_id is also set. If this is set, then a book will
+    # only be shown if it was seen on an appropriate list within this
+    # number of days. If the number is zero, the book must be
+    # _currently_ on an appropriate list.
+    list_seen_in_previous_days = Column(Integer, nullable=True)
```
aslagle: I don't understand what it means to be "seen" on a list here.

lane.py
```diff
@@ -387,897 +372,796 @@ def has_next_page(self):
...
+
+    # If this is set to True, then a book will show up in a lane only
+    # if it would _also_ show up in its parent lane.
+    #
+    # Currently this has no effect unless list_datasource_id or
+    # list_identifier_id is also set.
+    inherit_parent_restrictions = Column(Boolean, default=False, nullable=False)
```
aslagle: I'd like to discuss this option tomorrow.

lane.py
leonardr: WorkLists in general don't use CustomLists, so all that code is in Lane. I've added an explanation of this to the docstring, but I could also move the Lane code into here if I also move `Lane.customlists`. What do you think?

lane.py
leonardr: That's right. I've added a test for this.

lane.py
leonardr: NYT has about 30 different best-seller lists (paperback fiction, hardcover nonfiction, romance, etc.)

lane.py
leonardr: This is specific to CustomLists that monitor lists from external sources such as the NYT best-seller list. When a book leaves a best-seller list it's not removed from the corresponding CustomList. Instead, we stop updating its `.most_recent_appearance`. `list_seen_in_previous_days` lets us create a 'best-sellers' lane containing books that are _currently_ on the NYT best-seller list as well as books that were _recently_ on the best-seller list.

I've rewritten the comment in a way that hopefully clarifies things.

lane.py
leonardr: Sounds good. I believe the "currently this has no effect" comment is wrong so I took it out.

leonardr:
I think we could replace the whole idea of running the same query multiple times with different facets, with a single query that calculates a score for each work and orders by that score.

lane.py
```diff
@@ -387,897 +372,812 @@ def has_next_page(self):
...
         """
+        seen = set()
+        if not self.parent:
+            return
+        yield self.parent
+        for parent in self.parent.parentage:
+            if parent in seen:
+                raise ValueError("Lane parentage loop detected")
+            seen.add(parent)
+            yield parent
```
aslagle: should `self` and `self.parent` also be in `seen`?

lane.py
```diff
@@ -387,897 +372,812 @@ def has_next_page(self):
...
-            q = q.filter(work_model.fiction==None)
-        elif self.fiction != self.BOTH_FICTION_AND_NONFICTION:
-            q = q.filter(work_model.fiction==self.fiction)
+        If you set target_age to 16-18, you're saying that the audiences
+        are [Young Adult, Adult].
 
-        if self.media:
-            q = q.filter(edition_model.medium.in_(self.media))
+        If you set target_age 12-15, you're saying that the audiences are
+        [Adult, Children].
```
aslagle: not `[Young Adult, Children]`?

lane.py
```diff
@@ -387,897 +372,812 @@ def has_next_page(self):
...
+        if isinstance(value, int):
+            value = (value, value)
+        if isinstance(value, tuple):
+            value = tuple_to_numericrange(value)
+        self._target_age = value
+        if value.upper >= 18:
+            audiences.append(Classifier.AUDIENCE_ADULT)
+        if value.lower < Classifier.YOUNG_ADULT_AGE_CUTOFF:
+            audiences.append(Classifier.AUDIENCE_CHILDREN)
+        if value.upper >= Classifier.YOUNG_ADULT_AGE_CUTOFF:
```
aslagle: Should this also check that value.upper <= 18, or prevent setting target age > 18?

lane.py
```diff
@@ -1286,17 +1186,26 @@ def search(self, query, search_client, pagination=None):
                 # TODO: once we migrate all existing search indexes,
                 # the first argument should become the Library associated
                 # with this Lane.
```
aslagle: Not related to this PR, but this TODO could probably be done now?

lane.py
```diff
@@ -1306,364 +1215,127 @@ def search(self, query, search_client, pagination=None):
...
-        for sublane in self.visible_sublanes:
-            works = sublane.featured_works(
-                use_materialized_works=use_materialized_works
+        parent_distinct = False
+        if self.parent and self.inherit_parent_restrictions:
+            # In addition to the other restrictions imposed by this
+            # Lane, books will show up here only if they would
+            # also show up in the parent Lane.
+            qu, parent_distinct = self.parent.apply_bibliographic_filters(
+                _db, qu, work_model, featured
```
aslagle: I think I understand what this is doing, but the relationship between `apply_bibliographic_filters` and `apply_custom_filters` is confusing. `apply_bibliographic_filters` calls `apply_custom_filters`, but then `apply_custom_filters` calls `apply_bibliographic_filters` on the parent lane. I'm not sure why there are two methods.

lane.py
```diff
@@ -1306,364 +1215,127 @@ def search(self, query, search_client, pagination=None):
...
+            audience_has_no_target_age = False
+
+        # The lane's target age is an inclusive NumericRange --
+        # set_target_age makes sure of that. The work's target age
+        # must overlap that of the lane.
+        qu = qu.filter(
+            or_(
+                work_model.target_age.overlaps(self.target_age),
+                audience_has_no_target_age
+            )
```
aslagle: What happens if the audience is YA or children, but there's no target age?

opds.py
```diff
@@ -1001,7 +982,7 @@ def add_breadcrumbs(self, lane, annotator, include_lane=False):
             )
             
             # Add links for all visible ancestors that aren't root
-            for ancestor in reversed(lane.visible_ancestors()):
+            for ancestor in lane.parentage:
```
aslagle: I think this still needs to be reversed - the order of the breadcrumbs is important.

tests/test_external_search.py
```diff
@@ -945,38 +939,49 @@ class TestSearchFilterFromLane(DatabaseTest):
...
         filter = search.make_filter(
             collection_ids,
-            lane.media, lane.languages, lane.exclude_languages,
-            lane.fiction, list(lane.audiences), lane.age_range,
+            lane.media, lane.languages,
+            lane.fiction, list(lane.audiences), lane.target_age,
             lane.genre_ids,
         )
-        collection_filter, medium_filter = filter['and']
+        [collection_filter] = filter['and']
```
aslagle: why did this change?

tests/test_lane.py
```diff
@@ -165,45 +155,48 @@ def test_order_by(self):
...
                 order_ascending=ascending,
             )
-            return f.order_by(work, edition)[0]
+            return f.order_by(work)[0]
 
         def compare(a, b):
             assert(len(a) == len(b))
             for i in range(0, len(a)):
+                if not a[i].compare(b[i]):
+                    set_trace()
```
aslagle: you've got a leftover set_trace here

tests/test_lane.py
```diff
@@ -296,986 +293,1083 @@ def facetify(collection=Facets.COLLECTION_FULL,
...
+        self._works = []
+        self.works_calls = []
+
+    def queue_works(self, works):
+        """Set the next return value for works()."""
+        self._works.append(works)
+
+    def queue_featured_works(self, works):
+        """Set the next return value for featured_works()."""
+        self._featured_works.append(works)
```
aslagle: I thought this class didn't mock featured_works?

tests/test_lane.py
```diff
@@ -296,986 +293,1083 @@ def facetify(collection=Facets.COLLECTION_FULL,
...
+        gutenberg_only.license_datasource = DataSource.lookup(
+            self._db, DataSource.GUTENBERG
+        )
+
+        distinct = match_works(gutenberg_only, [nonfiction])
+        # No custom list is involved, so there's no need to make the query
+        # distinct.
+        eq_(False, distinct)
+
+        # A lane may
```
aslagle: Looks like you forgot to check the age range filter for this lane...

opds.py
leonardr: I took it out because I thought `parentage` returned the parents in the right order but you're right, it doesn't.

tests/test_external_search.py
leonardr: Previously, if you created a generic Lane it had a restriction to the ebook medium by default. That's no longer the case -- there's no medium restriction unless you specify one.

tests/test_lane.py
leonardr: Well, you've certainly got me there...

lane.py
leonardr: Nothing happens in `apply_age_range_filter` because it returns early if `self.target_age == None`. But in `WorkList.apply_audience_filter` the query is restricted to the appropriate audiences.

lane.py
leonardr: Yes, I think so.

tests/test_lane.py
leonardr: I'm not sure what this is referring to. There's no age range or audience filter in `fiction_lane`, `nonfiction_lane`, or `both_lane`, so they all find both children's and adult books.

aslagle: This comment ended up on the wrong line :(

I'll try to find it again.

lane.py
leonardr: The subclasses of `WorkList` in the circulation manager, like `SeriesLane`, need to override `apply_custom_filters` but leave `apply_bibliographic_filters` alone. I think I can replace this with one method where the subclasses call `super()` though.

tests/test_lane.py
```diff
@@ -296,986 +291,1086 @@ def facetify(collection=Facets.COLLECTION_FULL,
...
+        fiction_lane = self._lane()
+        fiction_lane.fiction = True
+        match_works(fiction_lane, [childrens_fiction])
+
+        nonfiction_lane = self._lane()
+        nonfiction_lane.fiction = False
+        match_works(nonfiction_lane, [nonfiction])
+
+        both_lane = self._lane()
+        bo
```
aslagle: This is where comment https://github.com/NYPL-Simplified/server_core/pull/683#discussion_r145752977 was supposed to be.

The line is `younger_ya.target_age = (12,13)` in case it moves again.

You create this lane but you don't test it like you do with the older_ya lane.

aslagle: It did move again :angry:

aslagle: It should be on line 1287 - github is off by about 100 lines for some reason.

leonardr: You're right, that test is just absent.

lane.py
```diff
@@ -387,916 +375,852 @@ def has_next_page(self):
...
-            age_range = [age_range]
+    def apply_audience_filter(self, _db, qu, work_model):
+        """Make sure that only Works classified under this lane's
+        allowed audiences are returned.
+        """
+        if not self.audiences:
+            return qu
+        qu = qu.filter(work_model.audience.in_(self.audiences))
+        if (Classifier.AUDIENCE_CHILDREN in self.audiences
+            or Classifier.AUDIENCE_YOUNG_ADULT in self.audiences):
```
aslagle: @leonardr Shouldn't this hack be applied if adult is not in the audiences, rather than if children or ya is?

If the WorkList is only for children or ya, you want to filter out the Gutenberg books, but if it's for any audience, they can stay.

leonardr: It's pretty theoretical, but I'm concerned about a scenario like the Open Ebooks "High School" lane, which has audience=(Young Adult, Adult) but which still shouldn't have these books.

migration/20171026-change-cachedfeeds-to-use-lane-id.sql
```diff
@@ -0,0 +1 @@
+drop table cachedfeeds;
```
aslagle: we should update the date of this migration before merging.

PR #685, opened 2017-10-19
Lane and LaneGenre modifications are configuration-altering events.

leonardr:
This branch adds listeners so that modifications to Lane and LaneGenre objects are treated as configuration changes.


leonardr:
That was my original plan, but I don't think it's necessary, since the Lane objects themselves will be blown away and recreated when this happens.

PR #686, opened 2017-10-19
Fix migration timestamping

### LINKED ISSUE #646: Track Python migration scripts separately from SQL migration scripts
Running SQL migration scripts before Python migration scripts is generally a good thing, but if there's a crash during the migration, the timestamp will be updated to the date of the most recent script that completed. This might result in SQL migration scripts running twice (because a Python migration crashed) or Python migration scripts not running at all (because a SQL migration crashed).

Keeping two different migration timestamps -- one for SQL, one for Python -- should address the problem.

courte:
This branch add a distinct timestamp for python script migrations, to keep python migrations from being skipped in cases when the later SQL or python migrations error or fail in some way. Now, if an error arises in a python script, `core/bin/migrate_database --python-only` will re-run the busted scripts.

Notes:
- Interactions with the database have been moved to the `DatabaseMigrationScript.TimestampInfo` class.
- As usual, there's a ton of twisting and turning in the tests to get things working without running Actual Real Migrations.

Fixes #646.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1648,13 +1648,53 @@ class DatabaseMigrationScript(Script):
 
     class TimestampInfo(object):
         """Act like a ORM Timestamp object, but with no database connection."""
-        def __init__(self, timestamp, counter):
+
+        @classmethod
+        def find(cls, _db, service):
+            """Find or create an existing timestamp representing the last
+            migration script that was run.
+
+            :return: A TimestampInfo object or None
+            """
+            sql = "SELECT timestamp, counter FROM timestamps WHERE service=:service LIMIT 1;"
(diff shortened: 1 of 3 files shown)


scripts.py
```diff
@@ -2130,7 +2136,7 @@ def run(self, cmd_args=None):
                     "%s timestamp already exists: %r. Use --force to update." %
                     (self.name, existing_timestamp))
 
-        # Initialize the required timestamps with the Space Jame release date.
+        # Initialize the required timestamps with the Space Jam release date.
```
leonardr: :duck: :basketball: :rabbit2:

PR #690, opened 2017-10-25
OneClickImportMonitor can be run multiple times, but does nothing after the first time

leonardr:
There are a number of minor changes in this branch, but the main one is changing OneClickImportMonitor so it can be run on a regular schedule without killing the RBdigital servers. The work of the monitor will only be done once for any given collection.

### DIFF
--- a/model.py
+++ b/model.py
@@ -6847,18 +6847,24 @@ class LicensePool(Base):
         logging.info or a similar method
         """
         edition = self.presentation_edition
-        message = 'CHANGED '
+        message = u'CHANGED '
         args = []
+        if self.identifier:
+            identifier_template = '%s/%s'
+            identifier_args = [self.identifier.type, self.identifier.identifier]
+        else:
+            identifier_template = '%s'
+            identifier_args = [self.identifier]
         if edition:
-            message += '%s "%s" %s (%s)'
+            message += u'%s "%s" %s (' + identifier_template + ')'
             args.extend([edition.medium, 
                          edition.title or "[NO TITLE]",
-                         edition.author or "[NO AUTHOR]",
-                         self.identifier]
+                         edition.author or "[NO AUTHOR]"]
                     )
+            args.extend(identifier_args)
         else:
-            message += '%s'
-            args.append(self.identifier)
+            message += identifier_template
+            args.extend(identifier_args)
 
         def _part(message, args, string, old_value, new_value):
             if old_value != new_value:
--- a/oneclick.py
+++ b/oneclick.py
@@ -961,7 +961,6 @@ class OneClickImportMonitor(OneClickSyncMonitor):
     
     def invoke(self):
         timestamp = self.timestamp()
-        set_trace()
         if timestamp.counter and timestamp.counter > 0:
             self.log.debug(
                 "Collection %s has already had its initial import; doing nothing.", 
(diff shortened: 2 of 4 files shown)


model.py
```diff
@@ -6847,18 +6847,24 @@ def circulation_changelog(self, old_licenses_owned, old_licenses_available,
...
-                         edition.author or "[NO AUTHOR]",
-                         self.identifier]
+                         edition.author or "[NO AUTHOR]"]
                     )
+            args.extend(identifier_args)
         else:
-            message += '%s'
-            args.append(self.identifier)
+            message += identifier_template
+            args.extend(identifier_args)
```
leonardr: This change is to fix an annoying issue where a book with a non-ASCII character in its title can't have its circulation information logged. The title itself is all right, but the representation of the book's Identifier _also_ includes the title, and turning the representation of the book's identifier into a '%s' causes a UnicodeDecodeError.

Since it's not necessary to include the title twice, this change simplifies things and eliminates the problem by only logging the identifier and identifier type.

oneclick.py
```diff
@@ -165,7 +165,15 @@ def request(self, url, method='get', extra_headers={}, data=None,
...
-        
+    
+        if (response.content
+            and 'Invalid Basic Token or permission denied' in response.content):
+            raise BadResponseException(
+                url, "Permission denied. This may be a temporary rate-limiting issue, or the credentials for this collection may be wrong.",
+                debug_message=response.content,
+                status_code=502
+            )
+
```
leonardr: This change is to treat rate-limiting and invalid credentials as an error condition rather than an unusually formatted API response.

oneclick.py
```diff
@@ -319,17 +327,13 @@ def get_ebook_availability_info(self, media_type='ebook'):
             resplist = response.json()
         except Exception, e:
             raise BadResponseException(url, "OneClick availability response not parseable.")
-
-        if not resplist:
-            raise IOError("OneClick availability response not parseable - has no resplist.")
-
```
leonardr: It's legitimate for this API call to return an empty list, if a collection is actually empty.

oneclick.py
```diff
@@ -957,7 +961,21 @@ class OneClickImportMonitor(OneClickSyncMonitor):
...
+            )
+            return 0, 0
+        result = self.api.populate_all_catalog()
+
+        # Record the work was done so it's not done again.
+        if not timestamp.counter:
+            timestamp.counter = 1
+        else:
+            timestamp.counter += 1
+        return result
```
leonardr: Here's the change that stops the OneClickImportMonitor from doing work more than once for a given collection. I toyed with adding a `--force` option but it was way too much work and we don't need it ATM.

oneclick.py
```diff
@@ -165,7 +165,15 @@ def request(self, url, method='get', extra_headers={}, data=None,
             allowed_response_codes=allowed_response_codes, 
             disallowed_response_codes=disallowed_response_codes
         )
-        
+    
+        if (response.content
+            and 'Invalid Basic Token or permission denied' in response.content):
```
aslagle: This is a weird message - are these supposed to be in the same string?

leonardr: Yeah, the server itself isn't sure what the problem is.

PR #691, opened 2017-10-25
Manage randomly-generated patron identifiers for identifying a patron with an external service

leonardr:
For external services like RBdigital, Bibliotheca, and Axis 360, we can send whatever string we want as a patron ID -- it doesn't have to be the patron's actual ID, which is good because their actual ID tends to change a lot.

This branch makes it easy to manage external-facing patron identifiers for different services. I'm about to start using this with RBdigital, but by adding a migration to handle people with active loans, we can also start using it for Bibliotheca and Axis 360.

### DIFF
--- a/model.py
+++ b/model.py
@@ -647,17 +647,25 @@ class Patron(Base):
     
     AUDIENCE_RESTRICTION_POLICY = 'audiences'
 
-    @property
-    def best_unique_identifier(self):
-        """Return the best available unique identifier for this Patron.
+    def identifier_to_remote_service(self, remote_data_source, generator=None):
+        """Find or randomly create an identifier to use when identifying
+        this patron to a remote service.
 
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -7504,6 +7524,14 @@ class Credential(Base):
...
     )
 
+
+    # A meaningless identifier used to identify this patron (and no other)
+    # to a remote service.
+    IDENTIFIER_TO_REMOTE_SERVICE = "Identifier Sent To Remote Service"
+
+    # An identifier used by a remote service to identify this patron.
+    IDENTIFIER_FROM_REMOTE_SERVICE = "Identifier Received From Remote Service"
+
```
leonardr: This is not used in this branch but it is used by the RBdigital branch I'm working on.

PR #693, opened 2017-10-26
Fixed OPDS and CachedFeed.refresh to handle WorkList and changed CachedFeed to store Lane id.

aslagle:
This fixes a few problems I ran into trying to use this branch in circulation.

I changed CachedFeed to store Lane.id instead of the lane name and languages, now that lanes are in the database, and made CachedFeed.refresh and AcquisitionFeed.page work if they get WorkLists instead of Lanes.

### DIFF
--- a/model.py
+++ b/model.py
@@ -6083,11 +6083,11 @@ class CachedFeed(Base):
     id = Column(Integer, primary_key=True)
 
     # Every feed is associated with a lane. If null, this is a feed
-    # for the top level.
-    lane_name = Column(Unicode, nullable=True)
-
-    # Every feed includes book from a subset of available languages
-    languages = Column(Unicode)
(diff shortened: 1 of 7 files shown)


model.py
```diff
@@ -6194,7 +6195,7 @@ def fetch(cls, _db, lane, type, facets, pagination, annotator,
                 # default page-type feed, which should be cheap to fetch.
                 cls.log.warn(
                     "Could not generate a groups feed for %s, falling back to a page feed.",
-                    lane.identifier
+                    lane.display_name
```
leonardr: I think lane.identifier is still better because display_name can be ambiguous.

aslagle: That's true. What should I use for a WorkList though?

leonardr: Good point. I think using display_name is okay for WorkList, at least for now.

PR #695, opened 2017-10-30
OdiloAPI Integration.

ejcepas:
Odilo Connector API Integration (discovery and circulation API), first version commit.

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -64,6 +64,7 @@ class Classifier(object):
 
     GUTENBERG_BOOKSHELF = "gutenberg:bookshelf"
     TOPIC = "schema:Topic"
+    TOPIC_TERM = "Topic term"
     PLACE = "schema:Place"
     PERSON = "schema:Person"
     ORGANIZATION = "schema:Organization"
--- a/classifier.py
+++ b/classifier.py
@@ -64,7 +64,6 @@ class Classifier(object):
 
     GUTENBERG_BOOKSHELF = "gutenberg:bookshelf"
     TOPIC = "schema:Topic"
-    TOPIC_TERM = "Topic term"
     PLACE = "schema:Place"
     PERSON = "schema:Person"
     ORGANIZATION = "schema:Organization"
--- a/model.py
+++ b/model.py
@@ -967,6 +967,7 @@ class DataSource(Base, HasFullTableCache):
 
     GUTENBERG = u"Gutenberg"
     OVERDRIVE = u"Overdrive"
+    ODILO = u"Odilo"
     PROJECT_GITENBERG = u"Project GITenberg"
     STANDARD_EBOOKS = u"Standard Ebooks"
     UNGLUE_IT = u"unglue.it"
@@ -1197,6 +1198,7 @@ class DataSource(Base, HasFullTableCache):
                 (cls.GUTENBERG, True, False, Identifier.GUTENBERG_ID, None),
                 (cls.RB_DIGITAL, True, True, Identifier.RB_DIGITAL_ID, None),
                 (cls.OVERDRIVE, True, False, Identifier.OVERDRIVE_ID, 0),
+                (cls.ODILO, True, False, Identifier.ODILO_ID, 0),
                 (cls.THREEM, True, False, Identifier.BIBLIOTHECA_ID, 60*60*6),
                 (cls.AXIS_360, True, False, Identifier.AXIS_360_ID, 0),
                 (cls.OCLC, False, False, None, None),
@@ -1601,6 +1603,7 @@ class Identifier(Base):
     
     # Common types of identifiers.
     OVERDRIVE_ID = u"Overdrive ID"
+    ODILO_ID = u"Odilo ID"
     BIBLIOTHECA_ID = u"Bibliotheca ID"
     GUTENBERG_ID = u"Gutenberg ID"
     AXIS_360_ID = u"Axis 360 ID"
@@ -1628,7 +1631,7 @@ class Identifier(Base):
     ONECLICK_ID = RB_DIGITAL_ID
 
     LICENSE_PROVIDING_IDENTIFIER_TYPES = [
-        BIBLIOTHECA_ID, OVERDRIVE_ID, AXIS_360_ID,
+        BIBLIOTHECA_ID, OVERDRIVE_ID, ODILO_ID, AXIS_360_ID,
         GUTENBERG_ID, ELIB_ID
     ]
 
@@ -2728,6 +2731,7 @@ class Edition(Base):
     AUDIO_MEDIUM = u"Audio"
     MUSIC_MEDIUM = u"Music"
     VIDEO_MEDIUM = u"Video"
+    IMAGE_MEDIUM = u"Image"
 
     ELECTRONIC_FORMAT = u"Electronic"
     CODEX_FORMAT = u"Codex"
@@ -5708,6 +5712,7 @@ class Subject(Base):
 
     GUTENBERG_BOOKSHELF = Classifier.GUTENBERG_BOOKSHELF
     TOPIC = Classifier.TOPIC
+    TOPIC_TERM = Classifier.TOPIC_TERM
     PLACE = Classifier.PLACE
     PERSON = Classifier.PERSON
     ORGANIZATION = Classifier.ORGANIZATION
@@ -7793,6 +7798,9 @@ class Representation(Base):
     GIF_MEDIA_TYPE = u"image/gif"
     SVG_MEDIA_TYPE = u"image/svg+xml"
     MP3_MEDIA_TYPE = u"audio/mpeg"
+    MP4_MEDIA_TYPE = u"video/mp4"
+    WMV_MEDIA_TYPE = u"video/x-ms-wmv"
+    ZIP_MEDIA_TYPE = u"application/zip"
     OCTET_STREAM_MEDIA_TYPE = u"application/octet-stream"
     TEXT_PLAIN = u"text/plain"
     AUDIOBOOK_MANIFEST_MEDIA_TYPE = u"application/audiobook+json"
@@ -7830,6 +7838,8 @@ class Representation(Base):
         MOBI_MEDIA_TYPE: "mobi",
         PDF_MEDIA_TYPE: "pdf",
         MP3_MEDIA_TYPE: "mp3",
+        MP4_MEDIA_TYPE: "mp4",
+        WMV_MEDIA_TYPE: "wmv",
         JPEG_MEDIA_TYPE: "jpg",
         PNG_MEDIA_TYPE: "png",
         SVG_MEDIA_TYPE: "svg",
@@ -7837,7 +7847,8 @@ class Representation(Base):
         TEXT_PLAIN: "txt",
         TEXT_HTML_MEDIA_TYPE: "html",
         APPLICATION_XML_MEDIA_TYPE: "xml",
-        AUDIOBOOK_MANIFEST_MEDIA_TYPE: "audiobook-manifest"
+        AUDIOBOOK_MANIFEST_MEDIA_TYPE: "audiobook-manifest",
+        ZIP_MEDIA_TYPE: "zip"
     }
 
     # Invert FILE_EXTENSIONS and add some extra guesses.
@@ -8687,6 +8698,7 @@ class DeliveryMechanism(Base, HasFullTableCache):
     STREAMING_TEXT_CONTENT_TYPE = u"Streaming Text"
     STREAMING_AUDIO_CONTENT_TYPE = u"Streaming Audio"
     STREAMING_VIDEO_CONTENT_TYPE = u"Streaming Video"
+    PERIODICAL_STREAMING_TEXT_CONTENT_TYPE = u"Periodical Streaming Text"
 
     NO_DRM = None
     ADOBE_DRM = u"application/vnd.adobe.adept+xml"
@@ -9710,6 +9722,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     # Supported protocols for ExternalIntegrations with LICENSE_GOAL.
     OPDS_IMPORT = u'OPDS Import'
     OVERDRIVE = DataSource.OVERDRIVE
+    ODILO = DataSource.ODILO
     BIBLIOTHECA = DataSource.BIBLIOTHECA
     AXIS_360 = DataSource.AXIS_360
     RB_DIGITAL = DataSource.RB_DIGITAL
@@ -9725,7 +9738,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     GUTENBERG = DataSource.GUTENBERG
 
     LICENSE_PROTOCOLS = [
-        OPDS_IMPORT, OVERDRIVE, BIBLIOTHECA, AXIS_360, RB_DIGITAL,
+        OPDS_IMPORT, OVERDRIVE, ODILO, BIBLIOTHECA, AXIS_360, RB_DIGITAL,
         DIRECTORY_IMPORT, GUTENBERG, ENKI,
     ]
 
@@ -9733,6 +9746,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     # licenses come from a specific data source.
     DATA_SOURCE_FOR_LICENSE_PROTOCOL = {
         OVERDRIVE : DataSource.OVERDRIVE,
+        ODILO : DataSource.ODILO,
         BIBLIOTHECA : DataSource.BIBLIOTHECA,
         AXIS_360 : DataSource.AXIS_360,
         RB_DIGITAL : DataSource.RB_DIGITAL,
--- a/model.py
+++ b/model.py
@@ -174,7 +174,7 @@ class HasFullTableCache(object):
     """A mixin class for ORM classes that maintain an in-memory cache of
     (hopefully) every item in the database table for performance reasons.
     """
-    
+
     RESET = object()
 
     # You MUST define your own class-specific '_cache' and '_id_cache'
@@ -182,15 +182,15 @@ class HasFullTableCache(object):
     #
     # _cache = HasFullTableCache.RESET
     # _id_cache = HasFullTableCache.RESET
-    
+
     @classmethod
     def reset_cache(cls):
         cls._cache = cls.RESET
         cls._id_cache = cls.RESET
-        
+
     def cache_key(self):
         raise NotImplementedError()
-        
+
     @classmethod
     def _cache_insert(cls, obj, cache, id_cache):
         """Cache an object for later retrieval, possibly by a different
@@ -208,7 +208,7 @@ class HasFullTableCache(object):
             # reset and the time we tried to put an object in the
             # cache. Stop trying to mess with the cache.
             pass
-                
+
     @classmethod
     def populate_cache(cls, _db):
         """Populate the in-memory caches from scratch with every single
@@ -220,7 +220,7 @@ class HasFullTableCache(object):
             cls._cache_insert(obj, cache, id_cache)
         cls._cache = cache
         cls._id_cache = id_cache
-        
+
     @classmethod
     def _cache_lookup(cls, _db, cache, cache_name, cache_key, lookup_hook):
         """Helper method used by both by_id and by_cache_key.
@@ -247,7 +247,7 @@ class HasFullTableCache(object):
                 # reset just now, we still have a copy of the 'old'
                 # cache which passed the 'cache != cls.RESET' test.
                 pass
-                
+
         if not obj:
             # Either this object didn't exist when the cache was
             # populated, or the cache was reset while we were trying
@@ -266,7 +266,7 @@ class HasFullTableCache(object):
             # Stick the object in the caches, assuming they're not
             # currently in a reset state.
             cls._cache_insert(obj, cls._cache, cls._id_cache)
-            
+
         if obj and obj not in _db:
             try:
                 obj = _db.merge(obj, load=False)
@@ -285,17 +285,17 @@ class HasFullTableCache(object):
                 logging.error("Unable to look up a fresh copy of %r", obj)
                 raise e
         return obj, new
-        
+
     @classmethod
     def by_id(cls, _db, id):
         """Look up an item by its unique database ID."""
         def lookup_hook():
-            return get_one(_db, cls, id=id), False        
+            return get_one(_db, cls, id=id), False
         obj, is_new = cls._cache_lookup(
             _db, cls._id_cache, '_id_cache', id, lookup_hook
         )
         return obj
-            
+
     @classmethod
     def by_cache_key(cls, _db, cache_key, lookup_hook):
         return cls._cache_lookup(
@@ -356,7 +356,7 @@ class SessionManager(object):
                 "Loading materialized view %s from %s.",
                 view_name, resource_file)
             sql = open(resource_file).read()
-            connection.execute(sql)                
+            connection.execute(sql)
 
         if not connection:
             connection = engine.connect()
@@ -385,8 +385,8 @@ class SessionManager(object):
 
         class MaterializedWorkWithGenre(Base, BaseMaterializedWork):
             __table__ = Table(
-                cls.MATERIALIZED_VIEW_WORKS_WORKGENRES, 
-                Base.metadata, 
+                cls.MATERIALIZED_VIEW_WORKS_WORKGENRES,
+                Base.metadata,
                 Column('works_id', Integer, primary_key=True),
                 Column('workgenres_id', Integer, primary_key=True),
                 Column('license_pool_id', Integer, ForeignKey('licensepools.id')),
@@ -394,21 +394,21 @@ class SessionManager(object):
                 autoload_with=engine
             )
             license_pool = relationship(
-                LicensePool, 
+                LicensePool,
                 primaryjoin="LicensePool.id==MaterializedWorkWithGenre.license_pool_id",
                 foreign_keys=LicensePool.id, lazy='joined', uselist=False)
 
         class MaterializedWork(Base, BaseMaterializedWork):
             __table__ = Table(
-                cls.MATERIALIZED_VIEW_WORKS, 
-                Base.metadata, 
+                cls.MATERIALIZED_VIEW_WORKS,
+                Base.metadata,
                 Column('works_id', Integer, primary_key=True),
                 Column('license_pool_id', Integer, ForeignKey('licensepools.id')),
               autoload=True,
                 autoload_with=engine
             )
             license_pool = relationship(
-                LicensePool, 
+                LicensePool,
                 primaryjoin="LicensePool.id==MaterializedWork.license_pool_id",
                 foreign_keys=LicensePool.id, lazy='joined', uselist=False)
 
@@ -446,7 +446,7 @@ class SessionManager(object):
 
         # Load all existing Genre objects.
         Genre.populate_cache(session)
-        
+
         # Create any genres not in the database.
         for g in classifier.genres.values():
             # TODO: On the very first startup this is rather expensive
@@ -467,7 +467,7 @@ class SessionManager(object):
         # If there is currently no 'site configuration change'
         # Timestamp in the database, create one.
         timestamp, is_new = get_one_or_create(
-            session, Timestamp, collection=None, 
+            session, Timestamp, collection=None,
             service=Configuration.SITE_CONFIGURATION_CHANGED,
             create_method_kwargs=dict(timestamp=datetime.datetime.utcnow())
         )
@@ -530,7 +530,7 @@ def get_one_or_create(db, model, create_method='',
             return obj
         except IntegrityError, e:
             logging.info(
-                "INTEGRITY ERROR on %r %r, %r: %r", model, create_method_kwargs, 
+                "INTEGRITY ERROR on %r %r, %r: %r", model, create_method_kwargs,
                 kwargs, e)
             __transaction.rollback()
             return db.query(model).filter_by(**kwargs).one(), False
@@ -572,7 +572,7 @@ class Patron(Base):
         Integer, ForeignKey('libraries.id'), index=True,
         nullable=False
     )
-    
+
     # The patron's permanent unique identifier in an external library
     # system, probably never seen by the patron.
     #
@@ -630,7 +630,7 @@ class Patron(Base):
     # library server).
     _synchronize_annotations = Column(Boolean, default=None,
                                       name="synchronize_annotations")
-    
+
     loans = relationship('Loan', backref='patron')
     holds = relationship('Hold', backref='patron')
 
@@ -644,7 +644,7 @@ class Patron(Base):
         UniqueConstraint('library_id', 'authorization_identifier'),
         UniqueConstraint('library_id', 'external_identifier'),
     )
-    
+
     AUDIENCE_RESTRICTION_POLICY = 'audiences'
 
     def identifier_to_remote_service(self, remote_data_source, generator=None):
@@ -666,7 +666,7 @@ class Patron(Base):
             self, refresh, allow_persistent_token=True
         )
         return credential.credential
-    
+
     def works_on_loan(self):
         db = Session.object_session(self)
         loans = db.query(Loan).filter(Loan.patron==self)
@@ -702,7 +702,7 @@ class Patron(Base):
     @hybrid_property
     def synchronize_annotations(self):
         return self._synchronize_annotations
-    
+
     @synchronize_annotations.setter
     def _set_synchronize_annotations(self, value):
         """When a patron says they don't want their annotations to be stored
@@ -725,7 +725,7 @@ Index("ix_patron_library_id_external_identifier", Patron.library_id, Patron.exte
 Index("ix_patron_library_id_authorization_identifier", Patron.library_id, Patron.authorization_identifier)
 Index("ix_patron_library_id_username", Patron.library_id, Patron.username)
 
-        
+
 class PatronProfileStorage(ProfileStorage):
     """Interface between a Patron object and the User Profile Management
     Protocol.
@@ -737,7 +737,7 @@ class PatronProfileStorage(ProfileStorage):
         :param patron: We are accessing the profile for this patron.
         """
         self.patron = patron
-    
+
     @property
     def writable_setting_names(self):
         """Return the subset of settings that are considered writable."""
@@ -783,7 +783,7 @@ class LoanAndHoldMixin(object):
             return license_pool.work
         if license_pool.presentation_edition and license_pool.presentation_edition.work:
             return license_pool.presentation_edition.work
-        return None        
+        return None
 
 
 class Loan(Base, LoanAndHoldMixin):
@@ -828,7 +828,7 @@ class Hold(Base, LoanAndHoldMixin):
 
         We have to wait for the available licenses to cycle a
         certain number of times before we get a turn.
-        
+
         Example: 4 licenses, queue position 21
         After 1 cycle: queue position 17
               2      : queue position 13
@@ -900,7 +900,7 @@ class Hold(Base, LoanAndHoldMixin):
         """When the book becomes available, position will be 0 and end will be
         set to the time at which point the patron will lose their place in
         line.
-        
+
         Otherwise, end is irrelevant and is set to None.
         """
         if start is not None:
@@ -923,7 +923,7 @@ class Annotation(Base):
 
     # We need to define some terms of our own.
     LS_NAMESPACE = u"http://librarysimplified.org/terms/annotation/"
-   
+
     IDLING = LS_NAMESPACE + u'idling'
     BOOKMARKING = OA_NAMESPACE + u'bookmarking'
 
@@ -951,7 +951,7 @@ class Annotation(Base):
             raise ValueError(
                 "Patron has opted out of synchronizing annotations."
             )
-    
+
         return get_one_or_create(
             _db, Annotation, patron=patron, *args, **kwargs
         )
@@ -1006,7 +1006,7 @@ class DataSource(Base, HasFullTableCache):
     }
     THREEM = BIBLIOTHECA
     ONECLICK = RB_DIGITAL
-    
+
     # Some sources of open-access ebooks are better than others. This
     # list shows which sources we prefer, in ascending order of
     # priority. unglue.it is lowest priority because it tends to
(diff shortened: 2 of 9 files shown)


model.py
```diff
@@ -8687,6 +8698,7 @@ class DeliveryMechanism(Base, HasFullTableCache):
     STREAMING_TEXT_CONTENT_TYPE = u"Streaming Text"
     STREAMING_AUDIO_CONTENT_TYPE = u"Streaming Audio"
     STREAMING_VIDEO_CONTENT_TYPE = u"Streaming Video"
+    PERIODICAL_STREAMING_TEXT_CONTENT_TYPE = u"Periodical Streaming Text"
```
leonardr: I'm pretty sure you should use STREAMING_TEXT_CONTENT_TYPE here. Whether the text being streamed is a periodical or something else isn't relevant here, just as `text/html` doesn't make a distinction between different types of web pages.

model.py
```diff
@@ -2728,6 +2731,7 @@ class Edition(Base):
     AUDIO_MEDIUM = u"Audio"
     MUSIC_MEDIUM = u"Music"
     VIDEO_MEDIUM = u"Video"
+    IMAGE_MEDIUM = u"Image"
```
leonardr: What's the goal of this? These constants are intended to go into `Edition.medium`. That implies a situation where a single image shows up in a catalog as a 'book' which may be 'borrowed' and downloaded for local reference. This isn't impossible, especially in an academic or archive context, but I don't think that's what Odilo offers.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+    ALL_PRODUCTS_ENDPOINT = BASE_URL + "/records"
+
+    RECORD_METADATA_ENDPOINT = BASE_URL + "/records/{recordId}"
+    RECORD_AVAILABILITY_ENDPOINT = BASE_URL + "/records/{recordId}/availability"
+
+    # --- Circulation API ---
+    CHECKOUT_ENDPOINT = BASE_URL + "/records/{recordId}/checkout"
+    CHECKOUT_GET = BASE_URL + "/checkouts/{checkoutId}"
+    CHECKIN_ENDPOINT = BASE_URL + "/checkouts/{checkoutId}/return"
+    # Downloads given checkout offering the url where to consume the digital resource.
```
leonardr: FYI we call this 'fulfillment' to make it clear that it's a separate operation from 'checkout'.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+    CHECKOUT_URL_ENDPOINT = BASE_URL + "/checkouts/{checkoutId}/download"
+
+    PLACE_HOLD_ENDPOINT = BASE_URL + "/records/{recordId}/hold"
+    HOLD_GET = BASE_URL + "/holds/{holdId}"
+    RELEASE_HOLD_ENDPOINT = BASE_URL + "/holds/{holdId}/cancel"
+
+    PATRON_CHECKOUTS_ENDPOINT = BASE_URL + "/patrons/{patronId}/checkouts"
+    PATRON_HOLDS_ENDPOINT = BASE_URL + "/patrons/{patronId}/holds"
+
+    def update_endpoints_url(self):
```
leonardr: Rather than redefining all these class constants whenever a new BASE_URL is set, I would define the constants once in the class without reference to BASE_URL, and then prepend the collection's base URL before making the request (this could go into do_get and do_post, for example).

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+    HTTP,
+    BadResponseException,
+)
+
+
+class OdiloAPI(object):
+    log = logging.getLogger("Odilo API")
+
+    LIBRARY_API_BASE_URL = u"library_api_base_url"
+    BASE_URL = "http://localhost:8080/api/v2"  # Debug value by default
```
leonardr: If you have one live API endpoint, I recommend renaming this to DEFAULT_BASE_URL and putting your live API endpoint here so it can serve as a default. If the API endpoing is different for every library, I recommend getting rid of this altogether.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+        credential.credential = odilo_data['token']
+        expires_in = (odilo_data['expiresIn'] * 0.9)
+        credential.expires = datetime.datetime.utcnow() + datetime.timedelta(seconds=expires_in)
+
+    def get_metadata(self, record_id):
+        identifier = record_id
+        if isinstance(record_id, Identifier):
+            identifier = record_id.identifier
+
+        url = self.RECORD_METADATA_ENDPOINT.replace('{recordId}', identifier)
```
leonardr: I expected you to use `string.format` here. Either change this to use `string.format(recordId=identifier)`, or use "%(recordId)s" instead of "{recordId} in your strings and format them with the `%` operator. The difference is cosmetic, but in most other places in the app we use the "%(recordId)s" style.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+
+        response = self.get(url)
+
+        if response and response.status == 200 and response.content:
+            return response.content
+        else:
+            self.log.warn('Cannot retrieve metadata for record: ' + record_id + ' response http ' + response.status)
+            if response.content:
+                self.log.warn(response.content)
```
leonardr: I would put all of this information in a single `warn` statement -- otherwise you're unlikely to find all the relevant information when grepping the error log.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+                self.log.warn(content)
+            return None
+
+    @staticmethod
+    def _do_get(url, headers, **kwargs):
+        # More time please
+        if not kwargs:
+            kwargs = {"timeout": 60}
+        else:
+            kwargs['timeout'] = 60
```
leonardr: kwargs is never going to be null, so you don't need the first clause. However if a specific value is passed in for `timeout` I would use that value instead of forcing it to 60. (the value passed in may be bigger than 60).

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+        url = self.RECORD_AVAILABILITY_ENDPOINT.replace('{recordId}', record_id)
+        status_code, headers, content = self.get(url)
+        content = json.loads(content)
+
+        if status_code == 200 and len(content) > 0:
+            return content
+        else:
+            self.log.warn('Cannot retrieve availability for record: ' + record_id + ' response http ' + status_code)
+            if content:
+                self.log.warn(content)
```
leonardr: Again, put the content in the first warn statement.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+    format_data_for_odilo_format = {
+        "ACSM": (
+            Representation.PDF_MEDIA_TYPE, DeliveryMechanism.ADOBE_DRM
+        ),
+        "PDF": (
+            Representation.PDF_MEDIA_TYPE, DeliveryMechanism.STREAMING_TEXT_CONTENT_TYPE
+        ),
+        "EPUB": (
+            Representation.EPUB_MEDIA_TYPE, DeliveryMechanism.STREAMING_TEXT_CONTENT_TYPE
+        ),
```
leonardr: I don't believe a PDF or EPUB file can be delivered through streaming. If this is a PDF or EPUB file that's been unpacked, rendered to HTML+images and delivered through a web application, then the media type here should be HTML_MEDIA_TYPE. (It will be delivered to the client as `text/html;profile=http://librarysimplified.org/terms/profiles/streaming-media` to distinguish it from a normal "book in a single web page".

If this is a PDF or EPUB that is downloaded in full onto the client and stored there, then the media type is correct but the DRM type is NO_DRM.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+        "MP4": (
+            Representation.MP4_MEDIA_TYPE, DeliveryMechanism.STREAMING_VIDEO_CONTENT_TYPE
+        ),
+        "WMV": (
+            Representation.WMV_MEDIA_TYPE, DeliveryMechanism.STREAMING_VIDEO_CONTENT_TYPE
+        ),
+        "JPG": (
+            Representation.JPEG_MEDIA_TYPE, DeliveryMechanism.NO_DRM
+        ),
```
leonardr: Again, I'm not sure what JPEGs you're delivering. I admit it's theoretically possible but I'd like to understand more so I can be sure about my recommendations.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+        "SCORM": (
+            Representation.ZIP_MEDIA_TYPE, DeliveryMechanism.NO_DRM
+        )
```
leonardr: This needs a more specific media type. An EPUB is a ZIP file but because it's served as `application/epub+zip` the client knows to render it as an ebook and not to open it in an archive manager.

I looked at the SCORM spec and it does not seem to define a media type, so I recommend we come up with one. How about `application/vnd.librarysimplified.scorm+zip`?

It may seem nitpicky to bring this up now, but if this isn't in place from the beginning you'll need to do an annoying database migration later, and if we never deal with it, a client will have no idea whether or not it can handle a given "ZIP" file, because so many things are packaged via ZIP.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+
+    odilo_medium_to_simplified_medium = {
+        "ACSM": Edition.BOOK_MEDIUM,
+        "PDF": Edition.BOOK_MEDIUM,
+        "EPUB": Edition.BOOK_MEDIUM,
+        "MP3": Edition.AUDIO_MEDIUM,
+        "MP4": Edition.VIDEO_MEDIUM,
+        "WMV": Edition.VIDEO_MEDIUM,
+        "JPG": Edition.IMAGE_MEDIUM,
+        "SCORM": Edition.ELECTRONIC_FORMAT
```
leonardr: ELECTRONIC_FORMAT is used to distinguish between ebooks and print books. SCORM needs a new medium constant. I think COURSEWARE_MEDIUM would be most appropriate.

odilo.py
```diff
@@ -0,0 +1,581 @@
...
+        record_id = availability['recordId']
+        primary_identifier = IdentifierData(Identifier.ODILO_ID, record_id)  # We own this availability.
+
+        licenses_owned = int(availability['totalCopies'])
+        licenses_available = int(availability['availableCopies'])
+        if 'numLoans' in availability:
+            licenses_checked_out = int(availability['numLoans'])
+        licenses_reserved = int(availability['holdsQueueSize'])
+        if 'numPatronsInHoldQueue' in availability:
+            patrons_in_hold_queue = int(availability['numPatronsInHoldQueue'])
```
leonardr: I want to be careful about this. `patrons_in_hold_queue` contains the number of patrons who are currently waiting for a copy of the book. `licenses_reserved` is the number of patrons who put the book on hold _earlier_, but who are now at the front of the queue and who could get the book right now if they wanted to. `licenses_reserved` should never be larger than `patrons_in_hold_queue`, because everyone with a reserved license is also in the hold queue.

It's not clear to me what `holdsQueueSize` is measuring or how it is different from `numPatronsInHoldQueue`, so I want to make sure you're mapping the information appropriately.

classifier.py
```diff
@@ -64,6 +64,7 @@ class Classifier(object):
 
     GUTENBERG_BOOKSHELF = "gutenberg:bookshelf"
     TOPIC = "schema:Topic"
+    TOPIC_TERM = "Topic term"
```
leonardr: If you classify all your books under `Subject`s of type TOPIC_TERM, but don't provide a way to map TOPIC_TERM classifications to `Genre`s, then your books won't be classified, even in terms of fiction/nonfiction, which means they probably won't show up in the lanes. In most cases I would recommend 

1. File your subjects as TAG instead of making up a new classification scheme. TAG is for free-text classifications and it should also work on English-language topic names.
2. Define an ODILO_TOPIC constant, and define a subclass of TagClassifier that handles it. If TagClassifier is good enough for your purposes, you don't need any code in the subclass, but this will keep your controlled vocabulary separate from other peoples'. (Assuming this is in fact a controlled vocabulary.)
3. If your topic terms are in Spanish, then you'll need to write the classification code yourself. It might be smart to refactor `TagClassifier` and do a Spanish version of the same code.

If you can show me a list of all your topic terms I can give more specific advice. You don't need to have the classifier in this branch but I want the `Subject`s you create to have the right `type` from the start.

model.py
ejcepas: ok, perfect

model.py
ejcepas: The goal of this is exactly what you say, we offer the possibility to catalogue images individually as a whole, as if books. Then, when a patron is logged in, he or she can consult the images, but no checkout or place a hold on them. We accept some image formats such as: jpg/jpeg, png y svg , bmp and gif to catalogue images.

odilo.py
ejcepas: ok, thank you, I have already changed it.

odilo.py
ejcepas: yes, we know the difference and have also the two steps: checkout a resource, and the fulfill it in order to consume it. Maybe we should rename the API endpoint in the future to avoid confusion.

odilo.py
ejcepas: Yes, you are right, we use a different API endpoint URL for each library, so I have deleted BASE_URL.

odilo.py
ejcepas: Ok, I have changed all string.replace for string.format

odilo.py
ejcepas: ok, changed.

odilo.py
ejcepas: ok

odilo.py
ejcepas: ok

odilo.py
ejcepas: We deliver the images as a whole as I explained in a previous comment.

odilo.py
ejcepas: Ok, you are right. It sounds good, we could use the media type you suggest for the SCORM files.

odilo.py
ejcepas: Ok, I created that constant. Also I've checked where Edition.BOOK_MEDIUM is used within the Edition class, in order to add Edition.IMAGE_MEDIUM and the new Edition.COURSEWARE_MEDIUM.

odilo.py
leonardr: OK, there's nothing wrong with this, it just seems a little weird to me, so go for it.

odilo.py
ejcepas: Ok, It's the first case you suppose. I have already correct the media type for both PDF and EPUB files.

odilo.py
ejcepas: ok, I have revised that, fixed it in the integration and updated the Connector availability API. Now we have:
- 'notifiedHolds' --> Patrons that placed a hold and have already been notified, so they can get the book when they want. 
- 'holdsQueueSize' --> all the patrons wainting for a copy of a specific book (notified or not).

classifier.py
ejcepas: We considered that and by the moment decided to use the first solution you suggest, just file our subjects as TAG.
Maybe in the future we could offer the BISAC subject code through Connector API withint the record metadata.

model.py
```diff
@@ -2738,27 +2743,29 @@ class Edition(Base):
         PERIODICAL_MEDIUM : u"http://schema.org/PublicationIssue",
         MUSIC_MEDIUM :  u"http://schema.org/MusicRecording",
         VIDEO_MEDIUM :  u"http://schema.org/VideoObject",
+        IMAGE_MEDIUM :  u"http://schema.org/ImageObject",
+        COURSEWARE_MEDIUM : u""
```
leonardr: Let's put `http://schema.org/Course` for this.

model.py
```diff
@@ -2738,27 +2743,29 @@ class Edition(Base):
...
 
     additional_type_to_medium = {}
     for k, v in medium_to_additional_type.items():
         additional_type_to_medium[v] = k
 
     medium = Column(
-        Enum(BOOK_MEDIUM, PERIODICAL_MEDIUM, AUDIO_MEDIUM,
-             MUSIC_MEDIUM, VIDEO_MEDIUM, name="medium"),
+        Enum(BOOK_MEDIUM, PERIODICAL_MEDIUM, AUDIO_MEDIUM, MUSIC_MEDIUM, VIDEO_MEDIUM, IMAGE_MEDIUM, COURSEWARE_MEDIUM,
+             name="medium"),
```
leonardr: This is going to need a migration script. See `core/migration/20170928-add-registered-coverage-status.sql` for an example. It would look something like this:

```
ALTER TYPE medium ADD VALUE IF NOT EXISTS 'image' AFTER 'video';
ALTER TYPE medium ADD VALUE IF NOT EXISTS 'courseware' AFTER 'image';
```

model.py
```diff
@@ -3347,7 +3358,7 @@ def choose_cover(self):
                     if not rep.mirrored_at and not rep.thumbnails:
                         logging.warn(
                             "Best cover for %r (%s) was never mirrored or thumbnailed!",
-                            self.primary_identiifer,
```
leonardr: I know this is just a spacing change but you've revealed a typo and fixing it would be helpful. This should be `self.primary_identifier`.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+        if response.status_code == 200:
+            self._update_credential(credential, data)
+            self.token = credential.credential
+            return 'OK'
+        elif response.status_code == 400:
+            response = response.json()
+            message = response['error']
+            if 'error_description' in response:
+                message += '/' + response['error_description']
+            return 'message'
```
leonardr: This should be the variable `message`, not the string `'message'`.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+            create_method_kwargs=dict(
+                external_account_id=u'library_id_123',
+            )
+        )
+        integration = collection.create_external_integration(
+            protocol=ExternalIntegration.ODILO
+        )
+        integration.username = u'username'
+        integration.password = u'password'
+        integration.setting(OdiloAPI.LIBRARY_API_BASE_URL).value = u'http://localhost:8080/api/v2'
```
leonardr: This should be a completely fake URL that will never work, to ensure that tests never actually try to make an HTTP request.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+
+class OdiloRepresentationExtractor(object):
+    """Extract useful information from Odilo's JSON representations."""
+
+    log = logging.getLogger("OdiloRepresentationExtractor")
+
+    format_data_for_odilo_format = {
+        "ACSM": (
+            Representation.PDF_MEDIA_TYPE, DeliveryMechanism.ADOBE_DRM
+        ),
```
leonardr: Do you know that the media type here is PDF, or might it be EPUB or something else?

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+            published = book.get('releaseDate')
+
+        if published:
+            try:
+                published = datetime.datetime.strptime(published, "%Y%m%d")
+            except ValueError as e:
+                cls.log.warn('Cannot parse publication date from: ' + published + ', message: ' + e.message)
+
+        # yyyyMMdd --> record last modification date
+        last_update = book['modificationDate']
```
leonardr: I noticed you didn't use `get` here, the way you did with the other dates. Do you know this field will always be present?

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+        if last_update:
+            try:
+                last_update = datetime.datetime.strptime(last_update, "%Y%m%d")
+            except ValueError as e:
+                cls.log.warn('Cannot parse last update date from: ' + last_update + ', message: ' + e.message)
+
+        language = book.get('language')
```
leonardr: What format are your languages in? We should be able to handle it but I want to make sure.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+
+        subjects = []
+        for subject in book.get('subjects', []):
+            subjects.append(SubjectData(type=Subject.TAG, identifier=subject, weight=100))
+
```
leonardr: It probably won't result in any changes to this branch but if you can send me a complete list of the subjects you use, it would be useful for my work.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+    log = logging.getLogger("OdiloRepresentationExtractor")
+
+    format_data_for_odilo_format = {
+        "ACSM": (
+            Representation.PDF_MEDIA_TYPE, DeliveryMechanism.ADOBE_DRM
+        ),
+        "EBOOK_STREAMING": (
+            Representation.TEXT_HTML_MEDIA_TYPE, DeliveryMechanism.STREAMING_TEXT_CONTENT_TYPE
+        ),
```
leonardr: Rather than defining a new constant here and mapping both 'PDF' and 'EPUB' to that constant so the lookup works, I think it would be simpler to map both 'PDF' and 'EPUB' to identical values in this dictionary.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+        if grade_level:
+            subject = SubjectData(type=Subject.GRADE_LEVEL, identifier=grade_level, weight=10)
+            subjects.append(subject)
+
+        medium = None
+        formats = []
+        for format_received in book.get('formats', []):
+            # Both formats (PDF and EPUB) behave the same, and are basically the same thing for the final users.
+            if format_received in ('PDF', 'EPUB'):
+                format_received = 'EBOOK_STREAMING'
```
leonardr: See comment above re: mapping 'PDF' and 'EPUB' separately.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+            if format_received in cls.format_data_for_odilo_format:
+                content_type, drm_scheme = cls.format_data_for_odilo_format.get(format_received)
+                formats.append(FormatData(content_type, drm_scheme))
+                if not medium:
+                    medium = cls.odilo_medium_to_simplified_medium.get(format_received)
```
leonardr: It probably doesn't matter but feel like this should be outside the `if` statement in line 470.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+            else:
+                cls.log.warn('Unrecognized format received: ' + format_received)
+
+        if not medium:
+            medium = Edition.BOOK_MEDIUM
+
+        identifiers = []
+        isbn = book.get('isbn')
+        if isbn:
+            if len(isbn) == 10:
```
leonardr: I use `if isbnlib.is_isbn10(isbn)` for this purpose.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+
+        Note:  The json data passed into this method is from a different file/stream
+        from the json data that goes into the book_info_to_circulation() method.
+        """
+        if 'id' not in book:
+            return None
+
+        odilo_id = book['id']
+        primary_identifier = IdentifierData(Identifier.ODILO_ID, odilo_id)
+        active = book.get('active')
```
leonardr: What does 'active' mean in this context? I noticed that you won't treat a metadata import as successful unless `active` is true. If a book is not 'active' because its metadata is incomplete then that's appropriate. If a book is not 'active' because it's no longer in the collection, then it would be better to allow the import to proceed but make sure the `CirculationData` makes it clear that the work is not available.

tests/test_odilo.py
```diff
@@ -0,0 +1,140 @@
...
+class TestOdiloBibliographicCoverageProvider(OdiloTest):
+    def setup(self):
+        super(TestOdiloBibliographicCoverageProvider, self).setup()
+        self.provider = OdiloBibliographicCoverageProvider(
+            self.collection, api_class=MockOdiloAPI
+        )
+        self.api = self.provider.api
+
+    def test_process_item(self):
+        print 'Testing process item...'
```
leonardr: All print statements should be removed or converted into `self.log` statements. This one shouldn't be necessary because making it clear which test is being run is the responsibility of the test infrastructure.

tests/test_odilo.py
```diff
@@ -0,0 +1,140 @@
...
+        record_metadata, record_metadata_json = self.sample_json("odilo_metadata.json")
+        self.api.queue_response(200, content=record_metadata_json)
+        availability, availability_json = self.sample_json("odilo_availability.json")
+        self.api.queue_response(200, content=availability)
+
+        identifier, made_new = self.provider.process_item('00010982')
+        ok_(identifier, msg="Problem testing process item !!!")
```
leonardr: You should also test that `process_item` did what it's supposed to do -- that metadata and availability information were imported properly and that `handle_success` was called -- as well as verifying that the Identifier returned has the right `.type` and `.identifier`.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+        language = book.get('language')
+
+        subjects = []
+        for subject in book.get('subjects', []):
+            subjects.append(SubjectData(type=Subject.TAG, identifier=subject, weight=100))
+
+        grade_level = book.get('gradeLevel')
+        if grade_level:
+            subject = SubjectData(type=Subject.GRADE_LEVEL, identifier=grade_level, weight=10)
+            subjects.append(subject)
```
leonardr: I noticed that the example grade level in your test is 'K-12'. I recognize that 'K-12' is _formatted_ like a grade level, but it covers every single grade in the US school system, making it seem more like an indicator of children-ness than a grade level per se. Is 'K-12' just an unfortunate example? Do you have more specific grade levels like '3-4'?

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+        title = book.get('title')
+        subtitle = book.get('subtitle')
+        series = book.get('series')
+        series_position = book.get('seriesPosition')
+
+        contributors = []
+        author = book.get('author')
+        if author:
+            roles = [Contributor.AUTHOR_ROLE]
+            contributor = ContributorData(sort_name=author, display_name=author, roles=roles, biography=None)
```
leonardr: Treating `sort_name` and `display_name` is a problem. I see from your test that your names are in the format 'E. Veale'. We need names in the format 'Veale, E.' (that's `sort_name`), with 'E. Veale' (that's `display_name`) being an optional addition. If you just don't have the `sort_name` then we'll figure something out, but it would be best to modify your API to provide it.

tests/test_odilo.py
```diff
@@ -0,0 +1,140 @@
...
+             ('Classics', Subject.TAG, 100),
+             ('Fantasy', Subject.TAG, 100),
+             ('K-12', Subject.GRADE_LEVEL, 10),
+             ],
+            [(x.identifier, x.type, x.weight) for x in subjects]
+            )
+
+        [author] = metadata.contributors
+        eq_("E. Veale", author.sort_name)
+        eq_("E. Veale", author.display_name)
```
leonardr: See above re: sort and display names.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+            if len(isbn) == 10:
+                isbn = isbnlib.to_isbn13(isbn)
+            identifiers.append(IdentifierData(Identifier.ISBN, isbn, 1))
+
+        # A cover
+        links = []
+        cover_image_url = book.get('coverImageUrl')
+        if cover_image_url:
+            image_data = cls.image_link_to_linkdata(cover_image_url, Hyperlink.IMAGE)
+            if image_data:
```
leonardr: Looking at your example I see that the cover image sent through the API (http://pruebasotk.odilotk.es/public/OdiloPlace_eduDistUS/pg54159_225x318.jpg) is about 200x300 pixels. This is approximately the size we use for thumbnails. There's a larger image at http://pruebasotk.odilotk.es/public/OdiloPlace_eduDistUS/pg54159.jpg but the API doesn't link to it.

The best thing to do here would be to modify the API to send both URLs. This code would then create two `LinkData` objects, one for the full image and one for the thumbnail. The thumbnail's `LinkData` would be set as `.thumbnail` on the full image's `LinkData`.

The full image is used by the metadata wrangler to scale down a cover image to the correct dimensions for the SimplyE app.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+                links.append(image_data)
+
+        # Descriptions become links.
+        description = book.get('description')
+        if description:
+            links.append(LinkData(rel=Hyperlink.DESCRIPTION, content=description, media_type="text/html"))
```
leonardr: Is this correct? The description provided in the example JSON looks like plain text, not HTML.

leonardr:
It would also be comforting to add tests for the basic `OdiloAPI` method like `get`, `credential_object`, and `check_creds`. You can see some example code to model your tests on in `core/tests/test_overdrive.api`, e.g. `test_401_on_get_refreshes_bearer_token` and `test_credential_refresh_success`.

odilo.py
```diff
@@ -0,0 +1,598 @@
...
+            )
+            data = response.json()
+        except (HTTPException, RequestNetworkException) as e:
+            self.log.error("Cannot connect with %s, error: %s" % (self.TOKEN_ENDPOINT, e.message))
+            return None
+
+        if response.status_code == 200:
+            self._update_credential(credential, data)
+            self.token = credential.credential
+            return 'OK'
```
leonardr: I see that you're checking this value in the calling method in `circulation`. I don't like the idea of passing specific string values back and forth. I would rather do one of the following:

* Return None on success and an error string if there's an error (this would mean changing line 155 above).
* Raise an exception if there's an error; return value is irrelevant.

model.py
ejcepas: Ok, thank you !

model.py
ejcepas: ok, fixed!

odilo.py
ejcepas: thank you, fixed !!

odilo.py
ejcepas: ok, changed !!

odilo.py
ejcepas: You are right, It's better to use 'get' just in case, I've changed it.

odilo.py
ejcepas: I think it's ok, because if 'format_received' it's in 'cls.format_data_for_odilo_format', implies that it's also in 'cls.odilo_medium_to_simplified_medium.get' and the other way round.

odilo.py
ejcepas: ok, changed !

tests/test_odilo.py
ejcepas: Ok, I have either deleted or changed them at both test files.

odilo.py
ejcepas: "active" means that the book it's no longer in the collection, so, I have corrected that as you advised me.

tests/test_odilo.py
ejcepas: Ok, I have revised the tests following your advice.

odilo.py
ejcepas: Yes, it is. Descriptions can contain HTML. I've updated the description field at the test file.

odilo.py
ejcepas: Ok, I have removed the 'return' clauses and raised an exception if needed instead.

tests/test_odilo.py
ejcepas: Also I have added new tests for the API as you request below.

model.py
ejcepas: Ok, created and added !!

odilo.py
ejcepas: ACSM resources could be either PDF or EPUB, so, I'am not sure about what representation media type should we use here, since in addition the actual files received are .acsm (which internally are xml-like).

odilo.py
ejcepas: OK, I have added the field 'originalImageUrl' to the record metadata retrieved from the API, and used it at the integration as the main image and the other 'coverImageUrl' as the thumbnail.
Note that "the bigger" image you said is actually the "original image" uploaded when cataloging the resource, which doesn't mean that it's always going to be bigger, although it is the most usual.

odilo.py
ejcepas: I have added a new field to the record metadata obtained through the API: "subjectsBisacCodes", and treated it when importing resources from the API. 
So the codes of the subjects are just the BISAC standard codes.

odilo.py
ejcepas: And what about 'delivery_mechanism_to_internal_format' which is just the inverted map?? There will be two identical keys and the second will overwrite the first one.

odilo.py
ejcepas: 'K-12' is just a metadata example. Yes, we have more specific grade levels as you said.

odilo.py
leonardr: DeliveryMechanism.ADOBE_DRM describes the ACSM file. The first item in this tuple conveys what type of document you get _after_ fulfilling the ACSM file. It's important to convey this information because some clients might not support every format. For example, the SimplyE app currently supports EPUB but not PDF.

We don't have a way to accommodate a situation where the content provider just doesn't know what the media type is. Even if we could accommodate this in the database, OPDS doesn't have a way to represent that you might get one media type or the other and there's no way to know without trying it. Even if we could represent that in OPDS, it wouldn't be very useful because the client needs to know whether it can actually render a book before offering the user the chance to borrow it.

I can think of four resolutions:

1. Set the first media type to Representation.PDF_MEDIA_TYPE. Your books are either PDF or EPUB, and SimplyE (the primary client that will be rendering these books) supports EPUB but not PDF. Once SimplyE supports PDF, we'll know that we can render all the books you offer through this mechanism, and they'll start showing up in the mobile client. We'll need to make sure SimplyE can handle a case where we thought we were getting PDF but we actually get EPUB, but as long as the document itself is served with the correct media type it should be okay.
2. Set the first media type to Representation.EPUB_MEDIA_TYPE. This will make your collection's titles show up in SimplyE immediately, but if a book turns out to be a PDF, there will be an unexplained rendering error when SimplyE actually fulfills the ACSM file and tries to render the document. If the vast majority of your titles are actually EPUBs, this might be good enough as a stopgap until we add PDF support to SimplyE.
3. Set the first media type in the 2-tuple as `application/octet-stream;type=https://bib.schema.org/EBook`. `application/octet-stream` is the most general media type and using it indicates that you don't know the 'real' media type. This has the advantage of always being accurate, but clients are generally not comfortable handling arbitrary binary data.
However, https://www.iana.org/assignments/media-types/application/octet-stream allows for the media type to be given a "type" parameter indicating "the general type or category of binary data". The type is not intended to be processed automatically, but that recommendation was written in 1996 before we had a set of well-known URIs describing every general type of thing that might be delivered over the Internet. My idea is to use the schema.org URI for an ebook (https://bib.schema.org/EBook) to indicate that the document is some kind of ebook. This will signal to a client that's confident it can handle _most_ types of ebooks to say to the user "I can _probably_ render this book but I won't know for sure until you borrow it and I'm able to download it."
We can then change SimplyE to understand `application/octet-stream;type=https://bib.schema.org/EBook` and to give that warning. If the book happens to be an EPUB, then SimplyE will be able to render it. If it's a PDF, or some other unknown data type, then SimplyE will fail to render the book and give an error that makes sense in terms of that initial warning.
4. If you're absolutely certain that PDF and EPUB are the only two things that can be behind an "ACSM" type in your system, we can do solution 3 but use something like `http://odilo.us/terms/epub-or-pdf` as the `type` parameter instead of `https://bib.schema.org/EBook`. This will let us get rid of the warning altogether once we add PDF support to SimplyE.

leonardr: I always forget to mention this. Even after SimplyE adds PDF support, we will not be able to render PDF documents that we obtained through fulfilling an ACSM file. This is a limitation imposed by our contract with Adobe, not a technical limitation. So even with solution #4, SimplyE will always need to issue a warning that it might not be able to open the book after the user borrows it.

odilo.py
ejcepas: We use just "display_name"s, we don't have the "sort_name". I could add it to the API, but it would be "manually", as we don't have this data cataloged... or whatever other solution you suggest.

odilo.py
leonardr: Just to make sure you hook up the two parts correctly, `Subject.identifier` would be the BISAC code, e.g. "ARC000000" and `Subject.name` would be the name of the code in a specific language, e.g. "ARCHITECTURE / General".

odilo.py
leonardr: OK, as I said this is a bit of a problem. We have a web API for guessing `sort_name` from `display_name`. You can activate this API by calling `MetadataWranglerOPDSLookup.from_config(_db, collection)` and then passing the resulting object into `Metadata.apply()` as the `metadata_client` argument.

Behind the scenes, this makes an API call like this:

http://metadata.librarysimplified.org/canonical-author-name?display_name=E.%20Veale&urn=urn:isbn:1234567890

It sends over the display_name and (if possible) the ISBN of the book. The metadata wrangler uses these two pieces of information to query OCLC and VIAF in hopes of finding the author's canonical name (e.g. "Veale, E" on http://viaf.org/viaf/49093834/).

The problem is that this process is quite slow and the metadata wrangler is already overloaded. In the long term this is the right solution but you won't be able to get it to work this week. In the short term, you can use the `display_name_to_sort_name` function defined in `core/util/personal_names.py`, which uses heuristics to convert "E. Veale" into "Veale, E." most of the time.

odilo.py
ejcepas: Yes, I think I'm doing it properly. FYI, I'm sending you through the API two arrays related with subjects:
- subjects --> which are just subject tags.
- subjectsBisacCodes --> which are the BISAC codes of the subjects.

This is so because when cataloging, it is possible to add subjects to a resource without indicating the BISAC code, just add subjects 'tags' in general, so it is possible to have resources with subjects tags without subject code, since them may not belong to the standard.

odilo.py
ejcepas: Ok, so, by the moment I will use the ```display_name_to_sort_name``` function as you suggest me, thank you!

odilo.py
leonardr: Yes, this sounds right.

model.py
```diff
@@ -3358,7 +3358,7 @@ def choose_cover(self):
...
+                            self.primary_identifer,
```
leonardr: This is still wrong; should be `primary_identifier`.

odilo.py
```diff
@@ -425,7 +419,8 @@ def record_info_to_metadata(cls, book, availability):
         author = book.get('author')
         if author:
             roles = [Contributor.AUTHOR_ROLE]
-            contributor = ContributorData(sort_name=author, display_name=author, roles=roles, biography=None)
+            sort_author = display_name_to_sort_name(author)
+            contributor = ContributorData(sort_name=sort_author, display_name=author, roles=roles, biography=None)
```
leonardr: This is all right for now but in the long run we should rely on the name canonicalizer from the metadata wrangler to do this work. I think I've mentioned this before.

model.py
ejcepas: Opsss ! Changed it !

odilo.py
ejcepas: Yes, you told me about it, thanks.

PR #697, opened 2017-10-30
Fix logging on unrecognized Axis format

swandog30:
This branch simply adds a missing error function call when logging an unrecognized Axis format.


leonardr:
What was the unrecognized format name you found? We should add explicit support for it.

swandog30:
The format was 'AxisNow'.  The library is contacting support about it... also have heard that B/T might be able to change this format in the feed.

leonardr:
That sounds like a streaming text format, which we can support (as we do with Overdrive Read).

PR #698, opened 2017-11-01
Create a new BISACClassifier that works against official BISAC subjects

leonardr:
We have a BISACClassifier that works on Bibliotheca subjects, and a subclass of it that work on Axis 360 subjects, but none of these are real, honest-to-goodness BISAC subjects (Axis 360 is close but it doesn't use the identifiers, just the names). This branch introduces a classifier that works on subjects that use the official BISAC identifiers. It's not used for anything yet, but I plan to do follow-up branches that specialize it for Axis and Bibliotheca subjects, and make it possible to understand BISAC subjects that show up in OPDS feeds.

BISAC is a hierarchical classification system, e.g. "Juvenile Nonfiction / Science & Nature / Earth Sciences / Weather". The code in this branch classifies a BISAC subject by splitting it up into pieces and running lists of rules against the pieces to determine genre, audience, etc. This should be generalizable to other hierarchical classification systems, if we have to deal with others in the future.

We do not yet have an official, up-to-date list of all the BISAC codes. I'm using a cobbled-together CSV file that has most of the latest BISAC codes but probably not all of them.

This branch includes two BISACClassifier classes, the one in classifier/__init__.py (which is used by Bibliotheca and Axis) and the one in classifier/bisac.py (which is not used by anyone yet). Eventually the one in __init__.py will be going away, replaced by classifiers in classifier/bisac.py (which should work for Axis) and classifier/bibliotheca.py (which should be able to use the same rules with a different `scrub_name` method).


classifier/bisac.py
```diff
@@ -0,0 +1,605 @@
...
+                    # We are out of subject tokens, and we never found a
+                    # candidate for the next rule token.
+                    return False, rules, subject
+                else:
+                    # That token didn't match, but maybe the next one will.
+                    pass
+
+            # We went through the entire remaining subject and didn't
+            # find a match for the rule token that follows 'anything'.
+            return False, must_match, subject
```
aslagle: I don't think this code is reachable, but if it was reached, `must_match` would be undefined.

classifier/bisac.py
```diff
@@ -0,0 +1,601 @@
...
+
+        # BISAC classifies these genres at the top level, which we
+        # treat as 'nonfiction', but we classify them as fiction. It
+        # doesn't matter because they're neither, really.
+        m(Drama, nonfiction, 'Drama'),
+	m(Poetry, nonfiction, 'Poetry'),
+
+        # Now on to nonfiction.
+
+        # Classify top-level fiction categories into fiction genres.
```
aslagle: this part is nonfiction but has the comment copied from fiction

classifier/bisac.py
```diff
@@ -0,0 +1,601 @@
...
+        m(Music, nonfiction, 'Music'),
+        m(Performing_Arts, nonfiction, 'Performing Arts'),
+
+        # Reference & Study Aids
+        m(Dictionaries, nonfiction, 'Reference', 'Dictionaries'),
+        m(Foreign_Language_Study, nonfiction, 'Foreign Language Study'),
+        m(Law, nonfiction, 'Law'),
+        m(Study_Aids, nonfiction, 'Study Aids'),
+        m(Reference_Study_Aids, nonfiction, 'Reference'),
+        # m(Reference_Study_Aids, nonfiction, 'Linguistics'),
```
aslagle: I think you can get rid of this line? or were you thinking of mapping linguistics to something else?

classifier/bisac.py
```diff
@@ -0,0 +1,601 @@
...
+
+        # Handle cases where Juvenile/YA uses different terms than
+        # would be used for the same books for adults.
+        m(Business, nonfiction, 'Careers'),
+        m(Christianity, nonfiction, "Religious", "Christian"),
+        m(Cooking, nonfiction, "Cooking & Food"),
+        m(Education, nonfiction, "School & Education"),
+        m(Family_Relationships, nonfiction, "Family"),
+        m(Fantasy, fiction, "Fantasy & Magic"),
+        m(Fantasy, fiction, 'Ghost Stories'),
```
aslagle: There is a ghost stories genre, that's what you used for the adult term.

tests/test_classifier_bisac.py
```diff
@@ -0,0 +1,279 @@
...
+        genre_is("Fiction / Science Fiction / Steampunk", "Steampunk")
+        genre_is("Fiction / African American / Urban", "Urban Fiction")
+        genre_is("Fiction / Urban", None)
+        genre_is("History / Modern / 17th Century", "Renaissance & Early Modern History")
+        genre_is("Juvenile Nonfiction / Science & Nature / Fossils", "Nature")
+        genre_is("Juvenile Nonfiction / Science & Nature / Physics", "Science")
+        genre_is("Juvenile Nonfiction / Science & Nature / General", "Science")
+        genre_is("Juvenile Fiction / Social Issues / General", "Life Strategies")
+        genre_is("Juvenile Nonfiction / Social Issues / Pregnancy", "Life Strategies")
+        genre_is("Juvenile Nonfiction / Social Issues / Pregnancy", "Life Strategies")
```
aslagle: you have the same test case twice here

classifier/bisac.py
leonardr: I was thinking of it, but it turned out to be handled by "Language Arts & Disciplines".

classifier/bisac.py
```diff
@@ -0,0 +1,601 @@
...
+            genre = ruleset.match(*name)
+            if genre is cls.stop:
+                return None
+            if genre is not None:
+                return genre
+
+    @classmethod
+    def scrub_name(cls, name):
+        """Split the name into a list of lowercase keywords."""
+        return [x.strip().lower() for x in name.split('/')]
```
aslagle: I didn't see anything that uses the NAMES mapping. Should this check if the name is in fact an identifier, and replace it with the name? Or remove the identifier if the name starts with an identifier?

leonardr: NAMES is used in the test. There is a lot more cleanup work that can happen here, like using different separators or removing stuff that was copied-and-pasted from the BISAC web site, but I haven't seen these specific problems yet.

aslagle: I thought Biblioboard used the identifiers rather than the names?

Feedbooks also uses the identifiers, but they prepend 'FB' to them, so they would need a special scrub_name.

aslagle (approved):

classifier/bisac.py
leonardr: The Feedbooks code would go into scrub_identifier. With them and Biblioboard we should be able to create a Subject that uses both the identifier and the name. 

I can write some code that replaces our built-in name for the identifier (in case the distributor gives the wrong name) but I think that code would happen prior to scrub_name being called, maybe instead of scrub_name being called.

PR #699, opened 2017-11-01
Parse URNs in batches

courte:
This branch supports an attempt to limit the number of repetitive database touches in the Metadata Wrangler by parsing URNs in batches. It also allows identifiers to be added to a collections catalog in batches.

### DIFF
--- a/model.py
+++ b/model.py
@@ -1380,6 +1380,10 @@ class CoverageRecord(Base, BaseCoverageRecord):
         else:
             raise ValueError(
                 "Cannot look up a coverage record for %r." % edition) 
+
+        if isinstance(data_source, basestring):
+            data_source = DataSource.lookup(_db, data_source)
+
         return get_one(
             _db, CoverageRecord,
             identifier=identifier,
--- a/model.py
+++ b/model.py
@@ -1859,6 +1859,45 @@ class Identifier(Base):
         return (type, identifier_string)
 
     @classmethod
+    def parse_urns(cls, _db, identifier_strings):
+        """Batch processes URNs"""
+        failures = list()
+        identifier_details = dict()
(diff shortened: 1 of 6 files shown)


model.py
```diff
@@ -1519,9 +1522,8 @@ def bulk_add(self, works, operation, timestamp=None,
         # The SELECT part of the INSERT...SELECT query.
         new_records = _db.query(
             Work.id.label('work_id'), 
-            literal(operation, type_=BaseCoverageRecord.status_enum).label('operation'),
+            literal(operation, type_=String(255)).label('operation'),
```
leonardr: What's the motivation behind this change?

model.py
```diff
@@ -1859,6 +1861,45 @@ def type_and_identifier_for_urn(cls, identifier_string):
         return (type, identifier_string)
 
     @classmethod
+    def parse_urns(cls, _db, identifier_strings):
```
leonardr: I see how this method saves time when you give it a bunch of identifiers that are already in the catalog. Is that all, or does it also save time when the identifiers are new?

model.py
courte: I made this change when I was exploring bulk inserts. Type is supposed to be the target column type, and the column type of CoverageRecord.operation isn't an status enum, it's String.

If it's working just fine as-is, I can remove it.

leonardr: You're right, I mixed up operation and status originally. Thanks for fixing my error!

model.py
courte: It doesn't save time when identifiers are new, but it could. Right now `Identifier.for_foreign_id` does some final evaluations on the identifier, but I'll refactor to apply them separately.

leonardr:
Looking good!

PR #700, opened 2017-11-02
Treat BISAC-like subjects as BISAC

leonardr:
This branch changes the BISACClassifier so that it can handle BISAC-like classifications from three other sources:

* Axis 360, which is mainly BISAC-compliant but which doesn't use the identifiers (only the names) and which has some copy-and-paste errors in their BISAC names.
* Bibliotheca, which uses an older version of BISAC, also doesn't use the identifiers, and uses pipes instead of slashes in its names, among other differences.
* Feedbooks, which uses standard BISAC identifiers prefixed with "FB", and also has its own custom classifications that are not in BISAC at all.

Apart from adding a lot of code to handle specific weird cases, I made three main changes to support this:

* I introduced `Classifier.scrub_identifier_and_name` to handle the case where `scrub_identifier` is able to recognize the identifier as being from a controlled vocabulary where each identifier has a canonical name. When that happens, `scrub_identifier` can return a 2-tuple (identifier, name) to ensure that the canonical name (once it itself is scrubbed) is used instead of whatever random name the distributor may have provided.

* I created `CustomMatchToken` to handle the increasingly common BISAC classification case where two or more strings should be treated as the same string. This is how I deal with changes in the names given to BISAC subjects.

* If BISACClassifier can't classify something, it runs it through the KeywordBasedClassifier before giving up.

I'm about to do a head-to-head test to make sure there are no regressions (or an acceptably small number of regressions) when running the existing Bibliotheca and BISAC classifications through the new system; then I'll do a branch that converts them over.

### DIFF
--- a/classifier/bisac.py
+++ b/classifier/bisac.py
@@ -629,15 +629,61 @@ class BISACClassifier(Classifier):
             identifier, keyword, fiction, audience
         )
 
+    # A BISAC name copied from the BISAC website may end with this
+    # human-readable note, which is not part of the official name.
+    see_also = re.compile('\(see also .*')
+
     @classmethod
     def scrub_identifier(cls, identifier):
         if identifier.startswith('FB'):
-            return identifier[2:]
+            identifier = identifier[2:]
+        if identifier in cls.NAMES:
+            # We know the canonical name for this BISAC identifier,
+            # and we are better equipped to classify the canonical
+            # names, so use the canonical name in preference to
+            # whatever name the distributor provided.
+            return (identifier, cls.NAMES[identifier])
         return identifier
 
     @classmethod
     def scrub_name(cls, name):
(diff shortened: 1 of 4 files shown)


classifier/bisac.py
```diff
@@ -5,13 +5,57 @@
...
+        return self.re.search(subject_token)
+
+class Interchangeable(CustomMatchToken):
+    """A token that matches a list of strings."""
+    def __init__(self, *choices):
+        """All of these strings are interchangeable for matching purposes."""
+        self.choices = set(map(string.lower, choices))
+
+    def matches(self,subject_token):
+        return subject_token in self.choices
```
aslagle: I feel like this should use subject_token.lower, rather than relying on outside code doing that. Especially since the other CustomMatchToken classes aren't case-sensitive.

PR #701, opened 2017-11-02
Database lanes updates

aslagle:
More changes to get database lanes working in circulation:

- Changed the way search works so that WorkList is searchable and has the implementation of search. When searching a Lane, it will use the `search_target`, which will be any ancestor that is the root for a patron type, or a WorkList created using some of the Lane's restrictions.
- Brought back a few things that were removed but still used - `language_key`, `depth`.
- Added code to handle a Lane like "Children's Informational Books", which has a fiction restriction and an excluded genre (Biography) but no included genres.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -289,8 +289,8 @@ class ExternalSearchIndex(object):
 
         return base_works_index
 
-    def query_works(self, library, query_string, media, languages, fiction, audience,
-                    age_range, in_any_of_these_genres=[], fields=None, size=30, offset=0):
+    def query_works(self, library, query_string, media, languages, fiction, audiences,
+                    target_age, in_any_of_these_genres=[], fields=None, size=30, offset=0):
         if not self.works_alias:
             return []
 
@@ -307,7 +307,7 @@ class ExternalSearchIndex(object):
 
         filter = self.make_filter(
             collection_ids, media, languages, fiction, 
-            audience, age_range, in_any_of_these_genres
+            audiences, target_age, in_any_of_these_genres
         )
         q = dict(
             filtered=dict(
@@ -551,7 +551,7 @@ class ExternalSearchIndex(object):
             }
         }
         
-    def make_filter(self, collection_ids, media, languages, fiction, audience, age_range, genres):
+    def make_filter(self, collection_ids, media, languages, fiction, audiences, target_age, genres):
         def _f(s):
             if not s:
                 return s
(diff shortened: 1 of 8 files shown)


external_search.py
```diff
@@ -289,8 +289,8 @@ def base_index_name(self, index_or_alias):
 
         return base_works_index
 
-    def query_works(self, library, query_string, media, languages, fiction, audience,
-                    age_range, in_any_of_these_genres=[], fields=None, size=30, offset=0):
+    def query_works(self, library, query_string, media, languages, fiction, audiences,
+                    target_age, in_any_of_these_genres=[], fields=None, size=30, offset=0):
```
aslagle: I changed a couple arguments to match how this was being called.

aslagle:
Also: added a unique key to CachedFeed for feeds associated with a contributor or series lane.

lane.py
```diff
@@ -471,6 +486,16 @@ def parentage(self):
         return []
 
     @property
+    def language_key(self):
+        """Return a string identifying the languages used in this WorkList.
+        This will usually be in the form of 'eng,spa' (English and Spanish).
+        """
+        key = ""
+        if self.languages:
+            key += ",".join(self.languages)
```
leonardr: I think this should be sorted.

lane.py
```diff
@@ -981,7 +1063,19 @@ def parentage(self):
...
+        """How deep is this lane in this site's hierarchy?
+        i.e. how many times do we have to follow .parent before we get None?
+        """
+        depth = 0
+        tmp = self
+        while tmp.parent:
+            depth += 1
+            tmp = tmp.parent
+        return depth    
+
```
leonardr: Would `len(self.parentage)` be a simpler way to write this?

aslagle: I think it would. I copied this from master without looking at it.

leonardr: Fair enough.

model.py
```diff
@@ -6095,6 +6095,10 @@ class CachedFeed(Base):
     # A feed is of a certain type--currently either 'page' or 'groups'.
     type = Column(Unicode, nullable=False)
 
+    # A feed associated with a WorkList can have a unique key.
+    # This should be null if the feed is associated with a Lane.
+    unique_key = Column(Unicode, nullable=True)
+
```
leonardr: Does this need a migration or will the new field be created automatically?

PR #704, opened 2017-11-03
Bibliotheca and Axis subjects are BISAC

leonardr:
This branch changes all existing BISAC identifiers (obtained from Axis 360) so that what used to be the 'identifier' is now the 'name'. It converts all 3M subjects (obtained from Bibliotheca) into BISAC subjects and gets rid of all the code that handles "3M" subjects (the new BISAC classifier does a much better job on those subjects).

BISAC and Feedbooks categories are classified as BISAC when found in OPDS feeds.

Bibliotheca and Axis 360 now register the subjects they find as BISAC subjects, with the textual description in the `name` and no `identifier` (since those data sources don't provide the identifier). I changed all `scrub_identifier` implementations to be able to handle a null `identifier`.

As I removed the ThreeMClassifier I took one more pass through its classifications and added a few more things to the BISACClassifier that seemed like good ideas.

Finally, this branch adds a script that reclassifies any Works whose original classification was based on Identifiers that have since been marked as unchecked. This gives us an easy way to mark Works as needing reclassification during a migration, without bogging down the migration itself with reclassifying thousands of works.


tests/test_model.py
```diff
@@ -3219,7 +3219,30 @@ def test_update_external_index(self):
...
+
+        # This Subject hasn't been checked, so the Work associated with
+        # it shows up.
+        ds = DataSource.lookup(self._db, DataSource.OVERDRIVE)
+        classification = identifier.classify(ds, Subject.TAG, "some tag")
+        classification2 = identifier.classify(ds, Subject.TAG, "another tag")
+        eq_([w1], qu.all())
+
+        # Mark it as checked, and the Work stops showing up.
+        classification.subject.checked = True
```
aslagle: I don't understand why you don't also have to mark classification2 as checked, since the identifier is under both.

leonardr: Maybe this is related to why my tests are failing... :|

aslagle: I'd like to see the test check that the work still shows up when only one subject is marked as checked, if that's what's supposed to happen.

leonardr:
This fixes https://github.com/NYPL-Simplified/circulation/issues/141.

PR #706, opened 2017-11-06
Renamed default_loan_period to ebook_loan_duration.

leonardr:
This branch replaces the `default_loan_period` setting with `audio_loan_duration` and `ebook_loan_duration`. The one used to estimate loan durations is `ebook_loan_duration`. In the circulation module, individual `CirculationAPI` subclasses will decide whether these settings are purely advisory, to be used only as estimates, or whether they can be used to _determine_ how long loans should be.

### DIFF
--- a/migration/20171106-rename-default-loan-period.sql
+++ b/migration/20171106-rename-default-loan-period.sql
@@ -1,0 +1,1 @@
+update collectionsettings set name='ebook_loan_duration' where name='default_loan_period';
--- a/migration/20171106-rename-default-loan-period.sql
+++ b/migration/20171106-rename-default-loan-period.sql
@@ -1,1 +1,0 @@
-update collectionsettings set name='ebook_loan_duration' where name='default_loan_period';
--- a/model.py
+++ b/model.py
@@ -10447,44 +10447,33 @@ class Collection(Base, HasFullTableCache):
     EBOOK_LOAN_DURATION_KEY = 'ebook_loan_duration'
     STANDARD_DEFAULT_LOAN_PERIOD = 21
         
-    @hybrid_property
-    def default_loan_period(self):
+    def default_loan_period(self, library):
         """Until we hear otherwise from the license provider, we assume
         that someone who borrows a non-open-access item from this
         collection has it for this number of days.
         """
         return (
-            self.external_integration.setting(
-                self.EBOOK_LOAN_DURATION_KEY).int_value
-            or self.STANDARD_DEFAULT_LOAN_PERIOD
+            ConfigurationSetting.for_library_and_externalintegration(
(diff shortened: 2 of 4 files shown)


migration/20171106-rename-default-loan-period.sql
```diff
@@ -0,0 +1 @@
+update collectionsettings set name='ebook_loan_duration' where name='default_loan_period';
```
leonardr: This won't actually do anything because ebook_loan_duration is a per-library per-collection setting, and default_loan_period is a per-collection setting. So I could just delete the old settings.

aslagle: If it matters you could copy the old setting for every library and then delete it. Not sure if anyone is using it.

model.py
```diff
@@ -10453,15 +10455,15 @@ def default_loan_period(self):
         """
         return (
             self.external_integration.setting(
-                self.DEFAULT_LOAN_PERIOD_KEY).int_value
+                self.EBOOK_LOAN_DURATION_KEY).int_value
```
aslagle: This doesn't look like it's per-library?

aslagle: It's the default for the collection but libraries can change it?

leonardr: Aaaargh.

I don't know how to solve this without ripping up the whole estimate system -- a system that we know would remain broken in many ways even if this were fixed.

migration/20171106-rename-default-loan-period.sql
leonardr: I think I would rather either ignore it or delete it because the old description was vague enough that people didn't know what it did.

leonardr:
Ok, I think this is ready again. The loan durations are now properly per-library per-collection settings and the OPDS feed generator passes in the appropriate library. The reservation duration is just a per-collection setting (which isn't used at all currently).

model.py
```diff
@@ -10437,35 +10437,32 @@ def set_protocol(self, new_protocol):
...
-        new_value = int(new_value)
-        self.external_integration.setting(
-            self.DEFAULT_LOAN_PERIOD_KEY).value = str(new_value)
-
     DEFAULT_RESERVATION_PERIOD_KEY = 'default_reservation_period'
     STANDARD_DEFAULT_RESERVATION_PERIOD = 3
-            
+
     @hybrid_property
     def default_reservation_period(self):
```
aslagle: so default loan period is per-library per-collection, but default reservation period is still per-collection?

leonardr: Yes, mainly because default reservation period is barely used and I don't want to expose it in the user interface at all, so there's no need to touch it. I think it can be eventually replaced by a per-distributor constant but I'm not sure.

PR #709, opened 2017-11-07
Improve RBdigital subject classification

leonardr:
This branch creates and registers proper classifiers for "RBdigital" and "RBdigital audience" type subjects. There are a small number of subjects in both categories so I was able to determine the classification rules by exhaustion, running spot checks whenever I was unsure about a subject like 'beginning readers'.

The RBdigital-specific subjects seem more accurate than the BISAC subjects also provided by RBdigital, so I've changed the weights of new RBdigital-specific subjects going forward.

The "RBdigital" classifier falls back to the standard KeywordBasedClassifier and this allowed me to find some errors in how the KeywordBasedClassifier classifies things. This should improve classifications for `tag` and other subject types that use the KeywordBasedClassifier.

### DIFF
--- a/classifier/__init__.py
+++ b/classifier/__init__.py
@@ -43,7 +43,7 @@ class Classifier(object):
     LCSH = "LCSH"
     FAST = "FAST"
     OVERDRIVE = "Overdrive"
-    ONECLICK = "OneClick"
+    RBDIGITAL = "RBdigital"
     BISAC = "BISAC"
     BIC = "BIC"
     TAG = "tag"   # Folksonomic tags.
@@ -54,7 +54,7 @@ class Classifier(object):
     GRADE_LEVEL = "Grade level" # "1-2", "Grade 4", "Kindergarten", etc.
     AGE_RANGE = "schema:typicalAgeRange" # "0-2", etc.
     AXIS_360_AUDIENCE = "Axis 360 Audience"
-    ONECLICK_AUDIENCE = "OneClick Audience"
+    RBDIGITAL_AUDIENCE = "RBdigital Audience"
 
     # We know this says something about the audience but we're not sure what.
     # Could be any of the values from GRADE_LEVEL or AGE_RANGE, plus
@@ -2235,9 +2235,6 @@ class KeywordBasedClassifier(AgeOrGradeClassifier):
                    Eg("handwriting"),
                    Eg("information sciences"),
                    Eg("journalism"),
-                   Eg("language arts & disciplines"),
-                   Eg("language arts and disciplines"),
-                   Eg("language arts"),
                    Eg("library & information sciences"),
                    Eg("linguistics"),
                    Eg("literacy"),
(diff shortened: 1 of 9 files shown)


classifier/__init__.py
```diff
@@ -2235,9 +2235,6 @@ class KeywordBasedClassifier(AgeOrGradeClassifier):
                    Eg("handwriting"),
                    Eg("information sciences"),
                    Eg("journalism"),
-                   Eg("language arts & disciplines"),
-                   Eg("language arts and disciplines"),
-                   Eg("language arts"),
```
leonardr: These weren't being triggered because 'art' was taking them.

classifier/__init__.py
```diff
@@ -2444,11 +2444,7 @@ class KeywordBasedClassifier(AgeOrGradeClassifier):
                    "travelers",
                    "description.*travel",
                ),
-               
-               True_Crime: match_kw(
-                   "true crime",
-               ),
-
```
leonardr: This wasn't being triggered because 'Mystery' was taking 'crime'.

classifier/__init__.py
```diff
@@ -2982,6 +2992,8 @@ def audience(cls, identifier, name):
 
     @classmethod
     def target_age(cls, identifier, name):
+        if identifier == 'beginning reader':
+            return cls.range_tuple(5,8)
```
leonardr: This isn't what RBdigital means by "beginning reader" -- it's the industry standard seen elsewhere in this file.

classifier/rbdigital.py
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    genres = {
+        # This isn't true in general but because RBdigital is heavy on
+        # audio content, 'arts and entertainment' is more likely to be
+        # entertainment (e.g. music) than arts.
+        'arts entertainment' : Entertainment,
+        'business economics' : Business,
+        'comics graphic novels' : Comics_Graphic_Novels,
+        'home garden' : Hobbies_Home,
```
leonardr: These are RBdigital subjects where I didn't feel comfortable adding the appropriate tokens to the KeywordBasedClassifier (as I did with e.g. "inspirational nonfiction").

classifier/rbdigital.py
```diff
@@ -0,0 +1,97 @@
...
+    @classmethod
+    def audience(cls, identifier, name):
+        """The two subjects here that imply an audience, juvenile-fiction and
+        juvenile-nonfiction, could be either Childrens or YA, and
+        RBdigital always provides a more specific audience
+        classification, so we don't try to derive audience information
+        from an RBDigital Subject.
+        """
+        if identifier == 'erotica':
+            return Classifier.AUDIENCE_ADULTS_ONLY
```
leonardr: This isn't necessary--I didn't even see any "erotica" in my RBdigital collection and it'll be handled by the caller anyway--but just to be double-safe.

model.py
```diff
@@ -10851,6 +10852,8 @@ def numericrange_to_tuple(r):
 
 def tuple_to_numericrange(t):
     """Helper method to convert a tuple to an inclusive NumericRange."""
+    if not t:
+        return None
```
leonardr: This allows `target_age` to return None if it wants a subject to not be considered for target age at all.

PR #716, opened 2017-11-09
Bibliotheca audiobooks have distinct delivery mechanism

leonardr:
This is support work for https://github.com/NYPL-Simplified/circulation/pull/771. Bibliotheca's "MP3" books are now recorded as having Findaway DRM, not Adobe DRM.

### DIFF
--- a/config.py
+++ b/config.py
@@ -10,6 +10,7 @@ from sqlalchemy.engine.url import make_url
 from flask_babel import lazy_gettext as _
 
 from facets import FacetConstants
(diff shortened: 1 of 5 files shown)


config.py
```diff
@@ -10,6 +10,7 @@
 from flask_babel import lazy_gettext as _
 
 from facets import FacetConstants
+from sqlalchemy.exc import ArgumentError
```
leonardr: An unrelated change to add a missing import. This error is caught if your database environment variable is in the wrong format.

PR #721, opened 2017-11-17
Site configuration only changed on direct modifications

leonardr:
This branch stops `site_configuration_was_changed` from being triggered on any random change to a collection associated with an affected object. We already explicitly monitor the collections whose membership is important for site configuration purposes, by listening for `append` and `remove` objects.


leonardr:
The existing event monitors take care of all the cases I was worried about, so I don't think it's necessary to also add application-level invalidation code at this point.

PR #723, opened 2017-11-20
Message doesnt mean failure

leonardr:
This branch changes the behavior of the OPDSImporter to allow for the possibility that a <simplified:message> should be treated as success rather than failure. A subclass of OPDSImporter can cause this behavior by setting SUCCESS_STATUS_CODES to a list of status codes that indicate success rather than failure. An OPDSImporter that finds a message that indicates success will return the Identifier that caused the message, rather than creating a CoverageFailure from the message (or returning None, in the case where the status code is 200).

This branch also adds a new attribute of CoverageProviders, `COVERAGE_COUNTS_FOR_EVERY_COLLECTION`. If this is true (the default), then an Identifier will be covered once and that's it, no matter how many Collections contain the Identifier. As soon as there's a CoverageRecord for the Identifier that has no associated collection, then the work is done. If `COVERAGE_COUNTS_FOR_EVERY_COLLECTION` is false, then an Identifier will get separate coverage for every Collection that contains it.

The metadata wrangler coverage providers have `COVERAGE_COUNTS_FOR_EVERY_COLLECTION` set to false because their job is to tell the metadata wrangler which Identifiers are in which local Collections. As far as I know, all other coverage providers should have  `COVERAGE_COUNTS_FOR_EVERY_COLLECTION` set to true.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -431,6 +431,13 @@ class IdentifierCoverageProvider(BaseCoverageProvider):
     # Identifier in the system, which is probably not what you want.
     NO_SPECIFIED_TYPES = object()
     INPUT_IDENTIFIER_TYPES = NO_SPECIFIED_TYPES
+
+    # Set this to True if a given Identifier needs to be run through
+    # this CoverageProvider once for every Collection that has this
+    # Identifier in its catalog. Otherwise, a given Identifier will be
+    # considered completely covered the first time it's run through
+    # this CoverageProvider.
+    COVERAGE_COUNTS_FOR_EVERY_COLLECTION = True
     
     def __init__(self, _db, collection=None, input_identifiers=None,
                  replacement_policy=None, preregistered_only=False, **kwargs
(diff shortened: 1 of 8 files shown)


model.py
```diff
@@ -1309,7 +1309,6 @@ class CoverageRecord(Base, BaseCoverageRecord):
 
     SET_EDITION_METADATA_OPERATION = u'set-edition-metadata'
     CHOOSE_COVER_OPERATION = u'choose-cover'
-    SYNC_OPERATION = u'sync'
```
leonardr: My circulation branch removes the CoverageProvider that provides this operation, but there's some other code (the reaper) that looked for the CoverageRecords it creates. Removing the constant ensures that I changed all of that code.

coverage.py
```diff
@@ -535,15 +552,18 @@ def register(cls, identifier, collection=None):
         operation = cls.OPERATION
 
         was_registered = False
-        existing_record = CoverageRecord.lookup(identifier, source, operation)
+        existing_record = CoverageRecord.lookup(
+            identifier, source, operation, collection=collection
+        )
```
leonardr: This doesn't use collection_or_not because it's a class method.

tests/test_model.py
```diff
@@ -5833,19 +5833,43 @@ def test_lookup(self):
         source = DataSource.lookup(self._db, DataSource.OCLC)
```
leonardr: NOTE: If there is no collection on a CoverageRecord, you can look it up by passing in collection=None but not by passing in some other thing.

PR #724, opened 2017-11-21
Allow URNLookupController.process_urns to return a ProblemDetail

courte:
This branch catches differences in the metadata wrangler, where `URNLookupController.process_urns` sometimes returns a ProblemDetail.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -301,6 +301,11 @@ class URNLookupController(object):
         return feed_response(opds_feed)
 
     def process_urns(self, urns, **process_urn_kwargs):
+        """Processes a list of URNs for a lookup request.
+
+        :return: None or, to override default feed behavior, a ProblemDetail
+        or Response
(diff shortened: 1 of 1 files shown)


app_server.py
```diff
@@ -270,9 +270,13 @@ def work_lookup(self, annotator, route_name='lookup', **process_urn_kwargs):
         urns = flask.request.args.getlist('urn')
 
         this_url = cdn_url_for(route_name, _external=True, urn=urns)
-        self.process_urns(urns)
+        response = self.process_urns(urns)
         self.post_lookup_hook()
 
+        if response:
+            # In a subclass, self.process_urns may return a ProblemDetail
+            return response
```
leonardr: Can you add a docstring to process_urns saying that it can `:return:` a ProblemDetail or a Response if it wants to override the default behavior?

PR #726, opened 2017-11-22
Fix some minor problems encountered when testing lookup

leonardr:


app_server.py
```diff
@@ -270,7 +270,7 @@ def work_lookup(self, annotator, route_name='lookup', **process_urn_kwargs):
         urns = flask.request.args.getlist('urn')
 
         this_url = cdn_url_for(route_name, _external=True, urn=urns)
-        response = self.process_urns(urns)
+        response = self.process_urns(urns, **process_urn_kwargs)
```
leonardr: Without this, collection information obtained from the URL is not propagated from the metadata wrangler to process_urns.

opds.py
```diff
@@ -848,6 +848,9 @@ def _create_entry(self, work, license_pool, edition, identifier,
 
     def _make_entry_xml(self, work, license_pool, edition, identifier):
 
+        if work and not edition:
+            edition = work.presentation_edition
+
```
leonardr: I think this only happens on a few titles. This is indicative of a failure elsewhere in the system, and this doesn't fix that failure. So if there's a downside to this change, it's that it pushes the point of eventual failure further away. But for the metadata wrangler it seems to push the point of failure so far away that it won't be encountered with our current functionality.

For the record, the work on the metadata wrangler that appears to have this problem is https://unglue.it/api/id/work/130412/, "It's Complicated" by danah boyd.

PR #738, opened 2017-11-30
Find featured works using a single query

leonardr:
This branch replaces the code for finding featured works. Previously this code used a sequence of `Facets` objects to try more and more desperately to find works worthy of being featured. The new code uses a single query managed by the new `FeaturedFacets` class. This query orders works by a dynamically calculated numeric score that puts more feature-worthy works at the front.

I don't think this branch will improve a normal site's performance all that much, since most libraries have enough featurable works that they're picked up in the first query. However, we're now only one step away from using a single query to get featurable works for _all_ lanes at once (or at least a large number of lanes at once).


lane.py
```diff
@@ -328,6 +329,87 @@ def order_by(self, work_model):
...
+
+        Using this field in an ORDER BY statement ensures that
+        higher-quality works show up at the beginning of the
+        results. But if there aren't enough high-quality works,
+        lower-quality works will show up later on in the results,
+        eliminating the need to find lower-quality works with a second
+        query.
+
+        :param featurable_quality: The quality score necessary for a
+        work to be considered 'featurable'.
```
aslagle: This is no longer a param of this method.

aslagle (approved):

Issue #482, opened 2017-03-23
Searches that don't work well


aslagle:
I'm making this issue to keep track of specific searches that don't work well.

Here's a report from an app review:
"I looked up books by M.J. Rose and it gave me one, but when I looked up specific titles by the author they had them! I looked up Steve Berry, it gave me a few of his books and then others by authors whose names were not even close to it."

jbdalton:
"python programming"

Lots of good results in the first two dozen, then hits a bumpy long tail with matches on titles "Phthor" and "Pythias" sorting above titles like "Think Python: How to think like a Computer,"  "Python Data Science Handbook," etc.

jbdalton:
Looking more closely at "Steve Berry" and other search examples, I think the following changes, combined or separately tried, might help filter out some of the louder noise in the Elasticsearch results. 

* Change the fuzzy_prefix_length from 0 to 1.

    A fuzzy_prefix_length=1 will stop "tennis" from matching Dennis Rodman or the title "Kenny" (currently ranked at 14 in a "tennis" search). It basically forces an edge-ngram by not including the first character in the ngram analysis.

* Change the maximum fuzziness parameter to 1 from its AUTO setting.

    This will stop "tennis" from matching on the author "Shyalpa Tenzin Rinpoche." And visa versa, a search for "Tenzin" won't return the non-relevant "tennis," since the Levenshtein distance will be calculated only 1 edit away from the original input for strings larger than 5 characters, instead of the default 2 edits when it is set to AUTO.

Reviewing some of the options for the tokenizers being used when indexing might help weed out other tangential query results. For example, from https://www.elastic.co/blog/found-fuzzy-search:

> If one were to use a fuzzy query over an ngram analyzed field, the results would likely be bizarre, as ngrams break words up into many small letter combinations, many of which are only an edit or two away, though the actual words involved are quite dissimilar.

leonardr:
Searches for book titles that include genre names don't work well. For example, searching for "modern romance" turns up a lot of romance books but not the book called _Modern Romance_.

leonardr:
`the mountain man`, `mountain man`, and `law of the mountain man` get dramatically different results

leonardr:
Searching for `game of thrones` generally gives good results but exposes a number of issues:

[...]

PR #739, opened 2017-11-30
Full title match is better than split match

leonardr:
This branch adds two new ways a book might show up in search results: its title or one of its authors might be a near-exact match against the search term.

This should resolve the 'modern romance' and 'law of the mountain man' problems described in https://github.com/NYPL-Simplified/server_core/issues/482, where a book whose title includes a genre loses out to books that are in that genre. It may also resolve some of the 'game of thrones' problems described in that issue.

This branch changes the Elasticsearch configuration used by core during Travis runs so that everything runs in a single shard; this eliminates the possibility that search result ordering will be unpredictable because of how documents are assigned to shards.

This branch also changes the way Elasticsearch is configured: instead of separate configuration settings for the alias and the index, a single configuration setting controls the prefix used for both the alias and the index. This means that changing the alias to point to some other index is currently an unsupported operation -- it's supported by the code (transfer_current_alias) but nobody ever calls that method.

This is a good time to make that configuration change because this branch introduces a v3 of the search document, which requires that the entire index be rebuilt anyway. (The migration script should take care of this.)

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -450,9 +450,9 @@ class ExternalSearchIndex(object):
 
         # An exact title or author match outweighs a match that is split
         # across fields.
-        match_title = make_phrase_query(query_string, ['title.standard'], 100)
+        match_title = make_phrase_query(query_string, ['title.standard'], 200)
         must_match_options.append(match_title)
-        match_author = make_phrase_query(query_string, ['author.standard'], 100)
+        match_author = make_phrase_query(query_string, ['author.standard'], 200)
         must_match_options.append(match_author)
 
         if not fuzzy_blacklist_re.search(query_string):
@@ -538,7 +538,7 @@ class ExternalSearchIndex(object):
             match_classification_and_rest_of_query = {
                 'bool': {
                     'must': classification_queries,
-                    'boost': 100
+                    'boost': 200
                 }
             }
 
--- a/external_search.py
+++ b/external_search.py
@@ -339,7 +339,7 @@ class ExternalSearchIndex(object):
                 }
             }
 
-        def make_phrase_query(query_string, fields):
+        def make_phrase_query(query_string, fields, boost=100):
             field_queries = []
             for field in fields:
                 field_query = {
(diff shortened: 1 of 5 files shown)


external_search.py
```diff
@@ -453,6 +453,13 @@ def make_target_age_query(target_age):
         match_phrase = make_phrase_query(query_string, ['title.minimal', 'author', 'series.minimal'])
         must_match_options.append(match_phrase)
 
+        # An exact title or author match outweighs a match that is split
+        # across fields.
+        match_title = make_phrase_query(query_string, ['title.minimal'], 150)
```
aslagle: What happened to `title.standard`?

aslagle:
I think you need a migration for this - for the configuration setting change, if nothing else.

leonardr:
I forgot to add the migration script. All it does is create a brand new index using the default prefix. I don't think there are any instances where this won't work, and if there are any, we don't know enough about their setup to automatically migrate them.

leonardr:
I could also add a migration script to _remove_ the old configuration settings, just to avoid any confusion when looking at the database in the future.

tests/test_external_search.py
```diff
@@ -778,6 +836,143 @@ def query(*args, **kwargs):
...
+            title="Gumby In Love",
+            authors="Pokey",
+            audience=Classifier.AUDIENCE_YOUNG_ADULT, genre="Romance"
+        )
+        self.ya_romance.presentation_edition.subtitle = (
+            "Modern Fairytale Series, Book 3"
+        )
+
+        # TODO: Uncomment these lines and the 'modern romance'
+        # test fails for some reason.
```
aslagle: you fixed this!

external_search.py
```diff
@@ -751,6 +753,64 @@ def map_fields(cls, fields, field_description):
...
+                "type": "string",
+                "analyzer": "en_analyzer",
+                "fields": {
+                    "minimal": {
+                        "type": "string",
+                        "analyzer": "en_minimal_analyzer"},
+                    "standard": {
+                        "type": "string",
+                        "analyzer": "standard"
+                    }
```
leonardr: This is the part that's different.

PR #741, opened 2017-12-04
Seemingly redundant migration script shouldn't crash entire migration process.

leonardr:
The migration script `20171004-add-customlists-library-id.sql` seems to have already run, or to not need running, on all extant circulation manager databases, but it's still sometimes run in the migration to 2.2.0.

On the BPL production database (which I haven't migrated yet), the database migration timestamp is "2017-10-04". Is it possible that the migration script was run once on October 4th, but running the migration again now will run it again?

This branch changes the migration script so that if the field it's trying to create already exists, it succeeds with a message to verify that the script has already been run, rather than failing and screwing up the whole migration.

This branch also adds a new migration script which associates any existing 'Staff Picks' lists with the default library unless they're already associated with some other library.

### DIFF
--- a/migration/20171004-add-customlists-library-id.sql
+++ b/migration/20171004-add-customlists-library-id.sql
@@ -1,12 +1,21 @@
-alter table customlists add column library_id integer;
+DO $$ 
+    BEGIN
+        BEGIN
+            ALTER TABLE customlists ADD COLUMN library_id integer;
+	    alter table customlists
+	        add constraint customlists_library_id_fkey
+		foreign key (library_id)
+		references libraries(id);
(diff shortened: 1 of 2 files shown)


courte (approved):
If the database migration timestamp is "2017-10-04", then the migration shouldn't be running at all. I wonder if this is a migration timestamp logic problem.

leonardr:
Could it be an edge case that happens when you migrate a site the same day you add a migration script?

courte:
That shouldn't be possible. Migration timestamping is based entirely on the filename and doesn't look at the real-time date at all.

PR #744, opened 2017-12-06
Cache lane sizes

leonardr:
This branch creates a new field, `Lane.size`, which contains the pre-calculated number of works that show up in a lane. This number is updated whenever the materialized views are refreshed.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1254,6 +1254,10 @@ class Lane(Base, WorkList):
             return True
         return False        
 
+    def update_size(self, _db):
+        """Update the stored estimate of the number of Works in this Lane."""
+        self.size = fast_query_count(self.works(_db).limit(None))
+
     @property
     def genre_ids(self):
         """Find the database ID of every Genre such that a Work classified in
--- a/lane.py
+++ b/lane.py
@@ -577,11 +577,17 @@ class WorkList(object):
 
     @property
     def full_identifier(self):
-        """A human-readable identifier for this Lane that
+        """A human-readable identifier for this WorkList that
         captures its position within the heirarchy.
         """
-        full_parentage = list(self.parentage) + [self]
-        return " / ".join([x.display_name for x in full_parentage])
+        lane_parentage = list(self.parentage) + [self]
+        full_parentage = [x.display_name for x in lane_parentage]
+        if getattr(self, 'library', None):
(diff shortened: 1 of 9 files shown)


lane.py
```diff
@@ -577,11 +577,17 @@ def parentage(self):
...
-        full_parentage = list(self.parentage) + [self]
-        return " / ".join([x.display_name for x in full_parentage])
+        lane_parentage = list(self.parentage) + [self]
+        full_parentage = [x.display_name for x in lane_parentage]
+        if getattr(self, 'library', None):
+            # This WorkList is associated with a specific library.
+            # incorporate the library's name to distinguish between it
+            # and other lanes in the same position in another library.
+            full_parentage.insert(0, self.library.short_name)
+        return " / ".join(full_parentage)
```
leonardr: This makes output less confusing; multiple libraries on a single circulation manager are likely to have lanes with identical display names.

scripts.py
```diff
@@ -601,6 +602,30 @@ def process_library(self, library):
         raise NotImplementedError()
 
 
+class LaneSweeperScript(LibraryInputScript):
```
leonardr: Most of this code is brought in from circulation, from the parent class for the scripts that cache OPDS feeds for each lane. My original intention was to have a separate script that updated the counts for each lane. But I think it's better to update the counts only when the materialized views are updated, since that's the only time the counts are likely to change significantly.

util/opds_writer.py
```diff
@@ -56,13 +56,13 @@ def add_link_to_feed(cls, feed, children=None, **kwargs):
 
     @classmethod
     def add_link_to_entry(cls, entry, children=None, **kwargs):
+        kwargs = dict([unicode(x), unicode(y)] for x, y in kwargs.items())
```
leonardr: I'm not sure about this change. In my circulation branch, an exception was being raised here because the title of a link was a localized string like `l"All Fiction"`. We should be able to handle that, but this is kind of a ham-handed way to manage this problem and I don't understand why my circulation branch was the first time I've ever seen this problem. It seems like `l"All Fiction"` gets passed into here a lot.

aslagle: We haven't been localizing strings in the OPDS feeds yet as far as I know. I don't think this will hurt anything but it might be better to limit it to the title.

PR #748, opened 2017-12-15
Single query featured feed

leonardr:
This branch makes it possible, in most cases, to build a groups feed for a Lane using a single database query. I'd like some code review before I fix the rest of the tests because I have concerns about whether this will work in a real scenario.

I realize this is a really complicated branch but I have tests of all the new code--tests I wouldn't normally write for a proof of concept--which should help.

Here's the code I wrote: we make a query that matches all works that belong in the lane. We add two fields to that query. The first is the 'quality tier' field which was introduced in a previous branch. The second is the 'lane id' field. The 'lane id' field assigns a work to one of the lanes that contains that book. To match a given lane, the work must match that lane's bibliographic filter clause, and its `.random` field must fall within a randomly selected window whose size is proportional to the size of the lane. If there are 1000 romance books and 100 horror books, the window for the 'Romance' lane will be 1/10 the size of the window for the 'horror' lane. This guarantees (? :grimacing:) that we get approximately the same number of results for each lane.

This branch introduces database schema changes:

* The `mv_works_editions_workgenres_datasources_identifiers` materialized view is rebuilt to do an outer join against `workgenres`. This means works that aren't classified under any particular genre can be picked up by a query against `mv_works_editions_workgenres_datasources_identifiers`. This is important because a single lane may have sublanes that are genre-based plus sublanes that aren't.

* `lanes.inherit_parent_restrictions` is now True by default. The single-query thing only saves time if the parent lane's query also matches works that belong to the vast majority of its sublanes. When `inherit_parent_restrictions` is False, we have to run a separate query just for that sublane, and we lose the performance improvement. Fortunately, the only case where `lanes.inherit_parent_restrictions` really needs to be False is when a librarian wants to make a lane of books that wouldn't normally be classified in a certain place. In all other cases we can assume that a sublane contains a proper subset of the works available to its parent.

Some things I'm not sure about:

* What happens when a work is classified under multiple genres? This shouldn't cause problems but I don't think I tested it.
* What happens when you try to build a groups feed for a lane based on lists? It should work, but I didn't try it.
* What is the performance of this code on a large dataset? It better be pretty good to be worth the extra complexity.
* Does it even work on a real dataset? I'm worried that putting the window restriction on the clause within the CASE statement (rather than on the query as a whole) might just mean that the vast majority of items that are picked up are classified in the present lane. But different lanes have different numbers of works, so we can't just apply a single window restriction to the query as a whole. Different lanes need to have different window sizes. I remember trying this out on a real dataset and it more or less worked but now that I've written the code I'm worried.

Some obvious next steps, for this branch (if they cause problems now) or a future branch (if they don't).

* I don't think we need `mv_works_editions_datasources_identifiers` now that the other materialized view contains books that aren't classified under any genre.
* I don't think we need to pass around `featured` because whether or not a work is featured on its CustomList is no longer a reason to include or exclude it from a query. Now that work shows up in a query either way, but if it's featured on a relevant list it's assigned to a much higher quality tier.
* I don't think we need to keep track of whether a query needs to be made distinct. I think they all need to be made distinct now.


lane.py
```diff
@@ -689,15 +691,18 @@ def works(self, _db, facets=None, pagination=None):
             MaterializedWork,
             MaterializedWorkWithGenre,
         )
-        if self.genre_ids:
+        if self.genre_ids or True:
```
leonardr: I'll get rid of this but it's pretty convincing evidence that we don't need MaterializedWork anymore.

testing.py
```diff
@@ -381,6 +381,8 @@ def _lane(self, display_name=None, library=None,
             parent=parent, display_name=display_name,
             create_method_kwargs=dict(fiction=fiction)
         )
+        if is_new and parent:
+            lane.priority = len(parent.sublanes)-1
```
leonardr: This guarantees that lanes act the way you would expect in a test -- their priority is the order they were created.

lane.py
```diff
@@ -1393,34 +1398,256 @@ def search_target(self):
...
-        acquisition feed for the Lane.
+    def featured_window(self, target_size):
+        """Randomly select an interval over `Work.random` that ought to
+        contain approximately `target_size` high-quality works from
+        this lane.
+
+        :param: A 2-tuple (low value, high value), or None if the
+        entire span should be considered.
+        """
+        if self.size < 100:
```
aslagle: Maybe base this on `target_size/self.size` rather than a hard-coded number?

aslagle: I guess there would still be a hard-coded number....

lane.py
```diff
@@ -1393,34 +1398,256 @@ def search_target(self):
...
+            # This lane is so small that there's a significant risk
+            # that the interval might not contain enough works. Just
+            # choose randomly from the entire lane.
+            return 0,1
+        width = target_size / (self.size * 0.2)
+        
+        maximum_offset = 1-width
+        start = min(random.random(), maximum_offset)
```
aslagle: I think it would be more random to do `random.random() * maximum_offset`.

lane.py
```diff
@@ -1393,34 +1398,256 @@ def search_target(self):
...
+        return start, start+width
+
+    def groups(self, _db, include_sublanes=True):
+        """Return a list of (Lane, MaterializedWork) 2-tuples
+        describing a sequence of featured items for this lane and
+        (optionally) its children.
```
aslagle: What's the use case for optional children? I was thinking this generated the featured books for the children and then the 'All X' lane, but I would think 'All X' is the optional part of that.

leonardr: When you're building a groups feed for a lane, and one of its sublanes needs to have its featured feed calculated separately, you call sublane.groups(_db, include_sublanes=False). You don't want the sublanes because you're not building a groups feed for the sublane, you're building the sublane's _portion_ of its parent's feed.

aslagle: Oh I see, that's for the lanes that don't inherit parent restrictions.

lane.py
```diff
@@ -1393,34 +1398,256 @@ def search_target(self):
...
+        if not additional_needed:
+            return
+        additional_found = 0
+        for by_tier in unused_by_tier, used_by_tier:
+            # Go through each tier in decreasing quality order.
+            for tier in sorted(by_tier.keys(), key=lambda x: -x):
+                mws = by_tier[tier]
+                random.shuffle(mws)
+                for mw in mws:
+                    if (by_tier is unused_by_tier
```
aslagle: should this be `in`?

migration/20171218-rebuild-materialized-work-with-genre.sql
```diff
@@ -0,0 +1,121 @@
+-- The default for lanes.inherit_parent_restrictions has changed
+-- to true. Change the setting for all lanes not based on custom lists.
+-- For all lanes based on custom lists, a human has made a decision
+-- to set the value one way or the other.
+update lanes set inherit_parent_restrictions=true where _list_datasource_id is null and id not in (select l.id from lanes l join lanes_customlists);
```
aslagle: not a big deal but in the clause at the end you could select lane id from lanes_customlists and not need the join

aslagle:
I think this makes sense and it seems like it will work. I didn't notice anything that would cause problems with custom lists or multiple genres or performance, but it would be good to test it with more real data. I also like that we can get rid of a materialized view.

lane.py
leonardr: No, `is` is correct. We're looking in both `unused_by_tier` and `used_by_tier`, and this code only runs when we're looking in `unused_by_tier`.

PR #749, opened 2017-12-15
Register for coverage with both Identifier & WorkCoverageProviders

courte:
Registering all the Things* to support a change in the metadata wrangler. Now WorkCoverageProviders can be set to only process works that have an existing CoverageRecord.

Supports NYPL-Simplified/metadata_wrangler#152.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -542,6 +542,9 @@ class IdentifierCoverageProvider(BaseCoverageProvider):
         `registered_only` flag to process items. It's currently only in use
         on the Metadata Wrangler.
 
+        :param force: Set to True to reset an existing CoverageRecord's status
+        "registered", regardless of its current status.
+
         TODO: Take identifier eligibility into account when registering.
         """
         name = cls.SERVICE_NAME or cls.__name__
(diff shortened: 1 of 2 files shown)


coverage.py
```diff
@@ -532,11 +535,11 @@ def _input_identifier_types(cls):
...
     @classmethod
-    def register(cls, identifier, collection=None):
-        """Registers an identifier for future coverage
+    def register(cls, identifier, collection=None, force=False):
+        """Registers an identifier for future coverage.
 
         This method is primarily for use with CoverageProviders that use the
-        `preregistered_only` flag to process items. It's currently only in use
+        `registered_only` flag to process items. It's currently only in use
         on the Metadata Wrangler.
```
leonardr: `force` doesn't do what I expected an argument called `force` to do. Add a docstring for it and maybe either:

1. call it `force_to_registered`
2. call it `force_status` and make it a string so the caller can choose which status to force

coverage.py
```diff
@@ -1086,7 +1096,23 @@ def handle_success(self, identifier):
 class WorkCoverageProvider(BaseCoverageProvider):
 
     """Perform coverage operations on Works rather than Identifiers."""
-    
+
+    @classmethod
+    def register(cls, work, force=False):
```
leonardr: Here, a boolean `force` makes sense.

coverage.py
leonardr: Never mind, I confused this method with `add_for`. Boolean `force` makes sense here. But it does need a docstring.

PR #750, opened 2017-12-18
Always use the new materialized view

leonardr:
Instead of modifying one of the two materialized views, this branch changes the "single query" branch to create a third materialized view. All lane code is changed to use this materialized view exclusively. Methods that used to take a materialized view-like object as one of their arguments, no longer do. Once we verify that everything works, we can simply drop the other two materialized views.

If it doesn't work, we can go back to old code without having to rebuild the views.


tests/test_lane.py
```diff
@@ -449,34 +437,51 @@ def test_apply(self):
...
 
-        # Passing in distinct=True makes the query distinct.
+        # Passing in distinct=True makes the query distinct by work ID.
         eq_(False, base_query._distinct)
-        distinct_query = facets.apply(self._db, base_query, Work, True)
-        eq_(True, distinct_query._distinct)
+        distinct_query = facets.apply(self._db, base_query, True)
+        eq_([work_model.works_id], distinct_query._distinct)
 
+#        # Passing in distinct=True makes the query distinct on
```
courte: You commented this test case out.

tests/test_lane.py
```diff
@@ -449,34 +437,51 @@ def test_apply(self):
...
+        )
+
         facets = FeaturedFacets(0.5, False)
-        base_query = self._db.query(Work).join(Work.license_pools)
+        from model import MaterializedWorkWithGenre as work_model
+        base_query = self._db.query(work_model).join(work_model.license_pool)
+
+        def expect(works, qu):
+            expect_ids = [x.works_id for x in qu]
+            actual_ids = [x.id for x in works]
```
courte: The variable naming here is swapped.

lane.py
```diff
@@ -225,13 +225,11 @@ def dy(new_value):
...
     @classmethod
-    def order_facet_to_database_field(cls, order_facet, work_model):
+    def order_facet_to_database_field(cls, order_facet):
         """Turn the name of an order facet into a materialized-view field
         for use in an ORDER BY clause.
-
-        :param work_model: Either MaterializedWork or
-        MaterializedWorkWithGenre.
         """
+        from model import MaterializedWorkWithGenre as work_model
```
courte: Why are you importing MaterializedWorkWithGenre inline? I can't see any model.py imports that would keep you from importing it once higher up in the file.

model.py
```diff
@@ -312,9 +312,11 @@ class SessionManager(object):
 
     MATERIALIZED_VIEW_WORKS = 'mv_works_editions_datasources_identifiers'
     MATERIALIZED_VIEW_WORKS_WORKGENRES = 'mv_works_editions_workgenres_datasources_identifiers'
+    MATERIALIZED_VIEW_LANES = 'mv_works_for_lanes'
     MATERIALIZED_VIEWS = {
         MATERIALIZED_VIEW_WORKS : 'materialized_view_works.sql',
```
courte: You can remove this file and the SessionManager references MaterializedWork. (Unless it's for rollback purposes?)

files/materialized_view_for_lanes.sql
```diff
@@ -0,0 +1,113 @@
+create materialized view mv_works_for_lanes
```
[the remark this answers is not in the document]
leonardr: It's the same except that the join against workgenres is an outer join. This ensures that works that aren't classified under any genres still make it into the materialized view. This is the key insight that lets us use a single materialized view instead of two.

lane.py
leonardr: MaterializedWorkWithGenre isn't actually defined inside model.py until SessionManager.initialize() runs. This is because I couldn't find any way of defining a SQLAlchemy class for a materialized view without having an active database connection so SQLAlchemy could see the view for itself.

model.py
leonardr: It's in case we try it this way and we need to go back, but it's also so we can test on a running site with a big dataset, without modifying the database tables used by the production code.

tests/test_lane.py
leonardr: I've fixed this in my worklist-featured-feed branch.

tests/test_lane.py
leonardr: There were two test cases and I went back on forth on which one was the 'right' way to do the code. I've removed the commented-out case in worklist-featured-feed.

files/materialized_view_for_lanes.sql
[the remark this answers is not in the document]
courte: Primo 👌

PR #751, opened 2017-12-19
Make Subject.lookup work in situations where name is provided but identifier is not, even if there are multiple Subjects with a given name and type.

leonardr:
With recent changes to the way BISAC subjects work, we can end up with a situation where Subjects have a name but no identifier. This isn't great, but this branch makes it possible to look up a subject even when you only know the name and there is no identifier.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -939,12 +939,12 @@ class TestSubject(DatabaseTest):
         """
         m = Subject._lookup
         assert_raises_regexp(
-            ValueError, "Cannot look up Subject with no type."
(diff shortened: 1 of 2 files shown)


tests/test_model.py
```diff
@@ -933,14 +933,31 @@ def test_default_fiction(self):
         
 class TestSubject(DatabaseTest):
 
+    def test_lookup_errors(self):
+        """Subject.lookup will complain if you don't give it
+        enough information to find a Subject.
+        """
+        m = Subject._lookup
```
courte: Doesn't look like you ended up using this.

PR #754, opened 2017-12-20
Miscellaneous fixes from a hotfix deployment

leonardr:
After an exciting morning I've made some improvements to the code. Here are the big ones; smaller ones are inline.

* Axis 360 sends book descriptions now, in the `axis:annotation` tag. Start picking it up.
* Recalculate the sizes of lanes after refreshing materialized views through the `RefreshMaterializedViewScript`, not just after calling `SessionManager.refresh_materialized_views`. (Yes, we have two ways of refreshing materialized views, and we only really need one, but it's a little complicated.)
* Try really hard not to define the CASE statement that calculates work quality tier twice within a single SQL statement (it doesn't seem to be working, but it doesn't hurt to store this object along with the Lane and reuse it rather than building it from scratch every time).
* If a lane is built from a single list, there's no need to make the query for that lane DISTINCT, because there's no way a book can show up multiple times on that list. The query only needs to be DISTINCT if we know there are multiple lists in play, or if there _might_ be multiple lists in play (because we're building the lane out of all lists from a given data source). This is important because currently there's a major performance problem when a lane query is made DISTINCT.


lane.py
```diff
@@ -581,7 +588,7 @@ def full_identifier(self):
         captures its position within the heirarchy.
         """
         lane_parentage = list(self.parentage) + [self]
-        full_parentage = [x.display_name for x in lane_parentage]
+        full_parentage = [unicode(x.display_name) for x in lane_parentage]
```
leonardr: This is necessary because x.display_name might be a LazyLocalizedString.

lane.py
```diff
@@ -1132,6 +1139,10 @@ def get_library(self, _db):
     def collection_ids(self):
         return [x.id for x in self.library.collections]
 
+    @property
+    def children(self):
+        return self.sublanes
+
```
leonardr: Previously `WorkList` defined `children` but `Lane` did not.

opds.py
```diff
@@ -830,7 +830,7 @@ def _create_entry(self, work, license_pool, edition, identifier,
         if group_uri:
             self.add_link_to_entry(
                 xml, rel=OPDSFeed.GROUP_REL, href=group_uri,
-                title=group_title)
+                title=unicode(group_title))
```
leonardr: This is necessary because group_title might be a LazyLocalizedString.

leonardr:
There's no reason why `l` might not be a lane at this point.

PR #756, opened 2017-12-21
Worklist featured feed

leonardr:
This branch moves functionality from `Lane.groups()` to `WorkList.groups()` so that the top level of the site (a WorkList that has a bunch of Lanes underneath it) can request a grouped feed using a single query. Without this code, only a Lane can request a grouped feed using a single query.

Most of the changes in here are caused by moving code from Lane to WorkList.

After getting this code to work I was able to confirm my fears about the chances that the single query might not pick up enough works to fill the entire lane. It's pretty likely that this will happen. I'm going to investigate going back to multiple queries, or many subqueries with a UNION clause, but I'd like to do that using this branch as a base rather than starting again from master. I think most of the code in this branch will still be useful, it just needs to be moved around a bit and it won't be quite as fast.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1047,18 +1047,18 @@ class WorkList(object):
                     break
         used_in_parent = set()
         for lane in relevant_lanes:
-            if not lane in queryable_lane_set:
+            if lane in queryable_lane_set:
+                # We found results for this lane through the main query.
+                # Yield those results.
+                for mw in by_lane_id.get(lane.id, []):
+                    yield (mw, lane)
+            else:
                 # We didn't try to use the main query to find results
                 # for this lane because we knew the results, if there
                 # were any, wouldn't be representative. Do a whole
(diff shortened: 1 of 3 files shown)


lane.py
```diff
@@ -1409,10 +1664,8 @@ def featured_window(self, target_size):
         :param: A 2-tuple (low value, high value), or None if the
         entire span should be considered.
         """
-        if self.size < 100:
-            # This lane is so small that there's a significant risk
-            # that the interval might not contain enough works. Just
-            # choose randomly from the entire lane.
+        if self.size < target_size:
+            # Don't bother -- we're returning the whole lane.
```
leonardr: Rather than sending the entire interval if the target size is less than a specific number, send the entire interval if the target size is as big or bigger than the lane. If the target size is just really big compared to the lane, the math will work out.

PR #757, opened 2017-12-21
Build the featured feed by running a lot of really fast queries, rather than one really fast query that doesn't always get enough results

leonardr:
This branch gives up on the idea of building a grouped feed using a single database query. It comes close to working, but it doesn't reliably find enough books in cases where a lane is based on a genre that doesn't have very many books in it (like poetry, or bibliography on Open Ebooks).

Instead, we go back to running a query for every lane. The query is much faster than before because we add an index to the materialized view (which we should have added anyway), and the lanes will be filled to capacity more reliably (though not 100% reliable -- we could still end up short by a statistical fluke, but it's extremely unlikely).

There is one minor change in the rules about which books might show up in a grouped feed. It's now more likely that a book that showed up in a previous lane will show up again later on. Since we're no longer getting all the books in a single query, it's more difficult to exclude duplicates. For any given lane, we have a small buffer that lets us discard works that were used in a previous lane, but if we need those works to get that lane up to `target_size`, we will stick them on the end.

This branch includes a fair amount of cleanup and there's more that could be done, but I want to test this code on an NYPL-sized dataset to make sure we've really solved the performance problems before continuing.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -24,6 +24,7 @@ from sqlalchemy import (
     not_,
     Integer,
     Table,
+    Unicode,
 )
 from sqlalchemy.ext.associationproxy import (
     association_proxy,
@@ -77,7 +78,6 @@ from sqlalchemy import (
     Column,
     ForeignKey,
     Integer,
-    Unicode,
     UniqueConstraint,
 )
 from sqlalchemy.dialects.postgresql import (
@@ -1000,6 +1000,20 @@ class WorkList(object):
             # for WorkLists not handled by the query.
             qu = []
 
+        def _done_with_lane(lane_id):
+            """Called when we're done with a lane, either because
+            the lane chances or we've reached the end of the list.
+            """
+            # Did we get enough items?
+            num_missing = target_size-len(by_lane_id[lane_id])
+            if num_missing > 0 and might_need_to_reuse:
+                # No, we need to use some works we used in a
+                # previous lane to fill out this lane. Stick
+                # them at the end.
+                by_lane_id[lane_id].extend(
+                    might_need_to_reuse.values()[:num_missing]
+                )
+
         used_works = set()
         by_lane_id = defaultdict(list)
         working_lane_id = None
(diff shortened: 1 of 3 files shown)


lane.py
```diff
@@ -991,105 +995,83 @@ def _groups_for_lanes(self, _db, relevant_lanes, queryable_lanes):
...
-                        # this to the 'unused' dictionary in case we
-                        # need to use it to fill in the parent lane.
-                        #
-                        # NOTE: Checking `used`, as we did above,
-                        # isn't totally reliable because this work
-                        # might show up again in another lane and get
-                        # used then, but it doesn't hurt.
-                        unused_by_tier[quality_tier].append(mw)
-                    continue
+        work_quality_tier_lane = list(
```
leonardr: This code is mainly changes to use Lane objects instead of lane IDs.

leonardr: This is new code that achieves the same basic result. It's simpler because there's no attempt to fill up the last lane with leftovers.

lane.py
```diff
@@ -1099,78 +1081,58 @@ def _groups_query(self, _db, lanes):
...
-        return qu
+        :param target_size: Try to get approximately this many
+        items. There may be more or less; this controls the size of
+        the window and the LIMIT on the query.
+        """
+        from model import MaterializedWorkWithGenre
+        work_model = MaterializedWorkWithGenre
 
-    @classmethod
-    def _add_lane_id_field(cls, _db, qu, lanes, target_size):
```
leonardr: This method isn't needed anymore because we no longer use a single query to pull from multiple lanes.

lane.py
```diff
@@ -1099,78 +1081,58 @@ def _groups_query(self, _db, lanes):
...
             self.uses_customlists
         )
-        qu = self.works(_db, facets=facets)
-
-        # Include a field that assigns a work to one of the sublanes,
-        # or to the parent lane if it doesn't match any of the
-        # sublanes.
-        qu = self._add_lane_id_field(_db, qu, lanes, target_size)
+        first_query = None
+        other_queries = []
```
aslagle: a couple unused variables here

PR #758, opened 2018-01-02
Work queries are always distinct

leonardr:
Back when we had two different materialized views (one for work and one for work+genre), some lane queries needed to be distinct and some didn't. We passed around a 'distinct' argument to a lot of methods which controlled whether or not the query needed to become distinct after we were done applying filters to it.

Now that there is one materialized view (work+genre), all lane queries need to be distinct and there's no need to track whether or not a given query is distinct.

However I also discovered a case where just making the query distinct isn't good enough. If you have a lane that takes from multiple custom lists, and a single work is featured on list A and present (but not featured) on list B, the 'featured works' query can pick up the work twice, because the two instances of the work have different quality tiers. I couldn't find a way to filter out the duplicates in the SQL query, so I changed WorkList.featured_works to always filter out duplicates.

### DIFF
--- a/files/materialized_view_works_workgenres.sql
+++ b/files/materialized_view_works_workgenres.sql
@@ -40,7 +40,7 @@ as
      JOIN licensepools ON editions.id = licensepools.presentation_edition_id
      JOIN datasources ON licensepools.data_source_id = datasources.id
      JOIN identifiers on licensepools.identifier_id = identifiers.id
-     LEFT JOIN workgenres ON works.id = workgenres.work_id
+     JOIN workgenres ON works.id = workgenres.work_id
   WHERE works.presentation_ready = true
     AND works.simple_opds_entry IS NOT NULL
 
--- a/lane.py
+++ b/lane.py
@@ -243,7 +243,7 @@ class Facets(FacetConstants):
         }
         return order_facet_to_database_field[order_facet]
 
-    def apply(self, _db, qu, distinct=False):
+    def apply(self, _db, qu):
         """Restrict a query against MaterializedWorkWithGenre so that it only
         matches works that fit the given facets, and the query is
         ordered appropriately.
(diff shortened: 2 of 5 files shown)


files/materialized_view_works_workgenres.sql
```diff
@@ -40,7 +40,7 @@ as
      JOIN licensepools ON editions.id = licensepools.presentation_edition_id
      JOIN datasources ON licensepools.data_source_id = datasources.id
      JOIN identifiers on licensepools.identifier_id = identifiers.id
-     LEFT JOIN workgenres ON works.id = workgenres.work_id
+     JOIN workgenres ON works.id = workgenres.work_id
```
leonardr: This is changing the SQL back to what it used to be. Rather than modifying an existing materialized view, this branch creates a new one.

lane.py
```diff
@@ -347,10 +349,9 @@ def apply(self, _db, qu, distinct):
         from model import MaterializedWorkWithGenre as work_model
         quality = self.quality_tier_field()
         qu = qu.order_by(
-            quality.desc(), work_model.random.desc()
+            quality.desc(), work_model.random.desc(), work_model.works_id
         )
-        if distinct:
-            qu = qu.distinct(work_model.works_id)
+        qu = qu.distinct(quality, work_model.random, work_model.works_id)
```
courte: Why are you setting the query distinct on multiple values here?

tests/test_lane.py
```diff
@@ -683,11 +687,21 @@ def test_groups(self):
 
     def test_featured_works(self):
         wl = MockWorks()
```
courte: Can this be called `work_list` instead of `wl`? It's tough to parse alongside `w1`.

courte: Although now I'm looking at the file, and it looks like wl is the common practice across tests. Feel free to leave it.

courte:
Oh, and for the tests, it'll need the updates to requirements.txt.

lane.py
leonardr: If you order a query by X, you can set it distinct on (X, Y) but you can't set it distinct solely on Y. The query here is ordered by quality, then random, so I have to put those first in the distinct() calls.

I think the code I'm replacing either doesn't work, or only works in a limited situation. Here's the corresponding code currently in master, which sets DISTINCT on the same fields used in ORDER BY.

```
        # Set the ORDER BY clause.
        order_by, order_distinct = self.order_by(work_model)
        qu = qu.order_by(*order_by)
        if distinct:
            qu = qu.distinct(*order_distinct)
```

courte: Gotcha, thanks for explaining.

PR #759, opened 2018-01-02
Make code changes necessary to support SQLAlchemy 1.2.0

leonardr:
This branch changes the way we set up hybrid properties for compatibility with SQLAlchemy 1.2.0. Previously, the 'setter' method associated with a hybrid property needed to have a different name, like 'set_property', to avoid collision with the 'getter' method. Now the 'setter' method needs to have the _same_ as the original method.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1177,7 +1177,7 @@ class Lane(Base, WorkList):
         return self._visible and (not self.parent or self.parent.visible)
 
     @visible.setter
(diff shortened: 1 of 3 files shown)


leonardr:
I figured this out by looking at this bug filed against SQLAlchemy: https://bitbucket.org/zzzeek/sqlalchemy/issues/4149/hybrid-propert-behavior-changed-between-1x

PR #760, opened 2018-01-03
Single query featured feed fixes

leonardr:
This branch introduces miscellaneous fixes I needed to make to the single-query-featured-lane code to get it to work on a real site.

### DIFF
--- a/migration/20180101-materialized-work-for-lanes.sql
+++ b/migration/20180101-materialized-work-for-lanes.sql
@@ -5,8 +5,7 @@
 update lanes set inherit_parent_restrictions=true where _list_datasource_id is null and id not in (select lane_id from lanes_customlists);
 ALTER TABLE lanes ALTER COLUMN inherit_parent_restrictions SET DEFAULT true;
 
-drop materialized view mv_works_editions_workgenres_datasources_identifiers;
-create materialized view mv_works_editions_workgenres_datasources_identifiers
+create materialized view mv_works_for_lanes
 as
  SELECT 
     works.id AS works_id,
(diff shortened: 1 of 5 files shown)


migration/20180101-materialized-work-for-lanes.sql
```diff
@@ -5,8 +5,7 @@
 update lanes set inherit_parent_restrictions=true where _list_datasource_id is null and id not in (select lane_id from lanes_customlists);
 ALTER TABLE lanes ALTER COLUMN inherit_parent_restrictions SET DEFAULT true;
 
-drop materialized view mv_works_editions_workgenres_datasources_identifiers;
```
leonardr: Beyond this point the file used to be a copy of `core/files/materialized_view_works_workgenres.sql`, because we were modifying an existing materialized view. Now it's a copy of `core/files/materialized_view_for_lanes.sql` because we're creating a new materialized view.

scripts.py
```diff
@@ -606,9 +606,15 @@ class LaneSweeperScript(LibraryInputScript):
...
-        queue = self._db.query(Lane).filter(
+        from lane import WorkList
+        top_level_lanes = self._db.query(Lane).filter(
             Lane.library==library).filter(
                 Lane.parent==None).all()
+        top_level = WorkList()
+        top_level.initialize(
+            library, library.name, children=list(top_level_lanes)
+        )
+        queue = [top_level]
```
leonardr: Without this code, running the script that makes OPDS grouped feeds will make grouped feeds for 'fiction', 'nonfiction', etc, but it won't make the top-level grouped feed.

util/opds_writer.py
```diff
@@ -126,7 +126,7 @@ def updated(cls, *args, **kwargs):
     def __init__(self, title, url):
         self.feed = self.E.feed(
             self.E.id(url),
-            self.E.title(str(title)),
+            self.E.title(unicode(title)),
```
leonardr: This fixes a crash when the name of the feed contains non-ASCII characters.

PR #761, opened 2018-01-04
Added delegatedholds table for ODL holds management.

aslagle:
With ODL, a distributor doesn't manage the holds queue and the circulation manager has full control. This branch adds a `delegatedholds` table for the circ manager to store holds for ODL or any other distributors that don't handle holds.

For now, this will be used by a single circ manager to handle holds for its patrons, but in the near future we would also like to support ODL holds queues across a consortium. The consortium will have to designate one of its circ managers to handle the holds for the entire consortium, and other circ managers will delegate their holds to that circ manager. I'm planning to use IntegrationClient to represent the circ manager the hold is for in that case, and I'd like things to work mostly the same way whether the hold is for a local patron or a remote patron.

The `holds` table is unchanged - when a local patron creates an ODL hold it will be represented in both tables.

The `delegatedholds` table uses the start date of the hold to determine the queue position. There is an open question of whether librarians should be able to manipulate the queue. I think Hadrien considers it a useful feature of ODL but Micah and I can't think of a use case. If it is necessary they could do it by changing the start date, but maybe that's not ideal.

### DIFF
--- a/model.py
+++ b/model.py
@@ -6439,6 +6439,9 @@ class LicensePool(Base):
     # One LicensePool can have many Holds.
     holds = relationship('Hold', backref='license_pool')
 
+    # One LicensePool can have many DelegatedHolds.
+    delegatedholds = relationship('DelegatedHold', backref='license_pool')
+
     # One LicensePool can have many CirculationEvents
     circulation_events = relationship(
         "CirculationEvent", backref="license_pool")
@@ -10953,6 +10956,9 @@ class IntegrationClient(Base):
     created = Column(DateTime)
     last_accessed = Column(DateTime)
 
+    # One IntegrationClient can have many DelegatedHolds.
+    delegatedholds = relationship('DelegatedHold', backref='integration_client')
+
     def __repr__(self):
         return (u"<IntegrationClient: URL=%s ID=%s>" % (self.url, self.id)).encode('utf8')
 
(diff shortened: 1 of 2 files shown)


leonardr:
I like where this is going. A couple thoughts:

Is it possible to have one table for both types of holds? That seems neater than potentially having entries in two different tables for a single hold. You would add `patron_id` to the `holds` table as `remote_patron_identifier`, 

Now that you've introduced this concept, I think it would be more natural to call the things currently in the `holds` table "delegated holds" or "remote holds", and use a term like "local holds" for the things this branch puts in the `delegatedholds` table. If you have a hold on an Overdrive book, then you've really got a reference to a hold on Overdrive's side.

I think you called the table `delegatedholds` to be consistent with `delegatedpatronidentifiers`, but I was never happy with that name, and  `delegatedpatronidentifiers` should go away soon, as that duty is taken on by the library registry, so I think it's okay to break from the pattern. If both types of holds are in one table then what we name them becomes a matter for the ORM and not the database schema.

It may not be necessary to connect a delegated hold directly to the ExternalIntegration that manages it. Can you get it through LicensePool -> Collection -> ExternalIntegration? That's how we know e.g. which Overdrive integration is responsible for an Overdrive hold.

The use case I see for hold manipulation is: I borrowed a book, something went wrong, I lost my loan, and now I have to get back in line at the end of the holds queue. In this case a librarian might bump the patron to the front of the holds queue so that they can be next in line for the book. I don't think there's a use case for arbitrary rearrangements of the holds queue but there is a case for bumping someone to the front.

If we decide that manipulating the holds queue is useful, then keeping everything in the `holds` table, which already has `position`, will let you start keeping track of queue position separately from the date the hold was created.

aslagle:
Using one table is better in some ways, I'll try it. I think the main reason I didn't want to do that was that I liked having position be a property so I didn't have to update every hold for a pool whenever something happened to it.

You're right that it's not necessary to connect the hold to the `ExternalIntegration`, but it is necessary to connect it to the circ manager when it could be from a remote circ manager. I put in the `IntegrationClient` for that and didn't use it yet. If I use the `holds` table I'll leave that out for now and put in a migration when I actually work on that.

aslagle:
I switched to the `holds` table so I didn't end up needing any core changes yet.

leonardr:
Nice.

PR #762, opened 2018-01-05
Single query featured feed fixes

leonardr:
This branch improves the performance of feed generation by bringing in information that would otherwise require additional database queries. It also improves the initial creation of the materialized view.

### DIFF
--- a/files/materialized_view_for_lanes.sql
+++ b/files/materialized_view_for_lanes.sql
@@ -1,3 +1,4 @@
+-- Create the materialized view with no data.
 create materialized view mv_works_for_lanes
 as
  SELECT 
@@ -44,7 +45,8 @@ as
   WHERE works.presentation_ready = true
     AND works.simple_opds_entry IS NOT NULL
 
-  ORDER BY (editions.sort_title, editions.sort_author, licensepools.availability_time);
+  ORDER BY (editions.sort_title, editions.sort_author, licensepools.availability_time)
+  WITH NO DATA;
 
 -- Create a work/genre lookup.
 create unique index mv_works_for_lanes_work_id_genre_id on mv_works_for_lanes (works_id, genre_id);
@@ -57,11 +59,9 @@ create index mv_works_for_lanes_by_availability on mv_works_for_lanes (availabil
 
 create index mv_works_for_lanes_by_random_and_genre on mv_works_for_lanes (random, language, genre_id);
 
--- Not sure we need these two.
+create index mv_works_for_lanes_by_random_audience_target_age on mv_works_for_lanes (random, language, audience, target_age);
 
--- create index mv_works_for_lanes_by_random_and_audience on mv_works_for_lanes (random, language, audience, target_age);
-
--- create index mv_works_for_lanes_by_random_and_fiction on mv_works_for_lanes (random, language, fiction);
+create index mv_works_for_lanes_by_random_fiction_audience_target_age on mv_works_for_lanes (random, language, fiction, audience, target_age);
 
 -- Similarly, an index on everything, sorted by descending update time.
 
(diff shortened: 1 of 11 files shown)


files/materialized_view_for_lanes.sql
```diff
@@ -44,7 +45,8 @@ as
   WHERE works.presentation_ready = true
     AND works.simple_opds_entry IS NOT NULL
 
-  ORDER BY (editions.sort_title, editions.sort_author, licensepools.availability_time);
+  ORDER BY (editions.sort_title, editions.sort_author, licensepools.availability_time)
+  WITH NO DATA;
```
leonardr: This improves initial setup because when we create the indexes there's no data in the materialized view.

lane.py
```diff
@@ -725,14 +725,29 @@ def works(self, _db, facets=None, pagination=None, include_quality_tier=False):
...
+            qu = qu.options(
+                contains_eager(mw.license_pool),
+                # TODO: Strictly speaking, these joinedload calls are 
+                # only needed by the circulation manager. This code could
+                # be moved to circulation and everyone else who uses this
+                # would be a little faster. (But right now there is no one
+                # else who uses this.)
+
+                # These speed up the process of generating acquisition links.
+                joinedload("license_pool", "delivery_mechanisms"),
```
leonardr: This line is why I had to make LicensePool.delivery_mechanisms into a proper relationship -- otherwise there was no way to tell SQLAlchemy to load in the LicensePoolDeliveryMechanism objects or any of that object's children.

migration/20171208-add-lane-size.sql
```diff
@@ -1 +1,9 @@
-alter table lanes add column size integer not null default 0;
+DO $$ 
+    BEGIN
+        BEGIN
+            ALTER TABLE lanes ADD COLUMN size integer not null default 0;
+        EXCEPTION
+            WHEN duplicate_column THEN RAISE NOTICE 'column lanes.size already exists, not creating it.';
+        END;
+    END $$;
+
```
leonardr: I've had a lot of problems 'migrating' sites that already have `lanes.size` by the time this script gets to them. I think it's because these sites don't have the `lanes` table to begin with, and it's created with `size` intact.

migration/20180105-2-materialized-work-for-lanes.sql
```diff
@@ -1,10 +1,4 @@
--- The default for lanes.inherit_parent_restrictions has changed
--- to true. Change the setting for all lanes not based on custom lists.
--- For all lanes based on custom lists, a human has made a decision
--- to set the value one way or the other.
-update lanes set inherit_parent_restrictions=true where _list_datasource_id is null and id not in (select lane_id from lanes_customlists);
-ALTER TABLE lanes ALTER COLUMN inherit_parent_restrictions SET DEFAULT true;
-
```
leonardr: I moved this into a separate file so that the rest of this file could be a copy of `materialized_view_for_lanes.sql`.

migration/20180105-2-materialized-work-for-lanes.sql
```diff
@@ -64,11 +59,9 @@ create index mv_works_for_lanes_by_availability on mv_works_for_lanes (availabil
 
 create index mv_works_for_lanes_by_random_and_genre on mv_works_for_lanes (random, language, genre_id);
 
--- Not sure we need these two.
-
--- create index mv_works_for_lanes_by_random_and_audience on mv_works_for_lanes (random, language, audience, target_age);
```
leonardr: We do, in fact, need these two.

model.py
```diff
@@ -387,22 +387,23 @@ def initialize(cls, url):
...
-                Column('workgenres_id', Integer, primary_key=True),
-                Column('license_pool_id', Integer, ForeignKey('licensepools.id')),
-                autoload=True,
-                autoload_with=engine
-            )
-            license_pool = relationship(
-                LicensePool,
-                primaryjoin="LicensePool.id==MaterializedWorkWithGenre.license_pool_id",
-                foreign_keys=LicensePool.id, lazy='joined', uselist=False)
+        if create_materialized_views:
```
leonardr: I had to add this in so that when we're migrating a site that doesn't have materialized_works_for_lane to a site that does, this code (which assumed the materialized view already exists) doesn't run before we have a chance to create the SQL.

migration/20180105-2-materialized-work-for-lanes.sql
courte: You're missing the file for this.

model.py
courte: Can you add a comment here that clarifies that this is for the database migration script?

leonardr:
I investigated the materialized view problem and I'm not sure why it was only happening during migration, but I was able to come up with a general fix. The SQL to create the materialized view would always run, but in some cases it was acting like the database session that created the materialized view wasn't committed. In these cases, when the time came to create the MaterializedWorkWithGenre class, it would appear as though the materialized view wasn't in the database.

I changed it so that materialized views are all created WITH NO DATA and then we immediately refresh them after creating them. Refreshing the materialized view forces everyone to acknowledge that the materialized view has been created and is in the database.

PR #763, opened 2018-01-05
Bulk register for coverage

courte:
I did all this work to support performance for a change in the metadata wrangler that it turns out isn't currently necessary. Shucks.

YAGNI aside, I'd like to keep the changes for future use -- possibly refactoring the `IdentifierResolutionRegistrar` to process batches of identifiers at a time, for example.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -536,7 +536,7 @@ class IdentifierCoverageProvider(BaseCoverageProvider):
 
     @classmethod
     def register(cls, identifier, data_source=None, collection=None,
-        force=False
+        force=False, autocreate=False
     ):
         """Registers an identifier for future coverage.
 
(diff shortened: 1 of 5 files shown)


coverage.py
```diff
@@ -535,57 +535,102 @@ def _input_identifier_types(cls):
...
+        """
+        if not data_source:
+            return DataSource.lookup(_db, cls.DATA_SOURCE_NAME)
+        if isinstance(data_source, DataSource):
+            return data_source
+        if isinstance(data_source, basestring):
+            # If a string is passed, we assume it's meant to be a
+            # DataSource. Because on the metadata wrangler new Collection
+            # names may be passed, this ensures there will always be a
+            # DataSource.
```
leonardr: I think this should take `autocreate` as an argument so that the responsibility for passing in `True` lies with the metadata wrangler.

courte:
I thought the `IntegrationClientCoverageProvider` was going to be necessary to get coverage for requested metadata for ISBN-sending datasources (Axis 360, Bibliotheca), but the provider is only for mirroring the cover images. And we only need that for the OPDS_FOR_DISTRIBUTORS data source (though it may change in the future).

I spent some time trying to make the provider more general before I realized it, and that changed the way registration work. It's been a whole dang circle, I tell ya. (NYPL-Simplified/metadata_wrangler#156)

PR #764, opened 2018-01-08
Set hold end date for position > 0.

aslagle:
This code was only setting the hold end date if the hold position was 0. That seems inconsistent with `Hold.until`, which wants to use the end date if the distributor provides one. I'm not sure if this will cause any problems - could a hold reach its end date and be deleted when it's still active? Maybe we shouldn't even use end date to store the estimated wait time.

### DIFF
--- a/model.py
+++ b/model.py
@@ -915,10 +915,8 @@ class Hold(Base, LoanAndHoldMixin):
         """
         if start is not None:
             self.start = start
-        if position == 0 and end is not None:
+        if end is not None:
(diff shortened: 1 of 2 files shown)


leonardr:
I think the thinking is that when `position>0`, 'end date' is an estimate and it could be way off. But when `position==0`, 'end date' is a specific time that is well known. However, I agree that's not a reason to prevent `end` from being set if the caller wants to set it.

PR #765, opened 2018-01-09
Helper method that creates a WorkList representing a library's collection.

leonardr:
This branch fixes two problems:

* Previously created lanes should have had their `medium` limited to `[Book]` but didn't. 
* We had two different ways of creating the WorkList corresponding to a library's entire collection: one in the circulation manager app server setup and one in the script setup. The WorkLists were slightly different and both might have included works that were not books.

### DIFF
--- a/tests/test_lane.py
+++ b/tests/test_lane.py
@@ -635,7 +635,6 @@ class TestWorkList(DatabaseTest):
         # The other library only has one top-level lane, so we use that lane.
         l = WorkList.top_level_for_library(self._db, other_library)
         eq_(other_library_lane, l)
-        eq_(Edition.FULFILLABLE_MEDIA, wl.media)
 
         # This library has no lanes configured at all.
         no_config_library = self._library(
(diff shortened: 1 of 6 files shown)


tests/test_lane.py
```diff
@@ -603,6 +603,49 @@ def test_initialize(self):
...
+
+        # The default library gets a WorkList with the two top-level lanes as children.
+        wl = WorkList.top_level_for_library(self._db, self._default_library)
+        eq_([lane1, lane2], wl.children)
+        eq_(Edition.FULFILLABLE_MEDIA, wl.media)
+
+        # The other library only has one top-level lane, so we use that lane.
+        l = WorkList.top_level_for_library(self._db, other_library)
+        eq_(other_library_lane, l)
+        eq_(Edition.FULFILLABLE_MEDIA, wl.media)
```
aslagle: is this line meant to be checking `l.media`? or left over from a copy and paste?

PR #767, opened 2018-01-11
Add `<contributor>` tags to OPDS entries for non-author contributors

leonardr:
This branch represents non-author contributors to a work using the Atom `<contributor>` tag. It uses the unofficial OPDS standard technique of putting the contributor role in the `opf:role` attribute, and as specified in 1.4.1.2.x of the [OPF spec](http://www.idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section1.4.1.2) it uses [MARC Relator Codes](https://www.loc.gov/marc/relators/relaterm.html) to explain what role a contributor had in the work.

The migration script in this branch removes the `opds-entry` WorkCoverageRecords for any works with non-author contributors, so that those entries can be regenerated. The `bin/opds_entry_coverage` script takes care of regenerating OPDS entries for any works missing their `opds-entry` coverage records.

We'll need a docker branch to run that script on a regular basis to clean up after this migration script and any future ones.

### DIFF
--- a/migration/20180112-redo-opds-for-non-author-contributors.sql
+++ b/migration/20180112-redo-opds-for-non-author-contributors.sql
@@ -1,0 +1,14 @@
+-- Remove the 'generate-opds` coverage record for any works
+-- that have an associated series.
+delete from workcoveragerecords where operation='generate-opds' and work_id in (
+       select w.id from works w join editions e on w.presentation_edition_id=e.id where e.series is not null
+);
+
+-- Remove the 'generate-opds` coverage record for any works
+-- that have contributors in non-author roles.
+delete from workcoveragerecords where operation='generate-opds' and work_id in (
+       select distinct w.id from works w join editions e on w.presentation_edition_id=e.id join contributions c on e.id=c.edition_id where c.role not in ('Author', 'Primary Author')
+);
+
+-- These records will be added back when the OPDSEntryWorkCoverageProvider
+-- regenerates their OPDS entries.
--- a/model.py
+++ b/model.py
@@ -2466,7 +2466,6 @@ class Contributor(Base):
     # Types of roles
     AUTHOR_ROLE = u"Author"
     PRIMARY_AUTHOR_ROLE = u"Primary Author"
-    PERFORMER_ROLE = u"Performer"
     EDITOR_ROLE = u"Editor"
     ARTIST_ROLE = u"Artist"
     PHOTOGRAPHER_ROLE = u"Photographer"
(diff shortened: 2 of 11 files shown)


migration/20180112-redo-opds-for-non-author-contributors.sql
```diff
@@ -0,0 +1,14 @@
+-- Remove the 'generate-opds` coverage record for any works
+-- that have an associated series.
+delete from workcoveragerecords where operation='generate-opds' and work_id in (
+       select w.id from works w join editions e on w.presentation_edition_id=e.id where e.series is not null
+);
+
```
leonardr: This 'series' thing should have happened a long time ago but we didn't have any way of automatically regenerating the OPDS entries; now we do.

model.py
```diff
@@ -2466,7 +2466,6 @@ class Contributor(Base):
     # Types of roles
     AUTHOR_ROLE = u"Author"
     PRIMARY_AUTHOR_ROLE = u"Primary Author"
-    PERFORMER_ROLE = u"Performer"
```
leonardr: This constant is in the class twice.

model.py
```diff
@@ -2497,12 +2496,49 @@ class Contributor(Base):
...
+    }
+
     # People from these roles can be put into the 'author' slot if no
     # author proper is given.
     AUTHOR_SUBSTITUTE_ROLES = [
         EDITOR_ROLE, COMPILER_ROLE, COMPOSER_ROLE, DIRECTOR_ROLE,
-         CONTRIBUTOR_ROLE, TRANSLATOR_ROLE, ADAPTER_ROLE, PHOTOGRAPHER_ROLE,
-         ARTIST_ROLE, LYRICIST_ROLE, COPYRIGHT_HOLDER_ROLE
+        CONTRIBUTOR_ROLE, TRANSLATOR_ROLE, ADAPTER_ROLE, PHOTOGRAPHER_ROLE,
+        ARTIST_ROLE, LYRICIST_ROLE, COPYRIGHT_HOLDER_ROLE
```
leonardr: This is just reformatting. Now that we can properly represent all these types of people we may want to switch to just putting in `<contributor>` tags with no `<author>` tags.

monitor.py
```diff
@@ -485,6 +485,10 @@ def item_query(self):
 class OPDSEntryCacheMonitor(PresentationReadyWorkSweepMonitor):
     """A Monitor that recalculates the OPDS entries for every
     presentation-ready Work.
+
+    This is different from the OPDSEntryWorkCoverageProvider,
+    which only processes works that are missing a WorkCoverageRecord
+    with the 'generate-opds' operation.
```
leonardr: The days of this Monitor may be numbered.

model.py
leonardr: It would require a change on the mobile side to display the new credits; otherwise it would look like there were no authors. But there are not a lot of titles in this situation so most people wouldn't notice. I think it's mostly translations of ancient books credited to traditional authors, like Aesop's Fables.

PR #768, opened 2018-01-12
Remove a migration script that's no longer necessary

leonardr:
After creating this migration script (which creates a new materialized view) I changed the startup process so that any missing materialized views are created automatically. That makes this migration script unnecessary, and in fact it will now stop a migration from running.


scripts.py
```diff
@@ -737,11 +737,11 @@ class BibliographicRefreshScript(RunCollectionCoverageProviderScript, Identifier
     )
 
     def __init__(self, _db=None, **metadata_replacement_args):
-        replacement_policy = ReplacementPolicy.from_metadata_source(
+        self.replacement_policy = ReplacementPolicy.from_metadata_source(
             **metadata_replacement_args
         )
-        kwargs = dict(replacement_policy=replacement_policy)
 
+        kwargs = dict(replacement_policy=self.replacement_policy)
```
leonardr: This is an unrelated change; I happened to run this script recently and its `do_run` method assumed the replacement policy had been saved on the Script object.

PR #769, opened 2018-01-15
Changed dc:created to dc:issued in OPDS, and added a staff rating scale.

aslagle:
As part of improving book metadata editing in the admin interface, I wanted to add the publication date to the editable fields. However, I discovered that there was a lot of inconsistency in where that date is.

The admin interface uses the book's OPDS entry to fill in the initial form values. According to the OPDS spec, dc:issued "SHOULD be used to indicate the first publication date of the Publication". Our server was putting it in dc:created, and there was a comment that dc:issued is commonly conflated with atom:published. That didn't seem like a good reason not to follow the spec though.

The web client and the iOS client were using atom:published as the publication date, when that is actually the date the publication became available in the catalog according to both our server and the spec. Based on that it seems like using dc:created didn't solve the problem.

So I tried making this change and when I went to fix the tests I found that it's actually _feedparser_ that conflates atom:published and dc:issued, so if we use both we can no longer use feedparser for that information. This mostly affects OPDS import, which was already not using that info, maybe because it was likely to be inaccurate if others' feeds were following the spec. If we want to start using it now, we'd just have to do that in the etree part.

It's annoying that we can't use feedparser but I still think this is the right thing to do. I'm updating the web client and the mobile apps should be changed too, but this PR won't actually affect old versions since they're already using the wrong values (at least iOS is, I didn't check Android).

I also made one other change in this PR so that staff can add a rating to a book from the admin interface. I picked 1-5 as the rating scale.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -929,7 +929,6 @@ class OPDSImporter(object):
                         )
                     )
         last_opds_update = cls._datetime(entry, 'updated_parsed')
-        added_to_collection_time = cls._datetime(entry, 'published_parsed')
             
         publisher = entry.get('publisher', None)
         if not publisher:
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -1208,7 +1208,7 @@ class Metadata(MetaToModelUtility):
         fields = self.BASIC_EDITION_FIELDS
         for field in fields:
             new_value = getattr(metadata, field)
-            if new_value:
+            if new_value != None and new_value != '':
                 setattr(self, field, new_value)
 
         new_value = getattr(metadata, 'contributors')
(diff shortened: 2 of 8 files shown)


opds_import.py
```diff
@@ -929,7 +929,6 @@ def _data_detail_for_feedparser_entry(cls, entry, metadata_data_source):
                         )
                     )
         last_opds_update = cls._datetime(entry, 'updated_parsed')
-        added_to_collection_time = cls._datetime(entry, 'published_parsed')
```
aslagle: Here's a line where we were getting the conflated atom:published/dc:issued value, but we were never using it.

leonardr:
This is so annoying. :( So much confusion. But we already bypass feedparser for a lot of our OPDS parsing so this isn't too much additional work.

leonardr:
Looks good, merge it.

aslagle:
I also made it possible to set series position to 0 as discussed in https://github.com/NYPL-Simplified/circulation-web/pull/135, but I'll just note here that it is still not possible to set it to -1, since that's the value we use to indicate no number.

PR #771, opened 2018-01-17
Stop using deprecated materialized views

leonardr:
A user ran into a problem where the deprecated materialized views couldn't be updated because the unique index stopped being valid. Rather than add a migration script to change the unique index, I wrote this branch, which makes the application completely forget about those materialized views and not try to create them, or keep them updated, anymore.

The materialized views will still be there in existing deployments, in case mv_works_for_lanes doesn't work out and we have to go back. But I'd rather stop looking at them than slow down migration with an expensive script to manage views that are no longer used.


leonardr:
Agreed. We don't use them now, so there's no need to refresh them. If we need to start using them again,  we' can make a migration script that creates them if they're not already in place.

PR #772, opened 2018-01-17
Fix miscellaneous issues with OPDS entries

leonardr:
https://github.com/NYPL-Simplified/server_core/pull/769 makes it necessary to rebuild all cached OPDS entries. Since that's happening anyway I figured now is a good time to fix two other issues that also make it necessary to rebuild all cached OPDS entries.

* I updated the URIs used in `additionalType` to indicate whether a book is an ebook or an audiobook. Those URIs were chosen a long time ago and better alternatives are now part of schema.org.
* I fixed a bug which caused the `<bibframe:distribution>` tag to show up twice in an OPDS entry. It was added once (incorrectly) while the cached entry was being created and then added again (correctly) when it was time to add licensing information.

The changes to sample OPDS feeds are just changing http://schema.org/Book to http://schema.org/EBook.

### DIFF
--- a/model.py
+++ b/model.py
@@ -2966,7 +2966,7 @@ class Edition(Base):
 
     medium_to_additional_type = {
         BOOK_MEDIUM : u"http://schema.org/EBook",
-        AUDIO_MEDIUM : u"http://bib.schema.org/AudioBook",
+        AUDIO_MEDIUM : u"http://bib.schema.org/Audiobook",
         PERIODICAL_MEDIUM : u"http://schema.org/PublicationIssue",
         MUSIC_MEDIUM :  u"http://schema.org/MusicRecording",
         VIDEO_MEDIUM :  u"http://schema.org/VideoObject",
--- a/model.py
+++ b/model.py
@@ -2965,13 +2965,13 @@ class Edition(Base):
     FULFILLABLE_MEDIA = [BOOK_MEDIUM]
 
     medium_to_additional_type = {
-        BOOK_MEDIUM : u"http://schema.org/Book",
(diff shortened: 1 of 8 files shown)


model.py
```diff
@@ -4648,14 +4648,13 @@ def calculate_opds_entries(self, verbose=True):
             VerboseAnnotator,
         )
         _db = Session.object_session(self)
-        simple = AcquisitionFeed.single_entry(_db, self, Annotator,
-                                              force_create=True)
-        if simple is not None:
-            self.simple_opds_entry = unicode(etree.tostring(simple))
```
leonardr: Work.simple_opds_entry is actually set within single_entry, to an intermediate value. The value actually returned by single_entry is a feed that has been further annotated with license information (in this case, the provider of the LicensePool). That's not what we want in Work.simple_opds_entry -- we want the intermediate value with no provider info.

model.py
```diff
@@ -2965,8 +2965,8 @@ class Edition(Base):
     FULFILLABLE_MEDIA = [BOOK_MEDIUM]
 
     medium_to_additional_type = {
-        BOOK_MEDIUM : u"http://schema.org/Book",
-        AUDIO_MEDIUM : u"http://schema.org/AudioObject",
+        BOOK_MEDIUM : u"http://schema.org/EBook",
+        AUDIO_MEDIUM : u"http://bib.schema.org/AudioBook",
```
aslagle: This should be http://bib.schema.org/Audiobook - the 'B' isn't capitalized.

leonardr: Thanks, I'm always confused that EBook has intercaps and Audiobook doesn't.

PR #777, opened 2018-01-19
Support metadata needed

courte:
This branch supports changes in NYPL-Simplified/circulation#827.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -803,8 +803,12 @@ class IdentifierCoverageProvider(BaseCoverageProvider):
             identifiers=self.input_identifiers, collection=self.collection_or_not,
             **kwargs
         )
+
         if identifiers:
             qu = qu.filter(Identifier.id.in_([x.id for x in identifiers]))
+        if not identifiers and identifiers != None:
+            # An empty list was provided. The returned query should be empty.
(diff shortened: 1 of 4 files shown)


coverage.py
```diff
@@ -803,8 +803,12 @@ def items_that_need_coverage(self, identifiers=None, **kwargs):
             identifiers=self.input_identifiers, collection=self.collection_or_not,
             **kwargs
         )
+
         if identifiers:
             qu = qu.filter(Identifier.id.in_([x.id for x in identifiers]))
+        if not identifiers and identifiers != None:
```
courte: Caught this when I was running tests. I'm assuming here that if an empty list is given, the provider shouldn't be run on all missing identifiers.

model.py
```diff
@@ -1969,7 +1969,7 @@ def type_and_identifier_for_urn(cls, identifier_string):
         return (type, identifier_string)
 
     @classmethod
-    def parse_urns(cls, _db, identifier_strings):
+    def parse_urns(cls, _db, identifier_strings, autocreate=True):
```
courte: This keeps the default assumptions of `parse_urns` in line with those of `parse_urn`.

PR #782, opened 2018-01-24
Remove SubjectAssignmentMonitor

leonardr:
This branch removes the SubjectAssignmentMonitor and the script that runs it. That monitor makes sure all unchecked subjects become checked--i.e. are classified by the classifier--but it doesn't reclassify any of the works that use those subjects, so from the perspective of a user it does nothing.

By contrast, `ReclassifyWorksForUncheckedSubjectsScript` reclassifies all works that are classified under an unchecked subject, and the subject becomes checked as a side effect. If a subject somehow gets in without being classified under any works, it will never be checked, but it doesn't matter.

While I was at it I made a number of other small classification-related changes to stop script crashes and improve performance.


classifier/__init__.py
```diff
@@ -973,7 +978,10 @@ def scrub_identifier(cls, identifier):
             identifier = int(identifier)
         except ValueError:
             pass
-        return identifier
+
+        # For our purposes, Dewey Decimal numbers are identifiers
+        # without names.
+        return identifier, None
```
leonardr: Without this, the identifier (which is an integer here) is used as the name, and `scrub_identifier` choked trying to put it into a `Lowercased`.

model.py
```diff
@@ -3991,7 +3991,7 @@ def for_unchecked_subjects(cls, _db):
             LicensePool.identifier).join(
                 Identifier.classifications).join(
                     Classification.subject)
-        return qu.filter(Subject.checked==False).distinct()
+        return qu.filter(Subject.checked==False).order_by(Subject.id)
```
leonardr: This improves both reliability and performance. Previously the goal was to check every work only once. Now the goal is to check all the works associated with a subject at once. If the script crashes, it's less likely that a bunch of subjects will have been marked as checked without all the associated works being reclassified.

Getting rid of the `distinct` improves performance of the query quite a bit, but it means that sometimes a work will be reclassified more than once (because it was classified under two unchecked subjects). That's okay because reclassification is fast.

model.py
```diff
@@ -6219,33 +6219,36 @@ def assign_to_genre(self):
...
+        # Create a shorthand way of referring to this Subject in log
+        # messages.
+        parts = [self.type, self.identifier, self.name]
+        shorthand = ":".join(x for x in parts if x)
+
         if genre != self.genre:
             log.info(
-                "%s:%s genre %r=>%r", self.type, self.identifier,
-                self.genre, genre
+                "%s genre %r=>%r", shorthand, self.genre, genre
```
leonardr: I made this change to show identifier and/or name, rather than just identifier, because BISAC clasisfications have a name but might not have an identifier.

leonardr:
I just discovered I never actually asked anyone to review this PR.

PR #783, opened 2018-01-25
Never show a total number of holds smaller than your current holds position

leonardr:
This branch addresses https://github.com/NYPL-Simplified/circulation/issues/835 by ensuring that in an opds:holds tag, opds:total is never smaller than opds:position.

This branch also changes the behavior of opds:holds so that copies in the 'reserved' state are counted towards `opds:total`.

`AcquisitionFeed.license_tags` was previously untested and this branch only tests the part that generates the `opds:holds` tag.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -1189,7 +1189,7 @@ class AcquisitionFeed(OPDSFeed):
             position = hold.position
         if hold:
             if position and position > 0:
-                holds_kw['position'] = str(hold.position)
+                holds_kw['position'] = str(position)
             if position > total:
                 # The patron's hold position appears larger than the total
                 # number of holds. This happens frequently because the
--- a/opds.py
+++ b/opds.py
@@ -1181,9 +1181,14 @@ class AcquisitionFeed(OPDSFeed):
 
         holds_kw = dict()
         total = license_pool.patrons_in_hold_queue or 0
-        position = hold.position or 0
+        if hold.position is None:
+            # This shouldn't happen, but if it does, assume we're last
+            # in the list.
+            position = total
+        else:
+            position = hold.position
         if hold:
-            if position:
+            if position and position > 0:
                 holds_kw['position'] = str(hold.position)
             if position > total:
                 # The patron's hold position appears larger than the total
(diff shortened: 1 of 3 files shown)


opds.py
```diff
@@ -1179,9 +1179,26 @@ def license_tags(cls, license_pool, loan, hold):
             return tags
 
 
-        holds_kw = dict(total=str(license_pool.patrons_in_hold_queue or 0))
-        if hold and hold.position:
-            holds_kw['position'] = str(hold.position)
+        holds_kw = dict()
+        total = ((license_pool.patrons_in_hold_queue or 0) + 
+                 (license_pool.licenses_reserved or 0))
```
aslagle: I agree that reserved licenses should count towards the total, but I thought that was already included in `patrons_in_hold_queue`. That's how I did it for ODL holds. Patrons with reserved licenses are at the beginning of the queue and have holds with position `0`.

leonardr: You're right, I remember this now.

opds.py
```diff
@@ -1179,9 +1179,27 @@ def license_tags(cls, license_pool, loan, hold):
             return tags
 
 
-        holds_kw = dict(total=str(license_pool.patrons_in_hold_queue or 0))
-        if hold and hold.position:
-            holds_kw['position'] = str(hold.position)
+        holds_kw = dict()
+        total = license_pool.patrons_in_hold_queue or 0
+        position = hold.position or 0
```
aslagle: I'm not sure when we'd be missing the hold position, but it might be safer to assume you're last rather than first if we don't know.

opds.py
```diff
@@ -1179,9 +1179,27 @@ def license_tags(cls, license_pool, loan, hold):
...
+        if hold:
+            if position:
```
aslagle: This won't work if position is 0.

leonardr: This is by design. [The spec](https://docs.google.com/document/d/1DZCupVm5EnvpIrqRq8ABJi-68vgqwA7CymO9PM7PyPE/edit#heading=h.6ffrgis88ere) shows an example where the patron is in position 0 and the `opds:position` field is missing. I'll change this code to be more explicit.

aslagle: Never mind, I checked that leaving it out for position 0 is the current behavior.

PR #785, opened 2018-01-25
Add version endpoint

courte:
The code side of NYPL-Simplified/circulation#834.

Fixes NYPL-Simplified/circulation#834.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -34,6 +34,7 @@ from model import (
 )
 from cdn import cdnify
 from classifier import Classifier
+from config import Configuration
 from lane import (
     Facets,
     Pagination,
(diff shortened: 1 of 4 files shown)


courte:
Review-ready, but it would be nice to house the version on the admin interface somewhere. This is an immediate fix, but it's something to consider for later.

leonardr:
I'd like to see a couple changes here. First, I'd like this to become part of the heartbeat controller rather than a separate controller. Second, since reading a file from the filesystem is 'complicated' I'd like to do it during initialization rather than every time the controller is activated.

For the second thing it's pretty easy -- put the file-reading code in `CirculationManager.load_settings` and store the value in the CirculationManager. The main complication here is you'll need to pass that information into the `HeartbeatController` so it can send it out with the 200.

What should the `HeartbeatController` do in this new regime? Well, here's [an Internet-Draft](https://inadarei.github.io/rfc-healthcheck/) that was recently published which I think would be useful. 

We don't need to implement that whole document right away. It would be enough to start sending the `application/vnd.health+json` media type and a static document that looks like this:

```
{
  "status": "pass",
  "releaseID": "2.1.5-58-g8c0bd99",
}
```

As time goes on we can move the stuff in ServiceStatusController into this document. Rather than create a custom visualizer for our service statuses we can make a generic `application/vnd.health+json` visualizer, or hopefully someone else will make one we can use.

So, for purposes of this branch, read from the file in `load_settings` and pass a static health document into the controller that processes `/heartbeat`. This will give us a starting point for future improvements.

courte:
Whoa, the details object looks made for our service_status endpoint. 😎 I'll clean this branch up today.

courte:
Okay, this is ready for another look. Because .version is so tied in with the system, I want to keep it in the `Configuration` instance. Even if the database somehow gets out of sync, Configuration will load the version from the filesystem.

app_server.py
```diff
@@ -246,8 +247,20 @@ def handle(self, exception):
...
+    VERSION_FILENAME = '.version'
+
+    def heartbeat(self, conf_class=None):
+        health_check_object = dict(status='pass')
+
+        Conf = conf_class or Configuration
+        app_version = Conf.app_version()
+        if app_version and app_version != Conf.NO_APP_VERSION_FOUND:
+            health_check_object['releaseID'] = app_version
+            health_check_object['version'] = app_version.split('-')[0]
```
leonardr: I think it would be useful to provide the whole thing here. The client can always decide to only show part of the version number.

Alternatively, I would change this code to `.split('-', 1)[0]`. That will do the same thing in the normal case, but will also work in the weird case where the full app version doesn't contain a dash.

courte: The Internet-Draft says that 'version' should represent the official release and 'releaseID' should represent any other changes. In the case of `git describe` the additional values would be the number of commits from that last official version and the commit sha at HEAD, which I think constitute patches / releaseID. (In production, releaseID and version will likely be the same. It's in the `development` containers or production hot-fixes that would show a difference.)

`split` returns an array with only one value even without the second parameter. (1 might be the default?)

leonardr: Well, you've got me there!

PR #787, opened 2018-01-26
A Collection may have a designated ExternalIntegration that's used to mirror things on OPDS import 

leonardr:
This branch adds a relationship between Collection and ExternalIntegration that says any open-access books or book covers discovered in the course of an OPDS import into the Collection should be mirrored externally as configured by the ExternalIntegration. As an example, you could set this up so that every time you do an OPDS import from Standard Ebooks, everything you find is mirrored to your own S3 bucket.

This makes it possible to duplicate a lot of the content server's functionality within the circulation manager.

I ripped up a lot of the S3 initialization code, which assumed there was a single S3 integration that was used site-wide and instantiated as necessary. S3Uploaders are no longer instantiated directly; you're supposed to go through MirrorUploader.sitewide or MirrorUploader.for_collection. This opens up the possibility of mirror techniques other than "upload to S3".

We still support the case where there is one sitewide integration (i.e. the metadata wrangler), but it relies on the database containing one and only one integration with GOAL="storage", so it's a little fragile.

I also made sure there was test coverage for every method in s3.py.

This branch is ready for review but I won't be merging it until I make any necessary changes to circulation and metadata. (I'm not sure about content -- this is the beginning of the end for that component and it might not be a good use of time to keep it compatible.)

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -2470,7 +2470,7 @@ class TestWork(DatabaseTest):
             lp3, complaint_type, "blah", "blah"
         )
 
-        eq_(set([complaint1, complaint2]), set(work.complaints))
+        eq_([complaint1, complaint2], work.complaints)
         assert complaint3 not in work.complaints
 
     def test_all_identifier_ids(self):
@@ -5078,17 +5078,6 @@ class TestRepresentation(DatabaseTest):
 
 
 class TestCoverResource(DatabaseTest):
-
-    def sample_cover_path(self, name):
-        base_path = os.path.split(__file__)[0]
-        resource_path = os.path.join(base_path, "files", "covers")
-        sample_cover_path = os.path.join(resource_path, name)
-        return sample_cover_path
-
-    def sample_cover_representation(self, name):
-        sample_cover_path = self.sample_cover_path(name)
-        return self._representation(
-            media_type="image/png", content=open(sample_cover_path).read())[0]
 
     def test_set_cover(self):
         edition, pool = self._edition(with_license_pool=True)
--- a/util/mirror.py
+++ b/util/mirror.py
@@ -47,7 +47,7 @@ class MirrorUploader(object):
         return cls.implementation(integration)
 
     @classmethod
-    def for_collection(cls, _db, collection):
+    def for_collection(cls, collection):
         """Create a MirrorUploader for the given Collection.
 
         :param collection: Use the mirror configuration for this Collection.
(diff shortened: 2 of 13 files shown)


tests/test_model.py
```diff
@@ -2470,7 +2470,7 @@ def test_complaints(self):
             lp3, complaint_type, "blah", "blah"
         )
 
-        eq_(set([complaint1, complaint2]), set(work.complaints))
+        eq_([complaint1, complaint2], work.complaints)
```
leonardr: This is an unrelated change to stop a nondeterministic test failure.

aslagle:
Does anything currently use separate buckets for cover images and books? I thought it did but I might be thinking of CDNs. I don't think it's necessary anyway.

leonardr:
NYPL's copies of open-access books are kept in the oabooks.nypl.org bucket and the book covers (open-access or not) are kept in the book-covers.nypl.org bucket. I don't know if there's a single S3Uploader that handles both books and book covers. But I also feel like I don't fully understand your question. Are you saying we could change an S3 integration so that it was only associated with a single bucket?

aslagle:
Yeah I just saw that in s3.py. I was wondering if having separate buckets would still be possible with this branch but I see that it is anyway. 

I do think the buckets should be configurable though - it seems less likely to me that anyone will want different S3 accounts (or different storage services) for different collections than that they will want different buckets within the same account.

aslagle:
To clarify my last comment, I think the buckets could be a collection + integration setting rather than an integration setting. But we don't currently have that type of setting, and it's not that hard to set up multiple integrations for the same S3 account, so it probably would not be worth it to make that change.

util/mirror.py
```diff
@@ -6,6 +6,88 @@ class MirrorUploader(object):
...
+
+        :param integration: An ExternalIntegration configuring the credentials
+           used to upload things.
+        """
+        from config import CannotLoadConfiguration
+        if integration.goal != self.STORAGE_GOAL:
+            # This collection's 'mirror integration' isn't intended to
+            # be used to mirror anything.
+            raise CannotLoadConfiguration(
+                "Cannot create an S3Uploader from an integration with goal=%s" %
```
aslagle: This might not be an S3Uploader.

leonardr:
Yeah, I created collections.mirror_integration because I didn't want to do more work than that to handle the concept of an integration associated with a collection.

leonardr:
Metadata branch is here: https://github.com/NYPL-Simplified/metadata_wrangler/pull/160

Circulation doesn't need anything but a submodule update.

Content is already broken from the lane refactoring we did a while back, so I say we forge ahead and do what it takes to get rid of that component altogether.

PR #788, opened 2018-01-29
Associate custom lists with collections, and add new works to the custom lists when they're imported

aslagle:


model.py
```diff
@@ -11083,6 +11094,39 @@ class CollectionIdentifier(object):
...
+    UniqueConstraint('collection_id', 'customlist_id'),
+)
+
+# When a pool gets a work and a presentation edition for the first time,
+# the work should be added to any custom lists associated with the pool's
+# collection.
+# In some cases, the work may be generated before the presentation edition.
+# Then we need to add it when the work gets a presentation edition.
+@event.listens_for(LicensePool.work_id, 'set')
+@event.listens_for(Work.presentation_edition_id, 'set')
```
aslagle: It was tricky to figure out what events would work here. `@event.listens_for(Collection.licensepools, 'append')` and `@event.listens_for(LicensePool, 'after_insert')` unfortunately don't work, since custom lists are associated with works rather than license pools.

tests/test_model.py
```diff
@@ -7821,6 +7821,22 @@ def assert_isbns(expected, result_query):
...
+        eq_(0, len(list2.entries))
+
+        # When a new pool is added to the collection and its presentation edition is
+        # calculated for the first time, it's automatically added to the lists.
+        work = self._work(collection=self.collection, with_license_pool=True)
+        eq_(1, len(list1.entries))
+        eq_(1, len(list2.entries))
+        eq_(work, list1.entries[0].work)
+        eq_(work, list2.entries[0].work)
+
```
leonardr: Suggest you also add a test that verifies nothing happens when a Work's presentation edition changes or a LicensePool changes works.

PR #789, opened 2018-01-30
Guess sort name based on display name in the metadata layer instead of leaving out the author.

aslagle:
When I import from an OPDS feed with no sort names and there's no metadata wrangler integration, the metadata layer should guess the sort names rather than leaving out the contributions.


aslagle:
@leonardr could you look at this again?

leonardr:
🎇

PR #790, opened 2018-01-31
Support for Feedbooks import

leonardr:
This branch adds the core support for https://github.com/NYPL-Simplified/circulation/pull/844. It also adds a migration script that I created earlier but neglected to `git add` in the branch that should have had it.

### DIFF
--- a/model.py
+++ b/model.py
@@ -10039,6 +10039,8 @@ class ExternalIntegration(Base, HasFullTableCache):
     ONE_CLICK = RB_DIGITAL
     OPDS_FOR_DISTRIBUTORS = u'OPDS for Distributors'
     ENKI = DataSource.ENKI
+    FEEDBOOKS = DataSource.FEEDBOOKS
+    UNGLUEIT = DataSource.UNGLUEIT
 
     # These protocols are only used on the Content Server when mirroring
     # content from a given directory or directly from Project
--- a/model.py
+++ b/model.py
@@ -10040,7 +10040,6 @@ class ExternalIntegration(Base, HasFullTableCache):
     OPDS_FOR_DISTRIBUTORS = u'OPDS for Distributors'
     ENKI = DataSource.ENKI
     FEEDBOOKS = DataSource.FEEDBOOKS
-    UNGLUEIT = DataSource.UNGLUEIT
 
     # These protocols are only used on the Content Server when mirroring
     # content from a given directory or directly from Project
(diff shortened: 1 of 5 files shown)


model.py
```diff
@@ -10039,6 +10039,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     ONE_CLICK = RB_DIGITAL
     OPDS_FOR_DISTRIBUTORS = u'OPDS for Distributors'
     ENKI = DataSource.ENKI
+    FEEDBOOKS = DataSource.FEEDBOOKS
```
leonardr: Although Feedbooks titles are imported via OPDS, the import procedure is significantly different from a normal OPDS import, to the extent that we treat it as a different protocol with its own configuration settings.

opds_import.py
```diff
@@ -1456,13 +1456,13 @@ def __init__(self, _db, collection, import_class,
                 )
             )
 
-        data_source = collection.data_source
+        data_source = self.data_source(collection)
```
leonardr: The 'Feedbooks' protocol has a hard-coded data source (Feedbooks) and, unlike other OPDS-type protocols, doesn't rely on a data source being configured on the collection.

opds_import.py
```diff
@@ -1456,13 +1456,13 @@ def __init__(self, _db, collection, import_class,
...
         if not data_source:
             raise ValueError(
                 "Collection %s has no associated data source." % collection.name
             )
 
-        self.feed_url = collection.external_account_id
+        self.feed_url = self.opds_url(collection)
```
leonardr: The 'Feedbooks' protocol derives its OPDS URL from a 'language' setting rather than by an explicit 'URL' setting.

scripts.py
```diff
@@ -1603,7 +1603,14 @@ class OPDSImportScript(CollectionInputScript):
...
     PROTOCOL = ExternalIntegration.OPDS_IMPORT
-    
+
+    def __init__(self, _db=None, importer_class=None, monitor_class=None, 
+                 protocol=None, *args, **kwargs):
+        super(OPDSImportScript, self).__init__(_db, *args, **kwargs)
+        self.importer_class = importer_class or self.IMPORTER_CLASS
+        self.monitor_class = monitor_class or self.MONITOR_CLASS
+        self.protocol = protocol or self.PROTOCOL
+
```
leonardr: This allows a script to invoke the OPDS import process using different rules for how the import is done, as well as which collections should have their imports refreshed under these rules.

Issue #480, opened 2017-03-21
Genre name must be unique


leonardr:
There's no reason to have two Genres with the same name, and a race condition can cause duplicate genres to be created. Putting a uniqueness constraint on Genre.name will flush out the race condition and stop the problem from affecting the app server.

PR #794, opened 2018-02-02
Add a unique index to genres.name.

leonardr:
Just a quick branch to fix an old issue (https://github.com/NYPL-Simplified/server_core/issues/480).

The migration script may cause problems on servers that have multiple genres with the same name, but none of NYPL's production servers are like that, so hopefully the race condition that called attention to this problem is also gone.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5876,7 +5876,7 @@ class Genre(Base, HasFullTableCache):
     """
     __tablename__ = 'genres'
     id = Column(Integer, primary_key=True)
-    name = Column(Unicode)
+    name = Column(Unicode, unique=True, index=True)
 
     # One Genre may have affinity with many Subjects.
     subjects = relationship("Subject", backref="genre")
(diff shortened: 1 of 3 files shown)


model.py
```diff
@@ -8708,7 +8708,7 @@ def cautious_http_get(cls, url, headers, **kwargs):
 
     # Sites that cause problems for us if we make automated
     # HTTP requests to them while trying to find free books.
-    AVOID_WHEN_CAUTIOUS_DOMAINS = ['gutenberg.org']
+    AVOID_WHEN_CAUTIOUS_DOMAINS = ['gutenberg.org', 'books.google.com']
```
leonardr: This is an unrelated change to make sure an OPDS import from unglue.it doesn't try to download PDFs from Google Books. (It's impossible for an automated client to get a PDF out of Google Books.)

courte (approved):

PR #795, opened 2018-02-02
Copy the MARCExtractor into core and clean it up slightly.

leonardr:
https://memegenerator.net/img/instances/500x/80680447/oh-hi-mark.jpg

This branch basically copies the MARCExtractor from content into core. I'm about to start work on a circulation branch that lets you build a collection out of a MARC file and a directory full of ebooks.

The code relating to cleaning up the incoming author names should be used more generally, but I didn't make any changes apart from refactoring the code so it could be tested separately, and adding a call to `name_tidy` to get rid of the annoying trailing periods after someone's name.


tests/test_metadata.py
```diff
@@ -1154,3 +1156,46 @@ def test_success(self):
...
+
+class TestMARCExtractor(DatabaseTest):
+
+    def setup(self):
+        super(TestMARCExtractor, self).setup()
+        base_path = os.path.split(__file__)[0]
+        self.resource_path = os.path.join(base_path, "files", "marc")
+
+    def sample_data(self, filename):
+        return open(os.path.join(self.resource_path, filename)).read()
```
courte: Can you return the file contents from within an `with open(...) as` block or use the `sample_data` method from `tests/__init__.py`, so that the file is closed after the fact?

PR #797, opened 2018-02-07
Crawlable acquisition feeds for lists

aslagle:
This branch makes a few changes to facilitate crawlable acquisition feeds for lists.

The main things are:
- I added CustomListEntry.list_id and CustomListEntry.first_appearance to the materialized view, in order to sort by a combination of Work.last_update_time and CustomListEntry.first_appearance.
- I changed the way updated is added to OPDS entries.
- I changed a unique constraint on CustomList so that names must be unique per library, rather than per library and data source. This was so I could have nicer URLs for the crawlable feeds that only need the library and the list name.

I don't think anyone would already have two lists with the same library and the same name, but if they do they'll get an error.

### DIFF
--- a/files/materialized_view_for_lanes.sql
+++ b/files/materialized_view_for_lanes.sql
@@ -52,7 +52,7 @@ as
   WITH NO DATA;
 
 -- First create an index that allows work/genre lookup. It's unique and incorporates license_pool_id so that the materialized view can be refreshed CONCURRENTLY.
-create unique index mv_works_for_lanes_unique on mv_works_for_lanes (works_id, genre_id, license_pool_id, list_id);
+create unique index mv_works_for_lanes_unique on mv_works_for_lanes (works_id, genre_id, list_id, license_pool_id);
 
 -- Create an index on everything, sorted by descending availability time, so that sync feeds are fast.
 
--- a/lane.py
+++ b/lane.py
@@ -42,7 +42,6 @@ from sqlalchemy.orm import (
     relationship,
 )
 from sqlalchemy.sql.expression import literal
-from sqlalchemy.sql.functions import Function
 
 from model import (
     get_one_or_create,
@@ -301,7 +300,7 @@ class Facets(FacetConstants):
         ]
     
         primary_order_by = self.order_facet_to_database_field(self.order)
-        if isinstance(primary_order_by, Function) or primary_order_by:
+        if primary_order_by is not None:
             # Promote the field designated by the sort facet to the top of
             # the order-by list.
             order_by = [primary_order_by]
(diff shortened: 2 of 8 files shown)


files/materialized_view_for_lanes.sql
```diff
@@ -49,12 +52,16 @@ as
   WITH NO DATA;
 
 -- First create an index that allows work/genre lookup. It's unique and incorporates license_pool_id so that the materialized view can be refreshed CONCURRENTLY.
-create unique index mv_works_for_lanes_unique on mv_works_for_lanes (works_id, genre_id, license_pool_id);
+create unique index mv_works_for_lanes_unique on mv_works_for_lanes (works_id, genre_id, license_pool_id, list_id);
```
leonardr: I'm not sure if it matters, but it seems to me that list_id should go first. It's more likely that you'd be doing a restriction based on list_id than that you'd be doing one based on license_pool_id.

lane.py
```diff
@@ -300,7 +301,7 @@ def order_by(self):
         ]
     
         primary_order_by = self.order_facet_to_database_field(self.order)
-        if primary_order_by:
+        if isinstance(primary_order_by, Function) or primary_order_by:
```
leonardr: Could you do this as `primary_order_by is not None`?

aslagle: That doesn't work when `primary_order_by` is a column - the normal case.

aslagle: Although, I tried `!=` rather than `is not`, I don't know if that would make a difference. I think `!=` tries to make it into a clause, which can't be used as a boolean.

migration/20180207-2-recreate-materialized-view.sql
```diff
@@ -0,0 +1,133 @@
+drop materialized view mv_works_for_lanes;
+create materialized view mv_works_for_lanes
```
leonardr: When I created `mv_works_for_lanes` I just let `SessionManager.initialize` create it from `materialized_view_for_lanes.sql` on startup. So it should be possible to just drop it and let it get recreated on next startup, rather than drop it and recreate it within the migration script.

This is risky though, because usually there are multiple processes 'starting up' at the same time, and each one might try to create the view.

Dropping and recreating the view within the migration script is also fine, but you'll need to explicitly refresh the view once it's recreated. It's no longer enough for a migration script like this to be a copy of `files/materialized_view_for_lanes.sql`, because `initialize` will only refresh a view that it creates. If the view exists but is empty (as it will be after this script runs), it will stay empty until the next time `refresh_materialized_views` runs.

lane.py
aslagle: Yes, `is not` seems to work.

PR #798, opened 2018-02-07
Upload script support

leonardr:
This is a small support branch for https://github.com/NYPL-Simplified/circulation/pull/852.

The biggest change is that the `util.mirror` module was moved into `mirror`. Most of the code remaining in that module dealt with the database model, and `util` is for things that don't deal with the database model.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -621,7 +621,9 @@ class MetaToModelUtility(object):
 
         # Determine the best URL to use when mirroring this
         # representation.
-        if title and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+        if link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+            if not title:
+                title = identifier.identifier
             extension = representation.extension()
             mirror_url = mirror.book_url(
                 identifier, data_source=data_source, title=title,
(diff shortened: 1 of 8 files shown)


metadata_layer.py
```diff
@@ -621,7 +621,9 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
 
         # Determine the best URL to use when mirroring this
         # representation.
-        if title and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+        if link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+            if not title:
+                title = identifier.identifier
```
leonardr: In my test this was causing the link to be ignored if the book had no title.

metadata_layer.py
```diff
@@ -621,10 +621,11 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
 
         # Determine the best URL to use when mirroring this
         # representation.
-        if title and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+        if link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+            url_title = title or identifier.identifier
```
leonardr: During development this was causing the link to not be mirrored if no title was provided. The book really needs to have a title to become a Work, but we can still come up with a URL for it even without a title.

PR #800, opened 2018-02-08
Fixed duplicate list entries when non-equivalent editions had the same work.

aslagle:


model.py
```diff
@@ -9380,7 +9380,7 @@ def remove_entry(self, work_or_edition):
         if isinstance(work_or_edition, Work):
```
leonardr: I don't think you need these three lines anymore since this method no longer uses `edition`.

PR #802, opened 2018-02-11
Added username and password options to OPDS Import.

aslagle:
I need to import from an OPDS feed that has open access content, but requires a username and password to access the feed (not the books and covers, just the feed itself). This branch makes it possible to configure a username and password for an OPDS feed.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -358,13 +358,13 @@ class OPDSImporter(object):
             "key": ExternalIntegration.USERNAME,
             "label": _("Username"),
             "optional": True,
-            "description": _("If authentication is required to access the OPDS feed, enter the username here."),
+            "description": _("If HTTP Basic authentication is required to access the OPDS feed (it usually isn't), enter the username here."),
         },
         {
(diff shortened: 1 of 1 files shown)


opds_import.py
```diff
@@ -353,7 +353,19 @@ class OPDSImporter(object):
...
+            "key": ExternalIntegration.USERNAME,
+            "label": _("Username"),
+            "optional": True,
+            "description": _("If authentication is required to access the OPDS feed, enter the username here."),
+        },
+        {
+            "key": ExternalIntegration.PASSWORD,
+            "label": _("Password"),
+            "optional": True,
+            "description": _("If authentication is required to access the OPDS feed, enter the password here."),
```
leonardr: I like that this is part of the OPDS import rather than a whole new protocol, but I think the description needs something to distinguish between this case and the OPDS For Distributors case. Something like this:

"If HTTP Basic authentication is required to access the OPDS feed (it usually isn't)..."

PR #803, opened 2018-02-12
Make the materialized view capable of handling a single Work that's on a list multiple times

leonardr:
This happens in some cases where multiple entries on a list are fuzzily matched against the same work. It primarily happens with the NYT bestsellers list, which is why we didn't find the problem in initial testing--it took a big preexisting dataset to expose the problem.


migration/20180207-2-recreate-materialized-view.sql
```diff
@@ -36,6 +38,7 @@ as
     licensepools.availability_time,
     licensepools.collection_id,
     customlistentries.list_id,
+    customlistentries.edition_id as list_edition_id,
```
aslagle: Changing this migration file means that someone who already deployed the development version with this migration won't have the update run automatically. I propose changing the date on the migration file whenever you change it, even if it's not in a release yet.

PR #804, opened 2018-02-15
Make LicensePoolDeliveryMechanism.set more robust

leonardr:
This is a support branch for https://github.com/NYPL-Simplified/circulation/pull/862

This branch makes it possible to avoid the `_db.commit()` call when calling `LicensePoolDeliveryMechanism.set()`, which will cause a problem if you call it from within a nested transaction.

The downside of doing this is that if you call `set(autocommit=False)`, it's possible that a `LicensePool`'s open-access status will change but `LicensePool.open_access` won't be updated. This isn't currently a problem with the code I've added to call `set(autocommit=False)`, since Overdrive is the only real-world distributor that will currently cause this code to run, and I don't think it will be a problem in general, but I'd really like to get rid of that _db.commit() call altogether.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5422,10 +5422,11 @@ class LicensePoolDeliveryMechanism(Base):
             a freely redistributable form.
         :param autocommit: Commit the database session immediately if
             anything changes in the database. If you're already inside
-            a nested session, pass in False here, but understand that if 
-            a LicensePool's open-access status changes as a result of 
-            calling this method, the change may not be properly reflected
-            in LicensePool.open_access.
+            a nested transaction, pass in False here to avoid
+            committing prematurely, but understand that if a
+            LicensePool's open-access status changes as a result of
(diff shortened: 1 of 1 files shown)


model.py
```diff
@@ -5420,32 +5420,45 @@ def set(cls, data_source, identifier, content_type, drm_scheme, rights_uri,
...
         )
-        lpdm.rights_status = rights_status
-
-        # TODO: We need to explicitly commit here so that
-        # LicensePool.delivery_mechanisms gets updated. It would be
-        # better if we didn't have to do this, but I haven't been able
-        # to get LicensePool.delivery_mechanisms to notice that it's
-        # out of date.
-        _db.commit()
+        if not lpdm.rights_status or rights_status.uri != RightsStatus.UNKNOWN:
```
aslagle: So if there was already a known rights status we can't change it?

PR #808, opened 2018-02-20
Logging configuration

EdwinGuzman:
This PR adds the appropriate data structure for the logging configurations.


leonardr:
I would expect to see some changes to `log.py:LogConfiguration.from_configuration` as well. That code assumes that configuration settings like LOG_LEVEL and LOG_FORMAT come from an `ExternalIntegration` with goal=LOGGING and protocol=INTERNAL_LOGGING. What you've done with the circulation branch is store those configuration settings stored in site-wide `ConfigurationSetting`s instead. So that `ExternalIntegration` doesn't exist anymore.

So `from_configuration` needs to be changed to look up those settings with `ConfigurationSetting.sitewide` instead of calling `setting` on a nonexistent `ExternalIntegration`.

I don't think we need a migration script because I can just reconfigure NYPL's sites once we release this code, but it should be pretty simple to write a migration script and I can walk you through it if you want.

config.py
```diff
@@ -130,7 +129,27 @@ class Configuration(object):
...
+
+    # Settings for the integration with protocol=INTERNAL_LOGGING
+    LOG_LEVEL = 'log_level'
+    LOG_APP_NAME = 'log_app'
+    DATABASE_LOG_LEVEL = 'database_log_level'
+    LOG_LEVEL_UI = [
+        { "key": DEBUG, "value": _("Debug") },
+        { "key": INFO, "value": _("Info") },
+        { "key": WARN, "value": _("Warn") },
+        { "key": ERROR, "value": _("Error") },
```
aslagle: should these be "label"?

config.py
```diff
@@ -144,6 +163,9 @@ class Configuration(object):
             "key": BASE_URL_KEY,
             "label": _("Base url of the application"),
         },
+        { "key": LOG_LEVEL, "label": _("Log Level"), "type": "select", "options": LOG_LEVEL_UI },
+        { "key": LOG_APP_NAME, "label": _("Log App") },
```
aslagle: This could use a better label or a description, I wouldn't know what it is.

log.py
```diff
@@ -52,7 +53,45 @@ def format(self, record):
...
+    LOG_MESSAGE_TEMPLATE = 'message_template'
+
+    SETTINGS = [
+        { "key": LOG_FORMAT, "label": _("Log Format"), "type": "select",
+            "options": [
+                { "key": JSON_LOG_FORMAT, "label": _("json") },
+                { "key": TEXT_LOG_FORMAT, "label": _("text") }
+            ]
+        },
+        { "key": LOG_MESSAGE_TEMPLATE, "label": _("template") }
```
aslagle: Maybe add a description that links to documentation on the message template format?

config.py
aslagle: Ah, I see there is one in LogConfiguration. I'm not sure why this setting is in two places though.

aslagle:
I'm not sure what happens now, but I think it's important for the SysLogger integration to be created automatically with default settings when the app first runs - otherwise you might not be able to see logs at all if the app doesn't work. Later, you could edit or remove it from the admin interface.

config.py
EdwinGuzman: Yes, will update.

config.py
EdwinGuzman: This may be from some confusion on my part. But I moved these values to LogConfiguration which is where it gets picked up from.

EdwinGuzman:
@aslagle updated based on feedback. I'm not sure how to set the SysLogger values by default automatically. Are those handled by the `default` property in the objects."

aslagle:
You'd probably need to add something that runs when the LogConfiguration is initialized and creates a SysLog ExternalIntegration if there aren't any logging integrations yet.

log.py
```diff
@@ -52,33 +53,139 @@ def format(self, record):
             data = data.encode("utf8")
         return data
 
-       
+class SysLogger():
```
leonardr: It's better to use `object` as the superclass of Python classes that would otherwise have no superclass.

log.py
```diff
@@ -162,27 +269,30 @@ def from_configuration(cls, _db, testing=False):
             loggly = ExternalIntegration.lookup(
                 _db, ExternalIntegration.LOGGLY, goal
             )
+
```
leonardr: This code works, but you haven't accomplished the refactoring I was hoping for. I believe this entire method can be replaced with something like this:

```
handlers = []
handlers.append(Syslog.from_configuration(_db))
loggly = Loggly.from_configuration(_db)
if loggly:
    handlers.append(loggly)
cls.replace_handlers(_db, handlers)
```

Most of the logic currently in this method would go into `Syslog.from_configuration`; most of the rest would go into `LogConfiguration.replace_handlers`. You've already got most of the Loggly code in `Loggly.from_configuration`, but this would include the part that finds the ExternalIntegration for a Loggly configuration and pulls configuration settings from the database.

Probably you would end up getting the sitewide settings multiple times -- once for the syslog, once for the Loggly configuration, and a third time to set the log levels -- but that's fine because this code is only called once, on startup -- it's more important that the code be modular and easy to read. Right now it's kind of a mess (which is my fault as I wrote the original code). You've made it work with the new system, but it still should be refactored so that all the code pertaining to the syslogger goes into `Syslogger`.

log.py
```diff
@@ -52,33 +53,221 @@ def format(self, record):
             data = data.encode("utf8")
         return data
 
-       
+class Logger(object):
+
+    JSON_LOG_FORMAT = 'json'
+    TEXT_LOG_FORMAT = 'text'
+
+    DEFAULT_MESSAGE_TEMPLATE = "%(asctime)s:%(name)s:%(levelname)s:%(filename)s:%(message)s"
```
leonardr: These three constants can live in SysLogger.

log.py
```diff
@@ -52,33 +53,221 @@ def format(self, record):
...
+    INFO = "INFO"
+    WARN = "WARN"
+    ERROR = "ERROR"
+
+    DEFAULT_APP_NAME = 'simplified'
+
+    # Settings for the integration with protocol=INTERNAL_LOGGING
+    LOG_FORMAT = 'log_format'
+    LOG_MESSAGE_TEMPLATE = 'message_template'
+
```
leonardr: These two can also live in SysLogger.

log.py
```diff
@@ -52,33 +53,221 @@ def format(self, record):
...
+        if _db and not testing:
+            goal = ExternalIntegration.LOGGING_GOAL
+            loggly = ExternalIntegration.lookup(
+                _db, ExternalIntegration.LOGGLY, goal
+            )
+            app_name = ConfigurationSetting.sitewide(_db, Configuration.LOG_APP_NAME).value or app_name
+
+        if loggly:
+            loggly = Loggly.loggly_handler(loggly)
+            cls.set_formatter(loggly, cls.LOG_FORMAT, cls.DEFAULT_MESSAGE_TEMPLATE, app_name)
```
leonardr: This class doesn't need any references to LOG_FORMAT or DEFAULT_MESSAGE_TEMPLATE. It's okay for this set_formatter() to have a different signature, because it's only called from from_configuration().

EdwinGuzman:
Got it. Updated again, thanks.

PR #810, opened 2018-02-22
Reduce the number of times the site configuration needs to change, and avoid contention when it does need to change

leonardr:
This branch fixes a bug that caused the site configuration timestamp to change a lot more than it needed to. It also introduces safeguards to stop different threads in a process from all changing that timestamp at once, and to reduce the risk that multiple different processes will try to change the timestamp at once.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1908,7 +1908,7 @@ lanes_customlists = Table(
 
 @event.listens_for(Lane, 'after_insert')
 @event.listens_for(Lane, 'after_delete')
-# @event.listens_for(Lane, 'after_update')
+@event.listens_for(Lane, 'after_update')
 @event.listens_for(LaneGenre, 'after_insert')
 @event.listens_for(LaneGenre, 'after_delete')
 @event.listens_for(LaneGenre, 'after_update')
--- a/lane.py
+++ b/lane.py
@@ -1908,7 +1908,7 @@ lanes_customlists = Table(
 
 @event.listens_for(Lane, 'after_insert')
 @event.listens_for(Lane, 'after_delete')
-@event.listens_for(Lane, 'after_update')
+# @event.listens_for(Lane, 'after_update')
 @event.listens_for(LaneGenre, 'after_insert')
 @event.listens_for(LaneGenre, 'after_delete')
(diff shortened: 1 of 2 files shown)


lane.py
```diff
@@ -1908,9 +1909,14 @@ def customlist_filter_clauses(
...
-@event.listens_for(LaneGenre, 'after_update')
 def configuration_relevant_lifecycle_event(mapper, connection, target):
     site_configuration_has_changed(target)
+
+
+@event.listens_for(Lane, 'after_update')
+@event.listens_for(LaneGenre, 'after_update')
+def configuration_relevant_update(mapper, connection, target):
+    if directly_modified(target):
+        site_configuration_has_changed(target)
```
leonardr: This is the bug fix. An "after_update" doesn't necessarily mean anything has actually been updated in database terms. We ran into this earlier with our other data model objects, which is why we created `directly_modified`.

model.py
```diff
@@ -11388,6 +11390,20 @@ def site_configuration_has_changed(_db, timeout=1):
     number of seconds since the last site configuration change was
     recorded.
     """
+    has_lock = site_configuration_has_changed_lock.acquire(blocking=False)
```
leonardr: Since blocking=False this will immediately return, True if we were able to get the thread lock and False if not.

Since the goal of this code is to make sure a timestamp is updated, it doesn't have to run if another thread is currently running it.

model.py
```diff
@@ -11402,11 +11418,12 @@ def site_configuration_has_changed(_db, timeout=1):
 
         # Update the timestamp.
         now = datetime.datetime.utcnow()
-        sql = "UPDATE timestamps SET timestamp=:timestamp WHERE service=:service AND collection_id IS NULL;"
+        earlier = now-datetime.timedelta(seconds=1)
+        sql = "UPDATE timestamps SET timestamp=:timestamp WHERE service=:service AND collection_id IS NULL AND timestamp<=:earlier;"
```
leonardr: Adding this clause to the SQL statement makes it likely that if two different processes run this code nearly simultaneously, the second one will end without doing anything because its UPDATE statement will match no rows.

PR #811, opened 2018-02-22
Support multithreaded provider script

courte:
Finally, finally, finally -- a PR that supports running CollectionCoverageProviders on multiple threads. Mimicking some of the work in https://github.com/shazow/workerpool, this branch introduces pools of threads and uses locks to protect OAuth credential refreshing for Overdrive.

There's some refactoring as well, which I'll explain in comments.


coverage.py
```diff
@@ -919,6 +921,17 @@ def _default_replacement_policy(self, _db):
         """
         return ReplacementPolicy.from_license_source(_db)
 
+    @classmethod
+    def collections(cls, _db):
```
courte: Pulled out of `CollectionCoverageProvider.all` below, to support any reordering work done by subclasses (in this case `IdentifierResolutionCoverageProvider`).

coverage.py
```diff
@@ -1104,6 +1111,22 @@ def set_presentation_ready(self, identifier):
         return identifier
 
 
+class CollectionCoverageProviderJob(Job):
+
+    def __init__(self, collection, provider_class, item_offset,
+        **provider_kwargs
+    ):
+        self.collection = collection
```
courte: I believe there's a way to create the provider at the thread/`DatabaseWorker` level, which may greatly simplify the work that each `Job` does. Initial tests lead to integrity errors, so I'm leaving it for a future PR.

model.py
```diff
@@ -328,13 +328,18 @@ class SessionManager(object):
...
 
     @classmethod
-    def sessionmaker(cls, url=None):
-        engine = cls.engine(url)
+    def sessionmaker(cls, url=None, session=None):
+        if not (url or session):
+            url = Configuration.database_url()
+        if url:
+            engine = cls.engine(url)
+        elif session:
```
courte: This addition allows a session factory to be created from an existing Session object.

tests/test_util_worker_pools.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # The database was expired for each job, then called one more
+        # time before finding out the queue was empty.
+        eq_(3, mock_db._expired)
+
+        # It was only committed for the working job.
+        eq_(1, mock_db._committed)
+
+        # The DatabaseWorker doesn't roll back the database. It
+        # trusts the task to manage that.
```
courte: I think this makes the most sense, given how coverage providers handle the work, but I'm open to other opinions.

model.py
```diff
@@ -328,13 +328,18 @@ class SessionManager(object):
     engine_for_url = {}
 
     @classmethod
-    def engine(cls, url=None):
+    def engine(cls, url=None, session=None):
```
leonardr: I doesn't look like session is used here.

scripts.py
```diff
@@ -292,12 +302,59 @@ def __init__(self, provider_class, _db=None, providers=None, **kwargs):
...
+
+        self.provider_class = provider_class
+        self.provider_kwargs = provider_kwargs
+
+    def worker_factory(self, job_queue):
+        """Returns a Worker with a database"""
+        return DatabaseWorker(job_queue, scoped_session(self.session_factory))
+
+    def run(self):
+        _db = scoped_session(self.session_factory)
```
leonardr: I don't understand scoped_session well enough to know if this is a problem, but I'm concerned that you're calling scoped_session on the same factory twice, once to make the DatabaseWorker and once within run(). Will both of these calls give the same scoped session or will there be two different scoped sessions?

tests/test_util_worker_pools.py
leonardr: Is the rationale that coverage providers are responsible for committing, so they are also responsible for rollback if something goes wrong?

scripts.py
courte: So, this is ultimately working because I'm creating different a new `scoped_session` for each thread, but yeah, it's not doing what I had hoped it was. I'm going to try to adjust it to use the factory directly, since scoping isn't giving me anything.

scripts.py
```diff
@@ -292,12 +302,59 @@ def __init__(self, provider_class, _db=None, providers=None, **kwargs):
...
+        if not collections:
+            return
+
+        with Pool(self.worker_size, self.worker_factory) as pool:
+            for collection in collections:
+                offset = 0
+                query_size, batch_size = self.get_query_and_batch_sizes(
+                    collection
+                )
+                _db.commit()
```
leonardr: `get_query_and_batch_sizes` doesn't seem like a method name that needs a commit. Presumably the commit is because it creates a provider object to estimate the collection size, and that  might do anything. Could you add a comment to this effect?

model.py
courte: Thanks - 'twas a throwback.

coverage.py
courte: I've attempted this twice now, and both resulted in lots of data integrity errors (usually around the 'id' set at db insert time for various tables).

leonardr:
What an adventure!

PR #812, opened 2018-03-01
Improve materialized view performance

courte:
This branch improves materialized view performance for search queries using indices and an improved join condition. Against the production database, these changes decrease query execution time from the 40-80 second range to a half-second or less. Planning time also decreases.

The foreign key indices were added at an earlier stage and may be unnecessary with the unique `mv_works_for_lanes_unique` index, but I saw slight performance improvements by adding the `mv_works_for_lanes_works_id` index. I also think it's a good practice to put an index on all foreign keys. I'm open to other thoughts on this, though.

Resolves NYPL-Simplified/circulation#872.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -810,8 +810,9 @@ class WorkList(object):
         qu = _db.query(mw).join(
             LicensePool, mw.license_pool_id==LicensePool.id
         ).filter(
-            mw.works_id.in_(work_ids)
-        )
+            mw.works_id.in_(work_ids),
+            LicensePool.work_id.in_(work_ids),
+        ).enable_eagerloads(False)
         qu = self._lazy_load(qu)
         qu = self._defer_unused_fields(qu)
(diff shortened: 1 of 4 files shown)


lane.py
```diff
@@ -810,8 +810,9 @@ def works_for_specific_ids(self, _db, work_ids):
         qu = _db.query(mw).join(
             LicensePool, mw.license_pool_id==LicensePool.id
         ).filter(
-            mw.works_id.in_(work_ids)
-        )
+            mw.works_id.in_(work_ids),
+            LicensePool.work_id.in_(work_ids),
```
aslagle: I'm curious about this part - does the materialized view have license pools that don't match the works somehow?

courte: This allows the index to be used to limit the LicensePools before the `only_deliverable_works` criteria are applied. Otherwise it attempts to filter with that criteria against the whole licensepools table without using the index.

aslagle: :+1:

PR #813, opened 2018-03-06
Manage pagination through the size of the current page rather than the total query size

leonardr:
This branch makes some index changes to the materialized view to improve performance of crawlable feeds. I've tested it on NYPL QA with the library-wide crawlable feed, the crawlable feed for the Overdrive collection, and the crawlable feed for NYPL's 'Staff Picks' list, which contains about 1500 books. In all cases I can now get the first page of a feed within a few seconds, which is hopefully fast enough for purposes of automated integration.

This branch also changes the way we determine whether or not a `Pagination` object represents the end of a list. Previously the act of `apply`ing the Pagination object to a query made it run a `count` query to count the total number of items in the result set. This was a slowdown in all cases and it didn't work at all for some of the crawlable feed queries, so I took it out altogether.

Now it's expected that the user of a `Pagination` object will explicitly tell the `Pagination` object either the total size of the result set that's being paginated, or the size of the current page. If the total size of the resultset is available, then "is there a next page?" works as before. Otherwise, if the total size of the current page is available, we say there's a next page if and only if the current page has items on it.

I changed `AcquisitionFeed.page` to set `this_page_size` after fetching a page of works. As far as I can tell, this is the only place `Pagination` objects are actually used, so this will eliminate the need to run a `count` query in all cases.

The performance downside of this is that sometimes we will incorrectly say there is a next page when there isn't. This can result in an extra request to the server. However the only client likely to get the last page of a feed is a client crawling a crawlable feed. On balance, clients will benefit from better performance on each page since we no longer run a `count` alongside each materialized view query.

### DIFF
--- a/files/materialized_view_for_lanes.sql
+++ b/files/materialized_view_for_lanes.sql
@@ -65,12 +65,14 @@ create index ix_mv_works_for_lanes_list_edition_id on mv_works_for_lanes (list_e
 create unique index mv_works_for_lanes_unique on mv_works_for_lanes (works_id, genre_id, list_id, list_edition_id, license_pool_id);
 
 -- Create an index on everything, sorted by descending availability time, so that sync feeds are fast.
-
+-- TODO: This index might not be necessary anymore.
 create index mv_works_for_lanes_by_availability on mv_works_for_lanes (availability_time DESC, sort_author, sort_title, works_id);
 
--- Create an index on everything, sorted by the maximum of entry first appearance, license pool availability time, and work update time, so that crawlable feeds are fast.
+-- Create an index on everything, sorted by 'last update time' (the thing used by crawlable feeds).
+create index mv_works_for_lanes_by_recently_updated on mv_works_for_lanes (GREATEST(availability_time, first_appearance, last_update_time) DESC, collection_id, works_id);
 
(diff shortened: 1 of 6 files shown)


files/materialized_view_for_lanes.sql
```diff
@@ -65,12 +65,14 @@ create index ix_mv_works_for_lanes_list_edition_id on mv_works_for_lanes (list_e
 create unique index mv_works_for_lanes_unique on mv_works_for_lanes (works_id, genre_id, list_id, list_edition_id, license_pool_id);
 
 -- Create an index on everything, sorted by descending availability time, so that sync feeds are fast.
-
+-- TODO: This index might not be necessary anymore.
 create index mv_works_for_lanes_by_availability on mv_works_for_lanes (availability_time DESC, sort_author, sort_title, works_id);
 
--- Create an index on everything, sorted by the maximum of entry first appearance, license pool availability time, and work update time, so that crawlable feeds are fast.
+-- Create an index on everything, sorted by 'last update time' (the thing used by crawlable feeds).
+create index mv_works_for_lanes_by_recently_updated on mv_works_for_lanes (GREATEST(availability_time, first_appearance, last_update_time) DESC, collection_id, works_id);
```
leonardr: Putting `collection_id` in here makes a crawlable feed for a single collection fast in a way that a standalone index on `collection_id` doesn't.

PR #814, opened 2018-03-07
Implement an algorithm for determining whether two delivery mechanisms are compatible

leonardr:
This branch encapsulates and improves an algorithm for determining whether two DeliveryMechanisms or LicensePoolDeliveryMechanisms are mutually compatible (such that fulfilling one doesn't restrict your ability to fulfill the other) or mutually exclusive.

The motivating use case is https://github.com/NYPL-Simplified/circulation/issues/876

### DIFF
--- a/model.py
+++ b/model.py
@@ -5504,10 +5504,11 @@ class LicensePoolDeliveryMechanism(Base):
         # In practice, this means that either the two
         # DeliveryMechanisms are the same or that one of them is a
         # streaming mechanism.
+        open_access_rules = self.is_open_access and other.is_open_access
         return (
             other.delivery_mechanism
-            and other.delivery_mechanism.compatible_with(
-                self.delivery_mechanism, self.is_open_access
+            and self.delivery_mechanism.compatible_with(
+                other.delivery_mechanism, open_access_rules
             )
         )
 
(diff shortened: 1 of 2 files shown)


aslagle:
I haven't looked at this code yet, but I have a comment on the use case. If the book is open access, you should never be locked in to a delivery mechanism.

aslagle:
If a delivery mechanism has no DRM, is there any reason we can't allow the patron to get multiple formats?

leonardr:
If an Overdrive book is available as DRM-free PDF and DRM-free EPUB, and you choose PDF, you're locked out of EPUB. That's a 'feature' imposed by Overdrive above and beyond the DRM system. So we can't say that all DRM-free mechanisms are compatible.

Open-access books are different in that they are fulfilled by redirecting you to a certain external URL for direct download. For those I think it's okay to assume that all DRM-free mechanisms are compatible.

leonardr:
It turned out we've had two TestDeliveryMechanism classes for a while, so some of the tests weren't running. I merged them.

model.py
```diff
@@ -5471,6 +5471,47 @@ def is_open_access(self):
...
+            # different places.
+            return True
+
+        # If the DeliveryMechanisms themselves are compatible, then the
+        # LicensePoolDeliveryMechanisms are compatible.
+        #
+        # In practice, this means that either the two
+        # DeliveryMechanisms are the same or that one of them is a
+        # streaming mechanism.
+        open_access_rules = self.is_open_access and other.is_open_access
```
aslagle: I feel like this could be `or`? At least if the locked in mechanism is not open access, other mechanisms that are open access should still be available. Maybe not the opposite though.

leonardr: It could be, but there are so many places where it depends on the vendor in a counterintuitive way that I think a vendor-independent algorithm like this one should be more conservative.

PR #815, opened 2018-03-08
Address problems seen when configuring and using real Odilo integrations.

leonardr:
This addresses the problems I mentioned in https://github.com/NYPL-Simplified/circulation/issues/817#issuecomment-371573178, so far as they can be addressed by changing the API client.


odilo.py
```diff
@@ -147,18 +147,38 @@ def refresh_creds(self, credential):
         response = self.token_post(
             self.TOKEN_ENDPOINT,
             dict(grant_type="client_credentials"),
-            allowed_response_codes=[200]
+            allowed_response_codes=[200, 400]
```
leonardr: We have special handling code for a 400 response which wasn't being run because token_post filtered it out.

odilo.py
```diff
@@ -147,18 +147,37 @@ def refresh_creds(self, credential):
...
+            self.library_api_base_url, response.content
+        )
+        generic_exception = BadResponseException(
+            self.TOKEN_ENDPOINT, generic_error
+        )
+
+        try:
+            data = response.json()
+        except ValueError:
+            raise generic_exception
```
leonardr: This handles a case I saw with a real user, who put in the URL to a library's Odilo OPAC instead of the library's Odilo API URL.

odilo.py
```diff
@@ -147,18 +147,37 @@ def refresh_creds(self, credential):
...
         elif response.status_code == 400:
             if data and 'errors' in data and len(data['errors']) > 0:
                 error = data['errors'][0]
-                message = ('description' in error and error['description']) or ''
-                raise BadResponseException(message)
+                if 'description' in error:
+                    message = error['description']
+                else:
+                    message = generic_error
+                raise BadResponseException(self.TOKEN_ENDPOINT, message)
```
leonardr: This is just to make sure the generic error is used instead of the empty string if Odilo (or whoever) sends back a 400 response code and JSON but no specific error message.

odilo.py
```diff
@@ -454,7 +476,7 @@ def record_info_to_metadata(cls, book, availability):
             except ValueError as e:
                 cls.log.warn('Cannot parse last update date from: ' + last_update + ', message: ' + e.message)
 
-        language = book.get('language')
+        language = book.get('language', 'spa')
```
leonardr: Nearly all Odilo books are in Spanish, so this is the Odilo equivalent of assuming an Overdrive book is in English -- not guaranteed to be right but a good guess.

odilo.py
```diff
@@ -147,18 +147,37 @@ def refresh_creds(self, credential):
...
             if data and 'errors' in data and len(data['errors']) > 0:
                 error = data['errors'][0]
-                message = ('description' in error and error['description']) or ''
-                raise BadResponseException(message)
+                if 'description' in error:
+                    message = error['description']
+                else:
+                    message = generic_error
+                raise BadResponseException(self.TOKEN_ENDPOINT, message)
+        raise generic_exception
```
courte: Is this a catchall for unknown errors? In this case, the message would still be about the provided base URL. Is that the behavior you want?

leonardr: Yeah, it's a catchall, but the most likely reason the problem (whatever it is) is happening is that the base URL is wrong.

PR #816, opened 2018-03-09
Allow loans and holds to be for an integration client instead of a patron

aslagle:
This branch supports a circulation manager sharing its ODL collection with libraries from its consortium that are hosted on other circ managers. I also fixed a couple bugs.


opds_import.py
```diff
@@ -1639,7 +1639,7 @@ def follow_one_link(self, url, do_get=None):
         """
         self.log.info("Following next link: %s", url)
         get = do_get or self._get
-        status_code, content_type, feed = get(url, None)
+        status_code, content_type, feed = get(url, {})
```
aslagle: This was a bug - `_update_headers` needs a dict.

util/http.py
```diff
@@ -315,9 +315,10 @@ def process_debuggable_response(cls, response, allowed_response_codes=None):
         """
 
         allowed_response_codes = allowed_response_codes or ['2xx', '3xx']
+        allowed_response_codes = map(str, allowed_response_codes)
         code = response.status_code
         series = cls.series(code)
-        if code in allowed_response_codes or series in allowed_response_codes:
+        if str(code) in allowed_response_codes or series in allowed_response_codes:
```
aslagle: This was another bug.

migration/20180310-add-loan-hold-integration-client.sql
```diff
@@ -0,0 +1,33 @@
+DO $$
```
leonardr: Please bump the date on this migration script; I landed two scripts yesterday and don't want them to run out of order.

model.py
```diff
@@ -810,11 +810,23 @@ def work(self):
...
 
+    @property
+    def library(self):
+        """Try to find the corresponding library for this Loan/Hold."""
+        if self.patron:
+            return self.patron.library
+        if self.integration_client:
+            # If this is a loan for a library outside this circulation manager,
+            # choose an arbitrary library from the collection.
+            return self.license_pool.collection.libraries[0]
```
leonardr: Is it still necessary to choose a library or is it now acceptable to return `None` when there is no patron?

PR #817, opened 2018-03-09
Dramatically speed up CustomListEntry.entries_for_work

leonardr:
Find matching entries in a CustomList using a single database query rather than pulling the entire list membership from the database and iterating over it to see which entries match.

Since `add_entry` and `remove_entry` both call `entries_for_work`, this dramatically speeds up those methods when a list is large. It doesn't even have to be that large -- I was having serious problems with the NYPL staff picks list, which only has 1400 entries.

### DIFF
--- a/model.py
+++ b/model.py
@@ -9511,11 +9511,28 @@ class CustomList(Base):
             edition = work_or_edition
             work = edition.work
 
-        equivalents = edition.equivalent_editions().all()
+        equivalent_ids = [x.id for x in edition.equivalent_editions()]
 
-        for entry in self.entries:
-            if (work and entry.work == work) or entry.edition in equivalents:
-                yield entry
+        clauses = []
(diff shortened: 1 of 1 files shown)


model.py
```diff
@@ -9511,11 +9511,28 @@ def entries_for_work(self, work_or_edition):
...
-                yield entry
+        clauses = []
+        if equivalent_ids:
+            clauses.append(CustomListEntry.edition_id.in_(equivalent_ids))
+        if work:
+            clauses.append(CustomListEntry.work==work)
+        if len(clauses) == 0:
+            # This shouldn't happen, but if it does, there can be
+            # no matching results.
+            return
```
courte: It doesn't look like the methods that depend on this one expect `None` as a response. Could you update them or return an empty query instead?

PR #819, opened 2018-03-13
Work.from_identifiers can filter against an ID other than Identifier.id

leonardr:
This branch is the core fix to https://github.com/NYPL-Simplified/circulation/issues/884. It makes it possible for Work.from_identifiers to filter against a field other than Identifier.id, such as the materialized view's `identifier_id`.

### DIFF
--- a/migration/20180313-add-identifier-id-to-materialized-view.sql
+++ b/migration/20180313-add-identifier-id-to-materialized-view.sql
@@ -1,0 +1,9 @@
+DO $$
+    BEGIN
+        BEGIN
+	    create index mv_works_for_lanes_identifier_id on mv_works_for_lanes (identifier_id);
+        EXCEPTION
+            WHEN duplicate_table THEN RAISE NOTICE 'Warning: mv_works_for_lanes_identifier_id already exists.';
+        END;
+    END
+$$;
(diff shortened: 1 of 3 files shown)


migration/20180313-add-identifier-id-to-materialized-view.sql
```diff
@@ -0,0 +1,9 @@
+DO $$
+    BEGIN
+        BEGIN
+	    create index mv_works_for_lanes_identifier_id on mv_works_for_lanes (identifier_id);
+        EXCEPTION
+            WHEN duplicate_table THEN RAISE NOTICE 'Warning: mv_works_for_lanes_identifier_id already exists.';
+        END;
+    END
+$$;
```
leonardr: In my tests this index was not totally necessary, in that the query was acceptably fast without it, but I kept it in because the performance improvement was significant.

model.py
```diff
@@ -9511,11 +9511,28 @@ def entries_for_work(self, work_or_edition):
             edition = work_or_edition
             work = edition.work
 
-        equivalents = edition.equivalent_editions().all()
+        equivalent_ids = [x.id for x in edition.equivalent_editions()]
```
courte: It's weird that these changes showed up again.

leonardr: Yeah, this was in a branch based on my previous branch, but I assumed the commits would disappear from this PR once I merged that branch.

PR #820, opened 2018-03-13
Miscellaneous small fixes

leonardr:
This branch fixes some obscure crashes I found while doing other work, mainly around the S3 integration and OPDS imports.


axis.py
```diff
@@ -364,6 +364,7 @@ class BibliographicParser(Axis360Parser):
     DELIVERY_DATA_FOR_AXIS_FORMAT = {
         "Blio" : None,
         "Acoustik" : None,
+        "AxisNow": None,
```
leonardr: Axis 360 added a new format we can't use. Putting it in here makes our Axis API client not worry about it.

axis.py
```diff
@@ -582,7 +582,7 @@ def extract_bibliographic(self, element, ns):
             informal_name = format_tag.text
             seen_formats.append(informal_name)
             if informal_name not in self.DELIVERY_DATA_FOR_AXIS_FORMAT:
-                self.log.error("Unrecognized Axis format name for %s: %s" % (
+                self.log.warn("Unrecognized Axis format name for %s: %s" % (
```
leonardr: This isn't really an error, just a sign that you need to add the new format to DELIVERY_DATA_FOR_AXIS_FORMAT.

classifier/__init__.py
```diff
@@ -856,6 +856,8 @@ def __new__(cls, value):
         if isinstance(value, Lowercased):
             # Nothing to do.
             return value
+        if not isinstance(value, basestring):
+            value = unicode(value)
```
leonardr: This fixes a crash reported by Robert Williams -- I'm not sure how this happened, but probably a Dewey Decimal number was being passed into this constructor.

model.py
```diff
@@ -9663,7 +9664,10 @@ def update(self, _db, equivalent_entries=None):
                 self.annotation = annotations[0]
 
         # Reset the entry's edition to be the Work's presentation edition.
-        best_edition = work.presentation_edition
+        if work:
+            best_edition = work.presentation_edition
+        else:
+            best_edition = None
```
leonardr: This happened in real usage with NYPL's staff picks list (we ended up with a Work that had no presentation edition, possibly due to an interaction with the best-seller list).

model.py
```diff
@@ -10321,7 +10325,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     CONTENT_SERVER = u'Content Server'
 
     # Integrations with STORAGE_GOAL
-    S3 = u'S3'
+    S3 = u'Amazon S3'
```
leonardr: It's simpler if this is consistent with the NAME used by S3Uploader. There are no existing S3 integrations, so changing this shouldn't affect any real sites.

opds_import.py
```diff
@@ -1508,7 +1508,7 @@ def _get(self, url, headers):
         return response.status_code, response.headers, response.content
 
     def _update_headers(self, headers):
-        dict(headers or {})
+        headers = dict(headers or {})
```
leonardr: This bug led to preexisting headers not being preserved.

courte (approved):

Issue #818, opened 2018-03-12
Use a different dependency for S3 uploads


courte:
TinyS3 now raises 403 errors on upload to our S3 buckets. Use `boto3` or a different dependency instead.

PR #822, opened 2018-03-13
Use boto3 instead of tinys3

leonardr:
The tinys3 project has been abandoned, so we're switching to [boto3](http://boto3.readthedocs.io). This branch changes `S3Uploader` to use `boto3.client` instead of `tinys3.Pool`.

This should fix https://github.com/NYPL-Simplified/server_core/issues/818.

The main downside of boto3 is that there's no built-in way to upload multiple items in parallel. The main upside is that it actually works with our S3 configuration.


s3.py
```diff
@@ -14,6 +18,7 @@
     ConnectionError,
     HTTPError,
 )
+from util.http import RemoteIntegrationException
```
courte: This isn't being used.

Issue #823, opened 2018-03-15
Lanes that involve customlists + something else are very slow to generate


leonardr:
NYPL production is having serious performance problems with "Best Sellers" and "Staff Picks" lanes -- not the lanes based solely on those lists, but the lanes that combine list membership with some other criteria like 'Fiction'. The problem is that although we can filter on list membership when querying the materialized view, we don't -- instead, we join against CustomListEntry, filter there, and then laboriously match the materialized view (which was filtered against "Fiction") against CustomListEntry.

I've got a hotfix on the site that makes those lanes work by filtering the materialized view directly, but I fixed it in a way that breaks an obscure feature which allows the creation of a lane based on the intersection of multiple lists. I have an idea of how to fix the problem without breaking that feature but it's going to require some experimentation.

PR #824, opened 2018-03-15
List based lane performance improvements

leonardr:
This branch addresses https://github.com/NYPL-Simplified/server_core/issues/823 by improving the performance of lanes whose membership is derived from custom lists.

The basic technique is the same as other times we did this -- put a restriction on the materialized view as well as putting it on the joined table -- but it's a little more complicated because the joined table (CustomListEntry, in this case) may be joined against the materialized view multiple times, and we can only put a restriction on the materialized view once.

The solution is to apply the first restriction in both places (materialized view and CustomListEntry), and subsequent restrictions only to the aliased CustomListEntry joins.

### DIFF
--- a/files/materialized_view_for_lanes.sql
+++ b/files/materialized_view_for_lanes.sql
@@ -72,7 +72,7 @@ create index mv_works_for_lanes_by_availability on mv_works_for_lanes (availabil
 create index mv_works_for_lanes_by_recently_updated on mv_works_for_lanes (GREATEST(availability_time, first_appearance, last_update_time) DESC, collection_id, works_id);
 
 -- This index quickly cuts down the number of rows considered when generating a crawlable feed for a custom list.
-create index mv_works_for_lanes_list_and_collection_id on mv_works_for_lanes (list_id, collection_id);
+create index mv_works_for_lanes_list_id_collection_id_language_medium on mv_works_for_lanes (list_id, collection_id, language, medium);
 
 -- Create indexes that are helpful in running the query to find featured works.
 
--- a/files/materialized_view_for_lanes.sql
+++ b/files/materialized_view_for_lanes.sql
@@ -71,7 +71,7 @@ create index mv_works_for_lanes_by_availability on mv_works_for_lanes (availabil
 -- Create an index on everything, sorted by 'last update time' (the thing used by crawlable feeds).
 create index mv_works_for_lanes_by_recently_updated on mv_works_for_lanes (GREATEST(availability_time, first_appearance, last_update_time) DESC, collection_id, works_id);
 
--- This index quickly cuts down the number of rows considered when generating a crawlable feed for a custom list.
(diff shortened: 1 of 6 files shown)


files/materialized_view_for_lanes.sql
```diff
@@ -71,8 +71,8 @@ create index mv_works_for_lanes_by_availability on mv_works_for_lanes (availabil
 -- Create an index on everything, sorted by 'last update time' (the thing used by crawlable feeds).
 create index mv_works_for_lanes_by_recently_updated on mv_works_for_lanes (GREATEST(availability_time, first_appearance, last_update_time) DESC, collection_id, works_id);
 
--- This index quickly cuts down the number of rows considered when generating a crawlable feed for a custom list.
-create index mv_works_for_lanes_list_and_collection_id on mv_works_for_lanes (list_id, collection_id);
+-- This index quickly cuts down the number of rows considered when generating feeds for a custom list or the intersection of multiple custom lists.
+create index mv_works_for_lanes_list_id_collection_id_language_medium on mv_works_for_lanes (list_id, collection_id, language, medium);
```
leonardr: Adding language and medium to this index isn't totally necessary but it gives a nice additional performance improvement for most lanes, because the database doesn't have to revisit the items in the CustomList and filter out the ones with the wrong language or medium.

lane.py
```diff
@@ -1878,37 +1879,64 @@ def customlist_filter_clauses(
...
         if must_be_featured:
             clauses.append(a_entry.featured==True)
         if self.list_seen_in_previous_days:
             cutoff = datetime.datetime.utcnow() - datetime.timedelta(
                 self.list_seen_in_previous_days
             )
             clauses.append(a_entry.most_recent_appearance >=cutoff)
-
-        if must_be_featured:
-            clauses.append(a_entry.featured==True)
```
leonardr: This is duplicate code; you can see the other copy of it on line 1902 above.

migration/20180307-replace-recently-updated-index.sql
```diff
@@ -7,11 +7,5 @@ DO $$
             WHEN OTHERS THEN RAISE NOTICE 'mv_works_for_lanes_by_recently_updated did not previously exist.';
         END;
 	create index mv_works_for_lanes_by_recently_updated on mv_works_for_lanes (GREATEST(availability_time, first_appearance, last_update_time) DESC, collection_id, works_id);
-
-        BEGIN
-	    create index mv_works_for_lanes_list_and_collection_id on mv_works_for_lanes (list_id, collection_id);
-        EXCEPTION
-            WHEN duplicate_table THEN RAISE NOTICE 'Warning: mv_works_for_lanes_list_and_collection_id already exists.';
-        END;
```
leonardr: This index is created differently now, and there's no need to create it when it will just be deleted in a future migration script.

PR #825, opened 2018-03-16
Allow shortened S3 URLs

leonardr:
This branch introduces a new configuration for the S3 integration which controls the presentation of the URLs of items mirrored to S3. When a file is uploaded to S3 we know that it is accessible at `http://s3.amazonaws.com/bucket/file`, and that's what's been put in so far, but this isn't always the best URL to show to people, since it's possible to set up DNS so that `https://bucket/` points to the S3 bucket.

This branch introduces a configuration option that lets you choose which template to use when determining the final `mirror_url` of a resource that was mirrored to S3.

* `https://s3.amazonaws.com/bucket/file` (default)
* `https://bucket/file`
* `http://bucket/file`

This branch also changes the default S3 URL to use HTTPS instead of HTTP.

Switching to boto3 has made me aware that the S3 code needs some work, but I think the eventual system will look more like this -- based on building URLs from known pieces -- and less like parsing pieces out of URLs.

### DIFF
--- a/s3.py
+++ b/s3.py
@@ -29,15 +29,15 @@ class S3Uploader(MirrorUploader):
     BOOK_COVERS_BUCKET_KEY = u'book_covers_bucket'
     OA_CONTENT_BUCKET_KEY = u'open_access_content_bucket'
 
-    URL_TRANSFORM_KEY = u'bucket_name_transform'
-    URL_TRANSFORM_HTTP = u'http'
-    URL_TRANSFORM_HTTPS = u'https'
-    URL_TRANSFORM_IDENTITY = u'identity'
+    URL_TEMPLATE_KEY = u'bucket_name_transform'
+    URL_TEMPLATE_HTTP = u'http'
+    URL_TEMPLATE_HTTPS = u'https'
+    URL_TEMPLATE_DEFAULT = u'identity'
 
-    URL_TEMPLATES_BY_TRANSFORM = {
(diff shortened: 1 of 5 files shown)


s3.py
```diff
@@ -23,17 +23,48 @@ class S3Uploader(MirrorUploader):
...
+        URL_TEMPLATE_DEFAULT: S3_BASE + u'%(bucket)s/%(key)s',
+    }
 
     SETTINGS = [
         { "key": ExternalIntegration.USERNAME, "label": _("Access Key") },
         { "key": ExternalIntegration.PASSWORD, "label": _("Secret Key") },
-        { "key": BOOK_COVERS_BUCKET_KEY, "label": _("Book Covers Bucket"), "optional": True },
-        { "key": OA_CONTENT_BUCKET_KEY, "label": _("Open Access Content Bucket"), "optional": True },
+        { "key": BOOK_COVERS_BUCKET_KEY, "label": _("Book Covers Bucket"), "optional": True,
+          "description" : _("All book cover images encountered will be mirrored to this S3 bucket. Large images will be scaled down, and the scaled-down copies will also be uploaded to this bucket. <p>The bucket must already exist&mdash;it will not be created automatically.</p>")
```
aslagle: I thought we were going to have something to turn this on or off per collection - is that still true?

leonardr: There can still only one S3 integration, but you can choose whether or not to associate a given collection with it.

PR #828, opened 2018-03-21
Create a general "database table" reaper framework, plus reapers for CachedFeeds and Credentials

leonardr:
This branch introduces `ReaperMonitor`, an abstract class that deletes rows from a database table if a timestamp field in that row is less than a certain date. I introduce specific ReaperMonitors for removing CachedFeeds that are more than a month old, and Credentials that expired more than a day ago. Currently rows are just piling up in these two tables with no way of removing them.

Some code is refactored from `RunCollectionMonitorScript` into a new class, `RunMultipleMonitorsScript`, for any script that needs to run one monitor after another. The new class `RunReaperMonitorsScript` runs all monitors registered in `ReaperMonitor.REGISTRY`.

This branch adds some indexes on timestamp fields that are necessary for the reapers to run in a reasonable time.

Loans and holds already have reapers, in circulation. I want to bring them into this more general system, but it needs to wait for another branch, because the logic of reaping loans and holds is more complex than "delete if this field is less than this date". This branch adds some indexes on timestamp fields that will be useful when I do make this change.

### DIFF
--- a/migration/20180321-add-timestamp-indexes.sql
+++ b/migration/20180321-add-timestamp-indexes.sql
@@ -1,0 +1,22 @@
+ DO $$
+    BEGIN
+	BEGIN
+	    create index ix_loans_start on loans (start);
+	EXCEPTION
+            WHEN duplicate_table THEN RAISE NOTICE 'Warning: is_loans_start already exists.';
+        END;
+
+	BEGIN
+	    create index ix_loans_end on loans ("end");
+	EXCEPTION
+            WHEN duplicate_table THEN RAISE NOTICE 'Warning: ix_loans_end already exists.';
+        END;
+
+	BEGIN
+	    create index ix_cachedfeeds_timestamp on cachedfeeds (timestamp);
+	EXCEPTION
+            WHEN duplicate_table THEN RAISE NOTICE 'Warning: ix_cachedfeed_timestamp already exists.';
+        END;
+    END;
+END
+$$;
--- a/migration/20180321-add-timestamp-indexes.sql
+++ b/migration/20180321-add-timestamp-indexes.sql
@@ -15,7 +15,7 @@
 	BEGIN
 	    create index ix_cachedfeeds_timestamp on cachedfeeds (timestamp);
 	EXCEPTION
-            WHEN duplicate_table THEN RAISE NOTICE 'Warning: ix_cachedfeed_timestamp already exists.';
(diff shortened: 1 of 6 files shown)


monitor.py
```diff
@@ -87,9 +89,12 @@ class Monitor(object):
     def __init__(self, _db, collection=None):
         self._db = _db
         cls = self.__class__
-        if not cls.SERVICE_NAME:
+        if self.SERVICE_NAME:
+            self.service_name = self.SERVICE_NAME
+        elif cls.SERVICE_NAME:
+            self.service_name = cls.SERVICE_NAME
```
leonardr: This allows a specific instance of a Monitor to define SERVICE_NAME rather than it needing to be defined ahead of time on the class level.

migration/20180321-add-timestamp-indexes.sql
```diff
@@ -0,0 +1,22 @@
+ DO $$
+    BEGIN
+	BEGIN
+	    create index ix_loans_start on loans (start);
+	EXCEPTION
+            WHEN duplicate_table THEN RAISE NOTICE 'Warning: is_loans_start already exists.';
```
aslagle: There is a typo in the warning message

monitor.py
aslagle: I thought `self.SERVICE_NAME` would already be `cls.SERVICE_NAME` if it wasn't defined on the instance.

leonardr: I'm talking about the opposite situation -- cls.SERVICE_NAME is not defined but self.SERVICE_NAME is.

aslagle: Right, but I don't think you need the `elif` here - if `cls.SERVICE_NAME` is defined it becomes `self.SERVICE_NAME` automatically.

monitor.py
```diff
@@ -692,10 +692,11 @@ def where_clause(self):
         return self.timestamp_field < self.cutoff
 
     def run_once(self, *args, **kwargs):
-        rows_deleted = self._db.query(self.MODEL_CLASS).filter(
-            self.where_clause).delete(synchronize_session=False)
+        row_deleted = self.query().delete(synchronize_session=False)
         self.log.info("Deleted %d row(s)", rows_deleted)
 
+    def query(self):
+        returnself._db.query(self.MODEL_CLASS).filter(self.where_clause)
```
aslagle: you missed a space

PR #832, opened 2018-03-27
Different language means different work

leonardr:
My old branch on this topic, https://github.com/NYPL-Simplified/circulation/pull/906, was an improvement but didn't fix the problem I was having. It fixed the problem when books with incompatible languages are both associated with a Work that's just been created (and thus does not yet have a presentation edition). It didn't fix the problem when a book with language A is associated with a Work whose presentation edition _is_ set and is in language B. I originally had code to fix this, but took it out because it was hard to test and didn't seem to make a difference. But it does make a difference.

This branch adds the code I should have added last time. It refactors the complex code that checks which LicensePools should be grouped into the same Work, and which Works are eligible for being 'the' work for that branch, so it can be tested separately.

### DIFF
--- a/model.py
+++ b/model.py
@@ -4071,7 +4071,9 @@ class Work(Base):
         # Work.
         licensepools_for_work = Counter()
         for lp in qu:
-            if lp.work and not licensepools_for_work[lp.work]:
+            if (lp.work
+                and lp.work.language in (None, language)
+                and not licensepools_for_work[lp.work]):
                 licensepools_for_work[lp.work] = len(lp.work.license_pools)
 
         work = None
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -4035,6 +4035,40 @@ def for_unchecked_subjects(cls, _db):
...
+                Edition.permanent_work_id==pwid
+            ).filter(
+                Edition.medium==medium
+            ).filter(
+                Edition.language==language
+            )
+        pools = set(qu.all())
+        for lp in pools:
+            if (lp.work
+                and lp.work.language in (None, language)
```
aslagle: Why don't you also check the medium here?

PR #833, opened 2018-03-27
Audiobook admin search

EdwinGuzman:
Adding media and languages parameters to the search method on Lane and opds feed.


lane.py
```diff
@@ -1047,6 +1047,14 @@ def search(self, _db, query, search_client, pagination=None):
         # Get the search results from Elasticsearch.
         results = None
 
+        if not media:
+            media = self.media
+        else:
+            media = [media]
```
leonardr: It's not an issue now, but it's better to check whether this is already a list before turning it into a list. Someone could pass a list into `search`.

```
if isinstance(media, basestring):
    media = [media]
```

tests/test_lane.py
```diff
@@ -1257,13 +1257,24 @@ def test_search(self):
...
-        
+
         # The single search result was converted to a MaterializedWorkWithGenre.
         [result] = results
         from model import MaterializedWorkWithGenre as mwg
         assert isinstance(result, mwg)
         eq_(work.id, result.works_id)
 
+        # Test that language and media are passed in
+        languages = ["fr"]
```
leonardr: It doesn't matter for purposes of this test, but we use the 3-letter language codes, so in a real scenario this would be 'fre'.

lane.py
```diff
@@ -1047,6 +1047,14 @@ def search(self, _db, query, search_client, pagination=None):
...
+
+        if not languages:
+            languages = self.languages
```
leonardr: Lane languages should take precedence over user-specified languages, not vice versa. If your browser preferences are set to English but you go into a Spanish lane and run a search, you should search the Spanish collection.

lane.py
```diff
@@ -1049,6 +1049,8 @@ def search(self, _db, query, search_client, media=None, pagination=None, languag
 
         if not media:
             media = self.media
+        elif media is Edition.ALL_MEDIUM:
+            media = Edition.medium_to_additional_type.keys()
```
leonardr: This will work but it would be simpler (and possibly more efficient) to set `media=None` here. `ExternalSearchIndex.make_filter` will do the right thing.

EdwinGuzman: Updated. I tried that before but didn't work. I may have tried `None` elsewhere instead.

leonardr:
OK, land it!

PR #840, opened 2018-04-03
Vacuum the database before _and_ after refreshing the materialized view.

leonardr:
This branch changes our strategy for refreshing the materialized view in a way that will hopefully eliminate problems I'm currently having on NYPL production. Our materialized view started out with great refresh performance which has deteriorated over time, indicating that improper vacuuming may be the problem. According to our DBA, this is the right way to refresh materialized views:

1. Vacuum the entire database
2. Refresh the materialized views
3. Vacuum just the materialized views

This branch also makes the refresh process use the normal logging method rather than printing to stdout. This will let me get better visibility into the time it takes to refresh.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1870,34 +1870,43 @@ class RefreshMaterializedViewsScript(Script):
             concurrently = ''
         else:
             concurrently = 'CONCURRENTLY'
-        # Initialize database
+
+        # Initialize the default database session. We can't use this
+        # session to run the VACUUM commands, because it wraps
+        # everything in a big transaction, and VACUUM can't be
+        # executed within a transaction block. But we will use it at
+        # the end, to recalculate the sizes of lanes.
         db = self._db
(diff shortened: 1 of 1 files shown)


scripts.py
```diff
@@ -1870,34 +1870,43 @@ def do_run(self):
...
+        # everything in a big transaction, and VACUUM can't be
+        # executed within a transaction block. But we will use it at
+        # the end, to recalculate the sizes of lanes.
         db = self._db
+
+        url = Configuration.database_url()
+        engine = create_engine(url, isolation_level="AUTOCOMMIT")
+        engine.autocommit = True
+        a = time.time()
+        engine.execute("VACUUM (VERBOSE, ANALYZE)")
```
aslagle: Is this verbose for debugging purposes? Does making it verbose affect performance?

leonardr: It is for debugging purposes, and it doesn't affect performance -- it prints a few lines for every affected table and index.

PR #842, opened 2018-04-04
Initial EntryPoint setup

leonardr:
This branch introduces the concept of 'EntryPoint', an additional filter that can be placed atop a library's lane structure. There are two EntryPoints to start with, for ebooks and audiobooks. Administrators can control which EntryPoints are enabled for their library.

In a future branch, a special facet group will be used to allow navigation between EntryPoints.

This branch adds coverage of the previously untested method `CachedFeed.fetch`.


entrypoint.py
```diff
@@ -0,0 +1,118 @@
...
+            )
+        if value in cls.INTERNAL_NAMES:
+            raise ValueError(
+                "Duplicate entry point internal name: %s" % value
+            )
+        cls.INTERNAL_NAMES.append(value)
+        cls.ENTRY_POINTS.append(entrypoint_class)
+        if default_enabled:
+            cls.DEFAULT_ENABLED.append(entrypoint_class)
+        cls.DISPLAY_TITLES[entrypoint_class] = display_title
```
aslagle: Display titles should also be unique.

lane.py
```diff
@@ -1066,20 +1073,34 @@ def search(self, _db, query, search_client, media=None, pagination=None, languag
...
+            kwargs = dict(
+                media=media,
+                languages=default_languages,
+                fiction=self.fiction,
+                audiences=self.audiences,
+                target_age=target_age,
+                in_any_of_these_genres=self.genre_ids,
+            )
+            if entrypoint:
+                kwargs = entrypoint.modified_search_arguments(**kwargs)
```
aslagle: It's possible that the entry point could conflict with media or languages passed in or defined in a lane. Not sure if this could happen in practice - maybe if you had the "World Languages" entry point, and then you went to a language lane under it? If you still had the entry point then it would replace the language of the lane you were in and search all the world languages.

leonardr: I think this is an issue to be resolved inside the entry point. `modified_search_arguments` can examine the existing search arguments and leave them alone if a scenario like this would happen.

We might eventually need to move to a system where the entry point can modify or override the search query rather than the arguments that go into the search query.

entrypoint.py
```diff
@@ -0,0 +1,123 @@
...
+    class by calling EntryPoint.register.
+
+    The list of entry points shows up as a facet group in a library's
+    top-level grouped feed, and in search results. The SimplyE client
+    renders entry points as a set of tabs.
+    """
+
+    # The name of the per-library setting that controls which entry points are
+    # enabled.
+    ENABLED_SETTING = "enabled_entry points"
```
EdwinGuzman: Are you perhaps missing an underscore here?

leonardr: Yes, definitely.

PR #843, opened 2018-04-09
Entry point is a facet group

leonardr:
This is for https://github.com/NYPL-Simplified/server_core/issues/844.

This branch gives OPDS feeds the ability to add links to a different entry point into the same feed. These links can be present in grouped feeds, search results and paginated feeds. (To start with, we will only be actually using this with grouped feeds and search results.)

The entry points available to a feed are determined by the WorkList associated with that feed. Currently there is no actual way to associate entry points with a WorkList, but the circulation manager will take care of that.

"Entry point" is treated as a special kind of facet group and the currently selected entry point is the active facet of that group. An extra attribute on the facet group link signals to the client that this particular facet group is an entry point (e.g. so it can be rendered as a set of tabs instead of a drop-down).

It might be easier to review this branch by comparing it against master. The `tab-configuration` branch added a lot of 'entrypoint' arguments to methods which this branch either removes or replaces with 'facets' arguments. By themselves these changes aren't large but I found it confusing to check whether I needed to achieve test coverage of a given method by porting a test from `tab-configuration` or by adding a new test.

The underlying rationale for that change is to reduce the number of times we have to change the method signatures. The methods that took facets before don't change signature, and the methods that didn't take facets now take a fully general object for adding additional filters to the database or search query.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -111,7 +111,7 @@ class FacetsWithEntryPoint(FacetConstants):
         """
         return "&".join("=".join(x) for x in sorted(self.items()))
 
-    def apply(self, qu):
+    def apply(self, _db, qu):
         if self.entrypoint:
             qu = self.entrypoint.apply(qu)
         return qu
@@ -192,7 +192,7 @@ class Facets(FacetsWithEntryPoint):
                       availability or self.availability,
                       order or self.order,
                       enabled_facets=self.facets_enabled_at_init,
-                      entrypoint or self.entrypoint
+                      entrypoint=(entrypoint or self.entrypoint)
         )
 
     def items(self):
@@ -304,7 +304,7 @@ class Facets(FacetsWithEntryPoint):
         matches works that fit the given facets, and the query is
         ordered appropriately.
         """
-        qu = super(Facets, self).apply(qu)
+        qu = super(Facets, self).apply(_db, qu)
         from model import MaterializedWorkWithGenre as work_model
         if self.availability == self.AVAILABLE_NOW:
             availability_clause = or_(
@@ -387,7 +387,7 @@ class FeaturedFacets(FacetsWithEntryPoint):
     AcquisitionFeed.groups().
     """
 
-    def __init__(self, minimum_featured_quality=None, uses_customlists=None,
+    def __init__(self, minimum_featured_quality=None, uses_customlists=False,
                  entrypoint=None):
         """Set up an object that finds featured books in a given
         WorkList.
@@ -417,7 +417,7 @@ class FeaturedFacets(FacetsWithEntryPoint):
         this will work.
         """
         from model import MaterializedWorkWithGenre as work_model
-        qu = super(FeaturedFacets, self).apply(qu)
+        qu = super(FeaturedFacets, self).apply(_db, qu)
         quality = self.quality_tier_field()
         qu = qu.order_by(
             quality.desc(), work_model.random.desc(), work_model.works_id
@@ -754,7 +754,7 @@ class WorkList(object):
             key += ','.join(audiences)
         return key
 
-    def groups(self, _db, include_sublanes=True, facets):
+    def groups(self, _db, include_sublanes=True, facets=None):
         """Extract a list of samples from each child of this WorkList.  This
         can be used to create a grouped acquisition feed for the WorkList.
 
@@ -1295,6 +1295,7 @@ class WorkList(object):
         library = self.get_library(_db)
         target_size = library.featured_lane_size
 
+        facets = facets or FeaturedFacets()
         facets = facets.navigate(
             library.minimum_featured_quality,
             self.uses_customlists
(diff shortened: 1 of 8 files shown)


model.py
```diff
@@ -6592,17 +6592,13 @@ def fetch(cls, _db, lane, type, facets, pagination, annotator,
...
         if isinstance(max_age, int):
             max_age = datetime.timedelta(seconds=max_age)
+
+        unique_key = None
         if lane and isinstance(lane, Lane):
             lane_id = lane.id
-            if entrypoint:
-                unique_key = entrypoint.INTERNAL_NAME
-            else:
-                unique_key = None
```
leonardr: Since entry point is now a facet group, the information necessary to distinguish between the same feed in two entry points is now kept in the facet field and there's no need for this code.

lane.py
```diff
@@ -139,13 +192,16 @@ def __init__(self, library, collection, availability, order,
         self.order_ascending = order_ascending
         self.facets_enabled_at_init = enabled_facets
 
-    def navigate(self, collection=None, availability=None, order=None):
+    def navigate(self, collection=None, availability=None, order=None,
+                 entrypoint=None):
```
aslagle: This appears to provide a way to navigate between entry points which contradicts the class's docstring.

opds.py
```diff
@@ -582,6 +602,21 @@ def page(cls, _db, title, url, lane, annotator,
             pagination.this_page_size = len(works)
         feed = cls(_db, title, url, works, annotator)
 
+        if lane.entrypoints and (not pagination or pagination.offset == 0):
+            # On the first page of a paginated feed there may be
+            # multiple entry points into the same dataset.
```
aslagle: This could lead to a weird UI where tabs disappear when you go to the page 2, depending on how the client works. If all the clients do infinite scroll instead of pagination it would be ok.

leonardr: I left that link out because I'm not sure where a tab on page 2 would go. Would it go to the first or second page of the other tab? Neither seems great to me.
 
I guess if it's like other facet groups, it would go to the top of the other tab. I can take out the `pagination.offset == 0` stuff if you'd like that behavior.

aslagle: I do think that would be better.

opds.py
```diff
@@ -607,6 +642,75 @@ def page(cls, _db, title, url, lane, annotator,
...
+        args = dict(href=href, title=title)
+        args['rel'] = cls.FACET_REL
+        args['{%s}facetGroup' % AtomFeed.OPDS_NS] = facet_group_name
+        if is_active:
+            args['{%s}activeFacet' % AtomFeed.OPDS_NS] = "true"
+        return args
+
+    @classmethod
+    def add_entrypoint_links(cls, feed, url_generator, entrypoints,
+                             selected_entrypoint, group_name='Start here'):
```
aslagle: If the client has tabs, I'm guessing this group name won't be displayed, except maybe for screen readers. But I'm not sure how this sounds if it shows up in a regular OPDS client that knows about facets but not entry points. I don't have another suggestion right now though.

opds.py
```diff
@@ -628,24 +732,42 @@ def add_breadcrumb_links(self, feed, lane, annotator):
...
         results = lane.search(
-            _db, query, search_engine, media, pagination=pagination, languages=languages
+            _db, query, search_engine, media, pagination=pagination, languages=languages, facets=facets
         )
         opds_feed = AcquisitionFeed(_db, title, url, results, annotator=annotator)
         AcquisitionFeed.add_link_to_feed(feed=opds_feed.feed, rel='start', href=annotator.default_lane_url(), title=annotator.top_level_title())
 
+        # The first page of a feed of search results may link to
+        # alternate entry points into those results.
+        if lane.entrypoints and (not pagination or pagination.offset == 0):
```
aslagle: this has the same possible issue for a client with pagination.

opds.py
leonardr: "Start here" is pretty bad but it's ultimately up to the application to decide what kind of thing an entry point is.

aslagle: I'd like generic OPDS clients to be able to do something reasonable without knowing about entry points, so maybe it does have to be up to the library. If they selected 'Ebooks' and 'Audio' as their entry points they could call it 'Formats'.

tests/test_opds.py
```diff
@@ -1632,3 +1785,173 @@ def test_unfilfullable_work(self):
             "I know about this work but can offer no way of fulfilling it."
         )
         eq_(expect, entry)
+
+
+class TestEntrypointLinkInsertion(DatabaseTest):
+    """Verify that the three main types of OPDS feeds -- grouped,
+    paginated, and search results -- will all
```
aslagle: will all what?

opds.py
leonardr: Since the only EntryPoints we have right now are for formats, I'll make "Formats" the default here, rather than have a default nobody actually wants to use.

PR #845, opened 2018-04-11
Load entry point into faceting objects during `load_facets`

leonardr:
This branch generalizes the `load_facets` function (used by `load_facets_from_request`) so that:

1. The caller can choose which faceting class to instantiate, and pass extra arguments into its constructor.
2. The caller can pass in a WorkList. If that WorkList has entry points, and the user requested a specific entry point, the faceting object will be instantiated with the corresponding EntryPoint class.

In a future branch we should change the `Facets` class to something like `ListFeedFacets`. Right now it sounds like a generic superclass, which isn't the case. A fair amount of `load_facets` only makes sense if the facets class being instantated is `Facets` rather than a different faceting class; we could make a faceting class more responsible for pulling its own information out of request arguments.

### DIFF
--- a/entrypoint.py
+++ b/entrypoint.py
@@ -120,4 +120,4 @@ EntryPoint.register(EbooksEntryPoint, "Books", default_enabled=True)
 
 class AudiobooksEntryPoint(MediumEntryPoint):
     INTERNAL_NAME = "Audio"
-EntryPoint.register(AudiobooksEntryPoint, "Audiobooks")
+EntryPoint.register(AudiobooksEntryPoint, "Audio")
--- a/lane.py
+++ b/lane.py
@@ -613,9 +613,10 @@ class WorkList(object):
         # This WorkList contains every title available to this library
         # in one of the media supported by the default client.
         wl = WorkList()
+
         wl.initialize(
             library, display_name=library.name, children=top_level_lanes,
-            media=Edition.FULFILLABLE_MEDIA
+            media=Edition.FULFILLABLE_MEDIA, entrypoints=library.entrypoints
         )
         return wl
 
@@ -653,7 +654,7 @@ class WorkList(object):
         show up in relation to its siblings when it is the child of
         some other WorkList.
 
-        :param entrypoints: A list of EntryPoint objects representing
+        :param entrypoints: A list of EntryPoint classes representing
         different ways of slicing up this WorkList.
         """
         self.library_id = None
--- a/lane.py
+++ b/lane.py
@@ -389,7 +389,7 @@ class FeaturedFacets(FacetsWithEntryPoint):
     AcquisitionFeed.groups().
     """
 
-    def __init__(self, minimum_featured_quality=None, uses_customlists=False,
+    def __init__(self, minimum_featured_quality, uses_customlists=False,
                  entrypoint=None):
         """Set up an object that finds featured books in a given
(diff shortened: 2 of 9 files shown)


model.py
```diff
@@ -3012,7 +3012,7 @@ class Edition(Base):
 
     # These are the media types currently fulfillable by the default
     # client.
-    FULFILLABLE_MEDIA = [BOOK_MEDIUM]
+    FULFILLABLE_MEDIA = [BOOK_MEDIUM, AUDIO_MEDIUM]
```
leonardr: This is a deceptively large change -- this means audio content will start showing up. It won't show up in the existing lane structure, because the existing lanes are hard-coded to only contain ebooks. But it will show up on the front page if the library administrator turns on that entry point (in which case it will have its own entry point) or turns off every single entry point (in which case it will be mixed with ebook content).

I'm OK with leaving this out for now but I'm using it to test on a real system.

The larger problem this causes is that Overdrive audio content starts showing up in feeds even though it's not currently fulfillable.

opds.py
```diff
@@ -675,14 +675,22 @@ def add_entrypoint_links(cls, feed, url_generator, entrypoints,
...
+
         is_default = True
         for entrypoint in entrypoints:
             link = cls._entrypoint_link(
                 url_generator, entrypoint, selected_entrypoint, is_default,
                 group_name
             )
             if link:
-                cls.add_link_to_feed(feed, **link)
+                cls.add_link_to_feed(feed.feed, **link)
```
leonardr: This is a bug that wasn't caught by the test because the test created the wrong mock object.

entrypoint.py
```diff
@@ -120,4 +120,4 @@ class EbooksEntryPoint(MediumEntryPoint):
 
 class AudiobooksEntryPoint(MediumEntryPoint):
     INTERNAL_NAME = "Audio"
-EntryPoint.register(AudiobooksEntryPoint, "Audiobooks")
+EntryPoint.register(AudiobooksEntryPoint, "Audio")
```
leonardr: This change matches our UI design for entry point navigation, where horizontal space is at a premium.

lane.py
```diff
@@ -1298,7 +1307,7 @@ def _featured_works_with_lanes(self, _db, lanes, facets):
         library = self.get_library(_db)
         target_size = library.featured_lane_size
 
-        facets = facets or FeaturedFacets()
+        facets = facets or self.default_featured_facets(_db)
         facets = facets.navigate(
             library.minimum_featured_quality,
             self.uses_customlists
```
aslagle: I think you can remove this navigate now

model.py
aslagle: I'm not sure what the implications of this are. I'm okay with having it in if the default is for only the ebook entry point to be enabled, and an admin would have to explicitly change something for audiobooks to show up.

opds.py
```diff
@@ -1516,7 +1524,12 @@ def group_uri(self, work, license_pool, identifier):
             if additional_lanes:
                 self.lanes_by_work[work] = additional_lanes
         else:
-            lane_name = str(work.id)
+            if isinstance(work, Work):
+                work_id = work.id
+            else:
+                # MaterialivedViewWithGenre
+                work_id = work.work_id
```
aslagle: isn't it `works_id`?

PR #846, opened 2018-04-13
Give faceting classes total responsibility over their own instantiation

leonardr:
This branch refactors `load_facets_from_request` so that almost all of the code runs inside the appropriate faceting class's implementation of `from_request`. 

Code that was specific to the `Facets` class has been moved to that class's `from_request` implementation, and test coverage has been added where before the coverage was a little spotty.

Despite the branch name, this does not add an extra EntryPoint for search, but that should be easy to add after this refactoring

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -139,29 +139,6 @@ def load_pagination(size, offset):
             return INVALID_INPUT.detailed(_("Invalid offset: %(offset)s", offset=offset))
     return Pagination(offset, size)
 
-def load_entrypoint(entrypoint, worklist):
-    """Turn user input into an EntryPoint class from the EntryPoint registry.
-
-    :param worklist: A WorkList.
-
-    :return: An EntryPoint class. This will be the requested
-    EntryPoint if possible. If a nonexistent or unusable EntryPoint is
-    requested, the WorkList's default EntryPoint will be returned. If
-    the WorkList has no EntryPoints, or no WorkList is provided, None
-    will be returned.
-    """
-    if not worklist or not worklist.entrypoints:
-        # This WorkList has no EntryPoints. No EntryPoint should ever
-        # be returned from this method.
-        return None
-    default = worklist.entrypoints[0]
-    cls = EntryPoint.BY_INTERNAL_NAME.get(entrypoint)
-    if not cls:
-        return default
-    if cls not in worklist.entrypoints:
-        return default
-    return cls
-
 def returns_problem_detail(f):
     @wraps(f)
     def decorated(*args, **kwargs):
--- a/app_server.py
+++ b/app_server.py
@@ -101,39 +101,21 @@ def load_facets_from_request(
     The active request must have the `library` member set to a Library
     object.
 
-    :param facet_config: An object to use instead of the request Library
-    when deciding which facets are enabled.
-
-    :param lane: An optional WorkList to use when checking which EntryPoints
-    are aviailable.
-
-    :param base_class: A facet class, such as FacetsWithEntryPoint or one of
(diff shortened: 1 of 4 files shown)


app_server.py
```diff
@@ -143,51 +125,6 @@ def load_pagination_from_request(default_size=Pagination.DEFAULT_SIZE):
...
-                entrypoint=None, worklist=None, base_class=Facets,
-                base_class_constructor_kwargs=None):
-    """Turn user input into a faceting object."""
-    config = facet_config or library
-    order_facets = config.enabled_facets(Facets.ORDER_FACET_GROUP_NAME)
-    if order and not order in order_facets:
-        return INVALID_INPUT.detailed(
-            _("I don't know how to order a feed by '%(order)s'", order=order),
-            400
-        )
```
leonardr: All this stuff has been moved into `Facets.from_request`.

app_server.py
```diff
@@ -202,29 +139,6 @@ def load_pagination(size, offset):
             return INVALID_INPUT.detailed(_("Invalid offset: %(offset)s", offset=offset))
     return Pagination(offset, size)
 
-def load_entrypoint(entrypoint, worklist):
```
leonardr: This now happens as part of `FacetsWithEntryPoint.from_request` -- there's no need to call it explicitly.

tests/test_lane.py
```diff
@@ -74,6 +75,101 @@ def apply(self, qu):
...
+    def test_from_request(self):
+        # from_request just calls _from_request.
+        expect = object()
+        @classmethod
+        def mock(cls, facet_config, get_argument, worklist, **extra_kwargs):
+            return expect
+        old = FacetsWithEntryPoint._from_request
+        FacetsWithEntryPoint._from_request = mock
+        eq_(expect, FacetsWithEntryPoint.from_request(None, None, None, None))
+        FacetsWithEntryPoint._from_request = old
```
leonardr: This test breaks the next one because `FacetsWithEntryPoint._from_request` isn't restored exactly how it was before. So we end up invoking the `FacetsWithEntryPoint` constructor instead of the `ExtraFacets` constructor. If you have any insight into this problem I'd love to hear it.

leonardr: Without this test:

```
>>> ExtraFacets.from_request
<bound method type.from_request of <class 'core.tests.test_lane.ExtraFacets'>>
```

With the test:

```
ExtraFacets._from_request
<bound method type._from_request of <class 'lane.FacetsWithEntryPoint'>>
```

aslagle: Maybe you could define a subclass of FacetsWithEntryPoint and override `_from_request` instead of changing the original method.

lane.py
```diff
@@ -100,6 +103,69 @@ def __init__(self, entrypoint=None, **kwargs):
...
+        EntryPoint is requested, the WorkList's default EntryPoint
+        will be returned. If the WorkList has no EntryPoints, or no
+        WorkList is provided, None will be returned.
+        """
+        if not worklist or not worklist.entrypoints:
+            return None
+        default = worklist.entrypoints[0]
+        ep = EntryPoint.BY_INTERNAL_NAME.get(name)
+        if not ep or ep not in worklist.entrypoints:
+            return default
```
aslagle: are there any cases that should return a problem detail?

tests/test_lane.py
leonardr: Good idea, that's simpler too.

lane.py
leonardr: Not right now, I'm leaving that space for subclasses that do weird things with entry points.

PR #847, opened 2018-04-16
Extra entrypoint for search

leonardr:
The overall goal of this branch is to add an 'All' entry point that puts no extra restrictions on which part of a lane is shown. You can turn on this entry point and show it everywhere, but it's not on by default, and even if it's off, it will show up in search results.

The centerpiece of this branch is `FacetsWithEntryPoint.available_entrypoints`, which allows a faceting object to modify the list of EntryPoints. The OPDS generation code uses `available_entrypoints` everywhere it used to look at `worklist.entrypoints`.

To make sure all relevant URL arguments are propagated, the OPDS generation code now relies more heavily on `navigate` and less on explicitly calling the constructor with different arguments. The `navigate` implementations have been changed to be more careful about creating a new object of the same class and passing in all information that was passed in to the previous faceting object.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -537,17 +537,18 @@ class AcquisitionFeed(OPDSFeed):
 
         # A grouped feed may link to alternate entry points into
         # the data.
-        if lane.entrypoints:
+        entrypoints = facets.available_entrypoints(lane)
+        if entrypoints:
             def make_link(ep, is_default):
                 if is_default:
                     # No need to clutter up the URL with the default
                     # entry point.
                     ep = None
                 return annotator.groups_url(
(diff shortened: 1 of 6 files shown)


opds.py
```diff
@@ -471,6 +471,7 @@ def groups(cls, _db, title, url, lane, annotator,
         """
         cached = None
         use_cache = cache_type != cls.NO_CACHE
+        facets = facets or lane.default_featured_facets(_db)
```
leonardr: Now that we're using `navigate` instead of the constructor we need to always have a faceting object to start with.

Issue #848, opened 2018-04-17
Don't set Representation.mirror_url unless you actually mirrored the representation, or know who did


leonardr:
Currently `Identifier.add_link` calls `Resource.set_mirrored_elsewhere` to store a resource's original URL in mirror_url. I believe we do this so that we can use the original URL while waiting for a better one to come along, e.g. through the metadata wrangler.

We can keep using the original URL while waiting for a better one to come along, but setting `mirror_url` creates confusion about whether the URL has really been "mirrored" or not. `mirror_url` should only be set when we either mirror the URL ourselves, or we hear from a source like the metadata wrangler which we trust to mirror things.

The bits of opds.py (and any similar bits elsewhere) which take `mirror_url` and put it in an OPDS entry should be changed to use a new property of `Resource` which uses `mirror_url` if it's available and `url` otherwise.

PR #849, opened 2018-04-17
Add a script to mirror unmirrored items if possible

leonardr:
This branch adds a script that, after the fact,  goes through unmirrored Resources that were not mirrored on initial import, and mirrors them.

It's not perfect because it doesn't capture the entire context in which we heard about the item that needs mirroring. In particular, it doesn't always work with FeedBooks books, because we might need to unzip the books and replace the CSS files, but outside of a FeedbooksImporter we don't know if we do, in fact, have to do this. But we do check the rights status of the books we're mirroring to make sure we have the rights to mirror them.

Sometimes Feedbooks OPDS feeds show a rights status that are incorrect from a US perspective, but by the time the information gets into the database (which is where this script works), the information has been translated to the US copyright regime.

A larger underlying problem is https://github.com/NYPL-Simplified/server_core/issues/848, which makes many resources look "mirrored" when they actually haven't been.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -552,7 +552,7 @@ class MetaToModelUtility(object):
         if edition and edition.title:
             title = edition.title
         else:
-            title = self.title or None
+            title = getattr(self, 'title', None) or None
 
         if ((not identifier) or (link_obj.identifier and identifier != link_obj.identifier)):
             # insanity found
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -500,6 +500,8 @@ class MetaToModelUtility(object):
     Contains functionality common to both CirculationData and Metadata.
     """
 
+    log = logging.getLogger("Abstract metadata layer - mirror code")
+
     def mirror_link(self, model_object, data_source, link, link_obj, policy):
         """Retrieve a copy of the given link and make sure it gets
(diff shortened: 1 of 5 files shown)


metadata_layer.py
```diff
@@ -550,7 +552,7 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
         if edition and edition.title:
             title = edition.title
         else:
-            title = self.title or None
+            title = getattr(self, 'title', None) or None
```
leonardr: This is necessary now that we're calling `mirror_link` directly on MetaToModelUtility, which unlike `Metadata` doesn't have a `.title`. In theory this could have been a problem with `CirculationData` as well.

model.py
```diff
@@ -10368,6 +10410,7 @@ class ExternalIntegration(Base, HasFullTableCache):
         AXIS_360 : DataSource.AXIS_360,
         RB_DIGITAL : DataSource.RB_DIGITAL,
         ENKI : DataSource.ENKI,
+        FEEDBOOKS : DataSource.FEEDBOOKS,
```
leonardr: Without this, calling `Collection.data_source` on a Feedbooks collection will return None.

Issue #852, opened 2018-04-18
Make Lane.search work for a lane based on a custom list


leonardr:
It should be possible to run a search from a lane based on a custom list.

In conversation with @aslagle we decided the best way to do this is to store a work's list IDs in its search document.

When a Work is added or removed from a list, it should be flagged for reindexing.

PR #854, opened 2018-04-19
Mention a work's customlists in its search document

leonardr:
This branch implements https://github.com/NYPL-Simplified/server_core/issues/852 by making a Work's list memberships part of its search document. When a Lane is searched, the CustomLists associated with that lane are passed into the search engine, a work must be on one of those lists to match the search query.

This makes it possible to search within lanes based on custom lists.

When a list is added to or removed from a custom list, it is flagged for reindexing.

I would especially value a close look at my changes to `make_filter`.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -263,7 +263,7 @@ class ExternalSearchIndex(object):
         return base_works_index
 
     def query_works(self, library, query_string, media, languages, fiction, audiences,
-                    target_age, in_any_of_these_genres=[], fields=None, size=30, offset=0):
+                    target_age, in_any_of_these_genres=[], on_any_of_these_lists=[], fields=None, size=30, offset=0):
         if not self.works_alias:
             return []
 
@@ -280,7 +280,8 @@ class ExternalSearchIndex(object):
 
         filter = self.make_filter(
             collection_ids, media, languages, fiction, 
-            audiences, target_age, in_any_of_these_genres
+            audiences, target_age, in_any_of_these_genres,
+            on_any_of_these_lists
         )
         q = dict(
             filtered=dict(
@@ -532,7 +533,7 @@ class ExternalSearchIndex(object):
             }
         }
         
-    def make_filter(self, collection_ids, media, languages, fiction, audiences, target_age, genres):
+    def make_filter(self, collection_ids, media, languages, fiction, audiences, target_age, genres, customlist_ids):
         def _f(s):
             if not s:
                 return s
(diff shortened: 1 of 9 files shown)


model.py
```diff
@@ -9560,7 +9549,16 @@ def featured_works(self):
...
+                  featured=None, update_external_index=True):
+        """Add a work to a CustomList.
+
+        :param update_external_index: When a Work is added to a list,
+        its external index needs to be updated. The only reason not to
+        do this is when the current database session already contains
+        a new WorkCoverageRecord for this purpose (e.g. because the
+        Work was just created) and creating another one would violate
+        the workcoveragerecords table's unique constraint.
+        """
```
leonardr: Not really happy about this particular change but 'automatically add every new work to a custom list' seems like a magnet for these problems.

model.py
```diff
@@ -4977,25 +4977,46 @@ def to_search_documents(cls, works):
...
+            works_alias.name + '.' + works_alias.c.work_id.name
+        )
+
+        def query_to_json(query):
+            """Convert the results of a query to a JSON object."""
+            return select(
+                [func.row_to_json(literal_column(query.name))]
+            ).select_from(query)
+
+        def query_to_json_array(query):
```
leonardr: This helper method lets me remove a lot of redundant code.

external_search.py
```diff
@@ -562,6 +563,8 @@ def _f(s):
                 # no longer happen but we'll handle it.
                 genre_ids = [genre.id for genre in genres]
             clauses.append(dict(terms={"genres.term": genre_ids}))
+        if customlist_ids is not None:
+            clauses.append(dict(terms={"list_id": customlist_ids}))
```
aslagle: This looks good to me although I'm not 100% sure how it will handle an empty list.

lane.py
```diff
@@ -1921,6 +1930,38 @@ def _gather_genre_ids(self):
...
+            )
+            ids = [x[0] for x in _db.execute(query)]
+        else:
+            # Find the IDs of some specific CustomLists.
+            ids = [x.id for x in self.customlists]
+        if len(ids) == 0:
+            # Return None to make it clear that we mean "there is no
+            # custom list restriction" rather than "exclude
+            # everything".
+            return None
```
aslagle: If the lane is based on a list data source and there are no lists with that data source yet, I think this should return an empty list rather than None.

model.py
aslagle: I don't see why `Work.external_index_needs_updating` isn't handling this case. Doesn't it return the existing coverage record if one already exists?

aslagle:
It would be nice to eventually update the index when a list is deleted, but I guess it's not particularly important since no one would be searching with that list id.

model.py
leonardr: It doesn't -- it tries to create a second new coverage record, and when the session is flushed, there's an IntegrityError. You can duplicate this by removing the `and update_internal_index` bit and running `nosetests tests/test_model.py:TestCollection.test_custom_lists`.

Behind the scenes the duplicate coverage records are being created by `get_one_or_create`. So the creation of the second record is succeeding inside the nested subtransaction, but then when it's time to commit for real there's a problem.

PR #867, opened 2018-05-08
Cap the amount of work done when finding classifications for a work for the VerboseAnnotator

leonardr:
This branch avoids a worst-case scenario where it can take several minutes to gather all the identifiers associated with an extremely popular book like _Alice in Wonderland_. This is fine when we're deciding how to classify such a book, but it's really bad when there's an active HTTP request hitting the metadata wrangler's `lookup` endpoint asking about 20 books and we're trying to find every single subject under which anyone has ever classified _Alice in Wonderland_.

The good news is it's not actually that important that we gather all those subjects. Circulation managers only need this information to improve search quality, so only the most frequently encountered subjects are important, and capping the number of identifiers at a few hundred will let us figure out the most common subjects. We could probably even take it down to 100.


files/recursive_equivalents.sql
```diff
@@ -8,9 +8,9 @@ $$
         WITH RECURSIVE
                 find_equivs(n, strength, input_id, output_id) AS
                 (
-                SELECT 1, 1::DOUBLE PRECISION, $1 as input_id, $1 as output_id
+                SELECT 1, 1::DOUBLE PRECISION, $1 as input_id, $1 as output_id, 0::BIGINT as r
                 UNION
-                SELECT fe.n + 1, fe.strength * e.strength, e.input_id, e.output_id
+                SELECT fe.n + 1, fe.strength * e.strength, e.input_id, e.output_id, row_number() over (order by null) as r
```
aslagle: I'm not sure how row_number works in a recursive function. Does it apply to the entire end result or each level?

lane.py
```diff
@@ -887,7 +887,7 @@ def full_identifier(self):
         """A human-readable identifier for this WorkList that
         captures its position within the heirarchy.
         """
-        lane_parentage = list(self.parentage) + [self]
+        lane_parentage = list(list(reversed(list(self.parentage)))) + [self]
```
aslagle: not sure you need quite this many `list`s

files/recursive_equivalents.sql
leonardr: It applies to each level. So you might get up to `cutoff * recursion_depth` rows.

files/recursive_equivalents.sql
```diff
@@ -1,16 +1,17 @@
-CREATE OR REPLACE FUNCTION fn_recursive_equivalents(parent INT, recursion_depth INT, strength_threshold DOUBLE PRECISION)
+CREATE OR REPLACE FUNCTION fn_recursive_equivalents(parent INT, recursion_depth INT, strength_threshold DOUBLE PRECISION, cutoff INT DEFAULT null)
 RETURNS TABLE
         (
-        recursive_equivalent INT
+        recursive_equivalent INT,
+	r BIGINT
```
aslagle: I don't think you need to return r here - that's probably why the tests are failing.

PR #868, opened 2018-05-08
Annotate OPDS <feed> tags with a URI for the currently selected EntryPoint.

leonardr:
This branch adds a `simplified:entryPoint` attribute to the `<atom:feed>` tag when an EntryPoint is selected.

I added test coverage for add_breadcrumb_links, which previously had no standalone test coverage. I was able to simplify the signature by making it an instance method -- previously it was a class method that took an instance as its first argument, so it was effectively an instance method already.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -558,7 +558,7 @@ class AcquisitionFeed(OPDSFeed):
                 feed, make_link, entrypoints, facets.entrypoint
             )
 
-        cls.add_breadcrumb_links(feed, lane, annotator)
+        cls.add_breadcrumb_links(feed, lane, facets.entrypoint)
         annotator.annotate_feed(feed, lane)
 
         content = unicode(feed)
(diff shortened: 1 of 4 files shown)


opds.py
```diff
@@ -558,7 +558,7 @@ def make_link(ep, is_default):
                 feed, make_link, entrypoints, facets.entrypoint
             )
 
-        cls.add_breadcrumb_links(feed, lane, annotator)
+        cls.add_breadcrumb_links(feed, lane, facets.entrypoint)
```
aslagle: did you mean to change this to `feed.add_breadcrumb_links`? if not, why didn't this one change?

opds.py
```diff
@@ -1151,9 +1176,34 @@ def _make_entry_xml(self, work, license_pool, edition, identifier):
...
+
+        if not entrypoint.URI:
+            return
+        self.feed.attrib[self.CURRENT_ENTRYPOINT_ATTRIBUTE] = entrypoint.URI
+
+    def add_breadcrumbs(self, lane, include_lane=False, entrypoint=None):
+        """Add list of ancestor links in a breadcrumbs element.
+
+        TODO: The switchover from "no entry point" to "entry point" needs
+        its own breadcrumb link.
```
aslagle: For some reason I thought "no entry point" would use a default entry point.

leonardr: When you visit the lane, you are assigned a default entry point for display purposes, but for breadcrumbs, "lane X" needs to come before "lane X but with entry point Y selected", even though they may be the same.

opds.py
leonardr: I did mean to, and it's weird that it works without the change, but i guess cls.method(instance) is the same as instance.method().

PR #873, opened 2018-05-16
Change _best_media_type to take the URL of the resource and use it to derive a default media type if none is provided.

leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-286


aslagle:
Interesting, I was doing it like this https://github.com/NYPL-Simplified/server_core/compare/mirroring?expand=1#diff-a590e76d63cb187ef6fa6a1a34d2eadeR569

aslagle (approved):

PR #875, opened 2018-05-17
Some lanes shouldn't show up in their own grouped feeds

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-287. It implements a boolean switch that turns the feature on or off; we can implement a custom name separately, later.

### DIFF
--- a/tests/test_lane.py
+++ b/tests/test_lane.py
@@ -2714,6 +2714,10 @@ class TestWorkListGroups(DatabaseTest):
                     (i, expect_item, actual_item,
                      expect, actual)
                 )
+            eq_(len(expect), len(actual),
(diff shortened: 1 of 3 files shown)


tests/test_lane.py
```diff
@@ -2771,6 +2771,24 @@ def assert_contents(g, expect):
...
+                (mq_sf, best_sellers),
+                (mq_sf, staff_picks),
+                (hq_sf, sf_lane),
+                (mq_sf, sf_lane),
+                (hq_ro, romance_lane),
+                (mq_ro, romance_lane),
+                (nonfiction, discredited_nonfiction),
+            ]
+        )
+        fiction.exclude_self_from_grouped_feed = False
```
aslagle: These should be switched to `include` - how is it working?

PR #876, opened 2018-05-21
Only set mirror_url when a Representation has actually been mirrored.

leonardr:
This is part of the work for https://jira.nypl.org/browse/SIMPLY-297. The goal is to stop using mirror_url as 'the URL that should go into OPDS feeds for this Representation' and only have it mean 'the URL to which this specific installation personally mirrored to S3'.

To this end, methods that mirror representations, or mark them as mirrored, now manage `mirror_url`. In general, other code is not supposed to touch `mirror_url`, unless (as in Representation.best_thumbnail) it needs to enforce a policy like 'a Representation that we mirrored is better than one we didn't.'

A new property, `Representation.public_url`, is designed to mean 'the URL that should go into OPDS feeds for this Representation'.

I'm working on a corresponding circulation branch.

### DIFF
--- a/model.py
+++ b/model.py
@@ -2205,18 +2205,15 @@ class Identifier(Base):
         if content or content_path:
             # We have content for this resource.
             resource.set_fetched_content(media_type, content, content_path)
-        elif (media_type and (
-                not resource.representation
-                or not resource.representation.mirrored_at)
-        ):
-            # There's a version of this resource stored elsewhere that we
-            # can use.
-            #
-            # TODO: just because we know the type and URL of the
-            # resource doesn't mean we can actually serve that URL
-            # to patrons. This needs some work.
-            resource.set_mirrored_elsewhere(media_type)
+        elif (media_type and not resource.representation):
+            # We know the type of the resource, so make a
+            # Representation for it.
+            resource.representation, is_new = get_one_or_create(
+                _db, Representation, url=self.url, media_type=media_type
+            )
 
+        # TODO: This is where we would mirror the resource if we 
+        # wanted to.
         return link, new_link
 
     def add_measurement(self, data_source, quantity_measured, value,
@@ -2338,8 +2335,6 @@ class Identifier(Base):
         images = cls.resources_for_identifier_ids(
             _db, identifier_ids, rel)
         images = images.join(Resource.representation)
-        images = images.filter(Representation.mirrored_at != None).\
-            filter(Representation.mirror_url != None)
         images = images.all()
 
         champions = Resource.best_covers_among(images)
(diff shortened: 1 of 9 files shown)


model.py
```diff
@@ -2205,18 +2205,15 @@ def add_link(self, rel, href, data_source, media_type=None, content=None,
...
-                not resource.representation
-                or not resource.representation.mirrored_at)
-        ):
-            # There's a version of this resource stored elsewhere that we
-            # can use.
-            #
-            # TODO: just because we know the type and URL of the
-            # resource doesn't mean we can actually serve that URL
-            # to patrons. This needs some work.
-            resource.set_mirrored_elsewhere(media_type)
```
leonardr: Important change -- we no longer act like a resource we just heard of is 'mirrored' at its original URL.

model.py
```diff
@@ -5836,17 +5839,6 @@ def as_delivery_mechanism_for(self, licensepool):
...
-        """We don't need our own copy of this resource's representation--
-        a copy of it has been mirrored already.
-        """
-        _db = Session.object_session(self)
-        if not self.representation:
-            self.representation, is_new = get_one_or_create(
-                _db, Representation, url=self.url, media_type=media_type)
-        self.representation.mirror_url = self.url
-        self.representation.set_as_mirrored()
-
```
leonardr: We no longer care whether a resource was 'mirrored' by someone else -- either we mirrored it ourselves or we're using someone else's URL.

model.py
```diff
@@ -5988,22 +5982,22 @@ def best_covers_among(cls, resources):
...
+            # In order, our criteria are: whether we
+            # mirrored the representation (which means we directly
+            # control it), image quality, and media type suitability.
+            #
+            # We invert media type suitability because it's given to us
+            # as a priority (where smaller is better), but we want to compare
+            # it as a quantity (where larger is better).
+            compare_key = (rep.mirror_url is not None, quality, -media_priority)
+            if not champion_key or (compare_key > champion_key):
+                # A new champion.
```
leonardr: The overall change in this method is to add 'did we personally mirror this representation' as a criterion for quality. To add this I got rid of the complexity of nested if-then statements to determine whether one representation is better than another. Instead we create a 3-tuple key such that a better Representation will > a worse Representation, and then compare the keys to see which is best.

s3.py
```diff
@@ -237,65 +238,49 @@ def final_mirror_url(self, bucket, key):
         template = templates.get(self.url_transform, default)
         return template % dict(bucket=bucket, key=self.key_join(key))
 
-    def mirror_one(self, representation):
-        """Mirror a single representation."""
-        return self.mirror_batch([representation])
-
-    def mirror_batch(self, representations):
```
leonardr: mirror_batch already became useless when we switched to boto3, which doesn't support simultaneous uploads.

aslagle:
Without actually having read this code yet, I think you should have a migration to delete mirror_urls that aren't actually mirrored. Maybe if url and mirror_url are the same and the representation is not a thumbnail?

leonardr:
I've got a planned script that will solve the problem more comprehensively (including mirroring resources if necessary) but I can start with a migration that wipes out mirror_url if mirrored_at is null.

aslagle:
I believe set_as_mirrored_elsewhere did set mirrored_at so I don't think that would remove everything.

PR #877, opened 2018-05-24
Dont store identifier in cached opds entry

leonardr:
This branch does the core work of https://jira.nypl.org/browse/SIMPLY-273. The OPDS entry that is cached under Work.simple_opds_entry or Work.verbose_opds_entry no longer contains any information that can be derived from a particular Identifier or LicensePool, only Work and Edition. This means the cached OPDS entry can be reused for any Identifier or LicensePool associated with that Work.

### DIFF
--- a/model.py
+++ b/model.py
@@ -11579,15 +11579,24 @@ class Collection(Base, HasFullTableCache):
         """Finds all works in a collection's catalog that have been updated
            since the timestamp. Used in the metadata wrangler.
 
-           :return: a Query
+           :return: a Query that yields (Work, LicensePool, Identifier,
+              WorkCoverageRecord, CollectionIdentifier) 5-tuples. With
+              the Work alone, the caller would not have the information
+              necessary to create full OPDS entries for the items.
         """
         opds_operation = WorkCoverageRecord.GENERATE_OPDS_OPERATION
-        qu = _db.query(Work).join(Work.coverage_records)\
-            .join(Work.license_pools).join(Identifier)\
-            .join(Identifier.collections).filter(
-                Collection.id==self.id,
-                WorkCoverageRecord.operation==opds_operation,
-            ).options(joinedload(Work.license_pools, LicensePool.identifier))
+        qu = _db.query(
+            Work, LicensePool, Identifier, WorkCoverageRecord,
+            CollectionIdentifier
+        )
+        qu = qu.filter(
+            Work.id==WorkCoverageRecord.work_id,
+            Work.id==LicensePool.work_id,
+            LicensePool.identifier_id==Identifier.id,
+            CollectionIdentifier.identifier_id==Identifier.id,
+            CollectionIdentifier.collection_id==self.id,
+            WorkCoverageRecord.operation==opds_operation,
+        ).options(joinedload(Work.license_pools, LicensePool.identifier))
 
         if timestamp:
             qu = qu.filter(
--- a/model.py
+++ b/model.py
@@ -4746,11 +4746,11 @@ class Work(Base):
         )
         _db = Session.object_session(self)
         simple = AcquisitionFeed.single_entry(
-            _db, self, Annotator(), force_create=True
+            _db, self, Annotator, force_create=True
         )
         if verbose is True:
             verbose = AcquisitionFeed.single_entry(
-                _db, self, VerboseAnnotator(), force_create=True
+                _db, self, VerboseAnnotator, force_create=True
             )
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -11579,15 +11579,24 @@ def works_updated_since(self, _db, timestamp):
         """Finds all works in a collection's catalog that have been updated
            since the timestamp. Used in the metadata wrangler.
 
-           :return: a Query
+           :return: a Query that yields (Work, LicensePool, Identifier,
+              WorkCoverageRecord, CollectionIdentifier) 5-tuples. With
+              the Work alone, the caller would not have the information
+              necessary to create full OPDS entries for the items.
```
leonardr: Without this, this branch will break the metadata wrangler. CatalogController.updates_feed will get a list of recently updated Works in a given Collection, but will have no way of knowing which Identifier associates the Work with the Collection. Previously this wasn't a problem because you didn't really need an Identifier to complete an OPDS entry, but now you do.

opds.py
```diff
@@ -79,12 +79,47 @@ class Annotator(object):
 
     opds_cache_field = Work.simple_opds_entry.name
 
-    @classmethod
-    def annotate_work_entry(cls, work, active_license_pool, edition,
+    def annotate_work_entry(self, work, active_license_pool, edition,
```
leonardr: Making this an instance method instead of a class method required a lot of follow-on changes. I made this change because _sometimes_ this method needs to be an instance method, and I wasn't able to work out a way to call a method that might be a class method or an instance method depending on the subclass.

It didn't really need to be a class method, so I don't think this is a big problem.

opds.py
```diff
@@ -79,12 +79,47 @@ class Annotator(object):
...
+        # First, try really hard to find an Identifier that we can
+        # use to make the <id> tag.
+        if not identifier:
+            if active_license_pool:
+                identifier = active_license_pool.identifier
+            elif edition:
+                identifier = edition.primary_identifier
+
+        if identifier:
+            entry.append(AtomFeed.id(identifier.urn))
```
leonardr: Previously the output of permalink_for() was used as the AtomFeed.id. But in the circulation manager we didn't actually want this behavior -- we always used the Identifier.urn because Annotator.permalink_for was always being called instead of LibraryAnnotator.permalink_for.

Some part of this change (I'm not sure what) made it so that the LibraryAnnotator was using its own permalink_for implementation, which revealed that we really didn't want the <atom:id> to be based on permalink_for(), we always want it to be Identifier.urn.

opds.py
leonardr: A lot of the additions to this class are moving the code that needs to know about a LicensePool or Identifier out of _make_entry_xml.

opds.py
```diff
@@ -219,7 +288,7 @@ def categories(cls, work):
         return categories
 
     @classmethod
-    def authors(cls, work, license_pool, edition, identifier):
+    def authors(cls, work, edition):
```
leonardr: Several methods have had LicensePool and Identifier removed from their signatures to guarantee they don't have access to that information.

opds.py
```diff
@@ -1048,15 +1165,20 @@ def _make_entry_xml(self, work, license_pool, edition, identifier):
...
+                # time.
                 image_type = "image/png"
                 if url.endswith(".jpeg") or url.endswith(".jpg"):
                     image_type = "image/jpeg"
                 elif url.endswith(".gif"):
                     image_type = "image/gif"
                 links.append(AtomFeed.link(rel=rel, href=url, type=image_type))
 
-
-        permalink = self.annotator.permalink_for(work, license_pool, identifier)
```
leonardr: The code removed from here went into annotate_work_entry above.

model.py
aslagle: Why do you need the CollectionIdentifier when you already know the collection?

opds.py
```diff
@@ -351,6 +429,7 @@ def active_licensepool_for(cls, work):
                 active_license_pool = p
         return active_license_pool
 
+    @classmethod
     def sort_works_for_groups_feed(self, works, **kwargs):
```
aslagle: Why did this become a class method?

model.py
leonardr: I don't need the CollectionIdentifier, but I couldn't find a way to use the CollectionIdentifier in the query but not return it in the results.

aslagle: You might be able to `.join` it separately instead of putting it in the first `query`.

PR #879, opened 2018-05-25
Make Identifier.parse_urns capable of refusing to consider Identifiers of certain types.

leonardr:
If a URN can be parsed but the application knows it can't handle Identifiers of a certain type, it's now easy to refuse to load unhandlable Identifiers into the database.

### DIFF
--- a/model.py
+++ b/model.py
@@ -2012,8 +2012,28 @@ class Identifier(Base):
         return (type, identifier_string)
 
     @classmethod
-    def parse_urns(cls, _db, identifier_strings, autocreate=True):
-        """Batch processes URNs"""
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -2012,8 +2012,28 @@ def type_and_identifier_for_urn(cls, identifier_string):
...
+        :param identifier_strings: A list of strings, each a URN
+           identifying some identifier.
+
+        :param autocreate: Create an Identifier for a URN if none
+            presently exists.
+
+        :param allowed_types: If this is a non-empty set of Identifier
+            types, only identifiers of those types may be looked
+            up. All other identifier types will be treated as though
+            they did not exist.
```
aslagle: I thnk this contradicts the code for an empty set

PR #880, opened 2018-05-29
Resource derivatives

aslagle:
This branch introduces the concept of a "resource derivation" which is when a resource is used to create another resource based on some settings. It also adds rights status and a free text explanation of rights to resources.

This is currently used for adding a title and author to an open-access image to create a cover. It would also make sense to use for thumbnails, but I didn't do that in this branch.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -46,6 +46,7 @@ from model import (
     PresentationCalculationPolicy,
     RightsStatus,
     Representation,
+    Resource,
     Work,
 )
 from classifier import NO_VALUE, NO_NUMBER
@@ -411,7 +412,8 @@ class IdentifierData(object):
 
 class LinkData(object):
     def __init__(self, rel, href=None, media_type=None, content=None,
-                 thumbnail=None, rights_uri=None):
+                 thumbnail=None, rights_uri=None, rights_explanation=None,
+                 original=None, derivation_settings=None):
         if not rel:
             raise ValueError("rel is required")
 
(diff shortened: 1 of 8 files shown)


model.py
```diff
@@ -5795,6 +5804,32 @@ class Resource(Base):
...
         Integer, ForeignKey('representations.id'), index=True)
 
+    # The rights status of this Resource.
+    rights_status_id = Column(Integer, ForeignKey('rightsstatus.id'))
+
+    # An optional explanation of the rights status.
+    rights_explanation = Column(Unicode)
+
+    # Many derivatives may be derived from a Resource.
+    derivative_derivations = relationship(
```
aslagle: This is a terrible name.

metadata_layer.py
```diff
@@ -1556,10 +1563,27 @@ def _key(classification):
 
         for link in self.links:
             if link.rel in Hyperlink.METADATA_ALLOWED:
+                original_resource = None
+                if link.original:
+                    rights_status = RightsStatus.lookup(_db, link.original.rights_uri)
+                    original_resource, ignore = get_one_or_create(
+                        _db, Resource, url=link.original.href,
+                    )
+                    original_resource.data_source = data_source
```
leonardr: Do we know that the original resource's data source is the same as the derivative?

model.py
leonardr: Would it help if we said 'ResourceTransformation' instead of 'ResourceDerivation'? Then a Resource would be derived through a transformation rather than through a derivation, and this could be called `transformations`.

model.py
```diff
@@ -5795,6 +5804,32 @@ class Resource(Base):
...
+        'ResourceDerivation',
+        primaryjoin="ResourceDerivation.original_id==Resource.id",
+        foreign_keys=id,
+        lazy="joined",
+        backref=backref('original', uselist=False),
+        uselist=True,
+    )
+
+    # A derivative resource may have one original.
+    derivation = relationship(
```
leonardr: Maybe derived_through?

model.py
```diff
@@ -7818,6 +7890,16 @@ class RightsStatus(Base):
         GENERIC_OPEN_ACCESS,
     ]
 
+    ALLOWS_DERIVATIVES = [
```
leonardr: A comment would be useful here for the future. I'm also not convinced that GENERIC_OPEN_ACCESS should be in this list. That's often a dumping ground for 'you can have a copy of this but there's been no formal relicensing'.

PR #882, opened 2018-05-29
Create an ORM relationship between Collection and ExternalIntegration

leonardr:
This is so https://github.com/NYPL-Simplified/metadata_wrangler/pull/171 can use that relationship in a join.

### DIFF
--- a/model.py
+++ b/model.py
@@ -5529,7 +5529,7 @@ class LicensePoolDeliveryMechanism(Base):
             identifier=identifier,
             data_source=data_source,
             delivery_mechanism=delivery_mechanism,
-            resource=resource
+            resource=resource,
         )
         if not lpdm.rights_status or rights_status.uri != RightsStatus.UNKNOWN:
             # We have better information available about the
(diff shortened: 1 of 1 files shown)


model.py
```diff
@@ -10685,6 +10685,13 @@ class ExternalIntegration(Base, HasFullTableCache):
         lazy="joined", cascade="all, delete-orphan",
     )
 
+    # Any number of Collections may designate an ExternalIntegration
+    # as the source of their configuration
+    collections = relationship(
+        "Collection", backref="_external_integration",
+        foreign_keys='Collection.external_integration_id',
```
aslagle: I thought foreign_keys listed columns in this table, not the other table.

leonardr: I think it's the other table. http://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.relationship.params.foreign_keys says: "if the primaryjoin condition of this relationship() is a.id == b.a_id, and the values in b.a_id are required to be present in a.id, then the “foreign key” column of this relationship() is b.a_id." In this case the primaryjoin is ExternalIntegration.id==Collection.external_integration_id.

leonardr: The comment describes the relationship from the Collection's POV so maybe that is the issue?

aslagle: I'm confused but I think foreign_keys isn't required at all in this case since Collection.external_integration_id is already a ForeignKey.

aslagle (approved):

PR #885, opened 2018-05-31
Opds entrypoint breadcrumbs

EdwinGuzman:
Adding the currently selected entry point in the breadcrumbs. Helps resolve https://jira.nypl.org/browse/SIMPLY-264

### DIFF
--- a/tests/test_opds.py
+++ b/tests/test_opds.py
@@ -1779,6 +1779,22 @@ class TestAcquisitionFeed(DatabaseTest):
         eq_(feed.links[2].href, TestAnnotator.default_lane_url() + "?entrypoint=" + EbooksEntryPoint.URI)
         eq_(feed.links[2].title, EbooksEntryPoint.INTERNAL_NAME)
 
+        # A two level lane feed with no entry point
+        wl = WorkList()
+        wl.initialize(library=self._default_library, display_name="wl")
+        wl.lane = self._lane(parent=self._lane())
+        multi_lane_feed_with_no_eps = AcquisitionFeed(
+            self._db, self._str, self._url, [], annotator=TestAnnotator
+        )
+        multi_lane_feed_with_no_eps.add_breadcrumbs(wl.lane)
+        feed = feedparser.parse(unicode(multi_lane_feed_with_no_eps))['feed']
+
+        eq_(len(feed.links), 3)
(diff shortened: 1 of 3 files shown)


tests/test_opds.py
```diff
@@ -1742,6 +1742,60 @@ def test_format_types(self):
...
+        entrypoints = [AudiobooksEntryPoint, EbooksEntryPoint]
+        with_entrypoints_wl = WorkList()
+        with_entrypoints_wl.initialize(library=self._default_library, display_name="wl",
+                           entrypoints=entrypoints)
+        with_entrypoints_wl.lane = self._lane()
+
+        single_lane_feed_with_eps = AcquisitionFeed(
+            self._db, self._str, self._url, [], annotator=TestAnnotator
+        )
+        single_lane_feed_with_eps.add_breadcrumbs(with_entrypoints_wl.lane, entrypoint=EbooksEntryPoint)
```
EdwinGuzman: I'm not sure if I'm doing this correctly, but with how I'm calling the `add_breadcrumbs` functions to test it, it doesn't seem like I need to create a `WorkList`. It looks like I can pass in `self._lane()` into this function instead of a lane from a `WorkList`.

leonardr: Yes, I think this is fine.

EdwinGuzman:
Tried the different approach. Didn't exactly get the `MockFeed` class right.

tests/test_opds.py
```diff
@@ -1743,73 +1743,104 @@ def test_format_types(self):
...
+        # Top Level Title > 2001 > 2002
+        feed = MockFeed()
+        feed.add_breadcrumbs(subsublane)
+        children = getElementChildren(feed)
+
+        eq_(len(children), 3)
+        eq_(children[0].attrib.get("href"), TestAnnotator.default_lane_url())
+        eq_(children[0].attrib.get("title"), TestAnnotator.top_level_title())
+        assert(("?entrypoint=" + ep.URI) not in children[1].attrib.get("href"))
+        eq_(children[1].attrib.get("title"), "2001")
```
leonardr: You can't rely on the names of objects you create with the helper functions -- the test might change slightly and then this would be "2002" instead of "2001". Instead, check this against the variable containing the value, e.g. `sublane.title`.

PR #886, opened 2018-06-05
Use media type when calling default filename

leonardr:
This is a support branch for https://github.com/NYPL-Simplified/metadata_wrangler/pull/174.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -646,7 +646,9 @@ class MetaToModelUtility(object):
                 extension=extension
             )
         else:
-            filename = representation.default_filename(link_obj)
+            filename = representation.default_filename(
+                link_obj, representation.media_type
+            )
             mirror_url = mirror.cover_image_url(
                 data_source, identifier, filename
             )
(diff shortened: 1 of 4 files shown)


metadata_layer.py
```diff
@@ -646,7 +646,9 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
                 extension=extension
             )
         else:
-            filename = representation.default_filename(link_obj)
+            filename = representation.default_filename(
+                link_obj, representation.media_type
+            )
```
leonardr: Without this, images mirrored from Content Cafe will have the ".aspx" extension even though that's obviously wrong. Passing in the known media type helps us guess better.

testing.py
```diff
@@ -1110,6 +1113,17 @@ def queue_response(self, response_code, media_type="text/html",
...
+            self, response_code, media_type="text/html",
+            other_headers=None, content=''
+    ):
+        """Queue a response of the type produced by HTTP.get_with_timeout."""
+        headers = dict(other_headers or {})
+        if media_type:
+            headers['Content-Type'] = media_type
+        response = MockRequestsResponse(response_code, headers, content)
+        self.responses.append(response)
+
```
leonardr: It's a little annoying that we have two different ways of making HTTP requests (Representation.simple_http_get and HTTP.get_with_timeout), but there's no reason DummyHTTPClient can't work for mocking either technique -- the only difference in most cases is the output that needs to be queued.

PR #888, opened 2018-06-07
Metadata support branch

leonardr:
This branch adds and fixes miscellaneous things in support of work I'm doing on the metadata wrangler for https://jira.nypl.org/browse/SIMPLY-332.

There may be another support branch before I'm done, but there are a lot of changes here already so let's get them in.

A number of the changes reflect a change in metadata wrangler strategy. We will always be creating presentation-ready works as soon as possible, even for ISBNs, and even without what the circulation manager would consider a minimal set of bibliographic information. We will then be serving the OPDS entries for those presentation-ready works rather than cobbling together a dynamically generated entry based on Resources associated with an ISBN.

* If we don't know anything about a Work's classifications, we don't say anything. Previously we assumed that a Work with absolutely no classification information was a nonfiction work for adults. This was okay for the circulation manager, since it made sure every book showed up _somewhere_, but it's misleading for ISBNs whose only coverage information comes from Content Cafe, and it turned out not to be a big problem on the circulation side.
* `CoverageProvider.get_operation` was turned into a property called `operation`, rather than `operation` being an attribute set in the constructor. This may require minor changes to circulation.
*  New method: `CoverageProvider.can_cover`, which answer the question of whether a CoverageProvider can handle a given Identifier.
* `CoverageProvider.ensure_coverage` sets the `.collection` of the newly created CoverageRecord if the CoverageProvider needs to cover the same Identifier separately for each collection.
* `CoverageProvider.license_pool` can create a LicensePool with a DataSource different from the one associated with the CoverageProvider. This is so (e.g.) the ContentCafeCoverageProvider can notice that an INTERNAL_PROCESSING LicensePool was never created for the Identifier it's processing, and create one so that it can create a Work.
* `CoverageProvider.work` can take a LicensePool to use as the basis of a Work rather than calling `license_pool` with the default arguments.
* Some code has been refactored from `CoverageRecord.__repr__` so that the same type of message can be included in an OPDS message. This will provide some visibility into "why isn't this ISBN ready?"
* If a book's LicensePool comes INTERNAL_PROCESSING, then it doesn't represent an actual copy of a book, and no <bibframe:distribution> tag will be present in the OPDS feed. (Another way to fix this would be to move this code out of core and into circulation, since circulation is the only component left that serves the actual books.)

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -137,10 +137,9 @@ class BaseCoverageProvider(object):
                 "%s must define SERVICE_NAME." % self.__class__.__name__
             )
         service_name = self.__class__.SERVICE_NAME
-        self.operation = self.get_operation()
-        
-        if self.operation:
-            service_name += ' (%s)' % self.operation
+        operation = self.get_operation()
+        if operation:
+            service_name += ' (%s)' % operation
         self.service_name = service_name
         if not batch_size or batch_size < 0:
             batch_size = self.DEFAULT_BATCH_SIZE
@@ -826,7 +825,7 @@ class IdentifierCoverageProvider(BaseCoverageProvider):
         """
         qu = Identifier.missing_coverage_from(
             self._db, self.input_identifier_types, self.data_source,
-            count_as_missing_before=self.cutoff_time, operation=self.operation,
+            count_as_missing_before=self.cutoff_time, operation=self.get_operation(),
             identifiers=self.input_identifiers, collection=self.collection_or_not,
             **kwargs
         )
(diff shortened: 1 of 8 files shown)


coverage.py
```diff
@@ -372,6 +372,16 @@ def finalize_batch(self):
...
 
+    def can_cover(self, identifier):
+        """Can this CoverageProvider do anything with the given Identifier?
+
+        This is not needed in the normal course of events, but a
+        caller may need to decide whether to pass an Identifier
+        into ensure_coverage() or register().
+        """
+        return (not self.input_identifier_types
+                or identifier.type in self.input_identifier_types)
```
aslagle: Should this be in IdentifierCoverageProvider instead of BaseCoverageProvider?

model.py
```diff
@@ -1395,6 +1395,11 @@ class CoverageRecord(Base, BaseCoverageRecord):
     )
 
     def __repr__(self):
+        template = '<CoverageRecord: %(timestamp)s identifier=%(identifier_type)s/%(identifier)s data_source="%(data_source)s"%(operation)s status="%(status)s" %(exception)s>'
```
aslagle: data source and operation could use a space between them

PR #889, opened 2018-06-08
Metadata support

leonardr:
This is a second support branch for https://github.com/NYPL-Simplified/metadata_wrangler/pull/175, and it's mostly stuff that turned out not to be necessary but is a good improvement.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -225,7 +225,7 @@ class ContributorData(object):
         :return: the possibly changed Contributor object and a flag of whether it's been changed.
         """
         log = logging.getLogger("Abstract metadata layer")
-        log.debug(u"Applying %r (%s) into %r (%s)", self, self.viaf, destination, destination.viaf)
+        log.debug("Applying %r (%s) into %r (%s)", self, self.viaf, destination, destination.viaf)
 
         made_changes = False
 
@@ -882,10 +882,10 @@ class CirculationData(MetaToModelUtility):
             raise ValueError(
                 "Cannot find license pool: CirculationData has no primary identifier."
             )
-        
-        data_source = self.data_source(_db)
+
+        data_source_obj = self.data_source(_db)
         license_pool, is_new = LicensePool.for_foreign_id(
-            _db, data_source=self.data_source_obj,
+            _db, data_source=data_source_obj,
             foreign_id_type=identifier.type, 
             foreign_id=identifier.identifier,
             collection=collection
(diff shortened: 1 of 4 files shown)


metadata_layer.py
```diff
@@ -225,7 +225,7 @@ def apply(self, destination, replace=None):
         :return: the possibly changed Contributor object and a flag of whether it's been changed.
         """
         log = logging.getLogger("Abstract metadata layer")
-        log.debug(u"Applying %r (%s) into %r (%s)", self, self.viaf, destination, destination.viaf)
+        log.debug("Applying %r (%s) into %r (%s)", self, self.viaf, destination, destination.viaf)
```
leonardr: This is necessary because repr() returns a base string, not a Unicode string.

metadata_layer.py
```diff
@@ -882,10 +882,10 @@ def license_pool(self, _db, collection, analytics=None):
             raise ValueError(
                 "Cannot find license pool: CirculationData has no primary identifier."
             )
-        
-        data_source = self.data_source(_db)
+
+        data_source_obj = self.data_source(_db)
         license_pool, is_new = LicensePool.for_foreign_id(
-            _db, data_source=self.data_source_obj,
+            _db, data_source=data_source_obj,
```
leonardr: This ended up not mattering but I don't see why you'd call data_source() and then not use the result.

model.py
```diff
@@ -4573,7 +4574,7 @@ def calculate_presentation_edition(self, policy=None):
 
     def calculate_presentation(
         self, policy=None, search_index_client=None, exclude_search=False,
-        default_fiction=False, default_audience=Classifier.AUDIENCE_ADULT
+        default_fiction=None, default_audience=None
```
leonardr: This _did_ matter -- the classifier itself didn't assume anything about fiction or audience status, but our actual entry point into that process did.

overdrive.py
```diff
@@ -1120,4 +1120,11 @@ def process_item(self, identifier):
...
             return self.failure(identifier, e)
 
+        self.metadata_pre_hook(metadata)
         return self.set_metadata(identifier, metadata)
+
+    def metadata_pre_hook(self, metadata):
+        """A hook method that allows subclasses to modify a Metadata
+        object derived from Overdrive before it's applied.
+        """
+        return metadata
```
leonardr: This allows the metadata wrangler's Overdrive coverage provider to strip out the CirculationData (assuming Overdrive provides any), so it doesn't think it should be serving books and create an Overdrive LicensePool.

PR #890, opened 2018-06-11
Allow a Work to be created in some circumstances even if no title is available.

leonardr:
This branch smooths out the process of creating Works for ISBNs looked up in the metadata wrangler. In some cases we have all the information necessary to create a Work for an ISBN, except for the book title. We don't actually need the book title because we're not displaying anything directly to readers. So it should be possible to create a Work with no title, just as it's possible to create a Work with no author.

### DIFF
--- a/model.py
+++ b/model.py
@@ -7550,7 +7550,8 @@ class LicensePool(Base):
 
 
     def calculate_work(
-        self, even_if_no_author=False, known_edition=None, exclude_search=False
+        self, even_if_no_author=False, known_edition=None, exclude_search=False,
+        even_if_no_title=False
     ):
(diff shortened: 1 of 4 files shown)


model.py
```diff
@@ -7560,15 +7561,23 @@ def calculate_work(
         of the LicensePool's presentation edition.
 
         :param even_if_no_author: Ordinarily this method will refuse
-        to create a Work for a LicensePool whose Edition has no title
-        or author. But sometimes a book just has no known author. If
+        to create a Work for a LicensePool whose Edition has
```
aslagle: You lost a 'no' in this edit.

aslagle (approved):

PR #892, opened 2018-06-12
Add fields novelist request

EdwinGuzman:
Adding dictionaries for http://schema.org/BookFormatType lookup. Related and used in [circulation #997](https://github.com/NYPL-Simplified/circulation/pull/977).

### DIFF
--- a/model.py
+++ b/model.py
@@ -2218,7 +2218,7 @@ class Identifier(Base):
             rights_status = RightsStatus.lookup(_db, rights_status_uri)
         resource, new_resource = get_one_or_create(
             _db, Resource, url=href,
-            create_method_kwargs=dict(data_source=data_source, 
+            create_method_kwargs=dict(data_source=data_source,
                                       rights_status=rights_status,
                                       rights_explanation=rights_explanation)
         )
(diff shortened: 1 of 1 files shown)


leonardr (changes requested):
I wouldn't put this in core for a few reasons.

NoveList has created their own controlled vocabulary -- it just happens to use the same terms as the schema.org terms. It's not the same because they're omitting the schema.org part of the URIs.

We don't use "hardcover" or "paperback" as possible values for Edition.medium, so it's misleading to define constants like HARDCOVER_FORMAT.

You don't use medium_to_book_format_type at all in the circulation branch.

NoveList uses `http://schema.org/AudiobookFormat` for 'audiobook format' where we use `http://bib.schema.org/Audiobook` for 'this is an audiobook'. Neither one of these is wrong but putting them both in the same file I think is likely to cause confusion.

leonardr:
I would put a single lookup table in novelist.py in circulation.

EdwinGuzman:
Thanks. That makes sense and I was thinking of going that route, but figured to have all those values together in one place. Will move it over.

PR #895, opened 2018-06-19
Define a common self-test interface for third-party integration classes

leonardr:
This branch defines a class called `HasSelfTests` which implements a class method `run_self_tests`. The job of this method is to instantiate the class and run a suite of self tests. The results are a sequence of `SelfTestResult` objects. Each `SelfTestResult` records how long some code took to run, whether it raised an exception, and what the return value was.

The overall idea is that the circulation manager's admin interface will eventually feature a self-test button for each configured integration. Clicking the self-test button will call `run_self_tests` on the appropriate class, and the `SelfTestResult` objects will be displayed to the user.

This is how self-tests differ from unit tests and integration tests -- they test that a running site is configured to communicate with a foreign server, not the correctness of the underlying code.

A new class, `IntegrationException`, represents a problem with a third-party integration. It's the new superclass of `RemoteIntegrationException`, which is used primarily for problems that happen during HTTP requests, and `CannotLoadConfiguration`, which is used for obviously missing data in an `ExternalIntegration`'s configuration. (In circulation, it's also the new superclass of `CirculationException`). The fundamental thing about an `IntegrationException` is that it has a `message`, like any Exception, but it also has a `debug_message`, which gives more specific information about the problem, including--potentially--how to fix the problem.

Although any `Exception` raised during a self-test will be understood as the reason why the test failed, using `IntegrationExceptions` instead and gradually improving their `debug_message`s will let us make the process of debugging a circulation manager configuration easier over time.

### DIFF
--- a/model.py
+++ b/model.py
@@ -8407,6 +8407,7 @@ class Representation(Base):
     EPUB_MEDIA_TYPE = u"application/epub+zip"
     PDF_MEDIA_TYPE = u"application/pdf"
     MOBI_MEDIA_TYPE = u"application/x-mobipocket-ebook"
+    AMAZON_KF8_MEDIA_TYPE = u"application/x-mobi8-ebook"
     TEXT_XML_MEDIA_TYPE = u"text/xml"
     TEXT_HTML_MEDIA_TYPE = u"text/html"
     APPLICATION_XML_MEDIA_TYPE = u"application/xml"
@@ -8428,6 +8429,7 @@ class Representation(Base):
         PDF_MEDIA_TYPE,
         MOBI_MEDIA_TYPE,
         MP3_MEDIA_TYPE,
+        AMAZON_KF8_MEDIA_TYPE,
     ]
 
     # These media types are in the order we would prefer to use them.
--- a/overdrive.py
+++ b/overdrive.py
@@ -320,10 +320,7 @@ class OverdriveAPI(object):
         """Get IDs for every book in the system, with the most recently added
         ones at the front.
         """
-        params = dict(collection_token=self.collection_token,
-                      sort="dateAdded:desc")
-        next_link = self.make_link_safe(
-            self.ALL_PRODUCTS_ENDPOINT % params)
+        next_link = self._all_products_link
         while next_link:
             page_inventory, next_link = self._get_book_list_page(
                 next_link, 'next'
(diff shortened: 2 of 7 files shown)


model.py
```diff
@@ -8407,6 +8407,7 @@ class Representation(Base):
     EPUB_MEDIA_TYPE = u"application/epub+zip"
     PDF_MEDIA_TYPE = u"application/pdf"
     MOBI_MEDIA_TYPE = u"application/x-mobipocket-ebook"
+    AMAZON_KF8_MEDIA_TYPE = u"application/x-mobi8-ebook"
```
leonardr: This is an unrelated change -- Standard Ebooks has started serving books in this media type and this makes it so we mirror them.

overdrive.py
```diff
@@ -332,6 +329,12 @@ def all_ids(self):
             for i in page_inventory:
                 yield i
 
+    @property
+    def _all_products_link(self):
+        params = dict(collection_token=self.collection_token,
+                      sort="dateAdded:desc")
+        return self.make_link_safe(self.ALL_PRODUCTS_ENDPOINT % params)
+
```
leonardr: This is used by the Overdrive self-test, which is located in circulation.

selftest.py
```diff
@@ -0,0 +1,136 @@
...
+    @classmethod
+    def run_self_tests(cls, _db, constructor_method=None, *args, **kwargs):
+        """Instantiate this class and call _run_self_tests on it.
+
+        :param _db: A database connection. Will be passed into
+        _run_self_tests as well as the first argument of the constructor
+        method.
+
+        :param constructor_method: Method to use to instantiate the
+        class, if different from the default constructor.
```
leonardr: We need this because some integration classes want to be instantiated with `from_config` or another similar method, and some want their normal constructor to be used.

selftest.py
```diff
@@ -0,0 +1,135 @@
...
+
+        :return: An iterator of SelfTestResult objects, starting with
+        the attempt to instantiate the test class in the first place.
+        """
+        constructor_method = constructor_method or cls
+        result = SelfTestResult("Initial setup.")
+        instance = None
+        try:
+            instance = constructor_method(_db, *args, **kwargs)
```
aslagle: I don’t think this will always work. For example, authentication providers have a library as their first argument.

selftest.py
```diff
@@ -0,0 +1,138 @@
...
+        if instance:
+            for result in instance._run_self_tests(_db):
+                yield result
+
+    def _run_self_tests(self, _db):
+        """Run a series of self-tests.
+
+        :return: A list of SelfTestResult objects.
+        """
+        raise NotImplementedError()
```
aslagle: I'm wondering if this should return a self test failure object that says "self tests have not been implemented for this integration". Otherwise, we'd likely get a 500 when trying to run tests from the admin interface, so we'd need some way to tell the admin interface in advance which integrations have self tests.

aslagle: I guess we need that anyway, since integrations that don't have them yet won't extend this class.

leonardr: I was thinking the admin interface wouldn't offer the self-test functionality for any class that didn't implement HasSelfTests. It needs to know how to instantiate the class, so there's no way to write fully general code anyway.

PR #898, opened 2018-06-20
Persist the most recent set of self-test results to the database

leonardr:
This branch addreses https://jira.nypl.org/browse/SIMPLY-677.

Implementations of HasSelfTests should implement a method with signature `external_integration(_db)`. This method returns an ExternalIntegration, and self-test results for a HasSelfTests object are stored as a configuration setting in that ExternalIntegration.

As you'll see in the corresponding circulation branch, implementations of `external_integration` vary but they're never very complicated.

`SelfTestResult` now has a dictionary representation that can be serialized as JSON.

`HasSelfTests.run_self_tests` uses these dictionary representations to create a JSON report on the self-test, which it stores in the appropriate ExternalIntegration.

`run_self_tests` now returns a 2-tuple instead of yielding a sequence of `SelfTestResult` objects. The first item in the tuple is the JSON report, which can be formatted for display to a human being. The second item is the list of SelfTestResults -- this is mainly useful during testing, to make sure the results were created properly.

### DIFF
--- a/selftest.py
+++ b/selftest.py
@@ -3,6 +3,8 @@
 from nose.tools import set_trace
 from util.http import IntegrationException
 import datetime
+import json
+from util.opds_writer import AtomFeed
 
 
 class SelfTestResult(object):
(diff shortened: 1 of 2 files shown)


selftest.py
```diff
@@ -82,11 +138,52 @@ def run_self_tests(cls, _db, constructor_method=None, *args, **kwargs):
...
+        # Store the formatted results in the database, if we can find
+        # a place to store them.
+        if integration:
+            integration.setting(
+                cls.SELF_TEST_RESULTS_SETTING
+            ).value = json.dumps(value)
+        return value, results
+
+    @classmethod
+    def prior_test_results(cls, _db, constructor_method=None, *args, **kwargs):
```
EdwinGuzman: Is this the method that I will use to display the results of the self tests from the db, or are you expecting the tests to re-run when an admin presses a "run tests" button?

aslagle: I don't know what it should look like, but I think you should always show the latest test results, with the button to re-run. The admin shouldn't have to explicitly request the results. Particularly if there's a failure, it should be prominently displayed up front.

leonardr: That's fine. The controller can call `prior_test_results` whenever it wants, even if no one has explicitly asked for the test results.

The only problem is that sometimes instantiating the object (which is necessary to call `external_integration` to retrieve the test results) triggers API requests. If the API configuration is broken, you might get a failure before you can even retrieve the test results. That's not a big problem because the controller can catch the exception and show that there's a problem.

However, a larger problem is that the API request might succeed but take a really long time (Odilo is currently in this situation). So now we're looking at a 30-second load time on the collections page even if all the tests technically pass.

The long-term solution is to make sure just instantiating the class doesn't make any network requests. I think we'd only have to change Odilo and Overdrive.

EdwinGuzman (approved):

PR #900, opened 2018-06-20
Add a self-test for ExternalIntegrations based on OPDS import.

leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-684.

This is a little different from the other HasSelfTest implementations because we generally think of an API integration class as something standalone, not a `Monitor`. But the self-test needs to exercise the code that makes an authenticated request to the OPDS feed, not just a random HTTP request, and `OPDSImportMonitor` is the class where that happens.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -60,6 +60,7 @@ from util.opds_writer import (
     OPDSMessage,
 )
 from mirror import MirrorUploader
+from selftest import HasSelfTests
 
 
 class AccessNotAuthenticated(Exception):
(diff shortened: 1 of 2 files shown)


opds_import.py
```diff
@@ -465,6 +467,74 @@ def data_source(self):
...
+        """Is `url` really an open-access link?
+
+        That is, can we make a normal GET request and get something
+        that looks like a book?
+        """
+        headers = {}
+        if type:
+            headers["Accept"] = type
+        status, headers, body = self.http_get(url, headers=headers)
+        if status == 200 and len(body) > 1024 * 10:
```
EdwinGuzman: Where does `1024 * 10` come from?

leonardr: I made it up. 10 kilobytes is smaller than any reasonably-sized EPUB but larger than most "uh-oh, there was an error" API error documents.

PR #901, opened 2018-07-09
Removing default variable when generating facet entrypoint links.

EdwinGuzman:
I believe this fixes the issue where the "All" default entrypoint facet link picks up the same facet link as what is currently selected. So, like the image below, when "Audio Books" was selected, the "All" tab link had `?entrypoint=Audio` in the url.

samelink

I noticed that down the line in `lane.py`, FeatureFacets.navigate line 560 is where the entrypoint would be passed as `None` if the value is a default. In that case, `self.entrypoint` is the value of the entrypoint facet from the url. I think there's a better way to fix this but I don't know where `FeatureFacets` is instantiated with its initial value of entrypoint to better fix this.

### DIFF
--- a/entrypoint.py
+++ b/entrypoint.py
@@ -55,7 +55,7 @@ class EntryPoint(object):
         if display_title in cls.DISPLAY_TITLES.values():
             raise ValueError(
                 "Duplicate entry point display name: %s" % display_title
-            )            
+            )
         cls.DISPLAY_TITLES[entrypoint_class] = display_title
         cls.BY_INTERNAL_NAME[value] = entrypoint_class
         cls.ENTRY_POINTS.append(entrypoint_class)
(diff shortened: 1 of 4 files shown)


leonardr:
I see. I think the best way to fix this would be to change `FacetsWithEntryPoint.query_string` so that it omits the `entrypoint=` bit whenever the entry point is the default one. But it's purely a cosmetic thing so I'm okay with just removing this code.

EdwinGuzman:
Thanks. I tried updating `FacetsWithEntryPoint.query_string` but it did not help me solve the issue.

entrypoint.py
```diff
@@ -141,9 +141,9 @@ def modified_search_arguments(cls, **kwargs):
...
     INTERNAL_NAME = "Book"
     URI = u"http://schema.org/EBook"
-EntryPoint.register(EbooksEntryPoint, "Books", default_enabled=True)
+EntryPoint.register(EbooksEntryPoint, "eBooks", default_enabled=True)
 
 class AudiobooksEntryPoint(MediumEntryPoint):
     INTERNAL_NAME = "Audio"
     URI = u"http://bib.schema.org/Audiobook"
-EntryPoint.register(AudiobooksEntryPoint, "Audio")
+EntryPoint.register(AudiobooksEntryPoint, "Audio Books")
```
aslagle: I think this is normally written as one word - "Audiobooks".

EdwinGuzman: Yeah, Ricardo has `Audio Books` in the mockup, but I think NYPL uses both conventions all over. It seems `Audiobooks` is more pervasive so I'll use that. Thanks.

PR #903, opened 2018-07-18
Added a method to check if a collection is visible to an admin.

aslagle:
This helps with some changes I'm making to the dashboard statistics where the totals will only include collections from libraries the admin has access to.

### DIFF
--- a/model.py
+++ b/model.py
@@ -10574,13 +10574,10 @@ class Admin(Base, HasFullTableCache):
         return False
 
     def can_see_collection(self, collection):
+        if self.is_system_admin():
+            return True
         for library in collection.libraries:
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -10573,6 +10573,17 @@ def lookup_hook():
...
 
+    def can_see_collection(self, collection):
+        for library in collection.libraries:
+            if self.is_librarian(library):
+                return True
+        if not collection.libraries:
+            # If the collection's not associated with any libraries, only system
+            # admins can see it.
+            if self.is_system_admin():
+                return True
```
leonardr: This is fine, but it would be a little simpler with the same results if you checked is_system_admin right off.

PR #904, opened 2018-07-20
Extract publication date from dc:issued during OPDS import.

aslagle:
We weren't extracting publication dates before, but now we have some OPDS feeds that have them.

See https://github.com/NYPL-Simplified/server_core/pull/769 for why this is done with elementtree instead of feedparser.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -6,6 +6,7 @@ from collections import (
 )
 import base64
 import datetime
+import dateutil
 import feedparser
 import logging
 import traceback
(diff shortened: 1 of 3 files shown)


opds_import.py
```diff
@@ -1306,6 +1307,18 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None):
...
             data['series'], data['series_position'] = cls.extract_series(series_tag[0])
 
+        issued_tag = parser._xpath(entry_tag, 'dcterms:issued')
+        if issued_tag:
+            date_string = issued_tag[0].text
+            # By default, the date for strings that only have a year will
+            # be set to January 1 rather than the current date.
+            default = datetime.datetime(datetime.datetime.now().year, 1, 1)
+            try:
+                data["published"] = dateutil.parser.parse(date_string, default=default)
```
aslagle: I'm not sure if I should be using issued or published for this. It might not be consistent between OPDS feeds.

aslagle: I meant if I should store it as Edition.issued or Edition.published. published seems safer - in our feeds we put either issued or published as dc:issued depending on what we have.

leonardr: I agree published is safer.

PR #905, opened 2018-07-23
Implemented and tested ExternalIntegration.with_setting_value.

leonardr:
This branch implements a new method that finds ExternalIntegrations based on their protocol and goal, _and_ on having a certain ConfigurationSetting set to a specific value. This is useful when you want to e.g. find an integration for a specific URL.


leonardr:
Used by https://github.com/NYPL-Simplified/circulation/pull/1000.

aslagle (approved):

PR #910, opened 2018-08-03
Update PatronInputScript to take a library

jonathangreen:
Update the PatronInputScript class to inheret from LibraryInputScript so that if patrons are associated with multiple libraries can be disambiguated.

These changes look bigger then they are. I had to move `PatronInputScript` to be below `LibraryInputScript` in the file so I could inherit from `LibraryInputScript`, and that made the diff a little messy.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -547,23 +547,23 @@ class IdentifierInputScript(InputScript):
 
 class LibraryInputScript(InputScript):
     """A script that operates on one or more Libraries."""
-
     @classmethod
-    def parse_command_line(cls, _db=None, cmd_args=None, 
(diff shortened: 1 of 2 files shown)


jonathangreen:
I rebased this to make it a little easier to code review. The first commit 
https://github.com/NYPL-Simplified/server_core/pull/910/commits/7c8156d6342e718b79030fda1f1d4902b9ba6b0a is just rearranging the functions and the second commit 
https://github.com/NYPL-Simplified/server_core/pull/910/commits/db6cfd8effda5add58bf25ef6d076015df67b4a0 is where all the changes happen.

PR #913, opened 2018-08-08
Remove trailing whitespace from all core Python files.

leonardr:
This will clean up the code generally and put a stop to future branches that include irrelevant whitespace changes.

I made this change with the following script:

find ./ -type f -name "*.py" -exec sed -ir 's/ \{1,\}$//g' {} \;


leonardr:
If there were a way of automatically tidying up code on the way in, I think that would be really useful, but it seems like it would have to run before the files are committed to version control (the way Atom does it). 

I don't like the idea of failing integration if there are minor problems with formatting. I've worked on a project like that and I think it dramatically cut project velocity for very little benefit.

My main concern is avoiding a ton of irrelevant whitespace in PRs. If there's just a little bit occasionally I'm OK with it.

So maybe we should set up linting as a pre-commit hook?

PR #917, opened 2018-08-15
Fix two flaws in hold time estimates.

leonardr:
This branch closes http://jira.nypl.org/browse/SIMPLY-1104. There are numerous problems with estimated wait times, but this branch closes three problems that are simply due to bad math.

* If there is no known default loan time, we can't make any estimate and we have to give up. This wouldn't have happened on any real site.
* If you are at the very front of the queue, such that the book is reserved to you, your estimated wait time is given as though you were at the end of the queue. I don't think this is shown to SimplyE users, since the book is available right now and you can borrow it, but if it were visible it would be very confusing.
* A `<` instead of a `<=` greatly increased the estimated wait time in the case where your position in line was an exact multiple of the number of available copies. This was especially noticeable when there was one copy of the book.

As part of this work I added test coverage for `Hold.until`, which was previously not covered (though the helper method it calls is covered pretty well).

### DIFF
--- a/model.py
+++ b/model.py
@@ -894,8 +894,7 @@ class Hold(Base, LoanAndHoldMixin):
         # Start with the default loan period to clear out everyone who
         # currently has the book checked out.
         duration = default_loan_period
-
-        if queue_position < total_licenses:
+        if queue_position <= total_licenses:
             # After that period, the book will be available to this patron.
             # Do nothing.
             pass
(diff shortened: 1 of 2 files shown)


model.py
```diff
@@ -894,8 +894,7 @@ def _calculate_until(
         # Start with the default loan period to clear out everyone who
         # currently has the book checked out.
         duration = default_loan_period
-
-        if queue_position < total_licenses:
+        if queue_position <= total_licenses:
```
aslagle: I think the only behavior that actually changed here is when `queue_position` and `total_licenses` are both 1. If `total_licenses` is > 1 and the same as `queue_position`, `cycles` was already 0 in the else condition.

That does seem like it was wrong before, but I think it might still be wrong - it was `default_loan_period + default_loan_period + default_reservation_period`, and now it's `default_loan_period`, but I think it should be `default_loan_period + default_reservation_period`. In the worst case, you're at position 1 in the queue and another patron has the book reserved but hasn't checked it out yet.

PR #918, opened 2018-08-16
Change 'spine' to 'readingOrder'

adriana-alter:
Partially resolves https://jira.nypl.org/browse/SIMPLY-1105.  For now, the code for "timeline" is still in for potential future use (even though we're not currently using it).  We're not using "sort as" (for author names), so no change was necessary there.


io7m:
Looks fine to me! :+1:

PR #920, opened 2018-08-16
Make language name lookups work consistently

leonardr:
This branch changes `LanguageCodes` to store its mappings a new class called `LookupTable` instead of  `defaultdict(lambda:None)`. That sort of `defaultdict` changes its contents in a way that can break the `in` operator in a nondeterministic way.

```
from collections import defaultdict
d = defaultdict(lambda:None)
'key' in d # False
d['key'] is None # True
'key' in d # True
```

What we really want here is a dictionary that returns `None` _without modifying the dictionary_ when we look up an item that's not there.

I could have fixed this by changing `LanguageCodes.string_to_alpha_3` to use `cls.three_to_two[s]` instead of `s in cls.three_to_two` but I figured it was safer to make both possible techniques work.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -20,7 +20,7 @@ from util import (
     Bigrams,
     english_bigrams,
     LanguageCodes,
-    LookupDict,
+    LookupTable,
     MetadataSimilarity,
     MoneyUtility,
     TitleProcessor,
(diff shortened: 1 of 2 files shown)


leonardr:
This fixes the cause of this test failure, which initially seemed completely baffling: https://travis-ci.org/NYPL-Simplified/circulation/builds/416890887

PR #923, opened 2018-08-17
Delete unused script

adriana-alter:
Partially resolves  https://jira.nypl.org/browse/SIMPLY-1111


leonardr:
The tests for this script in test_scripts.py also need to be removed.

PR #925, opened 2018-08-20
Fail gracefully when Elasticsearch server is inaccessible or not working as expected

leonardr:
This branch changes the behavior of the ExternalSearch class so that if there's a problem connecting to or communicating with the ElasticSearch server on startup, `CannotLoadConfiguration` is raised rather than one `ElasticSearchException` or another. The circulation manger is prepared to catch `CannotLoadConfiguration` here if there's a problem, but it shouldn't need to know about the particular search service in use.

The motivating case for this code is https://jira.nypl.org/browse/SIMPLY-1128. Without this code, it's possible to get your configuration into a state where you can't get into the admin interface.

Also, if there's a problem communicating with the server during a search, `Lane.search` now always returns an empty list; previously it could sometimes return `None`, which would cause other code to raise an uncaught exception.

This second change should hopefully resolve some mysterious ISEs seen in production.

### DIFF
--- a/tests/test_external_search.py
+++ b/tests/test_external_search.py
@@ -1,5 +1,6 @@
 from nose.tools import (
     assert_raises,
+    assert_raises_regexp,
     eq_,
     set_trace,
 )
@@ -11,6 +12,9 @@ from . import (
     DatabaseTest,
 )
 
+from elasticsearch.exceptions import ElasticsearchException
+
+from config import CannotLoadConfiguration
 from lane import Lane
 from model import (
     Edition,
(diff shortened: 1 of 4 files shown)


tests/test_external_search.py
```diff
@@ -863,6 +886,8 @@ class TestExactMatches(ExternalSearchTest):
 
     def setup(self):
         super(TestExactMatches, self).setup()
+        if not self.search:
+            return
```
leonardr: These classes should have been skipped when Elasticsearch wasn't configured, but they were being run.

PR #927, opened 2018-08-20
Remove Axis360 classifier

adriana-alter:


leonardr:
For the record, this class wasn't being used -- the generic "freeform audience classifier" happens to work for the audience strings provided by Axis 360. We are still tracking "Axis 360 audience" as a separate Subject type, though.

leonardr:
Resolves https://jira.nypl.org/browse/SIMPLY-1116

PR #928, opened 2018-08-20
Move Dewey Decimal classifier and its tests into separate files

adriana-alter:


leonardr:
Addresses https://jira.nypl.org/browse/SIMPLY-1117

PR #929, opened 2018-08-20
Separate LCCClassifier and its tests

adriana-alter:


leonardr:
Addresses https://jira.nypl.org/browse/SIMPLY-1118

PR #933, opened 2018-08-20
Move Simplified classifiers

adriana-alter:


leonardr:
Addresses https://github.com/NYPL-Simplified/server_core/pull/933

PR #935, opened 2018-08-21
Fix import statements

adriana-alter:


leonardr:
These erroneous imports didn't break the tests, but they prevented the actual app server from starting up.

PR #936, opened 2018-08-22
Boost non-fuzzy searches relative to fuzzy searches

leonardr:
This branch partially addresses https://jira.nypl.org/browse/SIMPLY-569 by slightly reducing the importance of fuzzy searches and the situations where a fuzzy search will be employed. The goal is to make some marginal improvements before the push to generating search documents with the Python DSL and the migration to Elasticsearch 6.

`prefix_length=1` exploits the fact that people usually don't misspell the first letter of a word. Fuzzy transforms will no longer affect the first letter of a word in a search query. This means that most typoes will be handled better, but if you transpose the first and second letters you'll get much worse results than before. (In a later version of Elasticsearch I think we can add `transpositions=true` to fix this case.)

The extra boost for the main `simple_query_string` search reflects the fact that most search terms aren't misspelled. This will give search terms that show up in a book's description or subjects a slight boost over similarly-spelled words in a book's title.

I would have liked to make the boost higher, but boosting the `simple_query_string` too much ruins a search for a term like `thomas pychon`, where one word is correctly spelled but that correctly spelled word returns terrible results on its own.

So, as usual, there's a lot more work that can be done -- this just makes a slight improvement without disrupting things too much.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -312,22 +312,26 @@ class ExternalSearchIndex(object):
 
     def make_query(self, query_string):
 
-        def _boost(query, boost):
+        def _boost(query, boost=1):
             """Boost a preexisting query."""
+            if boost == 1:
+                # No boost.
+                return query
             if 'bool' in query:
                 # It's already a boolean query, just boost it.
                 query['boost'] = boost
             else:
                 # Create a new boolean query with only one clause and
                 # a boost.
-                return {
+                query = {
                     'bool': {
                         'must': query,
                         'boost': boost
                     }
                 }
+            return query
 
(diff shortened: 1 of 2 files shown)


aslagle:
For a long time my opinion on this has been that we should replace the fuzzy matching with a suggester, but it requires client work.

leonardr:
I ran some experiments and didn't notice a big difference in limiting `max_expansions`. I _did_ notice a big difference in weighting matches from the fuzzy search significantly lower than matches from the non-fuzzy searches. We've been using weighting to give big boosts to exact title and author matches; we can use a scaled down version of the same technique to boost an exact word match over a fuzzy word match.

PR #939, opened 2018-08-27
Process debug response

leonardr:
Currently if you make a debuggable request to a server that returns a 5xx error, core creates a problem detail document describing the error. But any useful information sent in the response body (such as a problem detail document describing the error on the foreign server) is lost.

This branch changes things so that this information is stored in the `debug_message` of the new problem detail document.

This also simplifies the code path of a debuggable request -- previously `process_response` was being called in addition to `process_debuggable_response`.

This is part of the fix to https://jira.nypl.org/browse/SIMPLY-1150.

### DIFF
--- a/util/http.py
+++ b/util/http.py
@@ -207,7 +207,7 @@ class HTTP(object):
         :param args: Positional arguments for the request function.
         :param kwargs: Keyword arguments for the request function.
         """
-        process_response = kwargs.pop(
+        process_response_with = kwargs.pop(
             'process_response_with', cls._process_response
         )
         allowed_response_codes = kwargs.pop('allowed_response_codes', [])
@@ -252,7 +252,7 @@ class HTTP(object):
             # a generic RequestNetworkException.
             raise RequestNetworkException(url, e.message)
 
-        return process_response(
+        return process_response_with(
             url, response, allowed_response_codes, disallowed_response_codes
         )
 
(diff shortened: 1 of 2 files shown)


util/http.py
```diff
@@ -325,31 +334,29 @@ def debuggable_post(cls, url, payload, **kwargs):
...
-            allowed_response_codes = kwargs['allowed_response_codes']
-        else:
-            # We want request_with_timeout to allow through all
-            # response codes so that we can handle bad ones in a more
-            # helpful way.
-            kwargs['allowed_response_codes']=["1xx", "2xx", "3xx", "4xx", "5xx"]
-
-            # But we want to apply the normal rules when deciding whether
-            # a given response is 'bad'.
-            allowed_response_codes = None
```
leonardr: We can remove this code because we are using `process_debuggable_response` to handle the response, where previously we were calling `request_with_timeout` (which called `process_response`, enforcing the rules of that method) and _then_ calling `process_debuggable_response`.

util/http.py
```diff
@@ -325,31 +334,29 @@ def debuggable_post(cls, url, payload, **kwargs):
...
         logging.info("Making debuggable %s request to %s: kwargs %r",
                      http_method, url, kwargs)
-        response = HTTP.request_with_timeout(http_method, url, **kwargs)
-        return cls.process_debuggable_response(response, allowed_response_codes)
```
leonardr: Now we call `process_debuggable_response` instead of `process_response`, since `process_debuggable_response` is what's passed into `_request_with_timeout`.

PR #940, opened 2018-08-27
Move custom list filtering code from Lane to WorkList

leonardr:
This branch gives WorkList the capability to find only works that are on specific `CustomLists`. Previously this functionality was only available to `Lane`. The underlying code is the same, except that `WorkList.customlist_ids` are specified up-front in `initialize`, and `Lane.customlist_ids` are kept in the database.

To get this to work I had to not only move `customlist_ids` over to WorkList, but also `list_datasource_id` and `list_seen_in_previous_days`. That's kind of overkill, but it's not a problem, since theoretically a WorkList might want to use those features.

I believe there's no reason why other things used in `Lane.bibliographic_filter_clause` like `fiction` and `license_datasource` couldn't also be moved to `WorkList` -- we just haven't needed those features in a `WorkList` yet.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -785,9 +785,9 @@ class WorkList(object):
 
     def initialize(self, library, display_name=None, genres=None,
                    audiences=None, languages=None, media=None,
-                   children=None, priority=None, entrypoints=None,
                    customlist_ids=None, list_datasource_id=None,
-                   list_seen_in_previous_days=None
+                   list_seen_in_previous_days=None,
+                   children=None, priority=None, entrypoints=None,
     ):
         """Initialize with basic data.
 
--- a/lane.py
+++ b/lane.py
@@ -785,7 +785,10 @@ class WorkList(object):
 
     def initialize(self, library, display_name=None, genres=None,
                    audiences=None, languages=None, media=None,
-                   children=None, priority=None, entrypoints=None):
+                   children=None, priority=None, entrypoints=None,
+                   customlist_ids=None, list_datasource_id=None,
+                   list_seen_in_previous_days=None
+    ):
         """Initialize with basic data.
 
         This is not a constructor, to avoid conflicts with `Lane`, an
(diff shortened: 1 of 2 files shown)


lane.py
```diff
@@ -1223,6 +1250,95 @@ def audience_filter_clauses(self, _db, qu):
             clauses.append(LicensePool.data_source_id != gutenberg.id)
         return clauses
 
+    def customlist_filter_clauses(
```
leonardr: I moved this method from `Lane` to `WorkList` without, I think, changing any code.

lane.py
```diff
@@ -2186,14 +2306,7 @@ def bibliographic_filter_clause(self, _db, qu, featured, outer_join=False):
         if self.fiction is not None:
             clauses.append(work_model.fiction==self.fiction)
 
-        if self.media:
-            clauses.append(work_model.medium.in_(self.media))
-
```
leonardr: This same code is in `WorkList.bibliographic_filter_clause`, so I don't think we need it here as well.

tests/test_lane.py
```diff
@@ -1577,6 +1633,151 @@ def for_audiences(*audiences):
                  non_gutenberg_children.id]),
             set(for_audiences()))
 
+    def test_customlist_filter_clauses(self):
```
leonardr: This test was moved from TestLane to TestWorkList and I did have to change it a little bit to create WorkLists instead of Lanes.

leonardr:
The motivating case is https://jira.nypl.org/browse/SIMPLY-459. This will allow a controller to initialize a WorkList for a specific CustomList and then pass it into `AcquisitionFeed.page` to get a paginated feed of books in that CustomList.

leonardr:
`WorkList.search()` already sends in `self.customlist_ids` into `query_works` as `on_any_of_these_lists`, so it should work automatically. But, it looks like I should maybe move `Lane._gather_customlist_ids` into `WorkList`, so that `self.customlist_ids` will return the right thing in cases where a `WorkList` is defined by  `list_datasource_id`.

leonardr:
The handling of custom lists in `Lane` is a little awkward, but I got everything working and added some missing test coverage.

PR #941, opened 2018-08-28
Request bug fixes

leonardr:
This branch fixes a couple small bugs in our HTTP client, one of which only showed up in a real scenario and one of which was due to missing test coverage.

### DIFF
--- a/tests/test_util_http.py
+++ b/tests/test_util_http.py
@@ -28,11 +28,22 @@ class TestHTTP(object):
 
     def test_request_with_timeout_success(self):
 
+        called_with = None
         def fake_200_response(*args, **kwargs):
+            # The HTTP method and URL are passed in the order
+            # requests.request would expect.
+            eq_(("GET", "http://url/"), args)
+
+            # Keyword arguments to _request_with_timeout are passed in
+            # as-is.
+            eq_("value", kwargs["kwarg"])
+
+            # A default timeout is added.
+            eq_(20, kwargs['timeout'])
             return MockRequestsResponse(200, content="Success!")
 
         response = HTTP._request_with_timeout(
-            "the url", fake_200_response, "a", "b"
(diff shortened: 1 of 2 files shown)


tests/test_util_http.py
```diff
@@ -28,11 +28,22 @@ def test_series(self):
 
     def test_request_with_timeout_success(self):
 
+        called_with = None
         def fake_200_response(*args, **kwargs):
+            # The HTTP method and URL are passed in the order
+            # requests.request would expect.
+            eq_(("GET", "http://url/"), args)
```
leonardr: This is the missing test coverage.

util/http.py
```diff
@@ -349,6 +355,7 @@ def debuggable_request(cls, http_method, url, make_request_with=None,
         """
         logging.info("Making debuggable %s request to %s: kwargs %r",
                      http_method, url, kwargs)
+        make_request_with = make_request_with or requests.request
```
leonardr: This is the problem that only showed up in real usage. When you don't mock `make_request_with`, no real value was being set.

util/http.py
```diff
@@ -236,6 +236,12 @@ def _request_with_timeout(cls, url, make_request_with, *args, **kwargs):
             if verbose:
                 logging.info("Sending %s request to %s: kwargs %r",
                              http_method, url, kwargs)
+            if len(args) == 1:
+                # requests.request takes two positional arguments,
+                # an HTTP method and a URL. In most cases, the URL
+                # gets added on here. But if you do pass in both
+                # arguments, it will still work.
+                args = args + (url,)
```
leonardr: This is the fix to the problem hidden by missing test coverage. `url` was being passed into this method and never actually being used. We were relying on `url` showing up _again_ in `*args`, which was confusing.

PR #942, opened 2018-08-30
Generate query documents with elasticsearch-dsl instead of raw dictionary manipulation

leonardr:
This branch rewrites the code that runs searches against an Elasticsearch index, so that it uses the `elasticsearch-dsl` package instead of generating big JSON documents itself. The actual search documents have not changed significantly, but the code that generates them should be much better organized, and in the future it should be much easier to add new ways of searching or tweak the way we search things.

I'd like your help with a problem with search result weighting that I can't figure out on my own. See https://github.com/NYPL-Simplified/server_core/pull/942#pullrequestreview-154314967 for the details.

Now, the details of this branch:

A search has two parts: a `Query` and a `Filter`. 

The `Query` is derived entirely from what the user typed into the search box. It's generated by the `Query.query()` method. A number of hypotheses are generated, each one of which might explain what the user's query string "really" means. A `dis_max` query is used to find the most likely hypothesis for that book, and a book's score is the best score it gets for any of those hypotheses.

The `Filter` encapsulates everything about the user's current context: the current library, the current lane, the currently selected entry point, and the languages enabled on the user's device. It's used to exclude books that match the search query but don't fit for some other reason. Right now, the main other reasons are "wrong library" and "wrong language".

The `Filter` is generated by the `Filter.from_worklist` method. This method can generate a `Filter` from any WorkList, even a really complicated one like "books that are on both the best-seller list and the staff picks list". This wasn't possible before. To get this to work I defined a bunch of methods in `WorkList` for compatibility with lane, and defined `WorkList.inherited_value` and `WorkList.inherited_values` to abstract away the process by which lane configuration is inherited.

The `SearchFacets` class is now responsible for pulling the `Accept-Language` header out of an active HTTP request and using it as a possible value for `Filter.languages`. Together, a `WorkList` and a `SearchFacets` cover every reason why a book should not show up in search results even though it matches the search query.

This branch will require a corresponding circulation branch, but it shouldn't be huge, because there's only one controller that does any searching.


opds.py
```diff
@@ -870,16 +870,22 @@ def add_breadcrumb_links(self, lane, entrypoint=None):
         self.show_current_entrypoint(entrypoint)
 
     @classmethod
-    def search(cls, _db, title, url, lane, search_engine, query, media=None, pagination=None,
-               annotator=None, languages=None, facets=None
```
leonardr: It's no longer necessary to pass in `media` or `languages`, because that's now handled by either `lane` or `facets`.

testing.py
```diff
@@ -57,7 +57,7 @@
     WorkCoverageProvider,
 )
 
-from external_search import DummyExternalSearchIndex
+from external_search import MockExternalSearchIndex
```
leonardr: We use 'Mock' elsewhere in the application for this type of class, and this branch is a good time to change the usage here.

lane.py
```diff
@@ -638,6 +662,58 @@ def quality_tier_field(self):
...
+        # restrictions will use the languages defined in the
+        # Accept-Language header used by the client.
+        language_header = get_header("Accept-Language")
+        languages = None
+        if language_header:
+            languages = parse_header(language_header)
+            languages = map(str, languages)
+            languages = map(LanguageCodes.iso_639_2_for_locale, languages)
+            languages = [l for l in languages if l]
+        languages = languages or None
```
leonardr: This code is copied in from circulation; my circulation branch will remove it.

external_search.py
```diff
@@ -268,375 +276,59 @@ def base_index_name(self, index_or_alias):
 
         return base_works_index
 
-    def query_works(self, library, query_string, media, languages, fiction, audiences,
-                    target_age, in_any_of_these_genres=[], on_any_of_these_lists=None, fields=None, size=30, offset=0):
```
leonardr: All this filtering information (in_any_of_these_genres, etc.) now lives in the Filter object.

external_search.py
```diff
@@ -268,375 +276,59 @@ def base_index_name(self, index_or_alias):
...
-        )
-        if fields is not None:
-            search_args['fields'] = fields
-        # search_args['explain'] = True
-        # print "Args looks like: %r" % search_args
-        results = self.search(**search_args)
-        # print "Results: %r" % results
-        return results
-
-    def make_query(self, query_string):
```
leonardr: The major accomplishment of this branch is splitting up `make_query` into smaller, independently tested methods.

external_search.py
```diff
@@ -882,8 +574,738 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+        # standard searchable fields.
+        fuzzy = self.fuzzy_string_query(query_string)
+        self._hypothesize(hypotheses, fuzzy, 1)
+
+        # The query string might contain some specific field matches
+        # (e.g. a genre name or target age), with the remainder being
+        # the "real" query string.
+        with_field_matches = self._parsed_query_matches(query_string)
+        self._hypothesize(
+            hypotheses, with_field_matches, 20, all_must_match=True
```
leonardr: This weight of 20 was originally a weight of 200. Weighting this hypothesis the same as the _match_phrase hypotheses above causes "Gumby in Love" to show up before "Modern Romance" in a search for `modern romance`. Unfortunately, the way it's weighted now, "Modern Romance" shows up before "Gumby in Love" in a search for `romance`, which isn't a disaster but this wasn't originally the case. This is the one thing about this branch I haven't been able to resolve to my satisfaction -- I don't see any significant differences in the generated Elasticsearch documents.

external_search.py
```diff
@@ -882,8 +577,745 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+        v = worklist.inherited_value
+        media = v('media')
+        languages = v('languages')
+        fiction = v('fiction')
+        audiences = v('audiences')
+        target_age = v('target_age')
+
+        # For genre IDs and CustomList IDs, we might get a separate
+        # set of restrictions from every item in the WorkList hierarchy.
+        # _All_ restrictions must be met for a work to match the filter.
```
leonardr: This is the new bit. The filter may end up with multiple sets of restrictions on `genre_ids` or `customlist_ids`, and the way filters work, every restriction must be met for the book to show up. My goal is to use this to replace the multiple aliased joins against `Genre` and `CustomListEntry` in `WorkList.bibliographic_filter_clause` and `WorkList.customlist_filter_clause`.

tests/test_external_search.py
```diff
@@ -396,490 +453,275 @@ def test_query_works(self):
...
+        for q in ('grade 4', 'grade 4-6', 'age 9'):
+            # ages 9-10 is a better result because a book targeted
+            # toward a narrow range is a better match than a book
+            # targeted toward a wide range.
+            expect([self.age_9_10, self.age_2_10], q)
 
+        # TODO: I expected this to act like the other queries, but the
+        # 2-10 book shows up above the 9-10 book. This might indicate
+        # a bug.
+        expect([self.age_9_10, self.age_2_10], "age 10-12", ordered=False)
```
leonardr: I'm pretty sure this behavior is not a change from the previous behavior, but not 100%.

tests/test_external_search.py
```diff
@@ -396,490 +453,275 @@ def test_query_works(self):
...
-        # Test searching across collections.
-        #
 
-        # If we add the missing collection to the default library, "A
-        # Tiny Book" starts showing up in searches against that
-        # library.
-        self._default_library.collections.append(self.tiny_collection)
-        results = query("a tiny book", None, None, None, None, None, None, None)
-        hits = results["hits"]["hits"]
-        eq_(1, len(hits))
```
leonardr: A lot of this test code could be simplified because we "search the current library" by setting an appropriate value for `Filter.collection_ids` when creating the `Filter` object. It's no longer any special code run during the search itself.

external_search.py
```diff
@@ -268,375 +279,59 @@ def base_index_name(self, index_or_alias):
...
+        results = search[start:stop]
+        if debug:
+            b = time.time()
+            self.log.info("Elasticsearch query completed in %.2fsec", b-a)
+            for i, result in enumerate(results):
+                self.log.info(
+                    '%02d "%s" (%s) work=%s score=%.3f',
+                    i, result.title, result.author, result.meta['id'],
+                    result.meta['score']
+                )
```
aslagle: You could try `search.extra(explain=True)` for debug mode - this adds extra info about how the score for each item was computed and it's very useful for debugging. It's much too verbose to log though.

external_search.py
```diff
@@ -882,8 +577,748 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+        "away", "stay",
+    ]
+
+    # If this regular expression matches a query, we will not run
+    # a fuzzy match against that query, because it's likely to be
+    # counterproductive.
+    #
+    # TODO: Instead of this, avoid the fuzzy query or weigh it much
+    # lower if there don't appear to be any misspelled words in the
+    # query string.
```
aslagle: TODO: switch to a suggester

external_search.py
```diff
@@ -882,8 +577,748 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+
+        # The query string might be an exact match for title or
+        # author. Such a match would be boosted quite a lot.
+        self._hypothesize(
+            hypotheses,
+            self._match_phrase("title.standard", query_string), 200
+        )
+        self._hypothesize(
+            hypotheses,
+            self._match_phrase("author.standard", query_string), 200
```
aslagle: `title.standard` exists because we define it in the mapping (https://github.com/NYPL-Simplified/server_core/blob/b8023d3dc2f7399dd9b875593081799c8e086a8c/external_search.py#L502), but for `author` the standard analyzer is the default so you don't need the `.standard`, and actually I'm surprised if it works.

external_search.py
```diff
@@ -882,8 +577,748 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+        # original text.
+        'author^6',
+        'publisher',
+        'imprint'
+    ]
+
+    # When we look for a close match against title, author, or series,
+    # we apply minimal stemming, because we're handling the case where
+    # the user typed something in exactly as is.
+    MINIMAL_STEMMING_QUERY_FIELDS = [
```
aslagle: This is not quite accurate, author has no stemming rather than minimal stemming. I believe minimal stemming only removes plurals, and our minimal analyzer also removes stop words. If you really want exact matches, you should use the standard analyzer rather than minimal.

external_search.py
aslagle: I see this was in the old code too - I guess it does work.

external_search.py
```diff
@@ -882,8 +577,748 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+        # call Query() recursively.
+        #
+        # However, someone who does this kind of search is
+        # probably not looking for a specific book. They might be
+        # looking for an author (eg, 'science fiction iain
+        # banks'). But they're most likely searching for a _type_
+        # of book, which means a match against summary or subject
+        # ('asteroids') would be the most useful.
+        if (self.final_query_string
+            and self.final_query_string != self.original_query_string):
```
aslagle: it's possible these could be different because only the final string is stripped.

external_search.py
```diff
@@ -882,8 +577,748 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+            word_boundary_pattern % match.strip(), re.IGNORECASE
+        ).sub("", query_string)
+
+
+class Filter(SearchBase):
+    """A filter for search results.
+
+    This covers every reason you might want to not show a search
+    result that matches the query string -- wrong media, wrong
+    language, not available in the patron's library, etc.
```
aslagle: This is true but could be misleading - elasticsearch applies the filters first to reduce the number of items that have to be scored with the query, rather than finding all matches and then filtering them.

external_search.py
```diff
@@ -882,8 +577,748 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+        media = v('media')
+        languages = v('languages')
+        fiction = v('fiction')
+        audiences = v('audiences')
+        target_age = v('target_age')
+
+        # For genre IDs and CustomList IDs, we might get a separate
+        # set of restrictions from every item in the WorkList hierarchy.
+        # _All_ restrictions must be met for a work to match the filter.
+        v = worklist.inherited_values
```
aslagle: you defined `v` twice.

aslagle: and this `v` is different than the other `v`?

external_search.py
```diff
@@ -882,8 +577,748 @@ def create_new_version(cls, search_client, base_index_name, version=None):
...
+        if target_age:
+            if isinstance(target_age, int):
+                self.target_age = (target_age, target_age)
+            elif isinstance(target_age, tuple) and len(target_age) == 2:
+                self.target_age = target_age
+            else:
+                # It's a SQLAlchemy range object. Convert it to a tuple.
+                #
+                # NOTE: This assumes that the range is inclusive
+                # on both sides.
```
aslagle: Could you use `numeric_range_to_tuple` from model.py to remove this assumption? I recently fixed a bug due to inclusive ranges becoming exclusive on commit, so I wouldn't trust this.

lane.py
aslagle: I know this was existing functionality, but while we're here, this is a problem in the admin interface. I've gotten complaints that admins can't search for non-english books.

tests/test_external_search.py
```diff
@@ -396,490 +453,275 @@ def test_query_works(self):
...
-        eq_(unicode(self.age_9_10.id), hits[0]['_id'])
+        expect([self.moby_dick, self.moby_duck], "mo by dick")
 
-        results = query("age 10-12", None, None, None, None, None, None, None)
-        hits = results["hits"]["hits"]
-        eq_(2, len(hits))
-        eq_(unicode(self.age_9_10.id), hits[0]['_id'])
+        # A query without an apostrophe matches a word that contains one.
+        # (NOTE: it's not clear whether this is a feature of the index or
+        # done by the fuzzy match.)
```
aslagle: I can't remember for sure if the stemmer does this, but it's likely handled by both.

tests/test_external_search.py
```diff
@@ -396,490 +453,275 @@ def test_query_works(self):
...
+        expect(self.tess, "durbervilles")
+        expect(self.tiffany, "tiffanys")
 
+        # A query with an 'e' matches a word that contains an
+        # e-with-acute. (NOTE: it's not clear whether this is a
+        # feature of the index or done by the fuzzy match.)
```
aslagle: This is the "asciifolding" filter in the analyzers.

tests/test_external_search.py
aslagle: I don't think it's a change. The target age query only scores how big the overlap is, it doesn't look at how large the non-overlapping part of the range is. That could be a good thing to add.

tests/test_external_search.py
```diff
@@ -1018,283 +838,978 @@ def expect_ids(works, *query_args):
...
-        collection_filter, medium_filter = filter['and']
-        expect = dict(terms=dict(medium=[Edition.AUDIO_MEDIUM.lower()]))
-        eq_(expect, medium_filter)
 
-    def test_query_works_from_lane_definition_handles_age_range(self):
-        search = DummyExternalSearchIndex()
+        assert_parses_as(
+            'divorce ages 10 and up',
+            ((10, 14), 40),
+            'divorce  and up' # TODO: not ideal
```
aslagle: yeah that's not great - did this work before?

aslagle: also how did it get 14?

external_search.py
aslagle: When I run the tests locally with the old code I get a couple failures, and one is that the `modern romance` test puts Gumby first. I think it's one of those things where tf-idf is computed within the shard so if they get indexed in different shards it doesn't work. It's possible that's happening to you as well. That's why you really need to test with a realistic index - these small tests are nice but don't tell you much.

But when I run it locally with your branch it finds 50 results - I wonder if it's using the wrong index.

aslagle: It looks like the way you're creating the Search doesn't tell it what index to use, so it's searching across every index I have.

lane.py
```diff
@@ -118,17 +124,24 @@ def navigate(self, entrypoint):
...
     @classmethod
     def from_request(
-            cls, library, facet_config, get_argument, worklist, **extra_kwargs
+            cls, library, facet_config, get_argument, get_header, worklist,
+            **extra_kwargs
     ):
         """Load a faceting object from an HTTP request.
 
+        # TODO: Is the 'library' argument redundant?
+
```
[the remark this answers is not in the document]
aslagle: I think it's not, because certain feeds need a custom facet setup instead of using the library's defaults (e.g. a series feed can order by series position).

lane.py
leonardr: I filed https://jira.nypl.org/browse/SIMPLY-1213 to deal with this.

external_search.py
leonardr: I've fixed the index problem.

My local Elasticsearch and the one set up by Travis both have `number_of_shards:1`, but I'm comparing the results here and they have different values for `.meta.shard`.  I've added this to the debug portion of the code so it'll show up on Travis if the same problem exists there.

external_search.py
leonardr: It wasn't working! Exact author matches weren't being boosted. That's why in `test_exact_matches`, a book whose title was "He Is Peter Graves" (i.e. "Peter Graves" plus some other stuff) was showing up before a book whose author was the literal string "Peter Graves".

I assumed that outcome was consistent with both results being boosted, but if that was the case the closer match should have come out on top.

external_search.py
leonardr: I'll rename the variables to make this less confusing.

external_search.py
leonardr: Good idea.

leonardr: Wow, I think I uncovered a bug in `numericrange_to_tuple`.

lane.py
```diff
@@ -1311,80 +1483,41 @@ def search_target(self):
...
-                dict(
-                    library=self.get_library(_db),
-                    query_string=query,
-                    fields=["_id", "title", "author", "license_pool_id"],
-                    size=pagination.size,
-                    offset=pagination.offset,
-                )
+        except elasticsearch.exceptions.ElasticsearchException, e:
+            logging.error(
+                "Problem communicating with ElasticSearch. Returning empty list of search results.",
```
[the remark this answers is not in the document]
leonardr: Agreed, I filed https://jira.nypl.org/browse/SIMPLY-1214 to deal with this.

tests/test_external_search.py
leonardr: 14 is YOUNG_ADULT_AGE_CUTOFF, so default_audience_for_target_range interprets "10 and up" as "10-14".

I don't think the "and up" part was chopped out before.

external_search.py
```diff
@@ -268,375 +280,64 @@ def base_index_name(self, index_or_alias):
...
+        stop = start + pagination.size
+
+        a = time.time()
+        # NOTE: This is the code that actually executes the ElasticSearch
+        # request.
+        results = search[start:stop]
+        if debug:
+            b = time.time()
+            self.log.info("Elasticsearch query completed in %.2fsec", b-a)
+            shards = set()
```
aslagle: you didn't use this - were you going to keep track of all the shards?

leonardr: Yeah, but it was easier to just print out the shard for each result.

PR #943, opened 2018-09-12
Refactor model.py into multiple files (old)

adriana-alter:


aslagle:
This work seems like it would have high potential for merge conflicts if anyone else is working on the model, and each module should be able to work before the rest is done, so I'd rather see frequent merges to master than a big refactoring branch. +1 to releasing manage patrons first though.

PR #948, opened 2018-09-12
Refactor model.py into multiple files

leonardr:
This branch converts `model.py`, a huge monolithic file, into `model/`, a package containing a large number of files. Similarly for `tests/test_model.py`, which becomes `tests/models/`.

The major change that had to be made to support this is: we can no longer write code like this from the `tests` package or the `model` package:

```
from config import Configuration
```

Instead we have to do a relative import:

```
from ..config import Configuration
```

Or even (if we're in tests/models):

```
from ...config import Configuration
```

Previously we avoided this by adding the top-level module to `sys.path` when running tests. This caused problems in circulation, though. When we wrote code like this in circulation, we ended up with a class called `core.model.LicensePool`:

```
from core.model import LicensePool
```

But that circulation code could cause code like this to run within the model package, whereupon we end up with a class called `model.Edition`:

```
from model import Edition
```

This is what Nick Coghlan refers to as [the double import trap](https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-double-import-trap). SQLAlchemy was continually checking things like `if core.model.Edition is model.Edition`, and all of those checks were failing.

Instituting a "don't mess with `sys.path`" rule meant we had to change a lot of our imports inside core, but once that change was made, the changes necessary to circulation and metadata were very small.

Using a relative import ensures that when we're running the core tests we always have a `model.LicensePool` object, and when we're running circulation or metadata code we always have a `core.model.LicensePool` object.

Other changes:

* Test setup reliably destroys and recreates the entire database schema, including the materialized views.
* The `MaterializedWorkWithGenre` class now has a schema defined up front rather than obtained by looking at a real materialized view at runtime. This simplifies startup and lets us import `MaterializedWorkWithGenre` at the same time we import, say, `LicensePool`, instead of having to wait for later.


leonardr:
I was distressed to learn that renaming model.py and moving test_model.py loses the Git history, but there seems to be no way around it, and the history will be destroyed soon enough when we move everything out of those files.

model/cached_feed.py
```diff
@@ -0,0 +1,201 @@
...
+)
+from sqlalchemy import (
+    Column,
+    Date,
+    DateTime,
+    ForeignKey,
+    Index,
+    Integer,
+    Unicode,
+    UniqueConstraint,
```
leonardr: A lot of the time these basic imports aren't necessary -- they're used in the big model.py but not in these smaller files. Since this will happen many tiemes, the simplest way to fix this sort of problem will be to run this code through pylint afterwards.

model/hyperlink_constants.py
```diff
@@ -0,0 +1,25 @@
+# encoding: utf-8
+
+class HyperlinkConstants(object):
```
leonardr: Let's call this `LinkRelations` -- it's easier to remember and it's a good match with `MediaTypes`.

leonardr:
We're having some problems getting this code to work in both core and circulation. The underlying issue is that some classes get imported through two different paths (e.g. core.model.LicensePool and model.LicensePool), or possibly even more than two paths. For most classes this doesn't matter, but it matters for data model classes, because SQLAlchemy uses class identity to understand the relationships between data model objects. And it matters for the Configuration class, which includes a class variable that must be explicitly initialized -- importing it through a different path means it's technically a different class and the variable isn't initialized.

This has been an occasional problem in the past, but changing model from a module to a package has blown it wide open.

After studying the problem for a while, I think the simplest solution that will work is to change all import statements that mention `model`, `lane`, or `config` to use absolute imports (`core.model`, `core.lane`, `core.config`). (`lane` also contains data model objects, they're just not in `model/` yet). This is a 'bite the bullet' type solution, because we were hoping to avoid this, but it does work and 

I came up with an interesting way to make Configuration work no matter how many ways it's imported, but interesting solutions to peripheral problems can wait for another branch.

leonardr:
The reason this is a bite-the-bullet solution is that it's confusing to have a mix of relative and absolute imports in core. Consistency would demand that we always use one or the other. So we'd have to change _all_ the import statements in core. And I don't think we can make it so relative imports cause obvious failures, so we don't have a good way of enforcing this.

leonardr:
Here's my PR for changing the necessary imports: 
https://github.com/NYPL-Simplified/server_core/pull/952/

I still hold out hope for a better solution, so I didn't push to this branch.

I absolute-ized the imports for entrypoint and classifier because they are 'polluted' by being used in lane.py. Presumably other modules like facets have the same problem, but we don't do identity comparison on those classes (at least in core).

leonardr:
I'm also trying a branch that completely gets rid of absolute imports:
https://github.com/NYPL-Simplified/server_core/tree/no-absolute-imports

This causes a number of problems I don't understand, but if we can get it to work it will be a much better starting point.

analytics.py
```diff
@@ -23,8 +24,17 @@ def __init__(self, _db):
         integrations = _db.query(ExternalIntegration).filter(ExternalIntegration.goal==ExternalIntegration.ANALYTICS_GOAL)
         # Turn each integration into an analytics provider.
         for integration in integrations:
+            kwargs = {}
+            module = integration.protocol
```
aslagle: I think I was the one who set this up originally and my reason for using the module as the protocol was that I thought it would be nice if someone could write a module and use it by adding it to the python path, without needing to get it merged into the main repo. We didn't follow this pattern for any other integrations, and now that we have the admin interface and docker, it doesn't make much sense. I think we should ditch this approach and make analytics work more like other integrations.

config.py
```diff
@@ -288,8 +290,6 @@ def cdns_loaded_from_database(cls):
 
     @classmethod
     def get(cls, key, default=None):
-        if cls.instance is None:
-            raise ValueError("No configuration object loaded!")
```
aslagle: This error seems like it could still be useful, in case we break the code that loads the configuration.

model/__init__.py
```diff
@@ -0,0 +1,533 @@
...
+    return "%s-%s" % (lower,upper)
+
+def numericrange_to_tuple(r):
+    """Helper method to normalize NumericRange into a tuple."""
+    if r is None:
+        return (None, None)
+    lower = r.lower
+    upper = r.upper
+    if lower and not r.lower_inc:
+        lower -= 1
```
aslagle: @leonardr wasn't this a bug you fixed?

model/__init__.py
```diff
@@ -0,0 +1,533 @@
...
+from patron import (
+    Annotation,
+    Hold,
+    Loan,
+    LoanAndHoldMixin,
+    Patron,
+    PatronProfileStorage,
+)
+import listeners
+from listeners import *
```
aslagle: Why is this one imported two ways?

model/__init__.py
```diff
@@ -0,0 +1,533 @@
...
+    # with the right arguments.
+    from ..log import LogConfiguration
+    LogConfiguration.initialize(_db)
+    return _db
+
+from admin import (
+    Admin,
+    AdminRole,
+)
+from background import (
```
aslagle: I don't get the name of this module. Did we not use 'coverage' because we already have a coverage.py?

model/constants.py
```diff
@@ -0,0 +1,260 @@
+# encoding: utf-8
+# DataSourceConstants, EditionConstants, IdentifierConstants, LinkRelations,
+# MediaTypes
```
aslagle: It's not clear why some constants were moved here and others weren't. Are these the most widely used outside their own classes?

model/constants.py
```diff
@@ -0,0 +1,260 @@
...
+    URN_SCHEME_PREFIX = "urn:librarysimplified.org/terms/id/"
+    ISBN_URN_SCHEME_PREFIX = "urn:isbn:"
+    GUTENBERG_URN_SCHEME_PREFIX = "http://www.gutenberg.org/ebooks/"
+    GUTENBERG_URN_SCHEME_RE = re.compile(
+        GUTENBERG_URN_SCHEME_PREFIX + "([0-9]+)")
+    OTHER_URN_SCHEME_PREFIX = "urn:"
+
+    IDEAL_COVER_ASPECT_RATIO = 2.0/3
+    IDEAL_IMAGE_HEIGHT = 240
+    IDEAL_IMAGE_WIDTH = 160
```
aslagle: I'm not sure what uses these ones but I never noticed them when I was working on covers. Seems like a weird place for them...

model/work.py
```diff
@@ -0,0 +1,1603 @@
...
+Index("ix_works_audience_fiction_quality_random", Work.audience, Work.fiction, Work.quality, Work.random)
+
+
+class BaseMaterializedWork(object):
+    """A mixin class for materialized views that incorporate Work and Edition."""
+    pass
+
+
+class MaterializedWorkWithGenre(Base, BaseMaterializedWork):
+    p = dict(primary_key=True)
```
aslagle: We no longer need `index=True`?

tests/test_appeal.py
```diff
@@ -13,7 +13,7 @@
     DatabaseTest,
```
aslagle: This file seems like it should be under model.

tests/test_measurement.py
```diff
@@ -6,7 +6,7 @@
     set_trace,
```
aslagle: This also seems like it should be under model.

model/work.py
leonardr: I believe `index=True` is only used by SQLAlchemy when initially creating the table. Since SQLAlchemy doesn't create the materialized view, it's redundant.

We do need the `primary_key=True` stuff because that's how SQLAlchemy knows how to look up an item in this table.

model/constants.py
leonardr: Yeah, these are the ones where they had to be in a separate module to avoid circular imports. I agree that we should move all the constants in here eventually.

model/constants.py
leonardr: Yes, this is weird -- they're here because Identifier is where we find the best cover image for a given identifier. This can also be resolved in a 'fix the constants' branch.

model/__init__.py
leonardr: That's right, but we can change it now that we're not messing with `sys.path` anymore. There's no chance of confusing `coverage` with `model.coverage`

model/__init__.py
leonardr: Yeah, but I fixed it as a side thing in https://github.com/NYPL-Simplified/server_core/pull/942, which hasn't been merged yet. (Looking at that PR, I see I also broke out the list of media out of the definition of Edition.medium two different ways.) I'll resolve it in whichever branch gets merged second.

analytics.py
leonardr: I filed https://jira.nypl.org/browse/SIMPLY-1253 as a follow-up ticket.

model/constants.py
leonardr: I filed https://jira.nypl.org/browse/SIMPLY-1252 as a follow-up ticket.

PR #950, opened 2018-09-17
Customlists feed builder

EdwinGuzman:
Moved code from [Circ #1043](https://github.com/NYPL-Simplified/circulation/pull/1043) to here. I'm a bit concerned about all the variables I had to pass to make this work, but it isolates the work to generate the OPDS feed from a query specifically for custom lists in this case.


opds.py
```diff
@@ -747,6 +747,26 @@ def make_link(ep):
...
+        page_of_works = pagination.apply(query)
+
+        feed = cls(_db, list.name, url, page_of_works, annotator)
+
+        if len(list.entries) > 0 and pagination.has_next_page:
+            OPDSFeed.add_link_to_feed(feed=feed.feed, rel="next", href=url_fn("custom_list", after=pagination.next_page.offset, library_short_name=library.short_name, list_id=list.id))
+        if pagination.offset > 0:
+            OPDSFeed.add_link_to_feed(feed=feed.feed, rel="first", href=url_fn("custom_list", after=pagination.first_page.offset, library_short_name=library.short_name, list_id=list.id))
+        if pagination.previous_page:
+            OPDSFeed.add_link_to_feed(feed=feed.feed, rel="previous", href=url_fn("custom_list", after=pagination.previous_page.offset, library_short_name=library.short_name, list_id=list.id))
```
leonardr: For this to be a general method, `list` and `library` need to go. You need to explicitly pass in the name of the feed, and you can't have any assumptions about the name of the controller (currently it's always "custom_list") or the arguments passed into the controller (currently library_short_name and list_id are controller-specific).

I can think of a couple good solutions. The first is to replace `library` and `list` with some more general arguments.

```
def from_query(cls, _db, query, worklist, pagination, feed_name, feed_url, annotator, url_fn, controller_name, **url_kwargs):
```

In this case you'd call it like this:

```
from_query(_db, query, lane, pagination, list.name, url, annotator, url_for, "custom_list", library_short_name=library.short_name, list_id=list.id)
```

Another possibility is to change `url_fn` so it handles everything about generating the URL. It would be expected to take a Pagination object and generate the appropriate URL:

```
def from_query(cls, _db, query, lane, feed_name, feed_url, pagination, url_fn, annotator):
    ...
    OPDSFeed.add_link_to_feed(feed=feed.feed, rel="next", href=url_fn(pagination.next_page))
```

In this case you would define url_fn like this:

```
def url_for_custom_list(pagination):
    return url_for("custom_list", after=pagination.offset, library_short_name=library.short_name, list_id=list.id)
```

Then you'd pass that function in as `url_fn`.

The third option that comes to mind is to do this with a new type of `Annotator`. It's the job of `Annotator.annotate_feed` to stick extra links onto a feed, and that's basically all that's happening here.

So subclass `Annotator` (or whatever `Annotator` subclass you're using now) and change it to add those three links. Most of the code in this method can go into there, and what's left is basically a call to the AcquisitionFeed constructor. So you might not need this method at all.

opds.py
```diff
@@ -747,6 +747,26 @@ def make_link(ep):
...
+    @classmethod
+    def from_query(cls, query, _db, library, lane, list, url, pagination, url_fn, annotator):
+        """Build  a feed representing one page of a given list. Currently used for
+        creating an OPDS feed for a custom list and not cached.
+        """
+        page_of_works = pagination.apply(query)
+
+        feed = cls(_db, list.name, url, page_of_works, annotator)
+
+        if len(list.entries) > 0 and pagination.has_next_page:
```
leonardr: `Pagination.has_next_page` only works if `.total_size` or `.this_page_size` has been set on the Pagination object. It's your responsibility to set those values -- otherwise `has_next_page` will always return True.

Something like this should work:

```
page_of_works = pagination.apply(query).all()
pagination.this_page_size = len(page_of_works)
```

aslagle: +1, I noticed this bug when I was testing this.

Could you do `pagination.apply(query).count()` though?

EdwinGuzman: Thanks, `pagination.total_size = query.count()` seems to give me the more consistent result that's needed.

leonardr (approved):

aslagle (approved):

PR #951, opened 2018-09-20
Search code that works on both ES1 and ES6

leonardr:
This branch modifies the search code so that it does the same thing whether the Elasticsearch DSL library is version 2.x (for Elasticsearch 1) or version 6.x (for Elasticsearch 6).

However, both versions of the library can't be installed in a single virtual environment. So we can't have a release that supports both ES1 and ES6, unless we do something wild like include a copy of the ES1 library in the code tree, while using pip to install the ES6 library.

This branch still uses ES1, but every point where ES6 differs from ES1 in an incompatible way (whether it's the code we use or the results that are returned from a search) is now clearly labelled.


external_search.py
```diff
@@ -217,23 +231,26 @@ def _set_works_alias(name):
             # Work against the index instead of an alias.
             _set_works_alias(self.works_index)
```
aslagle: This line needs updating

aslagle:
This looks good but seems like there is still some work to understand the scoring differences between the versions.

PR #953, opened 2018-09-21
No absolute imports

leonardr:
This branch makes https://github.com/NYPL-Simplified/server_core/pull/948 work as a standalone core branch and as a component of an app server like circulation or metadata.

These are the main changes I made:

* The equivalent of https://github.com/NYPL-Simplified/server_core/pull/954
* All imports from 'core.x' are changed to imports from 'x'
* The test setup no longer modifies sys.path. This has made core tests easier to write for a long time, but it caused [the double import trap](https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-double-import-trap). The double import trap wasn't a big problem most of the time, but 948 puts all of the data model objects in their own files and then imports them from `model`. The resulting objects are known as `model.Foo` within core but `core.model.Foo` within circulation or metadata.
So basically, imports from 'x' are changed to imports from '..x' in tests.
* MaterializedWorkWithGenre is now defined with the rest of the other data model objects, rather than being defined by slurping the schema out of the database as soon as the database connection is initiated. This means we have to explicitly define every field in the materialized view in Python code as well as in SQL.

### DIFF
--- a/analytics.py
+++ b/analytics.py
@@ -25,11 +25,18 @@ class Analytics(object):
         # Turn each integration into an analytics provider.
         for integration in integrations:
             try:
-                # Import the module relative to the package this
-                # module is in.
-                provider_module = importlib.import_module(
-                    ".." + integration.protocol, package=__name__
-                )
+                try:
+                    # Import the module name as-is, relying on sys.path
+                    # to find it.
+                    provider_module = importlib.import_module(
+                        integration.protocol
+                    )
+                except ImportException, e:
+                    # Import the module relative to the package this
+                    # module is in.
+                    provider_module = importlib.import_module(
+                        '..' + integration.protocol, package=__name__
+                    )
                 provider_class = getattr(provider_module, "Provider", None)
                 if provider_class:
                     if not integration.libraries:
--- a/analytics.py
+++ b/analytics.py
@@ -4,7 +4,7 @@ import contextlib
 import datetime
 from collections import defaultdict
 from model import ExternalIntegration
-from core.config import CannotLoadConfiguration
+from config import CannotLoadConfiguration
 from sqlalchemy.orm.session import Session
 
 class Analytics(object):
--- a/analytics.py
+++ b/analytics.py
@@ -31,7 +31,7 @@ class Analytics(object):
                     provider_module = importlib.import_module(
                         integration.protocol
                     )
-                except ImportException, e:
+                except ImportError, e:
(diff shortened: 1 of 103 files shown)


model/__init__.py
```diff
@@ -335,73 +340,66 @@ def initialize(cls, url, create_materialized_work_class=True):
...
-                )
-                license_pool = relationship(
-                    LicensePool,
-                    primaryjoin="LicensePool.id==MaterializedWorkWithGenre.license_pool_id",
-                    foreign_keys=LicensePool.id, lazy='joined', uselist=False)
-
-            globals()['MaterializedWorkWithGenre'] = MaterializedWorkWithGenre
-            import core.model
-            core.model.MaterializedWorkWithGenre = MaterializedWorkWithGenre
-
```
leonardr: This chunk is no longer necessary because MaterializedWorkWithGenre is defined at compile time, in model/works.py

app_server.py
```diff
@@ -49,9 +49,9 @@ def cdn_url_for(*args, **kwargs):
 
 def load_lending_policy(policy):
     if not policy:
-        logging.info("No lending policy.")
         return {}
     if isinstance(policy, basestring):
+        logging.info("Lending policy: %s" % policy)
```
leonardr: This is an unrelated change that's been bugging me for a long time. Not having a lending policy isn't interesting; it's having one that's interesting.

leonardr: I put this into a separate branch since it's totally unrelated.

analytics.py
```diff
@@ -24,7 +25,18 @@ def __init__(self, _db):
         # Turn each integration into an analytics provider.
         for integration in integrations:
             try:
-                provider_module = importlib.import_module(integration.protocol)
+                try:
+                    # Import the module name as-is, relying on sys.path
+                    # to find it.
+                    provider_module = importlib.import_module(
+                        integration.protocol
+                    )
```
leonardr: This strategy works within the applications (circulation and metadata).

analytics.py
```diff
@@ -24,7 +25,18 @@ def __init__(self, _db):
...
+                except ImportException, e:
+                    # Import the module relative to the package this
+                    # module is in.
+                    provider_module = importlib.import_module(
+                        '..' + integration.protocol, package=__name__
+                    )
```
leonardr: This strategy works within core itself.

PR #956, opened 2018-09-26
Excluded audio sources

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-1144 by setting up a default list of data sources whose audiobooks should not show up in lanes. The default list will change as SimplyE and the circulation manager add support for more audiobook vendors, but it's stored as a site-wide configuration setting so you can explicitly set it to whatever you want.

The currently excluded data sources are these:

* Overdrive - No SimplyE support, no circ manager support
* Axis 360 - SimplyE support, no circ manager support
* RBdigital - No SimplyE support, circ manager support

The exclusion happens in `Collection.restrict_to_ready_deliverable_works`, which is the lowest-level opportunity we have to say "no, don't include that". So every piece of code that grabs books from the materialized view should respect it -- it's on the same level as "we don't own any licenses for this book".

### DIFF
--- a/config.py
+++ b/config.py
@@ -199,6 +199,7 @@ class Configuration(object):
         },
         {
             "key": EXCLUDED_AUDIO_DATA_SOURCES,
+            "label": _("Excluded audiobook sources"),
             "description": _("Audiobooks from these data sources will be hidden from the collection, even if they would otherwise show up as available."),
             "default": None,
         },
@@ -274,7 +275,7 @@ class Configuration(object):
         """Load configuration information from the filesystem, and
         (optionally) from the database.
         """
-        cls.instance = cls.load_from_file(_db)
+        cls.instance = cls.load_from_file()
         if _db:
             # Only do the database portion of the work if
             # a database connection was provided.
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -267,7 +275,7 @@ def load(cls, _db=None):
         """Load configuration information from the filesystem, and
         (optionally) from the database.
         """
-        cls.instance = cls.load_from_file(_db)
+        cls.instance = cls.load_from_file()
```
leonardr: An unrelated change, fallout from my earlier branch that changed the way `Configuration` loading worked.

model.py
```diff
@@ -11818,10 +11857,11 @@ def restrict_to_ready_deliverable_works(
                 or_(LicensePool.licenses_owned > 0, LicensePool.open_access)
         )
 
-        # Only find books in an appropriate collection.
-        query = query.filter(
-            LicensePool.collection_id.in_(collection_ids)
-        )
+        if collection_ids is not None:
```
aslagle: This seems like a significant and unrelated change if there's actually a case when `collection_ids` is None.

aslagle (approved):

PR #957, opened 2018-09-26
Remove records of patrons whose library cards have long expired

adriana-alter:
This addresses https://jira.nypl.org/browse/SIMPLY-1256, and is the first part of work to address the more general problem https://jira.nypl.org/browse/SIMPLY-1197.


model/__init__.py
```diff
@@ -659,13 +659,13 @@ class Patron(Base):
...
-    holds = relationship('Hold', backref='patron')
+    loans = relationship('Loan', backref='patron', cascade='delete')
+    holds = relationship('Hold', backref='patron', cascade='delete')
 
-    annotations = relationship('Annotation', backref='patron', order_by="desc(Annotation.timestamp)")
+    annotations = relationship('Annotation', backref='patron', order_by="desc(Annotation.timestamp)", cascade='delete')
 
     # One Patron can have many associated Credentials.
-    credentials = relationship("Credential", backref="patron")
+    credentials = relationship("Credential", backref="patron", cascade="delete")
```
leonardr: I checked that this is right, and I think you could slightly improve these by changing to `cascade="delete-orphan"`. This will delete (e.g.) a `Loan` if its `patron` is set to None. (Which shouldn't happen the way we wrote the code.)

monitor.py
```diff
@@ -701,3 +702,10 @@ class CredentialReaper(ReaperMonitor):
     TIMESTAMP_FIELD = 'expires'
     MAX_AGE = 1
 ReaperMonitor.REGISTRY.append(CredentialReaper)
+
+class PatronReaper(ReaperMonitor):
+    """Remove patron records that expired more than 30 days ago"""
+    MODEL_CLASS = Patron
+    TIMESTAMP_FIELD = 'authorization_expires'
+    MAX_AGE = 30
```
aslagle: Is this new or did it already exist? I would make it longer in case someone has holds but doesn't get around to renewing their card for a while.

model/__init__.py
leonardr: Adriana and I looked into this and there are tests that assume you can set `Loan.patron=None` without deleting the loan, so we decided the minimal benefit wasn't worth changing this.

monitor.py
aslagle: That seems fine, but I'm not sure how long people normally take to renew a card.

model/__init__.py
aslagle: I think for ODL there are in fact Loans with no patron.

monitor.py
leonardr: From what I've seen people either renew a card pretty quickly or they wait so long that the ILS has no record of them and they end up getting a new card.

The motivating case for NYPL is that people sign up for temporary cards, don't convert them to full cards, and are now signing up again 6-12 months later because of a big membership push on the library's part. Our ILS has already forgotten about those temporary cards.

aslagle: That's interesting. I'm mostly worried about ODL since the circ manager has the only record of a hold, but 60 days should be enough.

leonardr: If this is a problem we could make it a per-library configuration -- we'd just need to write a reaper that worked differently on a per-library basis.

leonardr:
We'll need a migration script to add the ON CASCADE DELETE bits to existing databases.

tests/models/test_patron.py
```diff
@@ -472,6 +475,44 @@ def try_to_set_none(patron):
...
+        # Create a patron and check that it has  been created
+        patron = self._patron()
+        eq_(len(self._db.query(Patron).all()), 1)
+
+        # Give the patron a loan, and check that it has been created
+        work_for_loan = self._work(with_license_pool=True)
+        pool = work_for_loan.license_pools[0]
+        loan, is_new = pool.loan_to(patron)
+        eq_(loan, patron.loans[0])
+        eq_(len(self._db.query(Loan).all()), 1)
```
leonardr: Some stylistic improvements are possible here:

```
eq_([loan], patron.loans) can replace lines 487-488.
```

tests/models/test_patron.py
```diff
@@ -472,6 +475,44 @@ def try_to_set_none(patron):
...
+
+        # Delete the patron and check that it has been deleted
+        self._db.delete(patron)
+        eq_(len(self._db.query(Patron).all()), 0)
+
+        # The patron's loan, hold, annotation, and credential should also be gone
+        eq_(len(self._db.query(Loan).all()), 0)
+        eq_(len(self._db.query(Hold).all()), 0)
+        eq_(len(self._db.query(Annotation).all()), 0)
+        eq_(len(self._db.query(Credential).all()), 0)
```
leonardr: Similarly, `eq_([], self._db.query(x).all())`

model/__init__.py
```diff
@@ -71,7 +71,6 @@ def create(db, model, create_method='',
 
 def get_one(db, model, on_multiple='error', constraint=None, **kwargs):
     """Gets an object from the database based on its attributes.
-
```
leonardr: A couple more blank lines removed by Atom.

leonardr:
This looks good apart from the removed blank lines.

PR #958, opened 2018-09-28
Default entrypoint

adriana-alter:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1261 and https://jira.nypl.org/browse/SIMPLY-1262. (I'm still having trouble believing the 1262 thing but it does seem to work.)

The underlying problem is that `SearchFacets` recognizes 'All' as a valid entrypoint, but the code to determine 'which facets are valid?' never calls `facets.selectable_entrypoints`, it just assumes that the facet config's `.entrypoints` is the same list.


leonardr:
This branch needs to go into both `audiobooks` and `master`. The same code should work in both places.

PR #959, opened 2018-10-01
Got scripts to work by using nose's Importer to do imports.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1266.

This discussion is the best explanation of the problem I've seen: https://stackoverflow.com/a/47030746/8671745

Since we no longer hack sys.path, there is no longer any module called 'core' outside of the circulation or metadata component. There is also no module above tests/ or model/ (or, in this case, bin/) for the ".." in `from .. import x` to be a reference to. If you want to get this to work, the only solution is... to hack sys.path. The thing we just got rid of because our refactoring broke it.

So how is it that the tests can run? It turns out nose has a class called `Importer`, and `Importer.importFromDir(path, module_name)` will load a package under a given name, no questions asked. This branch creates a module called `startup` which is imported by the individual scripts instead of each one messing with `sys.path`. The `startup` module uses `importFromDir` to make the core package appear as `core` -- exactly what we would see if this script code were located in the components that use it.

After that point, you still can't refer to `scripts`, but you can refer to `core.scripts`, so it's an easy change to make in the individual scripts.


bin/repair/startup.py
```diff
@@ -0,0 +1,11 @@
+import os
+from os.path import dirname
+from nose.importer import Importer
+bin_dir = os.path.split(__file__)[0]
+component_dir = os.path.join(bin_dir, "..", "..", "..")
```
aslagle: Scripts in repair can't use bin/startup because of this?

model/__init__.py
```diff
@@ -37,14 +37,16 @@
...
 from constants import (
     DataSourceConstants,
     EditionConstants,
     IdentifierConstants,
     LinkRelations,
     MediaTypes,
 )
-from .. import classifier
+from sys import path
+from os.path import dirname as dir
```
aslagle: Are these necessary?

bin/repair/startup.py
leonardr: They can't do `from .. import startup` because `bin` isn't a package, and I couldn't make `bin` into a package without the nose import trick.

I did getthis working by stacking the two import hacks, like this:

```
import os
import sys
bin_dir = os.path.split(__file__)[0]
package_dir = os.path.join(bin_dir, "..")
sys.path.append(os.path.abspath(package_dir))
import startup
from core.scripts import CheckContributorNamesInDB
CheckContributorNamesInDB().run()
```

I decided it was better to keep the startup code in another file, even though it's nearly a copy of the file in the parent directory.

PR #960, opened 2018-10-01
Default entrypoint from audiobooks

adriana-alter:
This branch ports https://github.com/NYPL-Simplified/server_core/pull/958 to the audiobooks branch.

### DIFF
--- a/tests/test_lane.py
+++ b/tests/test_lane.py
@@ -125,7 +125,7 @@ class TestFacetsWithEntryPoint(DatabaseTest):
         assert isinstance(facets, ExtraFacets)
         eq_("extra value", facets.extra)
 
-     def test__from_request(self):
+    def test__from_request(self):
         """_from_request calls load_entrypoint and instantiates the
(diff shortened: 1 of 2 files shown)


tests/test_lane.py
```diff
@@ -129,39 +129,60 @@ def test__from_request(self):
...
+        # faceting class.
+        MockFacetsWithEntryPoint.expect = INVALID_INPUT
         eq_(
-            self.expect,
-            FacetsWithEntryPoint._from_request(
-                config, get_argument, mock_worklist
+            MockFacetsWithEntryPoint.expect,
+            MockFacetsWithEntryPoint._from_request(
+                config, get_argument, mock_worklist,
+                extra="extra kwarg"
```
leonardr: This is slightly different because the `audiobooks` branch doesn't have the version of `_from_request` that takes `get_header` (used to pull the value of the `Accept-Language` header when creating a `SearchFacets` object).

PR #961, opened 2018-10-01
Added migration script that makes book-only lanes medium-agnostic.

leonardr:
Fixes https://jira.nypl.org/browse/SIMPLY-1260 with the least invasive migration script I could come up wtih.


aslagle:
For things like this it would probably be easier to merge it into master and then merge master into the audiobooks branch.

leonardr:
That's a really good idea.

aslagle:
No, in that case there's not much you could do. You can cherry pick the commit but it wouldn't really help.

PR #963, opened 2018-10-02
Track whether currently selected EntryPoint is a default.

leonardr:
This is a support branch for work on https://jira.nypl.org/browse/SIMPLY-1262. When loading a Facets object from a request, we keep track not only of the currently selected EntryPoint but of whether that EntryPoint was explicitly selected by the user, or whether it's a default.

`load_facets_from_request` new takes an EntryPoint to use as the default, so that you can specify a default EntryPoint other than the standard default of "the first enabled EntryPoint for the request library".

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -128,7 +128,7 @@ class FacetsWithEntryPoint(FacetConstants):
     @classmethod
     def from_request(
             cls, library, facet_config, get_argument, get_header, worklist,
-            **extra_kwargs
+            default_entrypoint=None, **extra_kwargs
     ):
         """Load a faceting object from an HTTP request.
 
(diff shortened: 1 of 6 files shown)


lane.py
```diff
@@ -107,28 +107,33 @@ class FacetsWithEntryPoint(FacetConstants):
...
+        self.entrypoint_is_default = entrypoint_is_default
         self.constructor_kwargs = kwargs
 
     def navigate(self, entrypoint):
         """Create a very similar FacetsWithEntryPoint that points to
         a different EntryPoint.
         """
         return self.__class__(
-            entrypoint=entrypoint, **self.constructor_kwargs
+            entrypoint=entrypoint, entrypoint_is_default=False,
```
leonardr: When you call navigate() you are making some kind of choice, so the default-ness is gone.

PR #965, opened 2018-10-03
A deliverymechanism can be a DRM scheme with no underlying content type, if the DRM license file is also the manifest file

leonardr:
This is the main work for https://jira.nypl.org/browse/SIMPLY-1275. The underlying goal is to represent the fact that a Findaway license document is both the DRM file and the manifest file for its delivery mechanism. I decided to represent this by leaving the `content_type` empty -- a situation that was supported by AcquisitionFeed.format_types, but was untested and not allowed by the database schema.


migration/20181002-deliverymechanism-media-type-can-be-null.sql
```diff
@@ -0,0 +1,8 @@
+-- If a DRM document is also the manifest document, a delivery
+-- mechanism has no separate content type.
+ALTER TABLE deliverymechanisms ALTER COLUMN content_type DROP NOT NULL;
+
+-- MP3 files are not a content type for purposes of describing
+-- fulfillment, because they are always obtained by processing a
+-- manifest. In that case the manifest document is the content type.
+update deliverymechanisms set content_type=NULL where content_type='audio/mpeg';
```
aslagle: You could have an audio/mpeg file with no DRM, but I guess no circulation managers will have those yet.

PR #973, opened 2018-10-15
Track lane size independently for every active entry point

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1305 by estimating the size of each lane separately for every enabled entry point. The overall lane size is till calculated and can be used (e.g. by the admin interface), but when generating featured feeds we now take the entry point into account.

I took an approach that degrades gracefully if the new `size_by_entrypoint` field is not populated or incompletely populated.

One thing I could do to make it even more graceful is to always calculate the size for every known entrypoint, not just the disabled entrypoints. This will make things work automatically when you enable an entrypoint, without having to first refresh the materialized views. The downside is a performance hit in lane size calculation so long as a given entrypoint is disabled.

### DIFF
--- a/entrypoint.py
+++ b/entrypoint.py
@@ -72,6 +72,8 @@ class EntryPoint(object):
         cls.ENTRY_POINTS.remove(entrypoint_class)
         del cls.BY_INTERNAL_NAME[entrypoint_class.INTERNAL_NAME]
         del cls.DISPLAY_TITLES[entrypoint_class]
+        if entrypoint_class in cls.DEFAULT_ENABLED:
+            cls.DEFAULT_ENABLED.remove(entrypoint_class)
 
     @classmethod
     def modified_materialized_view_query(cls, qu):
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -558,15 +558,15 @@ class TestLaneSweeperScript(DatabaseTest):
         worklist = script.considered.pop(0)
         eq_(self._default_library, worklist.get_library(self._db))
         eq_(self._default_library.name, worklist.display_name)
-        eq_([good, bad], worklist.children)
+        eq_(set([good, bad]), set(worklist.children))
 
         # After that, every lane was considered for processing, with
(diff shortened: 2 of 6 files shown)


tests/test_scripts.py
```diff
@@ -558,15 +558,15 @@ def process_lane(self, lane):
         worklist = script.considered.pop(0)
         eq_(self._default_library, worklist.get_library(self._db))
         eq_(self._default_library.name, worklist.display_name)
-        eq_([good, bad], worklist.children)
+        eq_(set([good, bad]), set(worklist.children))
```
leonardr: This branch changed the order of the children returned by `Lane.children`. My best guess is that something happened in the internals because I added a field to the `lanes` table.

entrypoint.py
```diff
@@ -72,6 +72,8 @@ def unregister(cls, entrypoint_class):
         cls.ENTRY_POINTS.remove(entrypoint_class)
         del cls.BY_INTERNAL_NAME[entrypoint_class.INTERNAL_NAME]
         del cls.DISPLAY_TITLES[entrypoint_class]
+        if entrypoint_class in cls.DEFAULT_ENABLED:
+            cls.DEFAULT_ENABLED.remove(entrypoint_class)
```
leonardr: A mock EntryPoint was sticking around DEFAULT_ENABLED after the test was over and causing trouble.

aslagle:
Could you recalculate sizes when the enabled entrypoints change? That would be pretty rare.

leonardr:
It takes long enough that it can't be done while a user is waiting for a response to their request. We could set something in the database, so that cache_opds_blocks knows it should recalculate the lane sizes before proceeding. But if we're going to fix this I'd rather just run the calculation on every entry point. In the long run it's unlikely that a library would have a bunch of audiobooks in the collection but not use them.

leonardr:
It was really easy to calculate the size for every entrypoint, and the headache saved is worth a slight performance hit in a script that runs in the middle of the night.

PR #974, opened 2018-10-16
Don't assume all incoming items are books.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1306 in detail.

* The metadata layer no longer sets a default medium. If a data source doesn't know the medium for a title, we don't assume it's an ebook.
* The book's license source is now trusted above the metadata wrangler for everything except cover image. (Previously the book's license source was supposed to be trusted just _below_ the metadata wrangler, but that didn't work because of a bug.)
* A migration script finds all LicensePools where the presentation edition says "ebook", but the license source edition says "not an ebook", recalculates the presentation edition for those LicensePools, and regenerates the OPDS entries for their Works.

I will also do something on the metadata wrangler to remove its unwarranted assumptions about `medium`, but this assumption is in a lot of Works and it'll be tough to totally wipe it out.


model/edition.py
```diff
@@ -344,26 +344,25 @@ def sort_key(edition):
...
+                return -1.5
+
             if source.name in DataSourceConstants.PRESENTATION_EDITION_PRIORITY:
-                id_type = edition.primary_identifier.type
-                if (id_type == Identifier.ISBN and
-                    source.name == DataSourceConstants.METADATA_WRANGLER):
-                    # This ISBN edition was pieced together from OCLC data.
-                    # To avoid overwriting better author and title data from
-                    # the license source, rank this edition lower.
-                    return -1.5
```
leonardr: It turns out we _always_ want to rank the metadata wrangler lower than the license source.

model/edition.py
```diff
@@ -331,7 +331,7 @@ def missing_coverage_from(
         return q2
 
     @classmethod
-    def sort_by_priority(self, editions):
+    def sort_by_priority(cls, editions, license_source=None):
```
leonardr: This is a class method but we were treating it as an instance method of `LicensePool` and accessing `self.data_source`. That's why the sorting didn't work correctly before.

tests/models/test_edition.py
```diff
@@ -131,25 +131,34 @@ def test_missing_coverage_from(self):
             self._db, gutenberg, web)]))
 
     def test_sort_by_priority(self):
-        edition_admin = self._edition(data_source_name=DataSource.LIBRARY_STAFF, with_license_pool=False)
```
leonardr: Apart from the bit at the end, this code change is just cleanup to make the test read easier.

tests/models/test_licensing.py
```diff
@@ -554,53 +554,70 @@ def test_set_presentation_edition(self):
...
+        assert_not_equal(od, presentation)
+
+        # Within the presentation edition, information from the
+        # library staff takes precedence over anything else.
+        eq_(presentation.title, u"AdminInterfaceTitle1")
+        eq_(admin.contributors, presentation.contributors)
+
+        # Where the library staff has no opinion, the license source
+        # takes precedence over the metadata wrangler.
+        eq_(Edition.AUDIO_MEDIUM, presentation.medium)
```
leonardr: Similar here -- this is the only new bit, everything else in this test is cleanup.

PR #982, opened 2018-10-30
Add new logger that allows logging to AWS Cloudwatch Logs

jonathangreen:
This commit makes it possible to output the circulation managers logs to AWS CloudWatch and an optional ExternalIntegration.

Some changes here: 
- Move `set_formatter` into the base class since its used by all the subclasses
- Move some of the settings for all loggers out of the `SysLogger` class and into `LogConfiguration`
- Add the `CloudwatchLogs` class
- Add some tests for the `CloudwatchLogs` class

Pinging @leonardr for code review.

### DIFF
--- a/log.py
+++ b/log.py
@@ -259,7 +259,7 @@ class CloudwatchLogs(Logger):
 
     @classmethod
     def get_handler(cls, settings, testing=False):
-        """Turn a Loggly ExternalIntegration into a log handler.
+        """Turn ExternalIntegration into a log handler.
         """
         group = settings.setting(cls.GROUP).value or cls.DEFAULT_APP_NAME
         stream = settings.setting(cls.STREAM).value or cls.DEFAULT_APP_NAME
@@ -419,6 +419,10 @@ class LogConfiguration(object):
 
 
 class MockCloudWatchLogHandler(CloudWatchLogHandler):
+    """
+    Mock handler used during unit testing.
+    """
+
     def __init__(self, log_group=__name__, stream_name=None, use_queues=True, send_interval=60,
                  max_batch_size=1024*1024, max_batch_count=10000):
         self.log_group = log_group
--- a/log.py
+++ b/log.py
@@ -8,7 +8,9 @@ from flask_babel import lazy_gettext as _
 from config import Configuration
 from StringIO import StringIO
 from loggly.handlers import HTTPSHandler as LogglyHandler
-
+from watchtower import CloudWatchLogHandler
+from config import CannotLoadConfiguration
+from model import ExternalIntegration, ConfigurationSetting
 
 class JSONFormatter(logging.Formatter):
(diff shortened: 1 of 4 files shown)


log.py
```diff
@@ -52,22 +54,40 @@ def format(self, record):
         return data
 
 class Logger(object):
-
-    DEBUG = "DEBUG"
-    INFO = "INFO"
-    WARN = "WARN"
-    ERROR = "ERROR"
```
leonardr: What happened to these variables?

leonardr: Specifically, with these gone, how does LOG_LEVEL_UI know what the options are?

log.py
```diff
@@ -224,6 +219,68 @@ def set_formatter(cls, handler, app_name):
...
+        { "key": INTERVAL, "label": _("Update Interval Seconds") },
+    ]
+
+    SITEWIDE = True
+
+    @classmethod
+    def from_configuration(cls, _db, testing=False):
+        settings = None
+        cloudwatch = None
+        from model import (ExternalIntegration, ConfigurationSetting)
```
leonardr: These classes are already imported from model at the top of the file.

log.py
leonardr: OK, I get it. These constants were redundant because they're also in LogConfiguration.

log.py
```diff
@@ -264,6 +324,12 @@ class LogConfiguration(object):
         },
     ]
 
+    @classmethod
+    def _defaults(cls, testing=False):
+        log_level = cls.DEFAULT_LOG_LEVEL
+        database_log_level = cls.DEFAULT_DATABASE_LOG_LEVEL
+        return log_level, database_log_level
+
```
leonardr: I think you can get rid of this method and just use the constants. I spent a while confused why there were two different methods called `_defaults` and we don't need this one at all.

requirements.txt
```diff
@@ -29,3 +29,5 @@ pymarc
 # for author name manipulations
 nameparser>=0.5.1
 fuzzywuzzy
+
+watchtower
```
leonardr: Add a comment explaining what this is for Cloudwatch integration.

log.py
jonathangreen: Good call. Copy pasta on my part. Removing this.

log.py
jonathangreen: Sure. I had that method to make it easier to test the defaults, but I'll pull it out.

jonathangreen:
I've been testing this on my servers, and I had to make a few changes based on that testing. Code review feedback is also incorporated. Ready for another review now @leonardr.

jonathangreen:
@leonardr updated fields to add required.

jonathangreen:
Looks like tests failed, the pass locally for me though. Seems like wget timed out? Maybe just restarting they will pass.

PR #983, opened 2018-11-07
Work presentation coverage providers

leonardr:
This branch adds two new WorkCoverageProviders modeled after OPDSEntryWorkCoverageProvider. These two scripts give us a way to selectively remove coverage of the presentation calculation process, e.g. during a migration script. This gives us a way to fix the fallout from problems like https://jira.nypl.org/browse/SIMPLY-1306 and https://jira.nypl.org/browse/SIMPLY-1364.

Now, if you ever discover that one or more works have the wrong presentation, you don't have to fix the problem immediately -- just remove their coverage and one of these CoverageProviders will fix the problem in the background.

This is the core work of https://jira.nypl.org/browse/SIMPLY-1397.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -1358,7 +1358,7 @@ class OPDSEntryWorkCoverageProvider(WorkPresentationProvider):
 
 
 class WorkPresentationEditionCoverageProvider(WorkPresentationProvider):
-    """Make sure all Works have up-to-date presentation edition.
+    """Make sure each Work has an up-to-date presentation edition.
 
     This basically means comparing all the Editions associated with the
     Work and building a composite Edition.
--- a/coverage.py
+++ b/coverage.py
@@ -1338,7 +1338,7 @@ class WorkPresentationProvider(PresentationReadyWorkCoverageProvider):
     needs to have some aspect of its presentation recalculated. These
     providers give back the 'missing' coverage.
     """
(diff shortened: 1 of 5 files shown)


migration/20181107-remove-coverage-of-non-books-incorrectly-classified-as-books.sql
```diff
@@ -0,0 +1,29 @@
...
+          -- with an associated identifier that says they are _not_ books.
+          select lp.id
+            from licensepools lp
+                 join editions e on (
+                   lp.data_source_id=e.data_source_id
+		   and lp.identifier_id=e.primary_identifier_id
+                 )
+	     where e.medium != 'Book'
+          )
+    )
```
leonardr: This migration serves as an example of the kind of thing we can do with these scripts.

coverage.py
```diff
@@ -1325,23 +1326,107 @@ def items_that_need_coverage(self, identifiers=None, **kwargs):
...
+        update_search_index=True,
+    )
+
+    def process_item(self, work):
+        """Recalculate the presentation for a Work."""
+
+        # Calling calculate_presentation_edition won't, on its own,
+        # regenerate the OPDS feeds or update the search index.
+        # So we call calculate_presentation in a way where the only actual
+        # change made will be
```
aslagle: you didn't finish this sentence

PR #984, opened 2018-11-08
Reap records one at a time so that the SQLAlchemy cascades will run.

leonardr:
This branch fixes a problem discovered while testing https://jira.nypl.org/browse/SIMPLY-1256. Deleting all the reaped records at once only works if there are no records in other tables that need to be deleted in a cascade. We do cascades at the SQLAlchemy level, not the database level, which means we need to call delete() on each individual item.

This isn't so bad because it gives us a chance to report each decision individually. I added a `__repr__` implementation to Patron so that we can log the patron records that get reaped in this way, in case there's any confusion or problems resulting from this work.

I also renamed `PatronReaper` to `PatronRecordReaper` because the class name sounded creepy.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -680,8 +680,12 @@ class ReaperMonitor(Monitor):
         return self.timestamp_field < self.cutoff
 
     def run_once(self, *args, **kwargs):
-        rows_deleted = self.query().delete(synchronize_session=False)
-        self.log.info("Deleted %d row(s)", rows_deleted)
+        rows_deleted = 0
+        qu = self.query()
+        self.log.info("Deleting %d row(s)", qu.count())
+        for i in qu:
+            self.log.info("Deleting %r", i)
+            self._db.delete(i)
 
     def query(self):
         return self._db.query(self.MODEL_CLASS).filter(self.where_clause)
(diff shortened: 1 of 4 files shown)


tests/test_monitor.py
```diff
@@ -759,16 +759,19 @@ def test_run_once(self):
         eq_(set([active, eternal]), remaining)
 
     def test_reap_patrons(self):
-        m = PatronReaper(self._db)
+        m = PatronRecordReaper(self._db)
         expired = self._patron()
+        credential = self._credential(patron=expired)
```
leonardr: This verifies that the one-at-a-time strategy works. The old strategy would have choked trying to delete the patron record without first deleting this Credential.

monitor.py
```diff
@@ -680,8 +680,12 @@ def where_clause(self):
         return self.timestamp_field < self.cutoff
 
     def run_once(self, *args, **kwargs):
-        rows_deleted = self.query().delete(synchronize_session=False)
-        self.log.info("Deleted %d row(s)", rows_deleted)
+        rows_deleted = 0
+        qu = self.query()
+        self.log.info("Deleting %d row(s)", qu.count())
+        for i in qu:
```
leonardr: This is the actual change; iterating over the query results instead of doing one big delete statement.

EdwinGuzman: Is this faster? or why?

EdwinGuzman: Never mind, I get it now.

leonardr: It's slower, because SQLAlchemy will use separate SQL statements to cascade the delete to every other object associated with this object (e.g. all of this patron's loans and credentials). 

Trying to delete all the patrons at once is faster, and it would work if we did the cascades on a database level, but it doesn't work here.

PR #985, opened 2018-11-08
Ignore Overdrive images that are obviously stand-ins.

leonardr:
This branch fixes the Overdrive portion of https://jira.nypl.org/browse/SIMPLY-340.

The migration script ensures that tainted editions and works will no longer use the bad images, and that (on the metadata wrangler) the bibliographic coverage provider will try to get them proper cover images the next time it runs.

Once these books have new cover images from the metadata wrangler, the circ managers will pick it up automatically. But even no cover image is better than an inaccurate stand-in.

It looks like about 5000 titles total have been affected.

### DIFF
--- a/migration/20181108-remove-overdrive-standin-ids.sql
+++ b/migration/20181108-remove-overdrive-standin-ids.sql
@@ -1,0 +1,17 @@
+-- This script takes a belt-and-suspenders approach to removing
+-- placeholder images mistakenly downloaded from Overdrive.
+
+-- Remove the Overdrive bibliographic coverage for affected editions. The
+-- Overdrive bibliographic coverage provider will try to find new images.
+delete from coveragerecords where identifier_id in (select i.id from identifiers i join editions e on e.primary_identifier_id=i.id where (e.cover_full_url like '%00000000-0000-0000-0000%' or e.cover_thumbnail_url like '%00000000-0000-0000-0000%')) and data_source_id in (select ds.id from datasources ds where ds.name='Overdrive');
+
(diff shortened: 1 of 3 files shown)


migration/20181108-remove-overdrive-standin-ids.sql
```diff
@@ -0,0 +1,17 @@
...
+
+-- Delete work coverage records for works whose presentation editions ended up with bad cover images. They'll be refreshed and end up with no cover image. Once a better image is located, their presentation will be refreshed again.
+delete from workcoveragerecords where operation='choose-edition' and work_id in (select w.id from works w where presentation_edition_id in (select id from editions where cover_full_url like '%00000000-0000-0000-0000%' or cover_thumbnail_url like '%00000000-0000-0000-0000%'));
+
+-- Remove all hyperlinks that caused an edition to end up with a bad cover or thumbnail.
+delete from hyperlinks where id in (select h.id from hyperlinks h join resources res on h.resource_id=res.id join representations rep on res.representation_id=rep.id join editions e on h.identifier_id=e.primary_identifier_id where (e.cover_full_url like '%00000000-0000-0000-0000%' or e.cover_thumbnail_url like '%00000000-0000-0000-0000%') and h.rel='http://opds-spec.org/image' and res.url like '%00000000-0000-0000-0000%');
```
aslagle: should the resource and representation be deleted too?

leonardr: I think it's okay if they exist -- those files are real and we did download and mirror them -- we just decided we don't want to use them.

PR #989, opened 2018-11-21
Reading language query in search and removing the Accept-Language...

EdwinGuzman:
…header from the browser if it doesn't want to be used.

Fixes [SIMPLY-1213](https://jira.nypl.org/browse/SIMPLY-1213).

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -727,6 +727,9 @@ class SearchFacets(FacetsWithEntryPoint):
             media = None
         extra['media'] = media
         languageQuery = get_argument("language", None)
+        # Currently, the only value passed to the language query from the client is
+        # `all`. This will remove the default browser's Accept-Language header value
+        # in the search request.
         if languageQuery != "all" :
             extra['languages'] = languages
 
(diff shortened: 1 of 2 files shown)


lane.py
```diff
@@ -726,7 +726,10 @@ def from_request(cls, library, config, get_argument, get_header, worklist,
         if media not in EditionConstants.KNOWN_MEDIA:
             media = None
         extra['media'] = media
-        extra['languages'] = languages
+        languageQuery = get_argument("language", None)
+        if languageQuery != "all" :
+            extra['languages'] = languages
```
aslagle: I'm not sure how you're using this, but I would think if the `languageQuery` had some languages that weren't "all", you should use them here instead of the languages from the header. If `languageQuery` is `None`, you could use the header, and if it's "all", you could not use anything.

aslagle: Maybe add a comment that the only possible value is "all" then. This is confusing if you look at this code by itself. Eventually I think we will want to have something like what I described.

PR #990, opened 2018-11-28
Mark work for presentation calculation when metadata changes

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1485.

Whenever metadata.apply() is called, we keep track of whether the metadata changes might change the presentation edition of any associated Work (for simple stuff like title and contributor changes) or whether it might require the Work's presentation to be completely and expensively recalculated (for changes to subjects, measurements, or the summary).

At the end of the method, we check whether such a `Work` actually exists. If it does, we call a new `Work` helper method to put the appropriate `WorkCoverageRecord` in the REGISTERED state.

Some pieces of code call metadata.apply() and then immediately call Work.calculate_presentation(), or a method that calls Work.calculate_presentation such as LicensePool.calculate_work(). This isn't wrong, but it's now potentially inefficient. We need to evaluate these places to see whether we offload the work of calculating presentations to the background scripts.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -1732,7 +1732,7 @@ class Metadata(MetaToModelUtility):
             # Any LicensePool will do here, since all LicensePools for
             # a given Identifier have the same Work.
             pool = get_one(
-                self._db, LicensePool, identifier=edition.primary_identifier,
+                _db, LicensePool, identifier=edition.primary_identifier,
                 on_multiple='interchangeable'
             )
             if pool.work:
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -41,6 +41,7 @@ from model import (
     Identifier,
     LicensePool,
     LicensePoolDeliveryMechanism,
+    LinkRelations,
     Subject,
     Hyperlink,
     PresentationCalculationPolicy,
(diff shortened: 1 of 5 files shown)


metadata_layer.py
```diff
@@ -1429,7 +1429,17 @@ def apply(self, edition, collection, metadata_client=None, replace=None,
         so work.simple_opds_feed refresh can be triggered.
         """
         _db = Session.object_session(edition)
-        made_core_changes = False
```
leonardr: made_core_changes is the same as work_requires_new_presentation_edition, with one exception (which I think we handle better now).

metadata_layer.py
```diff
@@ -1587,6 +1598,8 @@ def _key(classification):
                     original_resource=original_resource,
                     transformation_settings=link.transformation_settings,
                 )
+                work_requires_new_presentation_edition = True
```
leonardr: This is the exception. Previously, if you added links, we didn't set made_core_changes. But if you add links there's the possibility that the work's cover image has changed, so the work does require a new presentation edition.

I'm don't understand 100% how we used made_core_changes outside this method, but it looks like it was sometimes used to trigger an immediate presentation edition recalculation, so it _should_ have been set to true when new links were added.

aslagle: Isn't the summary also a link? So this could require a full recalculation.

PR #991, opened 2018-11-28
Identifier->Work lookup no longer calls calculate_presentation if there's already a presentation-ready work

leonardr:
This changes the two main places in core where we go from an Identifier to a Work, possibly creating a new Work in the process.

These two places are the OPDSImporter.update_work_for_edition (called immediately after metadata comes in from an OPDS import) and CollectionCoverageProvider.work() (called by CollectionCoverageProvider.set_metadata())

In both cases, I remove the call to LicensePool.calculate_presentation() except when a new work needs to be created, or a work exists but it's not presentation-ready. In all other cases, we skip the potentially expensive call and expect that a background process will eventually update the Work's presentation information with whatever new metadata just came in. We expect this will happen because when a presentation-ready Work already exists, the adjacent call to Metadata.apply() put that Work in the 'registered' state for a full or partial metadata update.

This should dramatically speed up the process of importing new metadata from the metadata wrangler (OPDSImporter) as well as from all other data sources (each of which, I think, creates Works using a BibliographicCoverageProvider, which is a CollectionCoverageProvider).

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -1044,19 +1044,28 @@ class CollectionCoverageProvider(IdentifierCoverageProvider):
         LicensePool to use is provided.) This method will not create
         new LicensePools.
 
+        If the work is newly created or an existing work is not
+        presentation-ready, a new Work will be created by calling
+        LicensePool.calculate_work(). If there is an existing
+        presentation-ready work, calculate_work() will not be called;
+        instead, the work will be slated for recalculation when its
+        metadata changes through Metadata.apply().
+
+        :param calculate_work_kwargs: Keyword arguments to pass into
+            calculate_work() if and when it is called.
+
         :return: A Work, if possible. Otherwise, a CoverageFailure explaining
         why no Work could be created.
+
         """
         work = identifier.work
-        if work:
-            # There is already a Work associated with this Identifier.
-            # Return it.
-            return identifier.work
+        if work and work.presentation_ready:
+            # There is already a presentation-ready Work associated
+            # with this Identifier.  Return it.
+            return work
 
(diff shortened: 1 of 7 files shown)


coverage.py
```diff
@@ -1044,19 +1044,28 @@ def work(self, identifier, license_pool=None, **calculate_work_kwargs):
...
         work = identifier.work
-        if work:
-            # There is already a Work associated with this Identifier.
-            # Return it.
-            return identifier.work
-
-        # There is no Work associated with this Identifier. This means
-        # we need to create one. Since we can only create a Work from
-        # a LicensePool, beyond this point the CoverageProvider
-        # needs to have a Collection associated with it.
```
leonardr: This comment is redundant because you can't instantiate a CollectionCoverageProvider without providing a Collection.

model/licensing.py
```diff
@@ -935,16 +934,11 @@ def calculate_work(
...
+        # Recalculate the display information for the Work. Either the
         # associated LicensePools have changed, which may have caused
-        # the Work's presentation Edition to change.
-        #
-        # TODO: In theory we can speed things up by only calling
-        # calculate_presentation if licensepools_changed is
-        # True. However, some bits of other code call calculate_work()
-        # under the assumption that it always calls
-        # calculate_presentation(), so we'd need to evaluate those
-        # call points first.
```
leonardr: We're now heavily relying on the fact that LicensePool.calculate_work() always calls Work.calculate_presentation(), but the tradeoff is that we should now be calling LicensePool.calculate_work() a lot less often.

aslagle (approved):

PR #992, opened 2018-11-29
Fix condition which handles invalid OverDrive ISBN values

alsrlw:
Fixes Jira ticket #1477


leonardr:
@aslagle see https://jira.nypl.org/browse/SIMPLY-1477 for context.

PR #993, opened 2018-11-30
MARC export

aslagle:
This branch adds an exporter for generating and caching MARC files for a lane.

I'm also planning to make it possible to generate additional MARC files for a lane based on time, e.g. a MARC file with all the updates since a certain date, but I might do that in a separate branch.


coverage.py
```diff
@@ -1373,6 +1373,23 @@ def process_item(self, work):
         return work
 
 
+class MARCRecordWorkCoverageProvider(WorkPresentationProvider):
+    """Make sure all presentation-ready works have an up-to-date MARC
+    record.
+
+    This is different from the MARCRecordCacheMonitor, which sweeps
+    over all presentation-ready works, even ones which are already
+    covered.
```
aslagle: This was based on the OPDS code, I may not actually need both of the monitor and the coverage provider.

leonardr: I agree; I think we don't need the monitor. IMO Our pattern where we strategically remove coverage when we want the work to be done again makes these monitors unnecessary.

marc.py
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+    # TODO: Add remaining formats. Maybe there's a better place to
+    # store this so it's easier to keep up-to-date.
+    # There doesn't seem to be any particular vocabulary for this.
+    FORMAT_TERMS = {
+        (Representation.EPUB_MEDIA_TYPE, DeliveryMechanism.NO_DRM): "EPUB eBook",
+        (Representation.EPUB_MEDIA_TYPE, DeliveryMechanism.ADOBE_DRM): "Adobe EPUB eBook",
+        (Representation.PDF_MEDIA_TYPE, DeliveryMechanism.NO_DRM): "PDF eBook",
+        (Representation.PDF_MEDIA_TYPE, DeliveryMechanism.ADOBE_DRM): "Adobe PDF eBook",
```
leonardr: I don't know if this is where you got them but these look like the names used by Overdrive to describe formats.

marc.py
```diff
@@ -0,0 +1,612 @@
...
+        isbn = None
+        if identifier.type == Identifier.ISBN:
+            isbn = identifier
+        if not isbn:
+            _db = Session.object_session(identifier)
+            for identifier_id in identifier.equivalent_identifier_ids()[identifier.id]:
+                equivalent = get_one(_db, Identifier, id=identifier_id)
+                if equivalent.type == Identifier.ISBN:
+                    isbn = equivalent
+                    break
```
leonardr: I don't think it will matter in real cases, since circ managers don't have a ton of equivalent identifiers, but it would be more efficient to do `_db.query(Identifier).filter(Identifier.type==Identifier.ISBN).filter(Identifier.id.in_(identifier_ids))`.

marc.py
```diff
@@ -0,0 +1,612 @@
...
+    @classmethod
+    def add_title(cls, record, edition):
+        # Non-filing characters are used to indicate when the beginning of a title
+        # should not be used in sorting. This code tries to identify them by comparing
+        # the title and the sort_title.
+        non_filing_characters = 0
+        if edition.title != edition.sort_title and ("," in edition.sort_title):
+            sort_title = edition.sort_title
+            last_comma = len(sort_title) - sort_title[::-1].index(",")
+            stemmed = edition.sort_title[:last_comma-1]
```
leonardr: I think these three lines could be replaced with `stemmed = edition.sort_title[:edition.sort_title.rindex(",")]`

marc.py
```diff
@@ -0,0 +1,612 @@
...
+            non_filing_characters = 0
+
+        record.add_field(
+            Field(
+                tag="245",
+                indicators=["0", non_filing_characters],
+                subfields=[
+                    "a", unicode(edition.title) or "",
+                    "b", unicode(edition.subtitle) or "",
+                    "c", unicode(edition.author) or "",
```
leonardr: I feel like the `or ""` should be inside the parentheses here.

marc.py
```diff
@@ -0,0 +1,612 @@
...
+                ]))
+
+    @classmethod
+    def add_contributors(cls, record, edition):
+        """Create contributor fields for this edition.
+
+        TODO: Use canonical names from LoC.
```
leonardr: Are these the MARC relators? If so you can use `Contributor.MARC_ROLE_CODES`.

aslagle: This is referring to the contributor's names, not the roles, I'll update the comment.

marc.py
```diff
@@ -0,0 +1,612 @@
...
+
+    SETTINGS = [
+        { "key": STORAGE_PROTOCOL,
+          "label": _("Storage service protocol"),
+          "description": _("Storage protocol to use for uploading generated MARC files. The service must already be configured under 'Storage Services'."),
+          "type": "select",
+          "options": [{ "key": ExternalIntegration.S3, "label": ExternalIntegration.S3 }],
+          "default": ExternalIntegration.S3,
+        },
+    ]
```
leonardr: Rather than naming a storage protocol I think this should name a particular storage integration. Right now there can only be one storage integration sitewide, and when we allow more than one storage integration it's pretty likely we'll allow multiple integrations with the same protocol (because I think we'd change the 'S3' integration to refer to a specific S3 bucket rather than an overall set of credentials).

So for now, it would work to have this be a yes-or-no setting: upload it to storage or don't. But I do would like a way to say "this is a drop-down that lets you choose any configured setting with a certain goal."

aslagle: That would be better, but how would you represent it in the database?

marc.py
```diff
@@ -0,0 +1,612 @@
...
+        content = output.getvalue()
+        output.close()
+
+        # Mirror the content.
+        storage_protocol = self.integration.setting(self.STORAGE_PROTOCOL).value
+        if storage_protocol == ExternalIntegration.S3:
+            mirror_integration = ExternalIntegration.lookup(
+                self._db, storage_protocol, ExternalIntegration.STORAGE_GOAL)
+            if not mirror_integration:
+                raise Exception("No mirror integration for configured protocol %s" % storage_protocol)
```
leonardr: This is what prompted me to make the comment in https://github.com/NYPL-Simplified/server_core/pull/993/files#r238435014

This is pretty close to what happens in `MirrorIntegration.sitewide_integration`, and I don't think this way of writing it is more future-proof than checking a yes-or-no setting that says whether we should call `sitewide_integration`.

leonardr: Yes, I think a configuration setting makes sense. We would either need to add a column to ConfigurationSetting or deal with the edge case where an ExternalIntegration is deleted but its ID was referred to by ConfigurationSettings.

leonardr: I should explicitly say that I think switching to yes-or-no is a better solution for now.

coverage.py
leonardr: Now that the monitor is gone, remove the second paragraph here.

marc.py
```diff
@@ -154,11 +155,10 @@ def add_isbn(cls, record, identifier):
...
             _db = Session.object_session(identifier)
-            for identifier_id in identifier.equivalent_identifier_ids()[identifier.id]:
-                equivalent = get_one(_db, Identifier, id=identifier_id)
-                if equivalent.type == Identifier.ISBN:
-                    isbn = equivalent
-                    break
+            identifier_ids = identifier.equivalent_identifier_ids()[identifier.id]
+            isbn = _db.query(Identifier).filter(
+                Identifier.type==Identifier.ISBN).filter(
+                Identifier.id.in_(identifier_ids)).first()
```
leonardr: Maybe add an `order_by(Identifier.id)` so the ISBN is less likely to jump around should the MARC record ever be regenerated.

leonardr:
This is fine overall. The main thing I'm worried about is the upcoming transition from having at most one S3 storage integration which is used for everything, to... not having that. Either having multiple storage integrations which are used for different purposes (e.g. a different integration for each S3 bucket) or adding support for multiple storage protocols but keeping the 'pick one and use it for everything' model.

The only effect that has on this branch is that if we allow a site to have multiple storage integrations, it implies a change to your implementation of the MARC configuration, and if we stick with the "pick a protocol and use it for everything" model then what you have is fine. But if possible I'd like to reach a decision on that now.

aslagle:
I think we should eventually allow multiple storage integrations, but I don't think it would be too hard to do a migration once we do that, regardless of how the MARC configuration is set up now. I'm also fine with going to the yes/no configuration for now.

leonardr:
OK, let's leave it for now.

PR #994, opened 2018-12-03
Change update_work_for_edition to always return the Work in question

leonardr:
Previously the Work was only being returned if it was created or made presentation-ready in the course of update_work_for_edition. This caused a bug that showed up in Adriana's branch https://github.com/NYPL-Simplified/circulation/pull/1124/.

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -651,6 +651,10 @@ class OPDSImporter(object):
             # There is no presentation-ready Work for this
             # LicensePool. Try to create one.
             work, ignore = pool.calculate_work()
+        else:
+            # There is a presentation-ready Work for this LicensePool.
+            # Use it.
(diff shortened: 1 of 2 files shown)


aslagle:
You have failing tests, but +1. I ran into the same bug on my branch.

opds_import.py
```diff
@@ -651,6 +651,10 @@ def update_work_for_edition(self, edition):
             # There is no presentation-ready Work for this
             # LicensePool. Try to create one.
             work, ignore = pool.calculate_work()
+        else:
+            # There is a presentation-ready Work for this LicensePool.
+            # Use it.
+            work = pool.work
```
aslagle: The failing test is probably because pool might be None here

PR #995, opened 2018-12-05
Generalize the code that scales measurements according to a percentile scale

leonardr:
This branch generalizes the behavior of Measurement.normalized_value so that we can add new types of measurements without changing any code, just data.

Rather than keeping separate dictionaries of percentile lists, POPULARITY_PERCENTILES, DOWNLOAD_PERCENTILES, etc., we keep a dictionary of dictionaries called PERCENTILE_SCALES. The outer dictionary is keyed on the quantity to be scaled: POPULARITY, DOWNLOADS, and so on.

More importantly, this branch adds two new percentile scales, one for PUBLISHED_EDITIONS as measured by OCLC Classify, and one for HOLDINGS as measured by OCLC Classify.

### DIFF
--- a/model/measurement.py
+++ b/model/measurement.py
@@ -35,36 +35,68 @@ class Measurement(Base):
 
     GUTENBERG_FAVORITE = u"http://librarysimplified.org/terms/rel/lists/gutenberg-favorite"
 
-    # If a book's popularity measurement is found between index n and
-    # index n+1 on this list, it is in the nth percentile for
-    # popularity and its 'popularity' value should be n * 0.01.
+    # We have a number of ways of measuring popularity: by an opaque
+    # number such as Amazon's Sales Rank, or by a directly measured
+    # quantity such as the number of downloads or published editions,
+    # or the number of libraries with a given book in their
+    # collection.
     #
-    # These values are empirically determined and may change over
-    # time.
-    POPULARITY_PERCENTILES = {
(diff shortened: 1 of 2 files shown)


leonardr:
This is side work for https://jira.nypl.org/browse/SIMPLY-1364. We will be gathering this information so we ought to make use of it.

model/measurement.py
```diff
@@ -35,36 +35,68 @@ class Measurement(Base):
...
-        # one-half the largest number of books ever ordered in a
-        # single month.
-        DataSourceConstants.CONTENT_CAFE : [0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 5, 5, 5, 5, 6, 6, 7, 8, 9, 10, 11, 14, 18, 25, 41, 125, 387]
-
-        # This is a percentile list of OCLC Work IDs and OCLC Numbers
-        # associated with Project Gutenberg texts via OCLC Linked
-        # Data.
+    # All of these ways of measuring popularity need to be scaled to a
+    # range between 0 and 1. This nested dictionary contains
+    # empirically determined measurements keyed by what they are measureing
```
aslagle: I don't think I've ever looked at this code closely and now I'm curious how these were determined.

leonardr: I take N measurements, lay them out in a sorted list, and take evenly spaced chunks from the list.

For OCLC Classify, N ~= 35,000 recently published books.

PR #996, opened 2018-12-06
Add categories

adriana-alter:


config.py
```diff
@@ -243,20 +244,23 @@ class Configuration(object):
               for entrypoint in EntryPoint.ENTRY_POINTS
           ],
           "default": [x.INTERNAL_NAME for x in EntryPoint.DEFAULT_ENABLED],
+          "category": "Filters",
```
aslagle: How about "Lanes and Facets" for this one. I'm not sure if people know what facets are though...

aslagle (approved):

config.py
adriana-alter: Sounds good!

PR #997, opened 2018-12-10
Added times to cached marc files so incremental updates can be generated

aslagle:


marc.py
```diff
@@ -456,6 +457,12 @@ class MARCExporter(object):
 
     DESCRIPTION = _("Export metadata into MARC files that can be imported into an ILS manually.")
 
+    # This controls how often MARC files should be automatically updated.
+    # Since the crontab in docker isn't easily configurable, we can run a script
+    # daily but check this to decide whether to do anything.
```
aslagle: Making this configurable was a request from Christine.

marc.py
```diff
@@ -539,7 +552,7 @@ def create_record(cls, work, annotator, force_create=False, integration=None):
         record = None
         existing_record = getattr(work, annotator.marc_cache_field)
         if existing_record and not force_create:
-            record = Record(data=existing_record, force_utf8=True)
+            record = Record(data=existing_record.encode('utf-8'), force_utf8=True)
```
aslagle: This is unrelated to the time stuff, but I found some encoding errors while testing.

testing.py
```diff
@@ -387,11 +388,11 @@ def _lane(self, display_name=None, library=None,
     ):
         display_name = display_name or self._str
         library = library or self._default_library
-        lane, is_new = get_one_or_create(
+        lane, is_new = create(
```
aslagle: Using this method to create two lanes with different genres gave you one lane with two genres.

marc.py
```diff
@@ -456,6 +457,12 @@ class MARCExporter(object):
 
     DESCRIPTION = _("Export metadata into MARC files that can be imported into an ILS manually.")
 
+    # This controls how often MARC files should be automatically updated.
+    # Since the crontab in docker isn't easily configurable, we can run a script
+    # daily but check this to decide whether to do anything.
+    UPDATE_FREQUENCY = "marc_update_frequency"
+    DEFAULT_UPDATE_FREQUENCY = 30
+
```
leonardr: Stick the unit (days) in the comment.

s3.py
```diff
@@ -220,12 +220,12 @@ def cover_image_url(self, data_source, identifier, filename,
...
         return root + self.key_join(parts)
 
-    def marc_file_url(self, library, lane):
+    def marc_file_url(self, library, lane, start_time=None, end_time=None):
         """The path to the hosted MARC file for the given library, lane,
         and date range."""
         bucket = self.get_bucket(self.MARC_BUCKET_KEY)
         root = self.marc_file_root(bucket, library)
-        parts = [lane.display_name]
+        parts = [start_time, end_time, lane.display_name]
```
leonardr: I'm not crazy about the way this creates URLS like None/None, as seen in the test below. It looks like there will be an end_time in every realistic case. So how about making end_time required and either doing /end_time/ or /start_time-end_time/ as one path segment?

s3.py
```diff
@@ -220,12 +220,16 @@ def cover_image_url(self, data_source, identifier, filename,
         parts = [identifier.type, identifier.identifier, filename]
         return root + self.key_join(parts)
 
-    def marc_file_url(self, library, lane):
+    def marc_file_url(self, library, lane, start_time=None, end_time=None):
```
leonardr: Your latest changes are great but now either end_time needs to be required, or we need a strategy for when it's `None`, or we need to be okay with `/bucket/None/Lane.mrc`.

PR #1000, opened 2018-12-13
Enable axis 360 audiobooks

leonardr:
This simple branch removes Axis 360 from the list of vendors whose audiobooks are unsupported. It's possible to do this because of the resolution of https://jira.nypl.org/browse/SIMPLY-1539.


leonardr:
:100:0

PR #1002, opened 2019-01-07
Added a MARCExporter setting for a web client URL for those who use a web client with no registry.

aslagle:
https://github.com/NYPL-Simplified/circulation-patron-web/pull/57 makes it possible to set up a web client with no library registry, so it's necessary to be able to configure the web client URL to use in MARC files without going through the registration process.

So I added a library setting to the MARC exporter for it. If we need to use the web client URL somewhere other than MARC files, this wouldn't be very good. Another option would be adding a new integration with the "discovery" goal that doesn't use a registration protocol, but lets you configure a web client URL. That would make the MARC-related code simpler but would be more complicated overall. Let me know if you think that would be better.

### DIFF
--- a/marc.py
+++ b/marc.py
@@ -469,6 +469,7 @@ class MARCExporter(object):
     # http://www.loc.gov/marc/organizations/org-search.php
     MARC_ORGANIZATION_CODE = "marc_organization_code"
 
+    WEB_CLIENT_URL = u'marc_web_client_url'
     INCLUDE_SUMMARY = u'include_summary'
     INCLUDE_SIMPLIFIED_GENRES = u'include_simplified_genres'
     STORAGE_PROTOCOL = u'storage_protocol'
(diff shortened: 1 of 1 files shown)


leonardr:
Do you have plans to use the web client URL in more scenarios than the MARC records? Like, in the authentication document? (That seems useful to me.) If so, then it should either be its own integration or it should be a configuration setting on the library.

This seems like a scenario similar to MirrorUploader, where there's a class that can be instantiated from a database connection. That class abstracts away the differences between different strategies for a certain goal. MirrorUploader was overkill when I originally designed it, but now that we're considering storing mirrored content on disk or in the database, those possibilities are a lot easier to think about.

leonardr:
I'm OK with this either way -- it shouldn't be hard to migrate from this to another system with a SQL script.

PR #1003, opened 2019-01-08
If an attempt to parse an XML document results in an XMLSyntaxError, remove known bad characters and try again.

leonardr:
This addresses the general case of https://jira.nypl.org/browse/SIMPLY-1174. If an XML document contains characters not valid in XML, they will be removed.

Since very few XML documents contain such characters, we only scrub XML documents if they have a problem, rather than doing it preemptively.


leonardr:
I'm running the test suite locally while the Travis CI integration is in limbo.

PR #1006, opened 2019-01-11
Enforce delivery mechanism-related uniqueness constraints even when one of the fields subject to the constraint is null.

leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-1621 for `licensepooldeliverymechanisms` and `deliverymechanisms`. It prohibits the creation of duplicate `LicensePoolDeliveryMechanism` and `DeliveryMechanism` objects even in a case where one of the fields is null--a situation where a Postgresql uniqueness constraint is not enforced.

Note that this branch removes some unique indexes. I [learned](https://www.postgresql.org/docs/9.4/indexes-unique.html) while working on this that:

"PostgreSQL automatically creates a unique index when a unique constraint or primary key is defined for a table. The index covers the columns that make up the primary key or unique constraint (a multicolumn index, if appropriate), and is the mechanism that enforces the constraint."

That implies that our explicit indexes are redundant--a unique constraint _is_ an index and it will be used in queries. I checked this by running SQL like `explain select * from licensepooldeliveries where identifier_id=10 and data_source_id=4;` and it did show it would be doing an index scan against the unique constraint.

I tested the migration script by migrating my local test database and running the tests against it.

### DIFF
--- a/model/licensing.py
+++ b/model/licensing.py
@@ -1246,13 +1246,19 @@ class LicensePoolDeliveryMechanism(Base):
                          'delivery_mechanism_id', 'resource_id'),
     )
 
+# The uniqueness constraint doesn't enforce uniqueness when one of the
+# fields is null, and one of these fields -- resource_id -- is
+# _usually_ null. So we also need a unique partial index to properly
+# enforce the constraint.
 Index(
-    "ix_licensepooldeliveries_datasource_identifier_mechanism",
+    'ix_licensepooldeliveries_unique_when_no_resource',
     LicensePoolDeliveryMechanism.data_source_id,
     LicensePoolDeliveryMechanism.identifier_id,
     LicensePoolDeliveryMechanism.delivery_mechanism_id,
-    LicensePoolDeliveryMechanism.resource_id,
+    unique=True,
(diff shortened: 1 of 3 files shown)


model/licensing.py
```diff
@@ -1306,6 +1312,10 @@ class DeliveryMechanism(Base, HasFullTableCache):
     _cache = HasFullTableCache.RESET
     _id_cache = HasFullTableCache.RESET
 
+    __table_args__ = (
+        UniqueConstraint('content_type', 'drm_scheme'),
+    )
+
```
leonardr: In this case a new UniqueConstraint is created to replace a unique index, and then a second unique partial index is created.

model/licensing.py
```diff
@@ -1246,14 +1246,20 @@ def __repr__(self):
                          'delivery_mechanism_id', 'resource_id'),
     )
 
+# The uniqueness constraint doesn't enforce uniqueness when one of the
+# fields is null, and one of these fields -- resource_id -- is
+# _usually_ null. So we also need a unique partial index to properly
+# enforce the constraint.
```
leonardr: In this case the UniqueConstraint already exists, so the redundant unique index is converted to a unique partial index.

aslagle (approved):

PR #1007, opened 2019-01-17
Remove the never-used WorkContribution class.

leonardr:
The only code we ever wrote using this class was code to cascade deletes.


leonardr:
Addresses https://jira.nypl.org/browse/SIMPLY-559

aslagle (approved):

PR #1009, opened 2019-01-18
Use the sitewide debug level to decide whether to provide tracebacks and other detailed information when reporting errors.

leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-1663. When you set the site into the DEBUG state, you're trying to fix a specific problem and you want all possible information about that problem. This is always intended to be a temporary state, and once you've fixed the problem you take the site out of the DEBUG state and get back the more secure behavior.

### DIFF
--- a/tests/test_app_server.py
+++ b/tests/test_app_server.py
@@ -20,7 +20,10 @@ from . import (
 
 from ..opds import TestAnnotator
 
-from ..model import Identifier
+from ..model import (
+    Identifier,
+    ConfigurationSetting,
+)
 
 from ..lane import (
     Facets,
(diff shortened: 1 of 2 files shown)


tests/test_app_server.py
```diff
@@ -393,9 +428,13 @@ def test_unhandled_error(self):
             eq_(500, response.status_code)
             eq_("An internal error occured", response.data)
 
-        # Try it again with debug=True to get a stack trace instead of
-        # a generic error message.
-        handler = ErrorHandler(self.app, debug=True)
+
+    def test_unhandled_error_debug(self):
```
leonardr: The exception in test_unhandled_error rolls back the database session, and this test needs a working database session to set the log level, so I moved it to a separate test method.

PR #1011, opened 2019-01-24
If we can't use the database session to determine the current log level, go with our previous guess rather than crashing.

leonardr:
This was written as part of the work to fix the test failures at https://github.com/NYPL-Simplified/circulation/pull/1162. 

By itself it doesn't address the test failures -- I did that by changing the test to mock the object that was causing the problem. But it seems reasonable that if there's a a database problem, we should report the underlying problem, and not the fact that the problem happened again while we were seeing how much detail we should use to report the problem.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -23,6 +23,7 @@ from util.opds_writer import (
     OPDSFeed,
     OPDSMessage,
 )
+from sqlalchemy.exc import SQLAlchemyError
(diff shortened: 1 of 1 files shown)


leonardr:
This finishes the work for https://jira.nypl.org/browse/SIMPLY-1649. I'll do metadata scripts as well, but that's less important.

PR #1012, opened 2019-01-24
Record a script's completion as a Timestamp. 

leonardr:
Scripts that do nothing but invoke Monitors and CoverageProviders don't get their own timestamps.

This completes https://jira.nypl.org/browse/SIMPLY-1651 and does some of the work for https://jira.nypl.org/browse/SIMPLY-1649. Using the class name as a fallback script name technically completes 1649, but for some scripts the class names are a little vague and I think a more detailed explanation would be better, so I'll tackle that in a separate circulation branch.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -187,30 +187,25 @@ class Script(object):
             Configuration.load(self._db)
 
     def update_timestamp(self):
-        """Update any appropriate Timestamp for this script.
+        """By default scripts have no timestamp of their own.
 
-        Scripts that run Monitors and CoverageProviders will override
-        this with a no-op, since Monitors and CoverageProviders have their
-        own logic for creating Timestamps.
-        """
-        Timestamp.stamp(self._db, self.script_name, collection=None)
-
-
-class NoTimestampScript(Script):
-
-    def update_timestamp(self):
-        """This script doesn't have its own timestamp.
-
-        Instead, it does work through Monitors or CoverageProviders,
-        which have their own logic for creating timestamps.
-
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -2692,6 +2741,8 @@ def register_problem(cls, source, contribution, computed_sort_name, error_messag
 class Explain(IdentifierInputScript):
     """Explain everything known about a given work."""
 
+    name = "Explain everything known about a given work"
+
```
leonardr: We could also say that scripts which are only intended to be run from the command line should be `NoTimestampScripts`.

aslagle: There might be some that would be useful to keep track of, like a manual directory import, but I agree about this one.

leonardr:
I've flipped the logic, since almost all of our scripts shouldn't have their own timestamps. This does mean some additional work in circulation and metadata to change some of the script superclasses.

scripts.py
```diff
@@ -1618,9 +1652,11 @@ def do_run(self):
         self._db.commit()
 
 
-class WorkPresentationScript(WorkProcessingScript):
+class WorkPresentationScript(TimestampScript, WorkProcessingScript):
     """Calculate the presentation for Work objects."""
 
+    name = "Recalculate the presentation for works that need it."""
```
aslagle: I didn't know you could use different numbers of quotes at the beginning and the end of a string.

leonardr: You can't really; this is equivalent to `"foo"bar" == "foobar"`. It works by accident.

PR #1013, opened 2019-01-25
Add new fields to the timestamps table and start using most of them

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-1650
The columns mentioned here are added to the `timestamps` table: https://jira.nypl.org/browse/SIMPLY-529?focusedCommentId=30722&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-30722

* `service_type`
* `start`
* `exception`
* `achievements`

I changed the signature of `Timestamp.stamp` to allow callers to set the appropriate values. There are four places where this method is called:

* Monitor
* CoverageProvider
* TimestampScript
* Various tests

In each case I changed the `stamp` calls to set the appropriate values, especially `service_type` which is now required when looking up a Timestamp.

I also started setting values for `start` (the start time of the work tracked by the `Timestamp`) and `exception`.  `achievements` is a lot more complicated and can be addressed later.

The migration script is also pretty complicated -- we don't have totally consistent naming conventions for the things that make timestamps, and it's important to set the `service_type` appropriately for everything that's already run, because otherwise it will appear as though the script hasn't been run before. (In most cases that's not a huge deal but it can lead to wasted resources.) I'm pretty sure I got all scripts, monitors, and coverage providers that are currently active in circulation or metadata.

### DIFF
--- a/config.py
+++ b/config.py
@@ -545,7 +545,8 @@ class Configuration(object):
         if not known_value:
             from model import Timestamp
             known_value = Timestamp.value(
-                _db, cls.SITE_CONFIGURATION_CHANGED, None
+                _db, cls.SITE_CONFIGURATION_CHANGED, service_type=None,
+                collection=None
             )
         if not known_value:
             # The site configuration has never changed.
--- a/coverage.py
+++ b/coverage.py
@@ -189,8 +189,9 @@ class BaseCoverageProvider(object):
         for covered_statuses in covered_status_lists:
             offset = 0
             while offset is not None:
-                # TODO: change run_once to return a number of achievements
-                # (successfully covered records).
+                # TODO: change run_once to return achievement
+                # information (successfully covered records, transient
+                # errors, permanent errors).
                 try:
                     offset = self.run_once(
                         offset, count_as_covered=covered_statuses
--- a/coverage.py
+++ b/coverage.py
@@ -199,14 +199,14 @@ class BaseCoverageProvider(object):
                         "CoverageProvider %s raised uncaught exception.",
                         service_name, exc_info=e
                     )
-                    exception = traceback.exc_info()
+                    exception = traceback.format_exc()
 
         self.update_timestamp(start=start_time, exception=exception)
 
     def update_timestamp(self, **kwargs):
         Timestamp.stamp(
             _db=self._db, service=self.service_name,
-            service_type=Timestamp.COVERAGE_PROVIDER,
+            service_type=Timestamp.COVERAGE_PROVIDER_TYPE,
             collection=self.collection, **kwargs
         )
         self._db.commit()
--- a/coverage.py
+++ b/coverage.py
@@ -193,7 +193,11 @@ class BaseCoverageProvider(object):
         self.update_timestamp()
 
     def update_timestamp(self):
-        Timestamp.stamp(self._db, self.service_name, self.collection)
+        Timestamp.stamp(
+            _db=self._db, service=self.service_name,
+            service_type=Timestamp.COVERAGE_PROVIDER,
+            collection=self.collection
+        )
         self._db.commit()
 
     def run_once(self, offset, count_as_covered=None):
--- a/coverage.py
+++ b/coverage.py
@@ -1,6 +1,7 @@
 from nose.tools import set_trace
 import datetime
 import logging
+import traceback
 
 from sqlalchemy.orm.session import Session
 from sqlalchemy.sql.functions import func
@@ -197,9 +198,10 @@ class BaseCoverageProvider(object):
                 except Exception, e:
                     logging.error(
                         "CoverageProvider %s raised uncaught exception.",
-                        service_name, exc_info=e
+                        self.service_name, exc_info=e
                     )
                     exception = traceback.format_exc()
+                    break
 
         self.update_timestamp(start=start_time, exception=exception)
 
--- a/coverage.py
+++ b/coverage.py
@@ -183,20 +183,31 @@ class BaseCoverageProvider(object):
             BaseCoverageRecord.PREVIOUSLY_ATTEMPTED,
             BaseCoverageRecord.DEFAULT_COUNT_AS_COVERED
         ]
+        start_time = datetime.datetime.utcnow()
+        exception = None
         for covered_statuses in covered_status_lists:
             offset = 0
             while offset is not None:
-                offset = self.run_once(
-                    offset, count_as_covered=covered_statuses
-                )
+                # TODO: change run_once to return a number of achievements
+                # (successfully covered records).
+                try:
+                    offset = self.run_once(
+                        offset, count_as_covered=covered_statuses
+                    )
+                except Exception, e:
+                    logging.error(
+                        "CoverageProvider %s raised uncaught exception.",
+                        service_name, exc_info=e
+                    )
+                    exception = traceback.exc_info()
 
-        self.update_timestamp()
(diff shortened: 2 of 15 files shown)


monitor.py
```diff
@@ -158,17 +159,30 @@ def run(self):
...
+        exception = None
+        started_running = datetime.datetime.utcnow()
+        try:
+            new_timestamp_value = (
+                self.run_once(start, started_running) or started_running
+            )
+            self.cleanup()
+        except Exception, e:
+            self.log.error("Error running %s monitor", exc_info=e)
+            exception = traceback.format_exc()
```
leonardr: It would be better to pass the exception itself around and get the stack trace later but I don't see an easy way to do that in Python 2.

coverage.py
```diff
@@ -183,17 +184,34 @@ def run_once_and_update_timestamp(self):
...
+        self.update_timestamp(start=start_time, exception=exception)
 
-    def update_timestamp(self):
-        Timestamp.stamp(self._db, self.service_name, self.collection)
+    def update_timestamp(self, **kwargs):
+        Timestamp.stamp(
+            _db=self._db, service=self.service_name,
+            service_type=Timestamp.COVERAGE_PROVIDER_TYPE,
+            collection=self.collection, **kwargs
+        )
```
leonardr: Changing this function ensures that any time a CoverageProvider updates its timestamp the appropriate `service_type` is included.

monitor.py
```diff
@@ -136,6 +136,7 @@ def timestamp(self):
         timestamp, new = get_one_or_create(
             self._db, Timestamp,
             service=self.service_name,
+            service_type=Timestamp.MONITOR_TYPE,
```
leonardr: Changing this method ensures that any time a Monitor finds 'its' timestamp it's using the appropriate `service_type`.

scripts.py
```diff
@@ -204,11 +212,15 @@ def __init__(self, *args, **kwargs):
...
-        Timestamp.stamp(self._db, self.script_name,
-                        collection=self.timestamp_collection)
-
+        Timestamp.stamp(
+            _db=self._db, service=self.script_name,
+            service_type=Timestamp.SCRIPT_TYPE,
+            collection=self.timestamp_collection,
+            start=start_time, exception=exception
+        )
+
```
leonardr: Changing this method ensures that all Timestamps created by scripts have the right `service_type`.

scripts.py
```diff
@@ -2532,10 +2543,12 @@ def run(self, cmd_args=None):
...
         init_timestamp = self.parse_time('1996-11-15')
         overall_timestamp = existing_timestamp or Timestamp.stamp(
-            self._db, self.SERVICE_NAME, None, date=init_timestamp
+            _db=self._db, service=self.SERVICE_NAME,
+            service_type=Timestamp.SCRIPT_TYPE, date=init_timestamp
         )
         python_timestamp = Timestamp.stamp(
-            self._db, self.PY_TIMESTAMP_SERVICE_NAME, None, date=init_timestamp
+            _db=self._db, service=self.PY_TIMESTAMP_SERVICE_NAME,
+            service_type=Timestamp.SCRIPT_TYPE, date=init_timestamp
```
leonardr: Since the migration scripts don't use Script.update_timestamp I had to specially add `service_type` to here.

tests/models/test_coverage.py
```diff
@@ -8,11 +8,94 @@
 from ...model.coverage import (
     BaseCoverageRecord,
     CoverageRecord,
+    Timestamp,
     WorkCoverageRecord,
 )
 from ...model.datasource import DataSource
 from ...model.identifier import Identifier
 
+class TestTimestamp(DatabaseTest):
```
leonardr: The `Timestamp` class had no test coverage at all.

config.py
```diff
@@ -545,7 +545,8 @@ def site_configuration_last_update(cls, _db, known_value=None,
         if not known_value:
             from model import Timestamp
             known_value = Timestamp.value(
-                _db, cls.SITE_CONFIGURATION_CHANGED, None
+                _db, cls.SITE_CONFIGURATION_CHANGED, service_type=None,
+                collection=None
```
leonardr: This is an internal process, not one of the three recognized service types, so we leave `service_type` null.

monitor.py
```diff
@@ -158,17 +159,30 @@ def run(self):
...
-        cutoff = datetime.datetime.utcnow()
-        new_timestamp_value = self.run_once(start, cutoff) or cutoff
-        duration = datetime.datetime.utcnow() - cutoff
-        self.cleanup()
+        exception = None
+        started_running = datetime.datetime.utcnow()
+        try:
+            new_timestamp_value = (
+                self.run_once(start, started_running) or started_running
+            )
```
leonardr: This is a little annoying; I wrote code assuming that `timestamp` was the end time of the previous run and `start` was the start time. But for Monitors, the `timestamp` is the _start_ time of the previous run.

I'm not sure what to do about this; maybe I should change `Monitor` to look for its time in `Timestamp.start` and use `Timestamp.timestamp` just for display purposes?

migration/20190125-add-timestamp-fields.sql
```diff
@@ -0,0 +1,74 @@
...
+-- );
+
+-- -- Add columns to the timestamp table, including service_type for the
+-- -- enumerated type.
+
+-- ALTER TABLE timestamps ADD COLUMN service_type service_type;
+-- CREATE INDEX ix_timestamps_service_type ON timestamps USING btree (service_type);
+-- ALTER TABLE timestamps ADD COLUMN start TIMESTAMP WITHOUT TIME ZONE;
+-- ALTER TABLE timestamps ADD COLUMN achievements CHARACTER VARYING;
+-- ALTER TABLE timestamps ADD COLUMN exception CHARACTER VARYING;
```
aslagle: why is all this commented out?

monitor.py
aslagle: Yeah, I would change Monitor to use Timestamp.start to be consistent with the other scripts.

leonardr:
OK, this should be ready again. Summary of my changes:

The `timestamp` field has been renamed to `finish`, for symmetry with `start`. This implied a lot of code changes but none of the functionality changed except as noted below.

The migration script starts all old scripts off by setting `start` to the same value as `finish`.

`Monitor` uses `start` instead of `finish` to determine what timestamp it needs to cover -- that is, it needs to cover everything since the last time it _started_ running.

If a `Monitor` does not run to completion, `Timestamp.exception` is updated but `start` and `finish` are left alone. This fixes https://jira.nypl.org/browse/SIMPLY-1687.

I removed two `Monitor` class variables that are no longer used, INTERVAL_SECONDS and KEEP_TIMESTAMP -- this clarifies what the job of a `Monitor` is and why we don't update the `Timestamp` after an unsuccessful run.

monitor.py
```diff
@@ -136,40 +123,64 @@ def timestamp(self):
...
+
+        # At this point `last_run_start` represents the first moment
+        # at which something might have happened that the Monitor
+        # hasn't taken into consideration. `this_run_start` represents
+        # the first moment of _this_ run, which is the _final_ moment
+        # the Monitor must take into consideration right
+        # now. Everything that happens after this time can be handled
+        # on the subsequent run.
+        this_run_start = datetime.datetime.utcnow()
+        update_timestamp_with = dict()
```
aslagle: you didn't use this.

PR #1014, opened 2019-01-30
The only invalid XML character we need to handle is the null character. XMLParser can handle all the others.

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1684. It takes a much simpler approach than we were using previously to parsing XML documents that contain XML characters -- take care of the null character ourselves and let `lxml.XMLParser` handle everything else.

### DIFF
--- a/util/http.py
+++ b/util/http.py
@@ -234,8 +234,8 @@ class HTTP(object):
 
         try:
             if verbose:
-                logging.info("Sending %s request to %s: kwargs %r",
-                             http_method, url, kwargs)
+                logging.info("Sending request to %s: args %r kwargs %r",
+                             url, args, kwargs)
             if len(args) == 1:
                 # requests.request takes two positional arguments,
                 # an HTTP method and a URL. In most cases, the URL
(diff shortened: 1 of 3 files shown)


util/http.py
```diff
@@ -234,8 +234,8 @@ def _request_with_timeout(cls, url, make_request_with, *args, **kwargs):
 
         try:
             if verbose:
-                logging.info("Sending %s request to %s: kwargs %r",
-                             http_method, url, kwargs)
+                logging.info("Sending request to %s: args %r kwargs %r",
+                             url, args, kwargs)
```
leonardr: This is an unrelated problem I found while debugging. There's no variable `http_method` in this method -- it's one of the `args`.

util/xmlparser.py
```diff
@@ -78,44 +48,20 @@ def int_of_optional_subtag(self, tag, name, namespaces=None):
             return v
         return int(v)
 
-    def scrub_xml(self, xml):
```
EdwinGuzman: "Scrub" is the xml term for cleaning the data, right?

leonardr: It's not official but yeah, it refers to taking some data that might contain invalid characters, and removing those characters.

PR #1015, opened 2019-01-31
Only remove null characters from incoming markup if it's actually markup

leonardr:
This fixes a bug introduced by https://github.com/NYPL-Simplified/server_core/pull/1014.

`XMLParser.process_all` can take either a string containing XML markup or a previously parsed etree tag object. Obviously, the code that removes null characters should only be run if the input is a string.

I didn't find this bug earlier because we have bad test coverage for XMLParser--it was one of the first things I wrote for this project. So I added an extra test to get us closer to full coverage.

This branch also fixes a problem introduced by https://github.com/NYPL-Simplified/server_core/pull/1013. In circulation, we've got a Monitor that sets `Timestamp.counter` to a custom value. But then the generic `Monitor.run` implementation calls `Timestamp.update()` without specifying a value for `counter`. Ideally this would be interpreted as meaning "leave the `counter` alone", but it's interpreted as meaning "set `counter` to `None`." I changed `update()` to interpret missing values as "leave it alone", except for `exception`.

We can assume that if `update()` is called with no exception being specified, it's because there was no exception. A custom `Monitor` won't deal with an exception by setting `Timestamp.exception` -- it will raise the exception and let the generic implementation figure out how to report it.

### DIFF
--- a/model/coverage.py
+++ b/model/coverage.py
@@ -228,10 +228,18 @@ class Timestamp(Base):
         """Use a single method to update all the fields that aren't
         used to identify a Timestamp.
         """
-        self.start = start
-        self.finish = finish
-        self.achievements = achievements
-        self.counter = counter
+
+        if start is not None:
+            self.start = start
+        if finish is not None:
+            self.finish = finish
+        if achievements is not None:
+            self.achievements = achievements
+        if counter is not None:
+            self.counter = counter
+
+        # Unlike the other fields, None is a realistic value for
+        # .exception
         self.exception = exception
 
 
--- a/util/xmlparser.py
+++ b/util/xmlparser.py
@@ -49,21 +49,22 @@ class XMLParser(object):
         return int(v)
 
     def process_all(self, xml, xpath, namespaces=None, handler=None, parser=None):
-        # XMLParser can handle most characters and entities that are
-        # invalid in XML but it will stop processing a document if it
-        # encounters the null character. Remove that character
-        # immediately and XMLParser will handle the rest.
-        xml = xml.replace(b"\x00", "")
         if not parser:
             parser = etree.XMLParser(recover=True)
         if not handler:
             handler = self.process_one
         if isinstance(xml, basestring):
             root = None
-            exception = None
+
+            # XMLParser can handle most characters and entities that are
+            # invalid in XML but it will stop processing a document if it
+            # encounters the null character. Remove that character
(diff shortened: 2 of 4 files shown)


util/xmlparser.py
```diff
@@ -49,21 +49,22 @@ def int_of_optional_subtag(self, tag, name, namespaces=None):
...
-        # encounters the null character. Remove that character
-        # immediately and XMLParser will handle the rest.
-        xml = xml.replace(b"\x00", "")
         if not parser:
             parser = etree.XMLParser(recover=True)
         if not handler:
             handler = self.process_one
         if isinstance(xml, basestring):
             root = None
-            exception = None
```
leonardr: This variable should have been removed by my previous branch along with that big chunk of code.

model/coverage.py
```diff
@@ -228,10 +228,18 @@ def update(self, start=None, finish=None, achievements=None,
         """Use a single method to update all the fields that aren't
         used to identify a Timestamp.
         """
-        self.start = start
-        self.finish = finish
-        self.achievements = achievements
-        self.counter = counter
+
+        if start is not None:
```
EdwinGuzman: What happens if any of these values is `None`?

util/xmlparser.py
```diff
@@ -49,21 +49,22 @@ def int_of_optional_subtag(self, tag, name, namespaces=None):
...
+
+            # XMLParser can handle most characters and entities that are
+            # invalid in XML but it will stop processing a document if it
+            # encounters the null character. Remove that character
+            # immediately and XMLParser will handle the rest.
+            xml = xml.replace(b"\x00", "")
```
EdwinGuzman: Gotcha, so the issue was that it was trying to update this assuming it was always a string and that's where it failed?

model/coverage.py
leonardr: The only one I can think of that would be `None` normally is `exception`. But if you really wanted to set `achievements` to None, you would have to do it by manipulating the `Timestamp` object directly. You can no longer do it by calling `update()`.

Another possibility is to introduce a constant `NO_VALUE` and make that the default. That way it'll be clear whether a value is missing because the caller wants it to be set to `None`, or whether it's missing because the caller didn't specify a value.

util/xmlparser.py
leonardr: Exactly.

PR #1016, opened 2019-01-31
Made TimestampInfo able to work around unmigrated changes to the timestamps table itself.

leonardr:
The migration script in https://github.com/NYPL-Simplified/server_core/pull/1013 will work if it's run as SQL, but the `DatabaseMigrationScript` itself won't run that script, because it can't get to that point unless the migration script has already run. Understanding which scripts need to be run requires a knowledge of the current schema of the `timestamps` table.

This branch changes `TimestampInfo` to know about the old and new versions of the `timestamps` table. If it tries to pull a timestamp out of the database and gets a ProgrammingError, it reinitializes the database session and tries again with a query that will work in the older schema.

To test this on a database that has already been migrated:

```
alter table timestamps rename column finish to timestamp;
alter table timestamps drop column exception;
update timestamps set timestamp='2019-01-24' where service='Database Migration';
```

Then run `core/bin/migrate_database`. The script will run, `exception` will be added, and `timestamp` will be renamed to `finish`.

I changed the SQL migration script to be idempotent so that it's easy to test by running the migration multiple times, and it's possible to run this specific migration in isolation before running `core/bin/migrate_database`.

It's still important that `20190125-add-timestamp-fields.sql` be the _first_ migration script to run, because the timestamps table is going to be used as soon as the first script runs (to update the timestamp), so by the time _that_ happens, the table needs to be in shape. So people will need to deploy to 2.2.14 to get the other migrations, and _then_ deploy to 2.3.0 -- or else run this particular migration script manually.

It's not too late to  back this out by changing `timestamp.finish` back to `timestamp.timestamp` -- that's the particular change that causes this problem.


model/__init__.py
```diff
@@ -430,12 +433,12 @@ def initialize_data(cls, session, set_site_configuration=True):
         # it was updated by cls.update_timestamps_table
         return session
 
-def production_session():
+def production_session(initialize_data=True):
```
leonardr: I don't know why this doesn't happen the _first_ time `DatabaseMigrationScript` calls production_session, but the second time it calls that method (because it screwed up the first database connection), it tries to initialize the dataset, which we don't want to do until the schema is in good shape.

aslagle:
Looks like there's one migration (20190117-drop-work-contributions.sql) that hasn't been in a release yet. Will that cause a problem?

leonardr:
It will cause a problem, good catch.

PR #1017, opened 2019-01-31
Make "how deep to go when looking for equivalent identifiers" part of the PresentationCalculationPolicy

leonardr:
This is a first pass at addressing https://jira.nypl.org/browse/SIMPLY-1690.

There are many places in our code where we pass around specific numbers with the goal of fine-tuning the behavior of the `fn_recursive_equivalents` SQL function. This branch puts the three relevant numbers -- levels, threshold, and cutoff -- into `PresentationCalculationPolicy`.

Theoretically, `fn_recursive_equivalents` can be used outside of the context of presentation calculation, but in real life, every time we need it it's because we're doing something that is done in that context, e.g. generating an OPDS entry or a search document.

By itself, this doesn't do anything to address 1690. However, I'm now in a position to change the defaults and have those defaults respected by all our code.

In particular, in this branch I'd like to change the default values for 'number of levels of recursion' from 5 to 3. This will cover the cases that are actually important for our cases: Proprietary ID -> ISBN (mainly used on the circulation manager), and ISBN -> OCLC Work ID -> OCLC Number (used on the metadata wrangler). 

The rationale behind setting this value to 5 was to handle Gutenberg ID -> OCLC Work ID -> OCLC Number -> ISBN -> Overdrive ID, which it turns out we don't care about. As is, I believe the  recursion level of 5 is getting the metadata wrangler stuck in ISBN -> ISBN -> ISBN -> ISBN equivalencies created by OCLC Linked Data, and that's the cause of 1690.

I'm not sure that even _this_ will fix 1690, but if not it will give me a better testbed for other fixes. For example, I'm thinking of changing the default cutoff from 1000 to 100.

A potentially simpler solution would be to wipe out all the data from OCLC Linked Data on the metadata wrangler -- it took a long time to gather but I think we might be better off getting rid of it and starting from scratch later. I'd like to hear what you think about that idea.

For context, there are 5.7 million equivalencies recorded in the metadata wrangler database, and 3.6 million of them came from OCLC Linked Data. We could record the 3.6 million records in case we need them later and then delete them. At any rate, this branch would still be useful because we still need these algorithms, we still need to tweak the knobs, and keeping all the knobs in the PresentationCalculationPolicy makes it easy to add new ones or change the defaults.


model/customlist.py
```diff
@@ -265,7 +269,8 @@ def set_work(self, metadata=None, metadata_client=None):
             # Try using the less reliable, more expensive method of
             # matching based on equivalent identifiers.
             equivalent_identifier_id_subquery = Identifier.recursively_equivalent_identifier_ids_query(
-                self.edition.primary_identifier.id, levels=3, threshold=0.5)
+                self.edition.primary_identifier.id, policy=policy
+            )
```
leonardr: Most of this diff is like this -- changing methods that call other methods so that they propagate (or create) a PresentationCalculationPolicy object instead of propagating (or making up) values for one or more of the three tuning variables.

model/edition.py
```diff
@@ -707,7 +721,10 @@ def choose_cover(self):
             # Identifier. Try to find a thumbnail the same way we'd
             # look for a cover.
             for distance in (0, 5):
-                best_thumbnail, thumbnails = self.best_cover_within_distance(distance, rel=LinkRelations.THUMBNAIL_IMAGE)
+                best_thumbnail, thumbnails = self.best_cover_within_distance(
+                    distance=distance, policy=policy,
+                    rel=LinkRelations.THUMBNAIL_IMAGE,
+                )
```
leonardr: `best_cover_within_distance` is weird because we're using part of the incoming policy but not all of it -- we're overriding the `levels` variable and going one level deeper every time we fail to find a cover. That's because we're only going to choose one cover. A cover associated with an Identifier that's closer to the LicensePool is better than a cover associated with an Identifier that's further away, even if the further-away cover is bigger or has a better aspect ratio, because there's less of a risk that it's a cover for a totally different book.

Now that I think about it, the situation is similar with descriptions, but we don't actually pick descriptions this way. I added https://jira.nypl.org/browse/SIMPLY-1696 to address that.

tests/models/test_identifier.py
```diff
@@ -358,28 +387,66 @@ def test_recursively_equivalent_identifier_ids(self):
...
             set(equivs[identifier.id]))
         eq_(set([level_2_equivalent.id,
                  level_3_equivalent.id]),
             set(equivs[level_3_equivalent.id]))
 
-        # The query uses the same db function, but returns equivalents
-        # for all identifiers together so it can be used as a subquery.
+        # By setting a cutoff, you can say to look deep in the tree,
+        # but stop looking as soon as you have a certain number of
+        # equivalents.
```
leonardr: Something weird is going on here. The `fn_recursive_equivalents` function, defined in model/files/recursive_equivalents.sql, takes `cutoff` as an argument but doesn't do anything with it. But these tests show that setting a higher `cutoff` does make a difference. It can't be some old version of  `fn_recursive_equivalents` embedded in my database because 1) I checked its definition and 2) the tests pass on Travis.

leonardr:
Here's another idea -- I could add a `enabled` column to `equivalents` and then disable all the OCLC Linked Data equivalents. That would let us test the hypothesis that OCLC Linked Data is the culprit without deleting the data.

model/work.py
```diff
@@ -101,6 +101,8 @@ class Work(Base):
     CURRENTLY_AVAILABLE = "currently_available"
     ALL = "all"
 
+    DEFAULT_IDENTIFIER_CUTOFF = 100
```
aslagle: I don't think you used this, you have another default constant in another place.

opds.py
```diff
@@ -455,21 +456,22 @@ def add_ratings(cls, work, entry):
...
-        for this work at a given level, cut off the query after
-        approximately this many results. This will improve
-        performance, at the expense of ignoring classifications for
-        identifiers that are distantly related to the work.
+        :param policy: A PresentationCalculationPolicy to
+           use when deciding how deep to go when finding equivalent
+           identifiers for the work.
         """
+        policy = policy or PresentationCalculationPolicy(
+            equivalent_identifier_cutoff=100
```
aslagle: Or you meant to use that constant here?

tests/models/test_identifier.py
aslagle: The cutoff is confusingly referred to as `$4` in the function.

model/work.py
leonardr: Yeah, this was the first thing I wrote before I figured it made more sense to put these things into PresentationCalculationPolicy.

opds.py
leonardr: No, this is OK, but if I reduce the cutoff to 100 everywhere I can get rid of this.

tests/models/test_identifier.py
leonardr: Whew!

PR #1020, opened 2019-02-06
Remove RBdigital from the list of audio sources excluded by default.

leonardr:
This simple branch makes it so if your library has RBdigital audiobooks, they will show up in OPDS feeds once you enable the "Audiobooks" entry point.

Addresses https://jira.nypl.org/browse/SIMPLY-1702


EdwinGuzman:
It'd be nice to be able to control this outside of the codebase.

leonardr:
You can, with a sitewide setting.

PR #1021, opened 2019-02-08
Use TimestampData to keep track of the progress of coverage providers, monitors, and scripts

leonardr:
This branch introduces a new object to the metadata layer: `TimestampData`. This object corresponds to a row in the `timestamps` table.

`Script`, `Monitor`, and `CoverageProvider` all have an architecture where the superclass calls a "do something" method that the subclass is supposed to define. The `TimestampData` object is now how the subclass communicates back to the superclass about what happened in the "do something" method. The superclass is responsible for filling in any missing information from the `TimestampData` and actually writing it to the database.

In the simplest case, where the "do something" method doesn't have any particular status to report, `Script`, `Monitor`, and `CoverageProvider` will capture the start and finish times, record any exceptions, and write the `TimestampData` to the database.

In more complicated cases, any subclass of `Script`, `Monitor`, or `CoverageProvider` has a significant ability to  customize what goes into its `timestamps` row by having its 'do something' method return a `TimestampData`.

`Monitor` is a little different because if there's an unhandled exception we don't want to modify the `start` or `finish` times. So that class still makes direct use of the `Timestamp` object a lot.

`CoverageProvider` is the complicated one. I subclassed `TimestampData` as `CoverageProviderProgress` so that I could use it to keep track of the variable `offset`. This is different from `TimestampData.counter`, which is written to the database and kept across runs. The `offset` is used _within_ a run to skip over certain database rows that are known to be ignorable _for this run_.

(This may seem like overkill, but my next branch is going to add to `CoverageProviderProgress` the ability to keep track of successes and failures _across batches_, so it's really useful to have a single object that's being passed back and forth between calls to `run_once`.)

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -177,19 +177,10 @@ class BaseCoverageProvider(object):
     def run(self):
         start = datetime.datetime.utcnow()
         result = self.run_once_and_update_timestamp()
-        result = result or TimestampData(
-            start=start, finish=datetime.datetime.utcnow()
-        )
-        self.finalize_timestamp(result)
-        return result
 
-    @property
-    def timestamp(self):
-        # Look up the Timestamp object for this CoverageProvider
-        return Timestamp.lookup(
-            self._db, self.service_name, Timestamp.COVERAGE_PROVIDER_TYPE,
-            collection=self.collection
-        )
+        result = result or TimestampData()
+        self.finalize_timestampdata(result, start=start)
+        return result
 
     def run_once_and_update_timestamp(self):
         # First prioritize items that have never had a coverage attempt before.
@@ -235,7 +226,7 @@ class BaseCoverageProvider(object):
                 # The next run_once() call might raise an exception,
                 # so let's write the work to the database as it's
                 # done.
-                self.finalize_timestamp(progress)
+                self.finalize_timestampdata(progress)
 
                 if progress.counter == old_counter:
                     # run_once() did not actually make any progress.
@@ -244,10 +235,21 @@ class BaseCoverageProvider(object):
                     break
         return progress
 
-    def finalize_timestamp(self, timestamp):
+    @property
+    def timestamp(self):
+        """Look up the Timestamp object for this CoverageProvider."""
+        return Timestamp.lookup(
+            self._db, self.service_name, Timestamp.COVERAGE_PROVIDER_TYPE,
+            collection=self.collection
+        )
+
+    def finalize_timestampdata(self, timestamp, **kwargs):
+        """Finalize the given TimestampData and write it to the
+        database.
+        """
         timestamp.finalize(
             self.service_name, Timestamp.COVERAGE_PROVIDER_TYPE,
-            collection=self.collection
+            collection=self.collection, **kwargs
         )
         timestamp.apply(self._db)
         self._db.commit()
--- a/coverage.py
+++ b/coverage.py
@@ -24,7 +24,8 @@ from model import (
     WorkCoverageRecord,
 )
 from metadata_layer import (
-    ReplacementPolicy
+    ReplacementPolicy,
+    TimestampData,
 )
 from util.worker_pools import DatabaseJob
 
@@ -174,7 +175,21 @@ class BaseCoverageProvider(object):
         return self.OPERATION
 
     def run(self):
-        self.run_once_and_update_timestamp()
+        start = datetime.datetime.utcnow()
+        result = self.run_once_and_update_timestamp()
+        result = result or TimestampData(
+            start=start, finish=datetime.datetime.utcnow()
+        )
+        self.finalize_timestamp(result)
+        return result
+
+    @property
+    def timestamp(self):
+        # Look up the Timestamp object for this CoverageProvider
+        return Timestamp.lookup(
+            self._db, self.service_name, Timestamp.COVERAGE_PROVIDER_TYPE,
+            collection=self.collection
+        )
 
     def run_once_and_update_timestamp(self):
         # First prioritize items that have never had a coverage attempt before.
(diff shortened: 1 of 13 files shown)


coverage.py
```diff
@@ -174,7 +189,12 @@ def operation(self):
         return self.OPERATION
 
     def run(self):
-        self.run_once_and_update_timestamp()
+        start = datetime.datetime.utcnow()
+        result = self.run_once_and_update_timestamp()
+
+        result = result or CoverageProviderProgress()
+        self.finalize_timestampdata(result, start=start)
+        return result
```
leonardr: This pattern will also show up in Monitor and Script. The generic method calls the "do something" method, expects a TimestampData back, and writes the TimestampData to the database.

metadata_layer.py
```diff
@@ -502,6 +504,128 @@ def __init__(self, content_type, drm_scheme, link=None, rights_uri=None):
             self.rights_uri = self.link.rights_uri
 
 
+class TimestampData(object):
```
leonardr: This is the object itself. Some pieces of information, like `service_type` and `collection_id`, are provided by the superclass service code -- some CoverageProvider subclass can't decide that it's actually a script. 

Other pieces of information, like `start` and `finish` and `counter`, can be provided by the subclassed "do something" method, but the superclass method will provide default values if they're not provided.

Other pieces of information, like `achievements`, are ignored unless the "do something" method sets them.

`exception` is set by the superclass if the "do something" method raises an exception, but theoretically the "do something method" can set it directly and everything should work.

monitor.py
```diff
@@ -293,49 +309,45 @@ def __init__(self, _db, collection=None, batch_size=None):
...
                 break
 
-        # We're done with this run, either because we completed
-        # a sweep or because an exception was raised.
-        run_ended_at = datetime.datetime.utcnow()
-        timestamp.update(
-            start=run_started_at, finish=run_ended_at,
-            achievements=achievements, counter=new_offset,
-            exception=exception
-        )
```
leonardr: We no longer need to update the timestamp or catch an exception in this method -- we return a TimestampData with the new `counter` if we succeed, and if we raise an exception, `run` will catch it.

monitor.py
```diff
@@ -293,49 +309,45 @@ def __init__(self, _db, collection=None, batch_size=None):
         self.model_class = cls.MODEL_CLASS
         super(SweepMonitor, self).__init__(_db, collection=collection)
 
-    def run(self):
+    def run_once(self, *ignore):
```
leonardr: Bu subclassing `run_once` (the "do something" method) instead of `run`, we get to reuse the code that manages the Timestamp and handles exceptions. That's why code to do those two things is removed from this method.

metadata_layer.py
```diff
@@ -502,6 +504,128 @@ def __init__(self, content_type, drm_scheme, link=None, rights_uri=None):
...
+            self.collection_id = None
+        else:
+            self.collection_id = collection.id
+        if self.start is self.NO_VALUE:
+            self.start = start
+        if self.finish is self.NO_VALUE:
+            if finish is self.NO_VALUE:
+                finish = datetime.datetime.utcnow()
+            self.finish = finish
+        if self.start is self.NO_VALUE:
```
EdwinGuzman: This check is being done above. This is the case where both instances of `start` are `NO_VALUE`? And why do you need to set it to `self.finish`?

monitor.py
```diff
@@ -293,49 +309,45 @@ def __init__(self, _db, collection=None, batch_size=None):
...
+        # we're not trying to capture all the events that happened
+        # since a certain time -- so we're going to make sure the
+        # timestamp is set from the start of the run to the end of the
+        # last _successful_ batch.
         run_started_at = datetime.datetime.utcnow()
-        exception = None
-        achievements = None
+        timestamp.start = run_started_at
+
         while True:
```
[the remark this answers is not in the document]
leonardr: It'll finish if there's an exception (we catch the exception and `break`) or if `process_batch` returns 0 (indicating that the sweep has completed and the database ID reset to zero).

metadata_layer.py
leonardr: Yes, exactly -- neither the constructor nor `finalize` explained  when the service started running. In that case we use the only date we have available -- the time when the service started running. If it's also the case that no one explained when the service _stopped_ running (which is pretty likely) then we use the current time for both dates.

The goal is just to have both dates set to some value for all `Timestamp` objects, so that we can do things like sort by one field or another in database queries.

leonardr:
Sorry for the mess; this ended up more complicated than I hoped.

* Created Timestamp.to_data, which converts a database Timestamp object to an unfinalized TimestampData.
* Defined DatabaseTest.time_eq, which makes it really easy to assert that two datetimes are _almost_ equivalent -- they don't have to be exactly the same.
* Created TimelineMonitor, which abstracts away a very common pattern in monitors.
* Moved some code out of Monitor.timestamp to Monitor.initial_start_time so that it could be used in TimelineMonitor.

Finally, this is the big one: I changed `Monitor.run_once` to take a TimestampData object instead of a `start` and `cutoff` time. If run_once returns a TimestampData, that's the one we finalize and write to the `timestamps` table. This gives a Monitor subclass a lot of flexibility to record its status in TimestampData.achievements, mess around with what the dates in the timestamps mean, etc.

TimelineMonitor brings back the start/cutoff behavior, which let me minimize the changes necessary in circulation. Instead of making a big change to the `run_once` implementations, I changed monitors to subclass `TimelineMonitor`, and renamed `run_once` to `catch_up_from`.

model/coverage.py
```diff
@@ -242,6 +242,13 @@ def update(self, start=None, finish=None, achievements=None,
         # .exception
         self.exception = exception
 
+    def to_data(self):
+        """Convert this Timestamp to an unfinalized TimestampData."""
+        from ..metadata_layer import TimestampData
```
EdwinGuzman: What does the double dots mean for this import statement?

model/coverage.py
```diff
@@ -242,6 +242,13 @@ def update(self, start=None, finish=None, achievements=None,
         # .exception
         self.exception = exception
 
+    def to_data(self):
+        """Convert this Timestamp to an unfinalized TimestampData."""
```
EdwinGuzman: Unfinalized means that it hasn't finished running so what values should it not have until it is finalized?

model/coverage.py
leonardr: We're in the `core.model` package, and we need to import something from the `core` package above. It's analogous to `..` in a shell referring to the parent directory.

This is also why the import statement is inside the function (executed at runtime), rather than at the top of the file (executed on startup). `metadata_layer.py` requires model code to run, and trying to import `metadata_layer` while the model code is loading would create a loop.

model/coverage.py
leonardr: 'Unfinalized' means it's missing the information about which service it's a timestamp for. In this case we're deliberately leaving that data out, and that's because I want to force `finalize` to be called on this timestamp before it's written to the database. And _that's_ because `finalize` does other stuff like fill in defaults for any missing values.

model/coverage.py
```diff
@@ -100,6 +100,11 @@ class Timestamp(Base):
     COVERAGE_PROVIDER_TYPE = "coverage_provider"
     SCRIPT_TYPE = "script"
 
+    # A stand-in value used to indicate that a field in the timestamps
+    # should be set to None. This is necessary because 'None' generally
+    # means 'use the default value'.
+    NO_VALUE = object()
```
EdwinGuzman: Why/when would you want a Timestamp.finish to be able to take on a null value?

leonardr: In most cases it's not necessary. The motivating case is `MWCollectionUpdateMonitor` in circulation, which uses `timestamps.finish` to ask for changes since a certain time. If that monitor gets an empty OPDS feed, we want to leave `timestamps.finish` alone, not update it to the current time.

This is to avoid the possibility that there are pending changes that didn't show up the last time we asked for changes since time X, but will show up the next time we ask.

In the case where `timestamps.finish` starts out None, we want to make sure it stays None.

The BibliothecaEventMonitor used to be the same way -- it used `timestamps.start` and `timestamps.finish` to track the timestamps of the first and last events it saw in a given run. However, that was only for informational purposes, and it wasn't really that interesting, so I was able to convert that monitor over to a TimelineMonitor.

(It might not be that bad to also convert MWCollectionUpdateMonitor, but missing an update from the metadata wrangler is worse than missing one from Bibliotheca, because we have no metadata wrangler equivalent of the BibliothecaCirculationSweep.)

PR #1022, opened 2019-02-10
Improved memory usage when generating large MARC files.

aslagle:
This branch uses multipart S3 uploads and batched queries to reduce the memory usage when creating MARC files for large collections.

I chose the batch sizes by testing with a large local collection, and the maximum memory use I got from this was around 3GB, which is still a lot but it was well over 10GB before. @jonathangreen does that seem good enough?

### DIFF
--- a/marc.py
+++ b/marc.py
@@ -592,52 +592,91 @@ class MARCExporter(object):
 
         return record
 
-    def records(self, lane, annotator, start_time=None, force_refresh=False, mirror=None):
-        output = StringIO()
+    def records(self, lane, annotator, start_time=None, force_refresh=False, mirror=None, query_batch_size=500, upload_batch_size=7500):
+        """
+        Create and export a MARC file for the books in a lane.
+
+        :param lane: The Lane to export books from.
+        :param annotator: The Annotator to use when creating MARC records.
+        :param start_time: Only include records that were created or modified after this time.
+        :param force_refresh: Create new records even when cached records are available.
+        :param mirror: Optional mirror to use instead of loading one from configuration.
+        :param query_batch_size: Number of works to retrieve with a single database query.
+        :param upload_batch_size: Number of records to mirror at a time. This is different
+          from query_batch_size because S3 enforces a minimum size of 5MB for all parts
+          of a multipart upload except the last, but 5MB of records would be too many
+          works for a single query.
+        """
+
+        # We mirror the content, if it's not empty. If it's empty, we create a CachedMARCFile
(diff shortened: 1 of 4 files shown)


marc.py
```diff
@@ -592,52 +592,91 @@ def create_record(cls, work, annotator, force_create=False, integration=None):
...
+        offset = 0
+
+        url = mirror.marc_file_url(self.library, lane, end_time, start_time)
+        representation, ignore = get_one_or_create(
+            self._db, Representation, url=url, media_type=Representation.MARC_MEDIA_TYPE)
+
+        with mirror.multipart_upload(representation, url) as upload:
+            output = StringIO()
+            current_count = 0
+            while offset < total:
```
leonardr: I've spent some time mentally fiddling around with the idea of turning this into a generator and passing the generator into `mirror`. But what I'm actually thinking is that I'd rather see the code inside this loop in a helper method that can be tested separately.

Your existing test coverage is fine, so don't worry aobut this.

marc.py
```diff
@@ -592,52 +592,91 @@ def create_record(cls, work, annotator, force_create=False, integration=None):
...
+
+        # We mirror the content, if it's not empty. If it's empty, we create a CachedMARCFile
+        # and Representation, but don't actually mirror it.
+        if not mirror:
+            storage_protocol = self.integration.setting(self.STORAGE_PROTOCOL).value
+            if storage_protocol == ExternalIntegration.S3:
+                mirror_integration = ExternalIntegration.lookup(
+                    self._db, storage_protocol, ExternalIntegration.STORAGE_GOAL)
+                if mirror_integration:
+                    mirror = mirror or S3Uploader(mirror_integration)
```
leonardr: This seems very similar to `MirrorUploader.sitewide`. Can you use that instead?

aslagle: I think we talked about this in my first MARC PR too. I had originally thought of having an option to store MARC files on disk so it wasn't necessary to set up S3, but I'm not sure if anyone will need that. I didn't like the assumption that there would only ever be one storage option configured though.

leonardr: When we support more than one storage option, this code will probably need to be changed anyway. This code implies that there can be multiple storage options, but only one per protocol.

Not a big deal -- we can revisit this when we add other storage options. My only concern is that when that happens I will probably go through the code looking for references to sitewide() and I might miss this one.

jonathangreen:
@aslagle 3gb seems doable to me, much better than it was before. 👍🏻

PR #1028, opened 2019-02-19
Added migration script that should have been added long ago.

leonardr:
This fixes https://jira.nypl.org/browse/SIMPLY-1727.

### DIFF
--- a/migration/20190219-add-licensepooldeliveries-constraint-if-not-exists.sql
+++ b/migration/20190219-add-licensepooldeliveries-constraint-if-not-exists.sql
@@ -1,0 +1,8 @@
+-- Make sure the unique constraint on licensepooldeliveries exists; it may not exist in older databases.
+DO $$
+  BEGIN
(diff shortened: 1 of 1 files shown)


migration/20190219-add-licensepooldeliveries-constraint-if-not-exists.sql
```diff
@@ -0,0 +1,8 @@
+-- Make sure the unique constraint on licensepooldeliveries exists; it may not exist in older databases.
+DO $$
+  BEGIN
+    alter table licensepooldeliveries add constraint if not exists licensepooldeliveries_data_source_id_identifier_id_delivery_key unique (data_source_id, identifier_id, delivery_mechanism_id, resource_id);
+  EXCEPTION
+    WHEN duplicate_object THEN RAISE NOTICE 'service_type already exists, not creating it.';
```
aslagle: I don't think this is the right message but otherwise looks good.

PR #1029, opened 2019-02-19
Ensure that works and presentation editions for a pool are created in CirculationData.apply.

leonardr:
Amy's work for https://jira.nypl.org/browse/SIMPLY-1722


leonardr:
This looks good.

PR #1030, opened 2019-02-19
Core elastic search self tests

adriana-alter:
https://jira.nypl.org/browse/SIMPLY-1714

Goes with https://github.com/NYPL-Simplified/circulation/pull/1184

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -129,8 +129,14 @@ class ExternalSearchIndex(HasSelfTests):
         """Look up the name of the search index alias."""
         return cls.works_prefixed(_db, cls.CURRENT_ALIAS_SUFFIX)
 
-    def __init__(self, _db, url=None, works_index=None):
+    def __init__(self, _db, url=None, works_index=None, test_search_term=None,
+                 in_testing=False):
+        """Constructor
 
+        :param in_testing: Set this to true if you don't want an
+        Elasticsearch client to be created, e.g. because you're
+        running a unit test of the constructor.
+        """
         self.log = logging.getLogger("External search index")
         self.works_index = None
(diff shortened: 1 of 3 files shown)


external_search.py
```diff
@@ -479,6 +483,18 @@ def remove_work(self, work):
         if self.exists(**args):
             self.delete(**args)
 
+    def _run_self_tests(self, _db, search_service):
+        def _search_for_term():
+            [search_term] = [x.value for x in search_service.settings if x.key == "test_search_term"]
```
leonardr: You should be able to do `search_service.setting("test_search_term").value_or_default("test")` here.

selftest.py
```diff
@@ -179,9 +178,9 @@ def prior_test_results(cls, _db, constructor_method=None, *args, **kwargs):
         constructor_method = constructor_method or cls
         integration = None
         instance = constructor_method(*args, **kwargs)
-        integration = instance.external_integration(_db)
+        integration = instance.external_integration(_db) or instance.search_integration(_db)
```
leonardr: This is another argument for eventually creating subclasses of HasSelfTests that can remove this case-by-case logic from the core HasSelfTests code.

For now, I think it's better to use `isinstance` to be explicit about the classes we are expecting. I think this will work:

```
if isinstance(instance, ExternalSearchIndex):
    integration = instance.search_integration(_db)
else:
    integration = instance.external_integration(_db)
```

leonardr (approved):
Handing over to @EdwinGuzman since both Adriana and I worked on this branch.

EdwinGuzman (approved):

PR #1031, opened 2019-02-22
Track the achivements of CoverageProviders and IdentifierSweepMonitors

leonardr:
This branch makes CoverageProvider and IdentifierSweepMonitor include a record of their accomplishments when they finish running, whether those accomplishments are recorded directly to the database (`timestamps.accomplishments`) or in `TimestampData.accomplishments`.

This branch also adds achievement tracking to `OPDSImportMonitor`, the only monitor in core that does something specific.

`CoverageProvider` already has `CoverageProviderProgress`, a custom subclass of `TimestampData`, and this is where it pays off. By recording the number of successes, transient failures, and persistent failures encountered thus far, a `CoverageProvider` ensures that a detailed record of the progress so far can be written to the database at any point, without having to do a lot of string manipulation.

For `IdentifierSweepMonitors` I changed `process_batch` to return the number of items actually processed (it's possible that a span of database IDs didn't actually contain any matching items). I keep track of this and use it to record a running tally of the number of items processed over the course of a run.

For `TimestampScript` and for `Monitor` in general, what counts as an 'achievement' is not well-defined. It's up to the initial script or monitor. So my next branch will be adding this support to most of the monitors in the circulation manager. I think I'm  going to leave scripts alone; we can add support for `achievements` in scripts as necessary.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -95,6 +95,31 @@ class CoverageProviderProgress(TimestampData):
         # a single run of the CoverageProvider.
         self.offset = 0
 
+        self.successes = 0
+        self.transient_failures = 0
+        self.persistent_failures = 0
+
+    @property
+    def achivements(self):
+        """Represent the achievements of a CoverageProvider as a
+        human-readable string.
+        """
+        template = "Processed %(total)s: %(successes)s, %(transient)s, %(persistent)s"
+        total = (self.successes + self.transient_failures
+                 + self.persistent_failures)
+        p = Timestamp.pluralize
+        return template % dict(
+            total=p(total, "record"),
+            successes=p(self.successes, "success", "successes"),
+            transient=p(self.transient_failures, "transient failure"),
+            persistent=p(self.persistent_failures, "persistent failure"),
+        )
+
+    @achivements.setter
+    def achievements(self, value):
+        # It's not possible to set .achivements directly. Do nothing.
+        pass
+
 
 class BaseCoverageProvider(object):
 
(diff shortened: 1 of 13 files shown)


monitor.py
```diff
@@ -211,6 +215,14 @@ def run_once(self, progress):
...
         :param progress: A TimestampData representing the
            work done by the Monitor up to this point.
+
+        :return: A TimestampData representing how you want the
+            Monitor's entry in the `timestamps` table to look like
+            from this point on.  NOTE: Modifying the incoming
+            `progress` and returning it is generally a bad idea,
+            because the incoming `progress` is full of old
+            data. Instead, return a new TimestampData containing data
+            for only the fields you want to set.
```
leonardr: This turned out to be an important distinction in my circulation branch. Inside `TimelineMonitor.catch_up_from`s it's easier to modify the `progress` object you were given and return it, because `TimelineMonitor.run_once` will always set the timestamps appropriately. But for `Monitor` in general that won't work; you need to create a new TimestampData if you want to record achievements.

leonardr: That's what I do in `OPDSImportMonitor` below.

coverage.py
```diff
@@ -95,6 +95,28 @@ def __init__(self, *args, **kwargs):
...
+                 + self.persistent_failures)
+        return template % (
+            total, self.successes, self.transient_failures,
+            self.persistent_failures
+        )
+
+    @achievements.setter
+    def achievements(self, value):
+        # It's not possible to set .achievements directly. Do nothing.
+        pass
```
aslagle: Wouldn't it be better to raise an error? Or is this needed because of TimestampData?

leonardr: Yeah, the TimestampData constructor sets .achievements to None.

PR #1034, opened 2019-02-26
Don't make HTTP requests from the Overdrive constructor

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1729. It defers the acquisition of the Overdrive OAuth token or collection token until that data is actually needed. Previously this information was gathered in the constructor. This meant that if an Overdrive collection was misconfigured, the site wouldn't start up, because the constructor would raise an unhandled exception.

This also hurt performance on sites with a large number of Overdrive and Overdrive Advantage collections.

### DIFF
--- a/model/credential.py
+++ b/model/credential.py
@@ -54,13 +54,15 @@ class Credential(Base):
 
     @classmethod
     def lookup(self, _db, data_source, type, patron, refresher_method,
-               allow_persistent_token=False):
+               allow_persistent_token=False, allow_empty_token=False):
         from datasource import DataSource
         if isinstance(data_source, basestring):
             data_source = DataSource.lookup(_db, data_source)
         credential, is_new = get_one_or_create(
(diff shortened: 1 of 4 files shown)


model/credential.py
```diff
@@ -54,13 +54,15 @@ class Credential(Base):
 
     @classmethod
     def lookup(self, _db, data_source, type, patron, refresher_method,
-               allow_persistent_token=False):
+               allow_persistent_token=False, allow_empty_token=False):
```
leonardr: This is an unrelated change stemming from my work on the RBdigital API. At one point due to a bug, my database ended up with a persistent Credential object where .credential was None. I think in general, if we see that this happens, the refresher method should be called -- the same as if the token had expired. Otherwise there's no way out of the problem.

However, theoretically `None` could be a valid value for a credential, so I added an argument that lets the user say so.

PR #1035, opened 2019-02-26
Tests for total number of documents, and for number of documents per …

adriana-alter:
…collection

### DIFF
--- a/tests/test_external_search.py
+++ b/tests/test_external_search.py
@@ -287,7 +287,7 @@ class TestExternalSearch(ExternalSearchTest):
 
         eq_("Generating search document for the specified term: 'a search term'", test_results[1].name)
         eq_(True, test_results[1].success)
-        eq_([], test_results[1].result)
+        eq_("[]", test_results[1].result)
 
         eq_("Retrieving raw results for the specified term: 'a search term'", test_results[2].name)
         eq_(True, test_results[2].success)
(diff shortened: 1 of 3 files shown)


tests/test_external_search.py
```diff
@@ -321,7 +321,10 @@ def test__run_self_tests(self):
 
         eq_("Generating search document for the specified term: 'a search term'", test_results[1].name)
         eq_(True, test_results[1].success)
-        eq_("{'author': 'author', 'meta': {'id': 'id'}, 'id': 'id', 'title': 'Sample Book Title'}", test_results[1].result)
+        eq_(
+            '{\n "author": "author", \n "meta": {\n  "id": "id"\n }, \n "id": "id", \n "title": "Sample Book Title"\n}',
+            test_results[1].result
+        )
```
leonardr: This might be easier to test if you parsed `test_results[1].result` as JSON and then looked at the JSON object.

tests/test_external_search.py
```diff
@@ -337,7 +340,7 @@ def test__run_self_tests(self):
 
         eq_("Total number of documents per collection", test_results[5].name)
         eq_(True, test_results[5].success)
-        eq_(("{u'%s': 1}" % collection.name), test_results[5].result)
+        eq_((u'{\n "%s": 1\n}' % collection.name), test_results[5].result)
```
leonardr: Same here.

PR #1039, opened 2019-03-04
Make it easy to find LicensePools with no delivery mechanisms

leonardr:
This is a core branch in support of https://jira.nypl.org/browse/SIMPLY-1711. I'll be using `Collection.pools_with_no_delivery_mechanisms` in an Axis 360 self-test.

### DIFF
--- a/model/licensing.py
+++ b/model/licensing.py
@@ -204,8 +204,14 @@ class LicensePool(Base):
 
     @classmethod
     def with_no_delivery_mechanisms(cls, _db):
-        """Find LicensePools that have no delivery mechanisms."""
-        
+        """Find LicensePools that have no delivery mechanisms.
+
+        :return: A query object.
+        """
+        return _db.query(LicensePool).outerjoin(
+            LicensePool.delivery_mechanisms).filter(
+                LicensePoolDeliveryMechanism.id==None
+            )
 
     @property
     def deliverable(self):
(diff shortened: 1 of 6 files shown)


model/licensing.py
```diff
@@ -1302,6 +1313,8 @@ class DeliveryMechanism(Base, HasFullTableCache):
         (MediaTypes.EPUB_MEDIA_TYPE, NO_DRM),
         (MediaTypes.EPUB_MEDIA_TYPE, ADOBE_DRM),
         (MediaTypes.EPUB_MEDIA_TYPE, BEARER_TOKEN),
+        (None, FINDAWAY_DRM),
+        (MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE, None),
```
aslagle: This seems unrelated. How were audiobooks working before?

leonardr: We don't actually use `default_client_can_fulfill_lookup` for very much, which is why audiobooks were working before. My original idea was to write a self-test that counts books in a collection that are not fulfillable by the default client. That would still be interesting, but it's more responsive to the problem people are having (https://jira.nypl.org/browse/SIMPLY-1711) to find books that have _no_ delivery mechanisms whatsoever.

So this did turn out to be unrelated, but unless we want to get rid of this list altogether (which wouldn't be that hard) I figured we should keep it up to date.

PR #1040, opened 2019-03-07
Better default facets

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1662. Default facets are created by calling the constructor with empty arguments (which fills in library-specific defaults), not by guessing at 'default' values that may be disabled for a given library.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -275,21 +275,7 @@ class Facets(FacetsWithEntryPoint):
     """
     @classmethod
     def default(cls, library):
-        default_collection = library.default_facet(
-            Facets.COLLECTION_FACET_GROUP_NAME
-        )
-        default_availability = library.default_facet(
-            Facets.AVAILABILITY_FACET_GROUP_NAME
-        )
-        default_order = library.default_facet(
-            Facets.ORDER_FACET_GROUP_NAME
-        )
-        return cls(
-            library,
-            collection=default_collection,
-            availability=default_availability,
-            order=default_order,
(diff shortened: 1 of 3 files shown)


lane.py
```diff
@@ -1806,7 +1802,7 @@ def works_in_window(self, _db, facets, target_size):
         )
 
         lane_query = lane_query.order_by(
-            "quality_tier desc", work_model.random.desc()
+            text("quality_tier desc"), work_model.random.desc()
```
leonardr: Without this change, the `default` change caused some test failures -- see https://travis-ci.com/NYPL-Simplified/server_core/jobs/183150692 for an example. The error message was:

`CompileError: Can't resolve label reference for ORDER BY / GROUP BY. Textual SQL expression 'quality_tier desc' should be explicitly declared as text('quality_tier desc')`

I couldn't figure out why this was happening, especially since the tests didn't fail in isolation -- only when you ran the whole suite. But the explicit declaration seemed like good advice anyway.

tests/test_lane.py
```diff
@@ -3099,7 +3111,7 @@ def setup(self):
         random.seed(42)
 
     def test_groups(self):
-        """A comprehensive test of WorkList.groups()"""
+        # A comprehensive test of WorkList.groups()
```
leonardr: I changed these to comments when trying to debug the mysterious test failures. These are the tests that were failing.

PR #1043, opened 2019-03-10
ODL with individual copies

aslagle:
This branch adds a licenses table for tracking individual copies instead of consolidated copies in ODL.

It can't be merged until we have an updated ODL spec, but could be reviewed now.


opds_import.py
```diff
@@ -734,7 +734,7 @@ def extract_feed_data(self, feed, feed_url=None):
         fp_metadata, fp_failures = self.extract_data_from_feedparser(feed=feed, data_source=data_source)
         # gets: medium, measurements, links, contributors, etc.
         xml_data_meta, xml_failures = self.extract_metadata_from_elementtree(
-            feed, data_source=data_source, feed_url=feed_url
+            feed, data_source=data_source, feed_url=feed_url, do_get=self.http_get
```
aslagle: I added this in order to get the ODL status document for each copy during import.

leonardr:
I've looked over this and I understand it, so tentative :+1:

PR #1044, opened 2019-03-21
Fix inconsistencies and remove duplicate code between the timestamp metadata layer and model

leonardr:
The `Timestamp.NO_VALUE` object is being used in inconsistent ways. In `Timestamp` this value means "set the value of this field to None, even if that violates the normal expectations of the Timestamp data model". In `TimestampData` this value means "I don't care what the value of this field is".

This branch renames NO_VALUE to CLEAR_VALUE so that there's no ambiguity in the name. Within `TimestampData` we go back to the standard technique of using `None` to stand in for "I don't care what the value of this field is". It turns out use of `CLEAR_VALUE` is pretty uncommon, so using the standard technique is less confusing than creating a second special constant for "don't care".

This change revealed that there's code in `TimestampData.apply` that duplicates code in `Timestamp.stamp` in an inconsistent way. Since `apply` calls `stamp`, this inconsistency actually made it impossible to use `apply` to force `Timestamp.finish` to be None -- it would be set to `None` in `apply` and then `stamp` would set it to the current time.

This was revealed by my investigations into https://jira.nypl.org/browse/SIMPLY-1796.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -506,6 +506,8 @@ class FormatData(object):
 
 class TimestampData(object):
 
+    CLEAR_VALUE = Timestamp.CLEAR_VALUE
+
     def __init__(self, start=None, finish=None, achievements=None,
                  counter=None, exception=None):
         """A constructor intended to be used by a service to customize its
@@ -542,7 +544,7 @@ class TimestampData(object):
     @property
     def is_failure(self):
         """Does this TimestampData represent an unrecoverable failure?"""
-        return self.exception not in (None, Timestamp.CLEAR_VALUE)
+        return self.exception not in (None, self.CLEAR_VALUE)
 
     @property
     def is_complete(self):
@@ -552,9 +554,7 @@ class TimestampData(object):
         An operation is completed if it has failed, or if the time of its
         completion is known.
         """
-        return self.is_failure or self.finish not in (
-            None, Timestamp.CLEAR_VALUE
-        )
+        return self.is_failure or self.finish not in (None, self.CLEAR_VALUE)
 
     def finalize(self, service, service_type, collection, start=None,
                  finish=None, achievements=None, counter=None,
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -597,18 +597,10 @@ class TimestampData(object):
                 "Not enough information to write TimestampData to the database."
             )
 
-        start = self.start
-        finish = self.finish
-        if start is None and finish is None:
-            start = finish = datetime.datetime.utcnow()
-        elif start is None:
-            start = finish
-        elif finish is None:
-            finish = start
-
         return Timestamp.stamp(
             _db, self.service, self.service_type, self.collection(_db),
-            start, finish, achievements, counter, exception
+            self.start, self.finish, self.achievements, self.counter,
+            self.exception
         )
 
 
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -506,10 +506,10 @@ class FormatData(object):
 
 class TimestampData(object):
 
-    NO_VALUE = Timestamp.NO_VALUE
+    CLEAR_VALUE = Timestamp.CLEAR_VALUE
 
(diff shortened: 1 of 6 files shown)


metadata_layer.py
```diff
@@ -574,56 +574,33 @@ def finalize(self, service, service_type, collection, start=NO_VALUE,
...
         else:
             self.collection_id = collection.id
-        if self.start is self.NO_VALUE:
+        if self.start is None:
             self.start = start
-        if self.finish is self.NO_VALUE:
-            if finish is self.NO_VALUE:
+        if self.finish is None:
+            if finish is None:
                 finish = datetime.datetime.utcnow()
```
leonardr: This bit is also duplicative of what happens in `Timestamp.stamp` and maybe we should take it out.

metadata_layer.py
```diff
@@ -574,56 +574,32 @@ def finalize(self, service, service_type, collection, start=NO_VALUE,
...
+        if self.exception is None:
             self.exception = exception
 
     def collection(self, _db):
         return get_one(_db, Collection, id=self.collection_id)
 
     def apply(self, _db):
-        if any(x is self.NO_VALUE for x in [self.service, self.service_type,
-                                            self.collection_id]):
+        if any(x is None for x in [self.service, self.service_type]):
```
leonardr: When we were using NO_VALUE to mean "value has not been set" it made sense to check collection_id. Now that we're using None, it's ambiguous whether `collection_id=None` means "collection_id set to None" or "collection_id has not been set", but I think it's not too bad -- you're supposed to call `finalize`, which takes `collection` as a required argument, so you have to explicitly set it to `None`.

metadata_layer.py
```diff
@@ -574,56 +574,32 @@ def finalize(self, service, service_type, collection, start=NO_VALUE,
...
-        def _v(x):
-            if x is self.NO_VALUE:
-                return None
-            return x
-
-        achievements = _v(self.achievements)
-        counter = _v(self.counter)
-        counter = _v(self.counter)
-        exception = _v(self.exception)
-
```
leonardr: All this stuff is either duplicated in `stamp`, or I moved it to `stamp`.

model/coverage.py
```diff
@@ -235,24 +239,32 @@ def update(self, start=None, finish=None, achievements=None,
...
             self.counter = counter
 
-        # Unlike the other fields, None is a realistic value for
-        # .exception
+        # Unlike the other fields, None is the default value for
+        # .exception, so passing in None to mean "use the default" and
+        # None to mean "no exception" mean the same thing. But we'll
+        # support CLEAR_VALUE anyway.
+        if exception is self.CLEAR_VALUE:
+            exception = None
```
leonardr: This doesn't seem like a super useful feature but I figured I'd add it to be consistent.

monitor.py
```diff
@@ -244,7 +246,7 @@ class TimelineMonitor(Monitor):
     OVERLAP = datetime.timedelta(minutes=5)
 
     def run_once(self, progress):
-        if progress.finish in (None, progress.NO_VALUE):
+        if progress.finish is None:
```
leonardr: Now that the default value is None, we're never going to start off with `finish` set to `CLEAR_VALUE` -- that's something that only happens on the way _out_ of `run_once`.

tests/test_metadata.py
```diff
@@ -1475,6 +1481,23 @@ def test_apply(self):
...
+        d.achievements = clear
+        d.exception = clear
+        d.apply(self._db)
+
+        eq_(None, timestamp.start)
+        eq_(None, timestamp.finish)
+        eq_(None, timestamp.counter)
+        eq_(None, timestamp.achievements)
+        eq_(None, timestamp.exception)
+
```
leonardr: This is technically redundant since `CLEAR_VALUE` is now dealt with in `Timestamp.stamp` and `apply` no longer touches it, but this is an easy way to verify  that the bug I found has been fixed.

PR #1046, opened 2019-03-27
If there's a RemoteIntegrationException communicating with the metadata wrangler, don't let that derail the process of determining an author's sort name.

leonardr:
This fixes https://jira.nypl.org/browse/SIMPLY-1817.

This is much deeper in the code than I expected based on where the problem manifests -- when refreshing the list of NYT best-sellers -- but it makes sense to me that the call to the metadata wrangler would hpapen in `ContributorData`.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -24,6 +24,7 @@ import re
 from pymarc import MARCReader
 
 from util import LanguageCodes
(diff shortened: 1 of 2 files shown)


leonardr:
Here's the exception as seen in our admin interface, killing the NYT script:

metadata error

PR #1047, opened 2019-04-04
Add column delete collections flag

EdwinGuzman:
Fixes [SIMPLY-1849](https://jira.nypl.org/browse/SIMPLY-1849).

### DIFF
--- a/model/collection.py
+++ b/model/collection.py
@@ -101,7 +101,7 @@ class Collection(Base, HasFullTableCache):
     # When deleting a collection, this flag is set to True so that the deletion
     # script can take care of deleting it in the background. This is
     # useful for deleting large collections which can timeout when deleting.
-    set_to_delete = Column(Boolean, default=False)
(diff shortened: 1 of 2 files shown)


leonardr:
The migration script looks good, but that field also needs to be defined in the `Collection` class so we can access it through SQLAlchemy.

model/collection.py
```diff
@@ -97,6 +98,11 @@ class Collection(Base, HasFullTableCache):
         Integer, ForeignKey('externalintegrations.id'), nullable=True
     )
 
+    # When deleting a collection, this flag is set to True so that the deletion
+    # script can take care of deleting it in the background. This is
+    # useful for deleting large collections which can timeout when deleting.
+    set_to_delete = Column(Boolean, default=False)
+
```
leonardr: This has what I think is your old name for the field -- change it to `marked_for_deletion` and we should be good.

EdwinGuzman: Missed that.

PR #1048, opened 2019-04-04
Make Collection deletion more robust, and implement a ReaperMonitor that deletes Collections flagged for deletion

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1847. It makes four major changes:

* Add cascading deletes to all one-to-many relationships between LicensePool and some other data model object. These are objects that should be deleted along with the LicensePool. Currently these objects are left dangling when a LicensePool is deleted. (We didn't really notice because `Collection` deletion is the only time we actually remove a LicensePool from the database.)
* Implement `Collection.delete()`, which deletes a `Collection`'s LicensePools in batches before deleting the collection itself. This is by far the most time-consuming part of collection deletion, and this change makes that process robust to interruptions.
* Change `Collection` methods that return all `Collection`s matching certain criteria, so that they exclude `Collection`s marked for deletion. This includes the method used by `CollectionMonitor` to find which collections to act on. This change means that scripts which keep a `Collection` up-to-date with some external data source will ignore any collection that's marked for deletion.
* Implement `CollectionReaper`, which looks for `Collection`s that are marked for deletion and calls `delete` on them. Doing this as a `ReaperMonitor` means we don't have to add a new script.

I also made this minor improvement:

* `ReaperMonitor.run_once` returns a `TimestampData` that explains how many items were reaped. Previously it returned nothing, which meant those monitors didn't keep track of their achiements.

### DIFF
--- a/model/collection.py
+++ b/model/collection.py
@@ -13,6 +13,7 @@ from coverage import (
     WorkCoverageRecord,
 )
 from configuration import (
+    CannotLoadConfiguration,
     ConfigurationSetting,
     ExternalIntegration,
 )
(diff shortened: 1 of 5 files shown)


model/collection.py
```diff
@@ -715,6 +725,37 @@ def restrict_to_ready_deliverable_works(
...
+        for i, pool in enumerate(self.licensepools):
+            _db.delete(pool)
+            if not i % 100:
+                _db.commit()
+            # TODO: If the LicensePool's Work has no other
+            # LicensePools, the Work should be removed from the search
+            # index. We can't do this now because we currently have no
+            # way of removing something from the search index. This
+            # only affects search engine performance -- it won't cause
+            # phantom books to show up in search results.
```
aslagle: There is a remove_work method in the search code. Probably won't make much difference but I don't think it would be that hard to do this.

PR #1049, opened 2019-04-05
Add optional new -d parameter to bin/run to delay script startup.

jonathangreen:
This PR is required for: 
https://github.com/NYPL-Simplified/circulation-docker/pull/106

This PR introduces a new optional parameter `-d TIME` that introduces a random delay in starting a script from 0 - TIME minutes. This is done before any other processing in this script.


leonardr (approved):
This makes sense.

PR #1051, opened 2019-04-16
Feature branch - generate OPDS feeds using Elasticsearch, not materialized view

leonardr:
This is a feature branch to implement https://jira.nypl.org/browse/SIMPLY-1533 and https://jira.nypl.org/browse/SIMPLY-1535.

In real-world tests, this work sped up the generation of paginated feeds by about 40%. Generation of grouped feeds was the same as before but overall scalability should benefit from shifting load from the database to Elasticsearch.

In the future we can implement https://jira.nypl.org/browse/SIMPLY-2015 to speed up generation of grouped feeds. But this won't help too much if the time is dominated by OPDS feed generation -- we might do better to keep the OPDS entries in the search index.

This branch breaks Elasticsearch 1 compatibility and will probably not be restoring it. Once we merge this branch into master, we're committed to dropping ES1 support in the next circulation manager release.

Here's the work currently merged into this branch:

* Create a 'v4' search body format that contains all information necessary to implement these two features. (https://jira.nypl.org/browse/SIMPLY-1828)
* Change `Work.to_search_documents` to generate documents in the new format. In particular, the list of `collections` has been replaced with a nested document called `licensepools` which contains information about each of the Work's LicensePools. It's much more complicated to apply an Elasticsearch filter on `licensepools.collection_id` than to apply a filter to `collections.id`, and I had to change the API around a bit to get this to work.
* Change the search classes to support sorting the list of search results by some factor other than relevance. All sort orders currently implemented for paginated feeds are supported. The difficult one is 'sort by availability time', which requires a nested filter on licensepools.collection_id.
* Have a LicensePool flag its Work for reindexing whenever it moves in or out of availability. (https://jira.nypl.org/browse/SIMPLY-1901)
* Flag a Work for reindexing whenever a LicensePool is added or removed (https://jira.nypl.org/browse/SIMPLY-1901)
* Make it possible to generate lists of Elasticsearch results without providing a query string: https://github.com/NYPL-Simplified/server_core/pull/1052 https://jira.nypl.org/browse/SIMPLY-1900
* Implement all of the circulation manager's facets using Elasticsearch filters.
* Add a nested document `customlists` to the schema. This will let us make feeds based on membership in custom lists. (This has to be a nested document because a work might be featured on a list that's not being used to build the feed. Without a nested document, a work that's featured on any list will appear to be featured on all lists.)
* Change `WorkList.works` to get a list of books with an Elasticsearch query instead of a database query. This requires finding any extra database code in `WorkList` or `Facets` that wasn't ported by the work up above, and creating an Elasticsearch equivalent.
* Generate grouped feeds (random samples of high-quality works for multiple lanes) through Elasticsearch. (https://jira.nypl.org/browse/SIMPLY-1990)

Already taken care of:

* Have a Work flag itself for reindexing whenever it's added to or removed from a custom list. (This may already be implemented as a listener.)

To be done in a separate branch:

* Circulation manager: use Elasticsearch filters to generate 'author' (ContributorLane) and 'series' (SeriesLane) feeds.


external_search.py
```diff
@@ -1297,7 +1490,7 @@ def from_worklist(cls, _db, worklist, facets):
     def __init__(self, collections=None, media=None, languages=None,
                  fiction=None, audiences=None, target_age=None,
                  genre_restriction_sets=None, customlist_restriction_sets=None,
-                 facets=None
+                 facets=None, order=None, order_ascending=False
```
leonardr: I think it might be better not to have these arguments -- it could be redundant with what's in `facets`.

external_search.py
```diff
@@ -612,12 +616,163 @@ def latest_body(cls):
...
+                field_description={
+                    "type": type,
+                    # TODO: In some cases we can get away with setting
+                    # index: False here, which would presumably lead
+                    # to a smaller index and faster updates. However,
+                    # it might hurt performance of searches. When this
+                    # code is more mature we can do a side-by-side
+                    # comparison.
+                    "index": True,
+                    "store": True,
```
aslagle: Do these need to be stored?

leonardr: It looks like it's not necessary.

external_search.py
```diff
@@ -612,12 +615,163 @@ def latest_body(cls):
...
+            }
+        )
+
+        # Series must be analyzed (so it can be used in searches) but also present as a keyword
+        # (so it can be used in a filter when listing books from a specific series).
+        basic_string_plus_keyword = dict(basic_string_fields)
+        basic_string_plus_keyword["keyword"] = {
+            "type": "keyword",
+            "index": False,
+            "store": True,
```
aslagle: You also have store here.

leonardr:
Sorry for the churn. I wrote a bunch of tests of 'if x happens then the Work needs to be reindexed' before discovering that some of them were redundant and that there was already a single test containing all such code: `TestWork.test_reindex_on_availability_change`

leonardr:
I'll be merging this into the `3.0beta` prerelease branch, not into master.

PR #1053, opened 2019-05-01
Implement facets as Elasticsearch filters

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1902

The `Facets` object passed in to the `Filter` constructor is now used to apply restrictions on availability and quality, not just to impose a sort order. This means we can now use ElasticSearch to manage every restriction imposed by the `WorkList` a patron may be browsing, _or_ by the `Facets` they may be using to browse.

This should give us everything we need to build paginated lists using ElasticSearch rather than the materialized view.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -1507,7 +1507,6 @@ class Filter(SearchBase):
                  subcollection_filter=None, order=None, order_ascending=False,
                  minimum_featured_quality=0.65
     ):
-
         if isinstance(collections, Library):
             # Find all works in this Library's collections.
             collections = collections.collections
@@ -1517,6 +1516,9 @@ class Filter(SearchBase):
             # 'minumum featured quality' has no meaning. This feature
             # should not be used outside a library context, but just
             # in case, we'll set it to 0 to include everything.
+            #
+            # The Facets object may have a better answer. If so, it
+            # will provide it when we call modify_search_filter().
             minimum_featured_quality = 0
         self.minimum_featured_quality = minimum_featured_quality
 
--- a/external_search.py
+++ b/external_search.py
@@ -1505,10 +1505,10 @@ class Filter(SearchBase):
                  genre_restriction_sets=None, customlist_restriction_sets=None,
                  facets=None,
     ):
+
         if isinstance(collections, Library):
             # Find all works in this Library's collections.
             collections = collections.collections
-
         self.collection_ids = self._filter_ids(collections)
 
         self.media = media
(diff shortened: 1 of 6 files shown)


model/work.py
```diff
@@ -1422,7 +1426,9 @@ def query_to_json_array(query):
                 LicensePool.id.label('licensepool_id'),
                 LicensePool.collection_id.label('collection_id'),
                 LicensePool.open_access.label('open_access'),
-                LicensePool.licenses_available.label('available') > 0,
+                (LicensePool.licenses_available > 0).label('available'),
+                (LicensePool.licenses_owned > 0).label('owned'),
+                work_quality_column,
```
leonardr: The 'main collection' facet uses a nested filter to apply a restriction that involves both LicensePool.open_access and Work.quality. The only way I could get this to work was to put a copy of Work.quality into each LicensePool sub-document. It looks like when you create a nested query on a subdocument, the main document is out of scope. I'm not 100% sure about this though.

external_search.py
```diff
@@ -1493,12 +1503,12 @@ def from_worklist(cls, _db, worklist, facets):
     def __init__(self, collections=None, media=None, languages=None,
                  fiction=None, audiences=None, target_age=None,
                  genre_restriction_sets=None, customlist_restriction_sets=None,
-                 facets=None, order=None, order_ascending=False
+                 facets=None,
```
leonardr: order and order_ascending were redundant since that information can be passed in to a Facets object.

tests/test_external_search.py
```diff
@@ -558,6 +560,12 @@ def setup(self):
             eq_(self.sherlock, sherlock_2)
             eq_(2, len(self.sherlock.license_pools))
 
+            # This book looks good for some search results, but we own
+            # no copies of it, so it will never show up.
+            self.no_copies = _work(title="Moby Dick 2")
+            self.no_copies.license_pools[0].licenses_owned = 0
+
```
leonardr: This tests that the 'ownership' filter works in real situations.

tests/test_external_search.py
```diff
@@ -943,7 +1027,8 @@ def assert_order(sort_field, order, **filter_kwargs):
             """
             expect = self._expect_results
             facets = Facets(
-                self._default_library, None, None, order=sort_field, order_ascending=True
+                self._default_library, Facets.COLLECTION_FULL,
+                Facets.AVAILABLE_ALL, order=sort_field, order_ascending=True
```
leonardr: The `Facets` object can now impose additional filters on test results, so when we create one to test sort order we need to set the facets to values that won't actually filter anything out.

tests/test_external_search.py
```diff
@@ -1221,14 +1306,35 @@ def query(self):
         # The return value is a new MockSearch object based on the one
         # that was passed in.
         assert isinstance(built, MockSearch)
-        eq_(search, built.parent)
+        eq_(search, built.parent.parent)
```
leonardr: There's now an extra step -- a nested filter that's always applied -- so the mock object ends up creating a child _and_ a grandchild, which is what we end up with as `built`.

PR #1054, opened 2019-05-03
Make it possible to paginate search results using Elasticsearch's search_after feature

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1923.

When used to paginate Elasticsearch hits, the normal `Pagination` object uses numeric offsets to reach further and further into the list of hits. Unfortunately, Elasticsearch has a limit on how far this can work -- by default it's 10000 items, or 200 pages the way we paginate things. We need a solution that doesn't cut out at an arbitrary point.

This branch introduces the `SortKeyPagination` class, a subclass of `Pagination` which takes advantage of Elasticsearch's [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html) parameter. It works by storing the sort key of the last item in the current page. When the current page becomes the next page, this sort key is used to tell Elasticsearch where to start fetching results.

Normal `Pagination` still works fine to paginate Elasticsearch hits -- that's how we paginate normal search results, which don't have sort fields since they're ordered by relevance to the search.

Getting this to work requires that the `Pagination` object be notified once the results have been loaded (so that it can store the sort fields for the last item) and that it get a chance to modify the Elasticsearch query before the query runs (similar to the way it currently gets a chance to modify a database query before that query runs).

The other big change in this branch: I introduced to Elasticsearch lists the multi-layered sort field approach currently used in queries against the materialized view. In both cases the default sort order for books is (sort author, sort title, work ID), and it's subject to amendment or modification by the specific field you ask to sort by. So if you sort books by availability time, you're actually sorting them by availability time, sort author, sort title, work ID. If you sort books by title, you're actually sorting them by sort title, sort author, work ID.

This guarantees that every book in the index has a unique sort key (because of the inclusion of work ID) and it guarantees that when you sort by author, you see a given author's books in an order that makes sense (sorted by title)

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -1019,7 +1019,7 @@ class Query(SearchBase):
         self.query_string = query_string
         self.filter = filter
 
-    def build(self, elasticsearch, pagination):
+    def build(self, elasticsearch, pagination=None):
         """Make an Elasticsearch-DSL Search object out of this query.
 
         :param elasticsearch: An Elasticsearch-DSL Search object. This
@@ -1077,8 +1077,12 @@ class Query(SearchBase):
                 search = search.sort(*order_fields)
 
         # Apply any necessary query restrictions imposed by the
-        # Pagination object.
-        pagination.modify_search_query(search)
+        # Pagination object. This may happen through modification or
+        # by returning an entirely new Search object.
+        if pagination:
+            result = pagination.modify_search_query(search)
+            if result is not None:
+                search = result
 
         # All done!
         return search
--- a/external_search.py
+++ b/external_search.py
@@ -1075,9 +1075,9 @@ class Query(SearchBase):
             order_fields = self.filter.sort_order
             if order_fields:
                 search = search.sort(*order_fields)
-            pagination.order_fields = [x.keys()[0] for x in order_fields]
 
-        # Apply any necessary pagination.
+        # Apply any necessary query restrictions imposed by the
+        # Pagination object.
         pagination.modify_search_query(search)
 
         # All done!
@@ -1659,6 +1659,11 @@ class Filter(SearchBase):
     def sort_order(self):
         """Create a description, for use in an Elasticsearch document,
         explaining how search results should be ordered.
+
+        :return: A list of dictionaries, each dictionary mapping a
+        field name to an explanation of how to sort that
+        field. Usually the explanation is a simple string, either
+        'asc' or 'desc'.
         """
         order_fields = []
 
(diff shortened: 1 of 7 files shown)


tests/test_external_search.py
```diff
@@ -924,7 +927,7 @@ def expect(availability, collection, works):
                 order=Facets.ORDER_TITLE
             )
             self._expect_results(
-                works, None, Filter(facets=facets), order=False
+                works, None, Filter(facets=facets), ordered=False
```
leonardr: This was always wrong but it didn't cause problems until I started giving `query_works` any leftover keyword arguments passed into `_expect_results`. "order" is a valid keyword argument for `query_works`, but that's not what's supposed to happen here.

external_search.py
```diff
@@ -406,6 +407,14 @@ def query_works(self, query_string, filter=None, pagination=None,
                     i, result.title, result.author, result.meta['id'],
                     result.meta['score'] or 0, result.meta['shard']
                 )
+
+        # Convert the Search object into a list of hits.
+        results = [x for x in results]
```
leonardr: This introduces a tiny inefficiency because in most cases we iterate over the list twice, but the list isn't that big.

external_search.py
```diff
@@ -1063,9 +1072,17 @@ def build(self, elasticsearch):
 
         # Apply any necessary sort order.
         if self.filter:
-            order = self.filter.sort_order
-            if order is not None:
-                search = search.sort(order)
+            order_fields = self.filter.sort_order
+            if order_fields:
+                search = search.sort(*order_fields)
```
leonardr: There's now always multiple order_fields, since at a minimum we order by author, title, and work ID.

model/work.py
```diff
@@ -1567,6 +1567,7 @@ def query_to_json_array(query):
         # search document.
         search_data = select(
             [works_alias.c.work_id.label("_id"),
+             works_alias.c.work_id.label("work_id"),
```
leonardr: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html recommends making a copy of the data in `_id` rather than using `_id` as the tiebreaker field.

aslagle (approved):

model/work.py
leonardr: I don't think this per se requires a new migration. The feature branch creates a new version of the index, which should automatically trigger a full index rebuild. All the changes I've been making to this method (with more to come) will take effect then.

We probably need some kind of migration to mark the dropping of ES1 support though.

PR #1056, opened 2019-05-07
Filter more works in search index

leonardr:
This is a precursor to the branch that will fix https://jira.nypl.org/browse/SIMPLY-1903. It brings to Elasticsearch most of the functionality found in `Collection.restrict_to_ready_deliverable_works`.

* If a library doesn't allow holds, books that are not currently available are excluded from search results.
* If the site filters out audiobooks from certain sources, those audiobooks are excluded from search results.
* LicensePools that are suppressed get indexed (because it's good to have the ElasticSearch index accurately reflect the database) but are excluded from search results.
* Works that are not presentation-ready get indexed (ditto) but are excluded from search results.

These features of materialized view queries are not yet implemented:

* The filter that removes books if they have no delivery mechanisms. I tried to get this working and it was too difficult -- hopefully we can come back to it later.
* The hack that hides children's books from Project Gutenberg (which are frequently not suitable for modern children). I don't think we have any of this content on live library sites anymore, which makes this less necessary. But I might implement it just to be safe.

Without this code, we're relying on `Collection.restrict_to_ready_deliverable_works` to filter out unsuitable works _after_ we get them from the search index. If we get unlucky, we might find 50 works in the index but they all get filtered out by  `Collection.restrict_to_ready_deliverable_works`, which would give the incorrect impression that the list has reached its end. So it's better to do as much filtering as possible inside the Elasticsearch query.

After much experimentation, I decided to implement this by defining some 'universal' filters that are always applied to every outgoing query. There are separate methods for generating a universal base filter (the filter on presentation-ready works) and universal nested filters (everything else). This makes it a little annoying to test `Filter.build`, but as we add more stuff to the class methods that generate the universal filters, we'll only have to change the tests of those methods.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -1737,7 +1737,7 @@ class Filter(SearchBase):
         # TODO: It would be great to be able to filter out
         # LicensePools that have no delivery mechanisms. That's the
         # only part of Collection.restrict_to_ready_deliverable_works
-        # not implemented here or in .
+        # not already implemented in this class.
 
         # TODO: If we wanted to implement the Gutenberg audience hack
         # for Elasticsearch (lane.py:WorkList.audience_filter_clauses)
--- a/external_search.py
+++ b/external_search.py
@@ -34,6 +34,8 @@ from facets import FacetConstants
 from model import (
     numericrange_to_tuple,
     Collection,
+    ConfigurationSetting,
+    DataSource,
     ExternalIntegration,
     Library,
     Work,
@@ -1514,7 +1516,7 @@ class Filter(SearchBase):
         excluded = (
             ConfigurationSetting.excluded_audio_data_sources(_db)
         )
-        excluded_audio_data_sources = [
+        excluded_audiobook_data_sources = [
             DataSource.lookup(_db, x) for x in excluded
         ]
 
@@ -1522,7 +1524,7 @@ class Filter(SearchBase):
             library, media, languages, fiction, audiences,
             target_age, genre_id_restrictions, customlist_id_restrictions,
             facets,
-            excluded_audio_data_sources=excluded_audio_data_sources,
+            excluded_audiobook_data_sources=excluded_audiobook_data_sources,
             allow_holds=library.allow_holds,
         )
 
--- a/external_search.py
+++ b/external_search.py
@@ -1039,11 +1039,15 @@ class Query(SearchBase):
         query = self.query()
         nested_filters = defaultdict(list)
 
-        # Do the initial build of the filter, and apply restrictions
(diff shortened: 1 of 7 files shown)


external_search.py
```diff
@@ -1656,9 +1709,83 @@ def build(self, _chain_filters=None):
             range_query = self._match_range(
                 'quality', 'gte', self.minimum_featured_quality
             )
-            f = chain(f, range_query)
+            f = chain(f, F('bool', must=range_query))
```
leonardr: When `range_query` is a dictionary and not a real `Filter`, chaining it with the universal base caused a problem. Rather than try to make `_match_range` return a real `Filter` (which seems really tough), I create a boolean `Filter` that just has one argument in `must`.

model/work.py
```diff
@@ -1441,10 +1453,11 @@ def query_to_json_array(query):
         ).where(
             and_(
                 LicensePool.work_id==work_id_column,
+                work_presentation_edition_id_column==Edition.id,
```
leonardr: The absence of this little bit causes a bunch of problems that aren't caught by test_search_document, because that method only tests one work. So I added the "just a basic check" test in test_external_search to check for overzealous joins in general.

tests/test_external_search.py
```diff
@@ -421,60 +432,47 @@ def setup(self):
             self.moby_dick.presentation_edition.series = "Classics"
             self.moby_dick.summary_text = "Ishmael"
             self.moby_dick.presentation_edition.publisher = "Project Gutenberg"
-            self.moby_dick.set_presentation_ready()
```
leonardr: This is never necessary because `_work()` sets the work as presentation-ready.

tests/test_external_search.py
```diff
@@ -563,11 +548,29 @@ def setup(self):
...
+            self.not_presentation_ready = _work(title="Moby Dick 2")
+            self.not_presentation_ready.presentation_ready = False
+
+        # Just a basic check to make sure the search document query
+        # doesn't contain over-zealous joins. This is the main place
+        # where we make a large number of works and generate search
+        # documents for them.
+        eq_(1, len(self.moby_dick.to_search_document()['licensepools']))
+        eq_("Audio",
+            self.pride_audio.to_search_document()['licensepools'][0]['medium'])
```
aslagle: It's a little weird to be testing something inside `setup`.

leonardr:
> I could imagine a case where we might not want the universal filter if we had a special admin-specific search, but we can deal with that if we ever do it.

Yeah, we can add an argument to exclude the universal filters.

PR #1057, opened 2019-05-14
Make works_for_specific_ids capable of returning Work objects

leonardr:
This is a precursor branch for https://jira.nypl.org/browse/SIMPLY-1903.

It changes WorkList.works_for_specific_ids so that it can return either Work objects or MaterializedWorkWithView objects. This restores behavior that was removed about a year ago when we thought we could use the materialized view for everything.

The (forthcoming) method that runs lane queries against Elasticsearch is going to use `works_for_specific_ids` to look up the works for search results, so it can turn them into OPDS feeds. But if it looks up those works in the materialized view, we lose the main advantage of Elasticsearch -- new works show up in the search index almost immediately.

Once we get rid of the materialized view we'll change this method again so that it _only_ works with `Work` as its work model, but this preserves backwards compatibility for now.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1324,7 +1324,6 @@ class WorkList(object):
         :return: A Query, or None if the WorkList is deemed to be a
            bad idea in the first place.
         """
-        mw = mw
         # apply_filters() will apply the genre
         # restrictions.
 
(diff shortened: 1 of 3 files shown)


lane.py
```diff
@@ -72,7 +72,7 @@
     Library,
     LicensePool,
     LicensePoolDeliveryMechanism,
-    MaterializedWorkWithGenre as work_model,
+    MaterializedWorkWithGenre as mw,
```
leonardr: Since `MaterializedWorkWithGenre` is no longer the only "work model" in play in this file, I changed the import name. This causes most of the line diffs in this branch.

PR #1059, opened 2019-05-17
Build paginated feeds from the search index instead of the materialized view

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-1972.

In some simple tests, I found that this change speeds up the creation of paginated OPDS feeds by about 40%.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -763,7 +763,7 @@ class ExternalSearchIndexVersions(object):
         licensepool_fields_by_type = {
             'integer': ['collection_id', 'data_source_id'],
             'date': ['availability_time'],
-            'boolean': ['availability', 'open_access', 'suppressed'],
+            'boolean': ['availability', 'open_access', 'suppressed', 'licensed'],
             'keyword': ['medium'],
         }
         licensepool_definition = cls.map_fields_by_type(
--- a/external_search.py
+++ b/external_search.py
@@ -1774,7 +1774,7 @@ class Filter(SearchBase):
         not_suppressed = F('term', **{'licensepools.suppressed' : False})
         nested_filters['licensepools'].append(not_suppressed)
 
-        owns_licenses = F('term', **{'licensepools.owned' : True})
+        owns_licenses = F('term', **{'licensepools.licensed' : True})
         open_access = F('term', **{'licensepools.open_access' : True})
         currently_owned = F('bool', should=[owns_licenses, open_access])
         nested_filters['licensepools'].append(currently_owned)
--- a/external_search.py
+++ b/external_search.py
@@ -396,7 +396,6 @@ class ExternalSearchIndex(HasSelfTests):
         search = self.create_search_doc(query_string, filter=filter, pagination=pagination, debug=debug, return_raw_results=return_raw_results)
         start = pagination.offset
         stop = start + pagination.size
-
         a = time.time()
         # NOTE: This is the code that actually executes the ElasticSearch
         # request.
--- a/external_search.py
+++ b/external_search.py
@@ -396,6 +396,7 @@ class ExternalSearchIndex(HasSelfTests):
         search = self.create_search_doc(query_string, filter=filter, pagination=pagination, debug=debug, return_raw_results=return_raw_results)
         start = pagination.offset
         stop = start + pagination.size
+
         a = time.time()
         # NOTE: This is the code that actually executes the ElasticSearch
         # request.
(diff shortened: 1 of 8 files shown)


external_search.py
```diff
@@ -1775,7 +1774,7 @@ def universal_nested_filters(cls):
         not_suppressed = F('term', **{'licensepools.suppressed' : False})
         nested_filters['licensepools'].append(not_suppressed)
 
-        owns_licenses = F('term', **{'licensepools.owned' : True})
+        owns_licenses = F('term', **{'licensepools.licensed' : True})
```
leonardr: While debugging this, I found the 'owned' language confusing -- open access books are not 'owned' -- so I renamed the field to 'licensed'.

lane.py
```diff
@@ -1356,27 +1353,12 @@ def works(self, _db, facets=None, pagination=None, include_quality_tier=False):
                 mw.collection_id.in_(self.collection_ids)
             )
         qu = self.apply_filters(_db, qu, facets, pagination)
-        if qu:
```
leonardr: I'm pretty sure this conditional was only necessary because of a bad mock implementation of `apply_filters` in a test. If this method returns None instead of a real query, the calling code isn't going to work.

lane.py
```diff
@@ -1356,27 +1353,12 @@ def works(self, _db, facets=None, pagination=None, include_quality_tier=False):
...
-                # else who uses this.)
-
-                # These speed up the process of generating acquisition links.
-                joinedload("license_pool", "delivery_mechanisms"),
-                joinedload("license_pool", "delivery_mechanisms", "delivery_mechanism"),
-                # These speed up the process of generating the open-access link
-                # for open-access works.
-                joinedload("license_pool", "delivery_mechanisms", "resource"),
-                joinedload("license_pool", "delivery_mechanisms", "resource", "representation"),
-            )
```
leonardr: Very similar code needs to run whether the objects being fetched are `MaterializedWorkWithGenre` or `Work`. I moved this code into `_modify_loading` (formerly `_lazy_load`).

lane.py
```diff
@@ -1385,9 +1367,9 @@ def works_from_search_index(
             Filter,
             ExternalSearchIndex,
         )
-        search_client = search_client or ExternalSearchIndex(_db)
+        search_engine = search_engine or ExternalSearchIndex(_db)
```
leonardr: It's called a 'search engine' in opds.py so I made it consistent here. It's probably not consistent throughout the code base though.

tests/test_external_search.py
```diff
@@ -890,6 +890,14 @@ def test_query_works(self):
...
         expect([self.pride], "pride and prejudice", f)
 
+        # Here, a different data source is excluded, and it shows up.
+        f = Filter(
+            excluded_audiobook_data_sources=[
+                DataSource.lookup(self._db, DataSource.BIBLIOTHECA)
+            ]
+        )
+        expect([self.pride, self.pride_audio], "pride and prejudice", f)
+
```
leonardr: This test isn't related to any of the code in this branch; I added it while debugging a problem that turned out to be unrelated, and keeping it seems like a good idea.

tests/test_lane.py
```diff
@@ -1709,6 +1709,7 @@ def test_works_propagates_facets(self):
         class Mock(WorkList):
             def apply_filters(self, _db, qu, facets, pagination):
                 self.apply_filters_called_with = facets
+                return qu
```
leonardr: This is the bad mock method I mentioned earlier.

tests/test_opds.py
```diff
@@ -921,14 +921,16 @@ def test_page_feed(self):
...
+        search_engine.bulk_update([work1, work2])
+
         facets = Facets.default(self._default_library)
         pagination = Pagination(size=1)
 
         def make_page(pagination):
             return AcquisitionFeed.page(
                 self._db, "test", self._url, lane, TestAnnotator,
-                pagination=pagination
+                pagination=pagination, search_engine=search_engine
```
leonardr: Changing these tests was surprisingly easy. Instead of adding the works to the materialized view, I add them to a mock search index. Everything else works as-is.

external_search.py
```diff
@@ -763,7 +763,7 @@ def v4_body(cls):
         licensepool_fields_by_type = {
             'integer': ['collection_id', 'data_source_id'],
             'date': ['availability_time'],
-            'boolean': ['availability', 'open_access', 'suppressed'],
+            'boolean': ['availability', 'open_access', 'suppressed', 'licensed'],
```
aslagle: What is "availability"? Shouldn't that be "available"?

PR #1060, opened 2019-05-20
Allow case-insensitive sort

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1986 by using the ICU collation algorithm to order `sort_title` and `sort_author` rather than relying on the default Elasticsearch sort. There are some differences between the default Elasticsearch algorithm and how feeds have been sorted up to this point, but only two of them are a big deal: 

1. Sorting is case-sensitive: `van Vogt` sorts after `Zelazny`.
2. `[Unknown]` used to sort near the end of a list and now it sorts at the very top.

To fix these problems I introduced a custom `analyzer` with a custom `char_filter` that removes square brackets. It also removes periods, so that "H.G. Wells" and "HG Wells" sort the same. ("H. G. Wells" is a whole other problem which I wasn't able to solve.) I used https://www.elastic.co/guide/en/elasticsearch/guide/current/sorting-collations.html and https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pattern-replace-charfilter.html as my guides.

There's a much simpler solution that _almost_ works. You can define a `keyword` field as a `icu_collation_keyword` instead and get that collation algorithm without having to define any custom behavior. https://www.elastic.co/guide/en/elasticsearch/plugins/6.4/analysis-icu-collation-keyword-field.html

But an `icu_collation_keyword` field can't have a custom `normalizer`, so to get the `char_filter` you have to define a custom `analyzer`. But an `icu_collation_keyword` also can't have a custom `analyzer`, so I had to also effectively reproduce what `icu_collation_keyword` does within a `text` field. It looks like this is  just an oversight.

A simpler way of doing this would be to remove the square brackets and any other undesirable characters while building the search document. Then we could use normal `icu_collation_keyword`. Let me know if you'd rather do things this way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ install:
   # a low-resource development environment.
   - "echo 'transport.host: 127.0.0.1' >> ./${ES_PATH}/config/elasticsearch.yml"
   - "echo 'http.host: 0.0.0.0' >> ./${ES_PATH}/config/elasticsearch.yml"
+  - ./${ES_PATH}/bin/elasticsearch-plugin install analysis-icu
   - ./${ES_PATH}/bin/elasticsearch &
   - pip install -r requirements.txt
   - python -m textblob.download_corpora
--- a/external_search.py
+++ b/external_search.py
@@ -641,20 +641,24 @@ class ExternalSearchIndexVersions(object):
             to field names.
         """
         for type, fields in fields_by_type.items():
+            description={
+                "type": type,
+                # TODO: In some cases we can get away with setting
+                # index: False here, which would presumably lead
+                # to a smaller index and faster updates. However,
+                # it might hurt performance of searches. When this
+                # code is more mature we can do a side-by-side
+                # comparison.
+                "index": True,
+                "store": False,
+            }
+            if type == 'keyword':
+                description['normalizer'] = "sortable"
+
             mapping = cls.map_fields(
                 fields=fields,
-                field_description={
-                    "type": type,
-                    # TODO: In some cases we can get away with setting
-                    # index: False here, which would presumably lead
-                    # to a smaller index and faster updates. However,
-                    # it might hurt performance of searches. When this
-                    # code is more mature we can do a side-by-side
-                    # comparison.
-                    "index": True,
-                    "store": False,
-                },
-                mapping=mapping
+                mapping=mapping,
+                field_description=description,
             )
         return mapping
 
@@ -688,6 +692,13 @@ class ExternalSearchIndexVersions(object):
                         "type": "stemmer",
                         "name": "english"
                     },
+                },
+                "normalizer": {
+                    "sortable" : {
+                        "type": "custom",
+                        "char_filter": [],
+                        "filter": ["lowercase", "asciifolding"],
+                    }
                 },
                 "analyzer" : {
                     "en_analyzer": {
@@ -733,6 +744,10 @@ class ExternalSearchIndexVersions(object):
             "type": "keyword",
             "index": False,
             "store": False,
+            # We don't sort on series, but we do filter, and this will
+            # group together books in the same series if the series
+            # name differs by capitalization or placement of accents.
+            "normalizer": "sortable"
         }
         mapping = cls.map_fields(
             fields=["series"],
--- a/external_search.py
+++ b/external_search.py
@@ -690,7 +690,8 @@ class ExternalSearchIndexVersions(object):
         # by title, not by the other contributors.
         ("primary_author_only", "\s+;.*", ""),
 
-        # Remove periods from consideration.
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -35,6 +35,7 @@ install:
   # a low-resource development environment.
   - "echo 'transport.host: 127.0.0.1' >> ./${ES_PATH}/config/elasticsearch.yml"
   - "echo 'http.host: 0.0.0.0' >> ./${ES_PATH}/config/elasticsearch.yml"
+  - ./${ES_PATH}/bin/elasticsearch-plugin install analysis-icu
```
leonardr: This plugin provides the `icu_collation` filter type and the `icu_collation_keyword` field type.
It looks like it's included in AWS Elasticsearch instances by default, but it wasn't on my laptop and obviously it's not here unless we put it here.

external_search.py
```diff
@@ -194,6 +197,10 @@ def __init__(self, _db, url=None, works_index=None, test_search_term=None,
         if works_index and integration:
             try:
                 self.set_works_index_and_alias(_db)
+            except RequestError, e:
+                # This is almost certainly a problem with our code,
+                # not a communications error.
+                raise e
```
leonardr: I made this change to make it easier to debug problems with the body document.

external_search.py
```diff
@@ -641,20 +648,31 @@ def map_fields_by_type(cls, fields_by_type, mapping=None):
...
+                # TODO: In some cases we can get away with setting
+                # index: False here, which would presumably lead
+                # to a smaller index and faster updates. However,
+                # it might hurt performance of searches. When this
+                # code is more mature we can do a side-by-side
+                # comparison.
+                "index": True,
+                "store": False,
+            }
+            if type == 'icu_collation_keyword':
```
leonardr: For the most part, this lets us pretend that `icu_collation_keyword` works when we're defining fields.

external_search.py
```diff
@@ -763,7 +809,7 @@ def v4_body(cls):
         licensepool_fields_by_type = {
             'integer': ['collection_id', 'data_source_id'],
             'date': ['availability_time'],
-            'boolean': ['availability', 'open_access', 'suppressed', 'licensed'],
+            'boolean': ['availabile', 'open_access', 'suppressed', 'licensed'],
```
leonardr: This is the change you mentioned earlier today; I put it in this branch by mistake.

tests/test_external_search.py
```diff
@@ -1035,22 +1037,20 @@ def setup(self):
...
             self.moby_dick.presentation_edition.series = "Classics"
             self.moby_dick.presentation_edition.series_position = 10
             self.moby_dick.summary_text = "Ishmael"
             self.moby_dick.presentation_edition.publisher = "Project Gutenberg"
-            self.moby_dick.set_presentation_ready()
             self.moby_dick.random = 0.1
             self.moby_dick.last_update_time = datetime.datetime.now()
 
-            self.moby_duck = _work(title="Moby Duck", authors="Donovan Hohn", fiction=False)
+            self.moby_duck = _work(title="Moby Duck", authors="donovan hohn", fiction=False)
```
leonardr: This changes prove that sorting is case-insensitive.

external_search.py
aslagle: Now you spelled it wrong :)

aslagle:
Would be good to add a test for the H.G. Wells case, otherwise looks good.

leonardr:
I split out the regular expressions from the Elasticsearch document definition so that I could test those separately.

Making them easy to test also made me comfortable adding some more sophisticated normalizing regexes.

But those regexes only apply to author names -- titles should probably be left alone -- so I made sort_author its own special analyzer.

https://jira.nypl.org/browse/SIMPLY-1986 shows some illustrations of how author name normalization works now.

external_search.py
```diff
@@ -641,23 +648,64 @@ def map_fields_by_type(cls, fields_by_type, mapping=None):
...
+    for name, pattern, replacement in [
+        # The special author name "[Unknown]" should sort after everything
+        # else. REPLACEMENT CHARACTER is the final valid Unicode character.
+        ("unknown_author", "\[Unknown\]", u"\N{REPLACEMENT CHARACTER}"),
+
+        # Works by a given primary author should be secondarily sorted
+        # by title, not by the other contributors.
+        ("primary_author_only", "\s+;.*", ""),
+
+        # Remove periods from consideration.
```
aslagle: This comment is incorrect.

leonardr:
> I wonder if you could use a subgroup with `(([A-Z] )+)` to capture an arbitrary number of initials.

That's a good idea but it doesn't look like Java's regex format supports subgroups.

PR #1062, opened 2019-05-23
Choose featureable works through Elasticsearch

leonardr:
This branch makes it possible to use `FeaturedFacets` with Elasticsearch. The complex SQL that creates the 'quality tier field' is duplicated with much simpler Elasticsearch code that creates a number of function scores (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html). All function scores are applied to each document and their sum is used as the score for that document.

We have function scores for work quality, current availability, being featured on a relevant custom list, and a random element that ensures variety.

Early tests against a full-sized index give results similar to what is currently seen in the circulation manager.

### DIFF
--- a/tests/test_external_search.py
+++ b/tests/test_external_search.py
@@ -1428,13 +1428,13 @@ class TestFeaturedFacets(EndToEndExternalSearchTest):
         self.hq_available.quality = 1
 
         self.hq_available_2 = _work(title="Also HQ and available")
-        self.hq_available.quality = 1
+        self.hq_available_2.quality = 1
 
         self.not_featured_on_list = _work(title="On a list but not featured")
-        self.not_featured_on_list.quality = 0.79999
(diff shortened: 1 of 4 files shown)


tests/test_external_search.py
```diff
@@ -1405,6 +1419,124 @@ def test_exact_matches(self):
...
+            "Works without considering quality",
+            worklist, all_featured_facets,
+            [self.hq_available, self.hq_available_2,
+             self.not_featured_on_list, self.hq_not_available,
+             self.featured_on_list],
+        )
+
+        # Up to this point we've been avoiding the random element,
+        # but we can introduce that now by passing in a numeric seed.
+        # In normal usage, the current time is the
```
aslagle: the seed?

aslagle (approved):

PR #1064, opened 2019-05-28
Create OPDS grouped feeds through Elasticsearch, not materialized view

leonardr:
This branch implements https://jira.nypl.org/browse/SIMPLY-1990.
Elasticsearch, not the materialized view, now drives `WorkList.groups`.
The code to generate grouped lists of works from the materialized view is still present and tested but no longer used.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -446,16 +446,9 @@ class ExternalSearchIndex(HasSelfTests):
         needs_add = []
         successes = []
         for work in works:
-            if work.presentation_ready:
-                needs_add.append(work)
-            else:
-                # Works are removed one at a time, which shouldn't
-                # pose a performance problem because works almost never
-                # stop being presentation ready.
-                self.remove_work(work)
-                successes.append(work)
+            needs_add.append(work)
 
-        # Add any works that need adding.
+        # Add/update any works that need adding/updating.
         docs = Work.to_search_documents(needs_add)
 
         for doc in docs:
@@ -471,8 +464,6 @@ class ExternalSearchIndex(HasSelfTests):
 
         # If the entire update failed, try it one more time before
         # giving up on the batch.
-        #
-        # Removed works were already removed, so no need to try them again.
         if len(errors) == len(docs):
             if retry_on_batch_failure:
                 self.log.info("Elasticsearch bulk update timed out, trying again.")
--- a/testing.py
+++ b/testing.py
@@ -1002,6 +1002,10 @@ class ExternalSearchTest(DatabaseTest):
     def setup(self):
         super(ExternalSearchTest, self).setup(mock_search=False)
 
+        # Track the indexes created so they can be torn down at the
+        # end of the test.
+        self.indexes = []
+
         self.integration = self._external_integration(
             ExternalIntegration.ELASTICSEARCH,
             goal=ExternalIntegration.SEARCH_GOAL,
@@ -1021,13 +1025,21 @@ class ExternalSearchTest(DatabaseTest):
                 exc_info=e
             )
 
+    def setup_index(self, new_index):
+        "Create an index and register it to be destroyed during teardown."
+        self.search.setup_index(new_index=new_index)
+        self.indexes.append(new_index)
+
     def teardown(self):
         if self.search:
+            # Delete the works_index, which is almost always created.
             if self.search.works_index:
-                self.search.indices.delete(self.search.works_index, ignore=[404])
-            self.search.indices.delete('the_other_index', ignore=[404])
-            self.search.indices.delete('test_index-v100', ignore=[404])
-            self.search.indices.delete('test_index-v9999', ignore=[404])
+                self.search.indices.delete(
+                    self.search.works_index, ignore=[404]
+                )
+            # Delete any other indexes created over the course of the test.
+            for index in self.indexes:
+                self.search.indices.delete(index, ignore=[404])
             ExternalSearchIndex.reset()
         super(ExternalSearchTest, self).teardown()
 
--- a/testing.py
+++ b/testing.py
@@ -446,7 +446,7 @@ class DatabaseTest(object):
         identifier = license_pool.identifier
         content_type = Representation.EPUB_MEDIA_TYPE
         drm_scheme = DeliveryMechanism.NO_DRM
-        LicensePoolDeliveryMechanism.set(
+        return LicensePoolDeliveryMechanism.set(
             data_source, identifier, content_type, drm_scheme,
             RightsStatus.IN_COPYRIGHT
         )
--- a/testing.py
+++ b/testing.py
@@ -1050,7 +1050,7 @@ class EndToEndSearchTest(ExternalSearchTest):
 
     def setup(self):
         super(EndToEndSearchTest, self).setup()
-        
+
         # Create some works.
         if not self.search:
             # No search index is configured -- nothing to do.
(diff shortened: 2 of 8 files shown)


external_search.py
```diff
@@ -446,16 +446,9 @@ def bulk_update(self, works, retry_on_batch_failure=True):
...
         for work in works:
-            if work.presentation_ready:
-                needs_add.append(work)
-            else:
-                # Works are removed one at a time, which shouldn't
-                # pose a performance problem because works almost never
-                # stop being presentation ready.
-                self.remove_work(work)
-                successes.append(work)
+            needs_add.append(work)
```
leonardr: This change is not strictly related to this branch; it could have been made before but I didn't notice earlier. We now store works that are not presentation-ready in the search index, so `bulk_update` should include them rather than trying to remove them.

testing.py
```diff
@@ -968,6 +976,161 @@ def sample_cover_representation(self, name):
...
+    This makes search results more predictable.
+    """
+
+    def setup_index(self, new_index=None):
+        return super(SearchClientForTesting, self).setup_index(
+            new_index, number_of_shards=1, number_of_replicas=0
+        )
+
+
+class ExternalSearchTest(DatabaseTest):
```
leonardr: This code was previously in tests/external_search.py. It worked out because there was nowhere else we needed to run an end-to-end test using a real Elasticsearch instance. But now the real-world test of `WorkList.groups()` uses an Elasticsearch instance, so I moved this class to testing.py, making it usable from any test file.

testing.py
```diff
@@ -968,6 +976,161 @@ def sample_cover_representation(self, name):
...
+
+    def setup(self):
+        super(EndToEndSearchTest, self).setup()
+
+        # Create some works.
+        if not self.search:
+            # No search index is configured -- nothing to do.
+            return
+
+        self.populate_works()
```
leonardr: This is the only change I made when moving the class between files. Putting all the code to make the works into `populate_works` means the works can be added to the search index within `setup` -- this saves code and indent levels in every subclass of this class.

tests/test_external_search.py
```diff
@@ -414,226 +352,156 @@ def filters_to(start, finish):
...
         _work = self.default_work
 
-        if self.search:
+        self.moby_dick = _work(
+            title="Moby Dick", authors="Herman Melville", fiction=True,
+        )
+        self.moby_dick.presentation_edition.subtitle = "Or, the Whale"
+        self.moby_dick.presentation_edition.series = "Classics"
+        self.moby_dick.summary_text = "Ishmael"
+        self.moby_dick.presentation_edition.publisher = "Project Gutenberg"
```
leonardr: Most of the changes in this file are removing an indent level because it's no longer necessary to check `self.search` before creating works. `populate_works` is only called if a search client could be created.

tests/test_external_search.py
```diff
@@ -3194,7 +3013,7 @@ def test_next_page(self):
 
 class TestBulkUpdate(DatabaseTest):
 
-    def test_works_not_presentation_ready_removed_from_index(self):
+    def test_works_not_presentation_ready_kept_in_index(self):
```
leonardr: This is the test of the incidental change I made in this branch, since we now keep works in the index even if they're not presentation-ready.

tests/test_opds.py
```diff
@@ -508,7 +508,7 @@ def test_acquisition_feed_includes_license_source(self):
         expect = '<bibframe:distribution bibframe:ProviderName="%s"/>' % (
             gutenberg.name
         )
-        assert (1, unicode(feed).count(expect))
+        assert 1 == unicode(feed).count(expect)
```
leonardr: Unrelated change -- the assertion was written wrong and it caused a warning while running tests which I happened to notice.

tests/test_lane.py
```diff
@@ -3367,76 +3366,86 @@ def assert_contents(g, expect):
...
-        # larger number. In the previous example, hq_ro showed up
-        # after hq_litfic because we knew we'd already shown hq_ro in
-        # a previous lane.
+        # we get only the subset of the books previously returned for
+        # 'fiction'.
         assert_contents(
-            fiction.groups(self._db, include_sublanes=False),
-            [(hq_ro, fiction), (hq_litfic, fiction)]
+            make_groups(fiction, include_sublanes=False),
+            [(self.hq_litfic, fiction), (self.hq_sf, fiction)]
```
leonardr: I believe using the deterministic setting makes the results  more consistent here across sub-tests.

tests/test_lane.py
```diff
@@ -3503,21 +3513,24 @@ def assert_contents(g, expect):
...
-        # Here's an entry point that ignores everything except one
-        # specific book.
+        # Here's an entry point that applies a language filter
+        # that only finds one book.
         class LQRomanceEntryPoint(object):
             URI = ""
             @classmethod
-            def apply(cls, qu):
-                return qu.filter(work_model.sort_title=='LQ Romance')
-        facets = FeaturedFacets(0, entrypoint=LQRomanceEntryPoint)
```
leonardr: I had to change the way this works because you can't add a title restriction to our `Filter` object (though you can with an elasticsearch-dsl `F` object).

tests/models/test_licensing.py
```diff
@@ -1174,8 +1174,8 @@ def test_compatible_with_calls_compatible_with_on_deliverymechanism(self):
         # media types.
         edition, pool = self._edition(with_license_pool=True,
                                       with_open_access_download=True)
-        self._add_generic_delivery_mechanism(pool)
-        [mech1, mech2] = pool.delivery_mechanisms
+        [mech1] = pool.delivery_mechanisms
+        mech2 = self._add_generic_delivery_mechanism(pool)
```
leonardr: This test failed on Travis but not locally; I removed what could have been some nondeterministic code.

testing.py
```diff
@@ -968,6 +976,161 @@ def sample_cover_representation(self, name):
...
+                exc_info=e
+            )
+
+    def teardown(self):
+        if self.search:
+            if self.search.works_index:
+                self.search.indices.delete(self.search.works_index, ignore=[404])
+            self.search.indices.delete('the_other_index', ignore=[404])
+            self.search.indices.delete('test_index-v100', ignore=[404])
+            self.search.indices.delete('test_index-v9999', ignore=[404])
```
aslagle: I'm not sure which test creates these but it might make sense to remove them there instead of the parent class?

tests/test_lane.py
```diff
@@ -3264,41 +3264,43 @@ def _w(**kwargs):
...
-        best_seller_list.add_entry(mq_sf)
+        self.best_seller_list, ignore = self._customlist(num_entries=0)
+        self.best_seller_list.add_entry(self.mq_sf)
 
-        staff_picks_list, ignore = self._customlist(num_entries=0)
-        staff_picks_list.add_entry(mq_sf)
+        self.staff_picks_list, ignore = self._customlist(num_entries=0)
+        self.staff_picks_list.add_entry(self.mq_sf)
+
+    def test_groups(self):
```
aslagle: Now that this is a search test should it check self.search to decide whether to run?

aslagle: Or, maybe it's important enough now that we should always require elasticsearch to run the tests.

testing.py
leonardr: It looks like -v100 wasn't being used anymore. I added a helper method that creates the index and registers it so it can be deleted during teardown.

PR #1065, opened 2019-05-29
`WorkList.works()` now goes to search engine, not database

leonardr:
This simple branch changes the default implementation of `WorkList.works()` to go against the search engine rather than the materialized view. The materialized view code is moved to `WorkList.works_from_database` and it will eventually be modified to go against the `works` table rather than the materialized view. At that point we'll be able to remove the materialized view altogether.

For now, this allows me to modify the various `DynamicLane` classes in `api/lanes.py` one at a time, reimplementing `works` if necessary, rather than having to figure them all out at once.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1384,13 +1384,45 @@ class WorkList(object):
                 work_ids.add(work.works_id)
         return works
 
-    def works(self, _db, facets=None, pagination=None, include_quality_tier=False):
+    def works(self, _db, facets=None, pagination=None, **kwargs):
+        """Obtain -- somehow -- Work or Work-like objects that belong
+        in this WorkList.
+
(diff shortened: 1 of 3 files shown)


lane.py
```diff
@@ -1503,7 +1542,7 @@ def works_for_specific_ids(self, _db, work_ids, work_model=mw):
 
         b = time.time()
         logging.info(
-            u"Obtained %s×%d in %.2fsec", work_model.__name__, len(results), b-a
+            u"Obtained %sx%d in %.2fsec", work_model.__name__, len(results), b-a
```
leonardr: This is an unrelated change. The × character displays as `\u00d7` when the log format is JSON, which is more annoying than seeing the letter x.

PR #1066, opened 2019-05-30
Generate MARC records through Elasticsearch, not materialized view

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-2017.

The loop that pulls works from the database and converts them to MARC records now uses Elasticsearch and `SortKeyPagination` rather than the materialized view and regular `Pagination`.

In addition to the unit tests, I've tested the loop code on a large, real dataset. This is not the full MARC mirroring code -- I made a copy of the loop code and changed it to print out the titles of the books it found in each batch.

I don't think we should change the `query_batch_size` because we still go to the database to look up the work objects based on the IDs we get from Elasticsearch. 500 Works from the database is still a lot.

In the future, I think we could dramatically speed this up by storing a Work's MARC record in the search index. It would be easier to do this kind of speedup than to speed up the generation of OPDS feeds.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -1659,7 +1659,7 @@ class Filter(SearchBase):
         :param allow_holds: If this is False, books with no available
         copies will be excluded from results.
 
-        :param last_update_time: If this is set to a datetime, only books
+        :param updated_after: If this is set to a datetime, only books
         whose Work records (~bibliographic metadata) have been updated since
         that time will be included in results.
         """
@@ -1709,7 +1709,7 @@ class Filter(SearchBase):
         )
         self.allow_holds = kwargs.pop('allow_holds', True)
 
-        self.last_update_time = kwargs.pop('last_update_time', None)
+        self.updated_after = kwargs.pop('updated_after', None)
 
         # At this point there should be no keyword arguments -- you can't pass
         # whatever you want into this method.
@@ -1842,9 +1842,9 @@ class Filter(SearchBase):
 
         # Perhaps only books whose bibliographic metadata was updated
         # recently should be included.
-        if self.last_update_time:
+        if self.updated_after:
             last_update_time_query = self._match_range(
-                'last_update_time', 'gte', self.last_update_time
+                'last_update_time', 'gte', self.updated_after
             )
             f = chain(f, F('bool', must=last_update_time_query))
 
--- a/external_search.py
+++ b/external_search.py
@@ -1652,12 +1652,16 @@ class Filter(SearchBase):
         These minor arguments were made into unnamed keyword arguments to
         avoid cluttering the method signature:
 
-        `param excluded_audiobook_data_sources`: A list of DataSources that
+        :param excluded_audiobook_data_sources: A list of DataSources that
         provide audiobooks known to be unsupported on this system.
         Such audiobooks will always be excluded from results.
 
-        `param allow_holds`: If this is False, books with no available
+        :param allow_holds: If this is False, books with no available
         copies will be excluded from results.
+
+        :param last_update_time: If this is set to a datetime, only books
+        whose Work records (~bibliographic metadata) have been updated since
+        that time will be included in results.
         """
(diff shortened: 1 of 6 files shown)


external_search.py
```diff
@@ -2203,32 +2217,58 @@ def create_search_doc(self, query_string, filter=None, pagination=None, debug=Fa
 
     def query_works(self, query_string, filter, pagination, debug=False, return_raw_results=False, search=None):
         self.queries.append((query_string, filter, pagination, debug, return_raw_results))
+        docs = self.docs.values()
+        if pagination:
+            start_at = 0
+            if isinstance(pagination, SortKeyPagination):
```
leonardr: This bit is pretty ugly but it lets the mocked `query_works` work with both regular `Pagination` and `SortKeyPagination`.

external_search.py
```diff
@@ -2203,32 +2217,58 @@ def create_search_doc(self, query_string, filter=None, pagination=None, debug=Fa
...
+            results = [x['_id'] for x in docs]
 
         if pagination:
-            start = pagination.offset
-            stop = start + pagination.size
-            doc_ids = doc_ids[start:stop]
+            result_objs = [
+                MockSearchResult("title", "author", {}, x['_id'])
+                for x in docs
+            ]
```
leonardr: And this is necessary for `SortKeyPagination.page_loaded` to work.

external_search.py
```diff
@@ -1652,12 +1652,16 @@ def __init__(self, collections=None, media=None, languages=None,
...
         provide audiobooks known to be unsupported on this system.
         Such audiobooks will always be excluded from results.
 
-        `param allow_holds`: If this is False, books with no available
+        :param allow_holds: If this is False, books with no available
         copies will be excluded from results.
+
+        :param updated_after: If this is set to a datetime, only books
+        whose Work records (~bibliographic metadata) have been updated since
+        that time will be included in results.
```
leonardr: I'm a little worried about adding more and more tiny changes to the `Filter` object and having custom `Facets` object program it, rather than giving the `Facets` object the ability to directly modify the underlying Elasticsearch filter. This is definitely easier to test, though, and it keeps almost all the Elasticsearch code in `external_search.py`.

external_search.py
```diff
@@ -1834,6 +1840,14 @@ def build(self, _chain_filters=None):
             currently_available = Bool(should=[licenses_available, open_access])
             nested_filters['licensepools'].append(currently_available)
 
+        # Perhaps only books whose bibliographic metadata was updated
+        # recently should be included.
+        if self.updated_after:
+            last_update_time_query = self._match_range(
+                'last_update_time', 'gte', self.updated_after
```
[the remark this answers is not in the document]
aslagle: I don't think this matters - but something like `updated_on_or_after` would be more accurate if it's bothering you.

marc.py
```diff
@@ -453,6 +458,22 @@ def add_ebooks_subject(cls, record):
...
+    it only works updated since a certain time.
+    """
+
+    def __init__(self, start_time):
+        self.start_time = start_time
+
+    def modify_search_filter(self, filter):
+        filter.order = self.SORT_ORDER_TO_ELASTICSEARCH_FIELD_NAME[
+            self.ORDER_LAST_UPDATE
+        ]
```
aslagle: The MARC file doesn't actually need to be sorted, only filtered. Does sorting make it easier to break the query into pages?

leonardr: The results need to be sorted because we're paginating with SortKeyPagination, but they don't need to be sorted by last update -- work ID would also be okay. But I don't think there's any harm in sorting them by last update, and it will ensure the inclusion of any updates that happen while we're building the MARC records.

aslagle: And why are we paginating with SortKeyPagination?

leonardr: Elasticsearch won't go beyond 100k records with normal pagination.

PR #1067, opened 2019-05-31
Fix the Elasticsearch filter for series

leonardr:
This is a support branch for https://jira.nypl.org/browse/SIMPLY-1982. There may be another one on the way but this is big and miscellaneous enough that review is called for.

These are the changes in this branch:

* Turn `series.keyword` into an actual `keyword` field so we can filter on it. Previously it was a `text` field that was tokenized for sorting, which didn't work for filtering.
* Make it possible for a `Filter` to specify a list of fields that should be used to sort results, not just one field. Any tiebreaker fields not used in the original list are added to the end of the list as usual.
* Added some more default implementations to `BaseFacets` so that a subclass can pass through circulation manager code without crashing.
* Add `Pagination.page_has_loaded` so we can check whether `page_loaded` was already called. This is necessary because some implementations of `WorkList.works` call `page_loaded` immediately (so that they have access to internal Elasticsearch data) and some don't (because they return a database query, not an actual list of works).

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -1659,7 +1659,7 @@ class Filter(SearchBase):
         :param allow_holds: If this is False, books with no available
         copies will be excluded from results.
 
-        :param last_update_time: If this is set to a datetime, only books
+        :param updated_after: If this is set to a datetime, only books
         whose Work records (~bibliographic metadata) have been updated since
         that time will be included in results.
         """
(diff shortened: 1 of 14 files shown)


leonardr:
I'm on the fence about whether to make `Pagination.page_loaded` a no-op if `page_has_loaded` is true. This would keep us from having to check `page_has_loaded` outside `Pagination` logic -- we could call `page_has_loaded` whenever we wanted. It would be impossible to "reload" a page once it was already loaded, but we don't do that.

PR #1068, opened 2019-06-06
Add hook methods to make it easier to customize lanes and facets

leonardr:
This is a helper branch in service of https://jira.nypl.org/browse/SIMPLY-1982.

In my circulation branch, I'm trying to take redundant code out of the `FeaturedSeriesFacets` and `SeriesLane` classes and the controller that uses them, making them work more like regular `Facets` and `WorkList` objects and making the controller work more like a normal OPDS feed generation controller.

The simplest way to do this is to add a number of hook methods to `Facets` and `WorkList`, which do nothing special by default, but which the series-specific code can override to plug in extra behavior:

* `FacetsWithEntryPoint.finalize_navigate` is called after navigating from one faceting object to another. In my `SeriesFacets` class, this is used to propagate the series between `SeriesFacets` objects.
* `FacetsWithEntryPoint.finalize_from_request` is called after instantiating a faceting object from request data. In `SeriesFacets`, this is where the series -- an extra bit of data -- is pulled from the request.
* `Facets.available_facets` and `Facets.default_facet` are class methods that let a subclass enable or disable facets for a given facet group, overriding the library's configuration. `SeriesFacets` uses this to add the option to sort a list by series position, and to make this option the default.
* `WorkList.append_child` is used to give a `WorkList` subclass the chance to modify its child `WorkList`s as they come in. In circulation, we can create a `RelatedBooksLane` for a specific book. This `WorkList` has children for different ways in which other books might be related to this one: `ContributorLane`, `SeriesLane`, `RecommendationLane`. We want the language and audience of the original book to put restrictions on all the child lanes. Extending `RelatedBooksLane.append_child` makes it easy to put that logic in `RelatedBooksLane` instead of duplicating it in every class that might be used as a child of such a `WorkList`.

The number of hook methods I had to add makes me think the whole faceting system needs to be redone. We can tackle this as part of our OPDS 2 work.

There are two other changes in this branch which are not hook methods:

1. When a feed is ordered by "series position", it's now ordered by series position _and then title_. Previously it was ordered by series position _and then author_. This ordering seemed random for series like "For Dummies", where there is no ordering and each book has a different author. Theoretically, it can also help for series like "Best Science Fiction", where the series position is implicit in the title as a year.
2. I added a context manager that lets you mock the search engine implementation used by `Lane.works_from_search_index`. This way you can test code that calls `works_from_search_index` (e.g. to generate an OPDS feed) without having to pass in the search engine through many layers of method calls. With this in place, we might be able to completely get rid of the `search_engine` arguments in  `works_for_search_index` and the methods that call it. But we might also be using this to improve performance by reusing search engine objects -- I'm not sure.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -448,14 +448,18 @@ class Facets(FacetsWithEntryPoint):
     def navigate(self, collection=None, availability=None, order=None,
                  entrypoint=None):
         """Create a slightly different Facets object from this one."""
-        return self.__class__(self.library,
-                              collection or self.collection,
-                              availability or self.availability,
-                              order or self.order,
-                              enabled_facets=self.facets_enabled_at_init,
-                              entrypoint=(entrypoint or self.entrypoint),
-                              entrypoint_is_default=False,
+        destination = self.__class__(
+            self.library,
+            collection or self.collection,
+            availability or self.availability,
+            order or self.order,
+            enabled_facets=self.facets_enabled_at_init,
+            entrypoint=(entrypoint or self.entrypoint),
+            entrypoint_is_default=False,
         )
+        destination.finalize_navigate(self)
+        return destination
+
 
     def items(self):
         for k,v in super(Facets, self).items():
@@ -1188,13 +1192,25 @@ class WorkList(object):
         self.fiction = None
         self.target_age = None
 
-        self.children = children or []
+        self.children = []
+        if children:
+            for child in children:
+                self.append_child(child)
         self.priority = priority or 0
 
         if entrypoints:
             self.entrypoints = list(entrypoints)
         else:
             self.entrypoints = []
+
+    def append_child(self, child):
+        """Add one child to the list of children in this WorkList.
+
+        This hook method can be overridden to modify the child's
+        configuration so as to make it fit with what the parent is
+        offering.
+        """
+        self.children.append(child)
 
     @property
     def customlist_ids(self):
--- a/lane.py
+++ b/lane.py
@@ -194,12 +194,10 @@ class FacetsWithEntryPoint(BaseFacets):
         """Create a very similar FacetsWithEntryPoint that points to
         a different EntryPoint.
         """
-        destination = self.__class__(
+        return self.__class__(
             entrypoint=entrypoint, entrypoint_is_default=False,
             **self.constructor_kwargs
         )
-        destination.finalize_navigate(self)
-        return destination
 
     @classmethod
     def from_request(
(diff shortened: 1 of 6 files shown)


lane.py
```diff
@@ -194,10 +194,12 @@ def navigate(self, entrypoint):
         """Create a very similar FacetsWithEntryPoint that points to
         a different EntryPoint.
         """
-        return self.__class__(
+        destination = self.__class__(
             entrypoint=entrypoint, entrypoint_is_default=False,
             **self.constructor_kwargs
         )
+        destination.finalize_navigate(self)
```
aslagle: `navigate` is not a class method, so couldn't a subclass just override that without needing to add another method?

leonardr: You absolutely can, but I've always disliked how inheritance works for class methods. However, I might be using Python wrong. I'll take another look and see if I can improve this.

aslagle: But this isn't a class method anyway - only the other one is.

leonardr: It looks like I was able to take out both of these hook methods. Not the ones that decide which facets are available, though.

lane.py
```diff
@@ -1394,7 +1373,7 @@ def groups(self, _db, include_sublanes=True, facets=None,
         if not include_sublanes:
             # We only need to find featured works for this lane,
             # not this lane plus its sublanes.
-            for work in self.featured_works(_db, facets=facets):
+            for work in self.works(_db, facets=facets):
```
leonardr: The work I'm doing in circulation is making me stop using `featured_works` earlier than I thought. I'm calling `groups` on a bunch of lanes, some of which get their data from the database and some of which get their data from the search index. `works()` is designed to handle either case, but `featured_works`, which is deprecated, can only handle the database case.

lane.py
```diff
@@ -2122,7 +2104,7 @@ def _featured_works_with_lanes(
 
         # Ask the search engine for works from every lane we're given.
         for lane in lanes:
-            for work in lane.works_from_search_index(
+            for work in lane.works(
```
leonardr: Same problem here, but in reverse -- this is being called for lanes that get works from the database, not just lanes that get works from the search index. The lane implementation controls which is used in a specific case.

opds.py
```diff
@@ -703,6 +703,12 @@ def page(cls, _db, title, url, lane, annotator,
             search_engine=search_engine, debug=search_debug
         )
 
+        if not isinstance(works, list):
+            # It's possible that works() returned a database query or
+            # other generator-like object, but at this point we want
+            # an actual list of Work objects.
+            works = [x for x in works]
+
```
leonardr: And this makes OPDS feed generation work for both database and search engine lanes. Earlier I changed it from only working for database to only working for search engine.

tests/test_lane.py
```diff
@@ -156,6 +159,11 @@ def load_entrypoint(cls, entrypoint_name, entrypoints, default=None):
                 cls.load_entrypoint_called_with = (entrypoint_name, entrypoints, default)
                 return cls.expect
 
+            def finalize_from_request(self, facet_config, worklist):
+                self.finalize_from_request_called_with = (
+                    facet_config, worklist
+                )
```
aslagle: I think you can get rid of this now.

PR #1069, opened 2019-06-07
Added a hack to re-associate lanes with a database session after they're disconnected.

leonardr:
Changes in https://github.com/NYPL-Simplified/circulation/pull/1243/ exposed what I think is a preexisting problem. We know that when you run one of the circ manager's `CacheOPDSGroupFeedPerLane` scripts, all the `Lane` objects get disconnected from their original database session as soon as the app server starts up. So we added a hack to, e.g. `WorkList.groups()` that merges each lane back into the current database session.

Apparently, by the time the app server starts up, each lane also has a cached value for `.parent`. When the app server starts up, those cached values are _also_ disconnected from their database session. When we try to access `Lane.hierarchy` after the app server starts up, we end up using those cached values, and we're back to trying to do stuff to `Lane` objects that aren't connected to the session.

This wasn't a problem before because the process of making OPDS feeds for a Lane didn't, apparently, involve looking at `Lane.parent`. Now it does look at `Lane.parent` while it's building the `Filter` for the search engine.

This branch extends the hack to `Lane.hierarchy` so that all outgoing `Lane`s are associated with the correct database session.

I would prefer to fix the underlying problem, but I don't know why that disconnect happens. I'll look into it a bit.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -2414,13 +2414,20 @@ class Lane(Base, WorkList):
         """
         if not self.parent:
             return
-        yield self.parent
-        seen = set([self, self.parent])
-        for parent in self.parent.parentage:
-            if parent in seen:
+        parent = self.parent
+        if Session.object_session(parent) is None:
+            # This lane's parent was disconnected from its database session,
+            # probably when an app server started up.
+            # Reattach it to the database session used by this lane.
+            parent = Session.object_session(self).merge(parent)
+
+        yield parent
+        seen = set([self, parent])
+        for grandparent in parent.parentage:
+            if grandparent in seen:
                 raise ValueError("Lane parentage loop detected")
-            seen.add(parent)
-            yield parent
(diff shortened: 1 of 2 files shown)


leonardr:
We expunge all the Lanes from the initial database session inside `load_lanes`, when initializing the circulation manager: https://github.com/NYPL-Simplified/circulation/blob/master/api/lanes.py#L64

I don't see a way around this hack unless we come up with some way of transparently adding the session back (seems very difficult), or changing it so we can initialize a circulation manager in a script context to make it not do this kind of thing (probably pretty easy).

leonardr:
After some experimentation I filed https://jira.nypl.org/browse/SIMPLY-2034 to deal with the long-term problem. I'm going to merge this branch and deal with 2034 later, for the sake of reducing the number of big changes that might only become visible on a real site.

PR #1070, opened 2019-06-10
Change the "last update time" facet so that addition to a relevant collection or customlist counts as an "update".

leonardr:
Core work for https://jira.nypl.org/browse/SIMPLY-1984.

This is completely different from my original attempt and should probably be reviewed from scratch.

When you sort by "last update time", the sort keys are calculated by a script in the ironically-named Painless language. The algorithm used here is the one used in circulation when generating a crawlable script: a book is "last updated" when Work.last_update_time changes, _or_ when a LicensePool for the Work is added to a relevant Collection, _or_ when the Work is added to a relevant CustomList. The "last update time" for a Work is the latest of all those times.

If there are no collections or custom lists associated with the filter, then the old behavior is preserved; it's simple and probably a lot faster. However, in almost all cases there will be collections or custom lists associated with the filter, so it's not a big optimization.

Dates that we index are now indexed as `long`, not `date`. Incoming dates are turned into seconds-since-epoch. This makes the sorting function easier to write, and probably faster.


external_search.py
```diff
@@ -1983,44 +1992,140 @@ def sort_order(self):
...
+        # part of the query ensures that we only match works present
+        # on one list in every restriction set. Here, we need to find
+        # the latest time a work was added to _any_ relevant list.
+        all_list_ids = set()
+        for restriction in self.customlist_restriction_sets:
+            all_list_ids.update(self._filter_ids(restriction))
+        nested = dict(
+            path="customlists",
+            filter=dict(
+                terms={"customlists.list_id": sorted(all_list_ids)}
```
aslagle: does it need to be sorted?

tests/test_external_search.py
```diff
@@ -991,48 +991,168 @@ class TestSearchOrder(EndToEndSearchTest):
...
+        #
+        # a, b, c - when no collections or custom lists are associated with
+        #           the Filter.
+        # a, c, b - when collection 1 is associated with the Filter.
+        # b, a, c - when collections 1 and 2 are associated with the Filter.
+        # b, c, a - when custom list 1 is associated with the Filter.
+        # c, a, b - when collection 1 and custom list 2 are associated with
+        #           the Filter.
+        # c, a - when two sets of custom list restrictions [1], [3]
+        #        are associated with the filter.
```
aslagle: This is pretty thorough. The only thing I can think of that's not covered is when the works are associated with collections or lists, but the works' last update times are more recent and control the order.

aslagle:
Still looks good :)

external_search.py
leonardr: No, I was thinking this would make it easier to test but I don't actually need it in the test.

PR #1071, opened 2019-06-13
When searching for works, return search results annotated with the original Elasticsearch Hit objects

leonardr:
This should complete https://jira.nypl.org/browse/SIMPLY-2048. It makes a number of changes -- two big, a few small -- necessary to support https://jira.nypl.org/browse/SIMPLY-1984 (getting the crawlable lanes from Elasticsearch rather than the materialized view).

The two big changes:

* The script for calculating the last update time of a work (in the context of whatever collections or custom lists are relevant) is now stored in the Elasticsearch server every time a client starts up. (So it's more important than ever not to create a million clients.) Once this script is a stored script with a name, we can use it both to order a list and as the one and only field in `script_fields` -- fields whose values are calculated for each search result and returned as part of the `Hit` object.

* When it's time to look up `Work`s in the database based on the work IDs found in the `Hit` objects, we don't always return the `Work`s. If the `Hit` contains values from script fields, we bundle each `Work` with its `Hit` in a new class called `WorkSearchResult`. These objects defer most attribute access to the underlying `Work`, but the search result data is also available. We only do this when script fields are present, for two reasons: 1) performance, 2) if you think you have a `Work` but you actually have a `WorkSearchResult`, any attempt to modify the supposed `Work` will fail.

Small changes are described inline.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -2387,6 +2387,10 @@ class WorkSearchResult(object):
     This object acts just like a Work (though isinstance(x, Work) will
     fail), with one exception: you can access the raw ElasticSearch Hit
     result as ._hit.
+
+    This is useful when a Work needs to be 'tagged' with information
+    obtained through Elasticsearch, such as its 'last modified' date
+    the context of a specific lane.
     """
     def __init__(self, work, hit):
         self._work = work
--- a/external_search.py
+++ b/external_search.py
@@ -216,6 +216,7 @@ class ExternalSearchIndex(HasSelfTests):
             self.index = self.__client.index
             self.delete = self.__client.delete
             self.exists = self.__client.exists
+            self.put_script = self.__client.put_script
 
         # Sets self.works_index and self.works_alias values.
         # Document upload runs against the works_index.
@@ -223,6 +224,7 @@ class ExternalSearchIndex(HasSelfTests):
         if works_index and integration and not in_testing:
             try:
                 self.set_works_index_and_alias(_db)
+                self.set_stored_scripts()
             except RequestError, e:
                 # This is almost certainly a problem with our code,
                 # not a communications error.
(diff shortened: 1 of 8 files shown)


lane.py
```diff
@@ -406,19 +406,18 @@ def __init__(self, library, collection, availability, order,
...
         )
         availability = availability or self.default_facet(
             library, self.AVAILABILITY_FACET_GROUP_NAME
         )
         order = order or self.default_facet(library, self.ORDER_FACET_GROUP_NAME)
+        if order_ascending is None:
+            if order in Facets.ORDER_DESCENDING_BY_DEFAULT:
+                order_ascending = self.ORDER_DESCENDING
+            else:
+                order_ascending = self.ORDER_ASCENDING
```
leonardr: There is currently special code in circulation for changing the order of a crawlable feed. This establishes a general rule that feeds ordered by dates should be ordered "newest first" by default, which lets me get rid of that special code.

lane.py
```diff
@@ -944,6 +943,7 @@ class Pagination(object):
     DEFAULT_SIZE = 50
     DEFAULT_SEARCH_SIZE = 10
     DEFAULT_FEATURED_SIZE = 10
+    DEFAULT_CRAWLABLE_SIZE = 100
```
leonardr: It takes a lot longer to run a crawlable feed query against Elasticsearch than the materialized view, so I double the size of the feed to compensate.

I could see bumping this up to 200 or even 500, but I don't know how frequently someone importing from a crawlable feed encounters an error and loses a page worth of work.

opds.py
```diff
@@ -819,7 +821,7 @@ def add_entrypoint_links(cls, feed, url_generator, entrypoints,
                 url_generator, entrypoint, selected_entrypoint, is_default,
                 group_name
             )
-            if link:
+            if link is not None:
```
leonardr: This gets rid of a Python warning.

external_search.py
```diff
@@ -372,7 +425,7 @@ def base_index_name(self, index_or_alias):
         return base_works_index
 
     def create_search_doc(self, query_string, filter, pagination,
-                          debug, return_raw_results):
```
leonardr: This argument no longer makes sense since searches must now always return `Hit` objects (so they can be turned into `WorkSearchResult` objects).

external_search.py
```diff
@@ -381,15 +434,18 @@ def create_search_doc(self, query_string, filter, pagination,
             search = search.extra(explain=True)
 
         fields = None
-        if debug or return_raw_results:
+        if debug:
             # Don't restrict the fields at all -- get everything.
             # This makes it easy to investigate everything about the
             # results we do get.
-            fields = None
+            fields = ['*']
```
leonardr: Setting this to `None` was including all the regular fields but there was no way to add script fields. This way we get all the regular fields, plis we have a list to which to append the script fields.

external_search.py
```diff
@@ -1655,6 +1715,7 @@ def from_worklist(cls, _db, worklist, facets):
         fiction = inherit_one('fiction')
         audiences = inherit_one('audiences')
         target_age = inherit_one('target_age')
+        collections = inherit_one('collection_ids') or library
```
leonardr: Changes in this method are necessary to support synthetic lanes that have no associated Library -- crawlable lanes, basically -- without special code.

external_search.py
```diff
@@ -2080,53 +2142,37 @@ def _last_update_time_order_by(self):
...
-                "script": {
-                    "lang": "painless",
-                    "params": params,
-                    "source": script,
-                },
-                "order": self.asc,
-            },
-        }
-        return order
+        return { "last_update": {"order": self.asc, } }
```
aslagle: This line seems to be left over from something and no longer used.

opds.py
```diff
@@ -162,7 +162,9 @@ def annotate_work_entry(self, work, active_license_pool, edition,
                 title=unicode(group_title)
             )
 
-        # TODO: maybe we can do better than this in calculating updated()
+        # NOTE: This is a default that works in most cases. When
+        # ordering ElasticSearch results by last update time, you get
+        # a more reliable value that you can use instead.
```
aslagle: The code to use that is in circulation then?

leonardr: I think I should move this comment inside the `if` statement. The code for doing this is in core, it's just that search results are rarely sorted this way.

aslagle: Ah, I was confused because it didn't need to change in this PR.

external_search.py
aslagle: This is still here.

PR #1072, opened 2019-06-14
Pass in url_for

adriana-alter:


model/patron.py
```diff
@@ -422,11 +422,12 @@ class PatronProfileStorage(ProfileStorage):
     Protocol.
     """
 
-    def __init__(self, patron):
+    def __init__(self, patron, url_for):
```
leonardr: I think most of your test failures will go away if you have this take `None` as the default value.

PR #1073, opened 2019-06-17
Cached list size to improve admin interface performance.

aslagle:
This is an optimization for Open Bookshelf, which uses lists heavily. The admin interface makes a lot of requests that get the sizes of all lists.

### DIFF
--- a/model/customlist.py
+++ b/model/customlist.py
@@ -44,6 +44,10 @@ class CustomList(Base):
     responsible_party = Column(Unicode)
     library_id = Column(Integer, ForeignKey('libraries.id'), index=True, nullable=True)
 
+    # How many titles are in this list? This is calculated and
+    # cached when the list contents change.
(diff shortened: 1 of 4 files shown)


model/customlist.py
```diff
@@ -215,6 +221,9 @@ def entries_for_work(self, work_or_edition):
             )
         return qu
 
+    def update_size(self):
+        self.size = len(self.entries)
```
aslagle: I'm not sure if this method is actually needed, but could be useful for repair purposes.

aslagle:
I can't think of anything similar that would affect lists. We only display the total size of each list, and there aren't any additional facets or filtering. They're only used directly in the admin interface and the crawlable OPDS feeds (which currently don't need the size, but even if they did, they don't have any additional facets).

I can definitely add a script.

PR #1074, opened 2019-06-19
Make it possible to filter on author identity while searching

leonardr:
This completes https://jira.nypl.org/browse/SIMPLY-1985, the core portion of https://jira.nypl.org/browse/SIMPLY-1983.

This branch adds some fields to the 'contributors' subdocument indexed as part of a Work's ElasticSearch document. It also adds a subdocument definition for 'contributors' to the document body definition, so that the document is indexed _as_ a subdocument and not as a bunch of unrelated fields. This lets us create nested filters based on the 'contributors' subdocument, to find only those works whose contributors meet certain criteria.

This functionality is exposed when you pass an `author` into the `Filter` constructor. The `author` can be a `Contributor` or `ContributorData`, and the `Filter` will restrict the query to works with that individual listed in an author-type role.

A month ago I was afraid of doing this, but other ElasticSearch stuff I've done since then was a lot worse, so it was pretty easy.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -39,6 +39,7 @@ from facets import FacetConstants
 from model import (
     numericrange_to_tuple,
     Collection,
+    Contributor,
     ConfigurationSetting,
     DataSource,
     Edition,
@@ -749,6 +750,12 @@ class ExternalSearchIndexVersions(object):
                 description['type'] = 'text'
                 description['fielddata'] = True
                 description['analyzer'] = 'en_sortable_analyzer'
+            elif type == 'normalized_keyword':
+                # A normalized_keyword acts as a regular keyword but
+                # also has a simple normalizer attached to it to
+                # improve filtering.
+                description['type'] = 'keyword'
+                description['normalizer'] = 'filterable_string'
             mapping = cls.map_fields(
                 fields=fields,
                 mapping=mapping,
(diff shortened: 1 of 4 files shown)


external_search.py
```diff
@@ -1702,6 +1719,13 @@ class Filter(SearchBase):
     # the useful information from the search engine isn't lost.
     KNOWN_SCRIPT_FIELDS = ['last_update']
 
+    # In general, someone looking for things "by this person" is
+    # probably looking for one of these roles.
+    AUTHOR_MATCH_ROLES = list(Contributor.AUTHOR_ROLES) + [
+        Contributor.EDITOR_ROLE, Contributor.DIRECTOR_ROLE,
+        Contributor.ACTOR_ROLE
+    ]
+
```
leonardr: I don't think we want to change AUTHOR_ROLES here. The editor of a book is not the "author", but if you're looking for "books by this person" then books that they edited should quality.

aslagle: Aren't the contributor feeds only used for authors anyway? This seems like it would be more useful for the regular keyword search. Narrator might also be a good one.

leonardr: I think we're just a couple branches away from being able to use the contributors subdocument in searches. Like, we could add a hypothesis now that gives a big boost if an author-role contributor is an exact match for the search term, with an even bigger boost for primary author.

Then we could change the name fields so they're indexed twice, once as a keyword and once as a text field (this part seems really annoying). At that point we don't need to search on sort_author anymore; we just need it for searching.

PR #1076, opened 2019-06-21
Add a lookup method to ContributorData which performs basic data remediation

leonardr:
This branch moves some code from circulation to core, expands it a bit, and adds standalone tests. The functionality is the ability to go from a person's name, to a `ContributorData` containing the most complete information we can accurately get out of the database. This process may go through multiple `Contributor` objects representing different people with the same name, or multiple inconsistent records of a single person.

I don't know how much of this code we could do without if we had proper contributor reconciliation -- I think some but not all of it. We'd still need a way of taking a name that may refer to multiple people, and (if possible) figuring out which person we're actually talking about.

This whole thing is necessary because `CirculationLane` in the circulation manager takes author name as its input, not contributor ID. For a while I thought it would be easy to just start using contributor ID instead, and I refactored some code in `Annotator` into a hook method that I thought I could subclass. This didn't work out, but I've left the refactored code in place; it doesn't change functionality at all and the new code is better since much of the functionality could have standalone tests written for it.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -176,6 +176,7 @@ class SubjectData(object):
 
 
 class ContributorData(object):
+
     def __init__(self, sort_name=None, display_name=None,
                  family_name=None, wikipedia_name=None, roles=None,
                  lc=None, viaf=None, biography=None, aliases=None, extra=None):
@@ -183,7 +184,8 @@ class ContributorData(object):
         self.display_name = display_name
         self.family_name = family_name
         self.wikipedia_name = wikipedia_name
-        roles = roles or Contributor.AUTHOR_ROLE
+        if roles is None:
+            roles = Contributor.AUTHOR_ROLE
         if not isinstance(roles, list):
             roles = [roles]
         self.roles = roles
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -220,6 +220,67 @@ class ContributorData(object):
             roles=[contribution.role]
         )
 
+    @classmethod
+    def lookup(cls, _db, sort_name=None, display_name=None, lc=None,
+               viaf=None):
+        """Create a (potentially synthetic) ContributorData based on
+        the best available information in the database.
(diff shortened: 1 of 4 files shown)


metadata_layer.py
```diff
@@ -176,14 +176,16 @@ def __repr__(self):
...
     def __init__(self, sort_name=None, display_name=None,
                  family_name=None, wikipedia_name=None, roles=None,
                  lc=None, viaf=None, biography=None, aliases=None, extra=None):
         self.sort_name = sort_name
         self.display_name = display_name
         self.family_name = family_name
         self.wikipedia_name = wikipedia_name
-        roles = roles or Contributor.AUTHOR_ROLE
+        if roles is None:
+            roles = Contributor.AUTHOR_ROLE
```
leonardr: This doesn't make a difference now, but I'm creating synthetic `ContributorData` objects representing individuals outside of whatever they may have contributed to a Work, and I want to make sure those objects don't have any roles associated with them.

opds.py
```diff
@@ -300,38 +300,77 @@ def categories(cls, work):
     @classmethod
     def authors(cls, work, edition):
         """Create one or more <author> and <contributor> tags for the given
-        work."""
+        Work.
+
+        :param entry: A Tag representing the OPDS entry in progress.
```
aslagle: This param does not exist.

opds.py
```diff
@@ -300,38 +300,77 @@ def categories(cls, work):
...
-                if not role:
-                    # This contribution is not one that we publish as
-                    # a <atom:contributor> tag. Skip it.
-                    continue
-
-            name_key = name.lower()
-            if name_key in listed_by_role[role]:
+            tag = cls.contributor_tag(contribution, state)
+            if tag is None:
+                # author_tag decided that this contribution doesn't
```
aslagle: contributor_tag now?

PR #1077, opened 2019-06-24
Move all WorkList/Lane code that touches the database into a separate DatabaseBackedWorkList class.

leonardr:
This branch completes https://jira.nypl.org/browse/SIMPLY-2076. It's primarily a refactoring branch and should not change user-visible functionality.

This branch takes all of the code in lane.py that deals with the database (as opposed to the search engine) and either removes it, or puts it in new classes: `DatabaseBackedFacets` and `DatabaseBackedWorkList`. It may be easier to review these classes as though they were new, rather than trying to make sense of those parts of the diffs.

The new facets class is necessary because only a subset of the 'standard' sort orders can be implemented inside the database without using the materialized view.

Speaking of which, all of the lane code to do with the materialized view is gone. All of the database-specific code relating to building a list of featured works is gone. But a lot of code I was hoping to remove is still here. It would be nice to get rid of  `DatabaseBackedWorkList` altogether, and here's what's currently blocking that:

* `WorkList.works()` (the search index method) does its search index thing and calls `works_for_hits`, which creates a `SpecificWorkList`, a subclass of `DatabaseBackedWorkList`. That object does a database lookup to find Works for specific work IDs. This allowed me to eliminate some duplicate code around setting up a basic database query suitable for use in an OPDS feed.
* `Lane` subclasses `DatabaseBackedWorkList`, and we call `works_from_database` to count the total size of a lane. This requires that `DatabaseBackedWorkList` be able to handle every feature of a 'normal' lane, e.g. multiple sets of custom list restrictions. We could probably use the search index instead.
* The `DatabaseBackedWorkList` is the basis for the `RecommendationLane`, which finds works associated with specific ISBNs -- a feature that would be very difficult to reliably add to the search index. Like all WorkLists associated with a specific Work, RecommendationLane has language and audience filters that need to be taken into account. But it doesn't need the super-complicated stuff.
* The database code is already tested and it would be nice to have it around for future requirements. The flip side of this is that if we add a new feature to the `lanes` table we have to implement the filter twice -- once as a search engine filter in `WorkList` and once as a database filter in `DatabaseBackedWorkList`.

This branch also defines two hook methods that a subclass of `WorkList` or `DatabaseBackedWorkList` can implement to hard-code a modification to an outgoing search or database query. This should let me get rid of the awkward code in circulation where every special lane in `api/lanes.py` needs to have its own faceting class, because that's the only way to modify the `Filter` object.

### DIFF
--- a/entrypoint.py
+++ b/entrypoint.py
@@ -93,7 +93,7 @@ class EntryPoint(object):
         return filter
 
     @classmethod
-    def modify_database_query(cls, qu):
+    def modify_database_query(cls, _db, qu):
         """Default behavior is to not change a database query at all."""
         return qu
 
@@ -115,7 +115,7 @@ class MediumEntryPoint(EntryPoint):
     """
 
     @classmethod
-    def modify_database_query(cls, qu):
+    def modify_database_query(cls, _db, qu):
         """Modify a query against Work+LicensePool+Edition
         to match only items with the right medium.
         """
--- a/entrypoint.py
+++ b/entrypoint.py
@@ -76,13 +76,6 @@ class EntryPoint(object):
             cls.DEFAULT_ENABLED.remove(entrypoint_class)
 
     @classmethod
-    def modified_materialized_view_query(cls, qu):
-        """Modify a query against the mv_works_for_lanes materialized view
-        so it matches only items that belong in this entry point.
-        """
-        raise NotImplementedError()
-
-    @classmethod
     def modify_search_filter(cls, filter):
         """If necessary modify an ElasticSearch Filter object so that it
         restricts results to items shown through this entry point.
--- a/entrypoint.py
+++ b/entrypoint.py
@@ -100,8 +100,8 @@ class EntryPoint(object):
         return filter
 
     @classmethod
-    def apply(cls, qu):
-        """Default behavior is to not change a query at all."""
+    def modify_database_query(cls, qu):
+        """Default behavior is to not change a database query at all."""
         return qu
 
 
@@ -122,12 +122,12 @@ class MediumEntryPoint(EntryPoint):
     """
 
     @classmethod
-    def apply(cls, qu):
-        """Modify a query against the mv_works_for_lanes materialized view
+    def modify_database_query(cls, qu):
+        """Modify a query against Work+LicensePool+Edition
         to match only items with the right medium.
         """
-        from model import MaterializedWorkWithGenre as mv
-        return qu.filter(mv.medium==cls.INTERNAL_NAME)
+        from model import Edition
+        return qu.filter(Edition.medium==cls.INTERNAL_NAME)
 
     @classmethod
     def modify_search_filter(cls, filter):
--- a/external_search.py
+++ b/external_search.py
@@ -1764,7 +1764,7 @@ class Filter(SearchBase):
             target_age, genre_id_restrictions, customlist_id_restrictions,
             facets,
             excluded_audiobook_data_sources=excluded_audiobook_data_sources,
-            allow_holds=allow_holds, license_datasource_id=license_datasource_id
+            allow_holds=allow_holds, license_datasource=license_datasource_id
         )
 
     def __init__(self, collections=None, media=None, languages=None,
@@ -1906,10 +1906,10 @@ class Filter(SearchBase):
             )
             nested_filters['licensepools'].append(collection_match)
 
-        license_datasource = filter_ids(self.license_datasource)
-        if license_datasource:
+        license_datasources = filter_ids(self.license_datasources)
+        if license_datasources:
             datasource_match = F(
-                'terms', **{'licensepools.data_source_id' : license_datasource}
+                'terms', **{'licensepools.data_source_id' : license_datasources}
             )
             nested_filters['licensepools'].append(datasource_match)
 
--- a/external_search.py
+++ b/external_search.py
@@ -1738,6 +1738,8 @@ class Filter(SearchBase):
         target_age = inherit_one('target_age')
         collections = inherit_one('collection_ids') or library
 
+        license_datasource_id = inherit_one('license_datasource_id')
+
         # For genre IDs and CustomList IDs, we might get a separate
         # set of restrictions from every item in the WorkList hierarchy.
         # _All_ restrictions must be met for a work to match the filter.
@@ -1762,17 +1764,15 @@ class Filter(SearchBase):
             target_age, genre_id_restrictions, customlist_id_restrictions,
             facets,
             excluded_audiobook_data_sources=excluded_audiobook_data_sources,
-            allow_holds=allow_holds,
+            allow_holds=allow_holds, license_datasource_id=license_datasource_id
         )
-
 
     def __init__(self, collections=None, media=None, languages=None,
                  fiction=None, audiences=None, target_age=None,
                  genre_restriction_sets=None, customlist_restriction_sets=None,
                  facets=None, script_fields=None, **kwargs
     ):
-        """
-        These minor arguments were made into unnamed keyword arguments to
+        """These minor arguments were made into unnamed keyword arguments to
         avoid cluttering the method signature:
 
         :param excluded_audiobook_data_sources: A list of DataSources that
@@ -1788,6 +1788,10 @@ class Filter(SearchBase):
         :param author: If this is set to a Contributor or
         ContributorData, then only books where this person had an
         authorship role will be included.
+
+        :param license_datasource: If this is set to a DataSource,
+        only books with LicensePools from that DataSource will be
+        included.
 
         :param updated_after: If this is set to a datetime, only books
         whose Work records (~bibliographic metadata) have been updated since
@@ -1845,6 +1849,9 @@ class Filter(SearchBase):
 
         self.author = kwargs.pop('author', None)
 
+        license_datasources = kwargs.pop('license_datasource', None)
+        self.license_datasources = self._filter_ids(license_datasources)
+
         # At this point there should be no keyword arguments -- you can't pass
         # whatever you want into this method.
         if kwargs:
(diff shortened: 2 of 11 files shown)


entrypoint.py
```diff
@@ -100,8 +93,8 @@ def modify_search_filter(cls, filter):
         return filter
 
     @classmethod
-    def apply(cls, qu):
-        """Default behavior is to not change a query at all."""
+    def modify_database_query(cls, _db, qu):
+        """Default behavior is to not change a database query at all."""
```
leonardr: `Facets` and `Pagination` now have separate methods for modifying a `Query` vs. a`Filter`, so I renamed `apply` to make it clear what the input is.

external_search.py
```diff
@@ -1738,6 +1738,8 @@ def from_worklist(cls, _db, worklist, facets):
         target_age = inherit_one('target_age')
         collections = inherit_one('collection_ids') or library
 
+        license_datasource_id = inherit_one('license_datasource_id')
+
```
leonardr: I discovered a circulation feature I'd forgotten about where you can get a feed of the books provided by a specific license source. We already index this information in ElasticSearch (in order to filter out undeliverable audiobooks) so it was easy to add this capability to `Filter`.

lane.py
```diff
@@ -669,8 +690,8 @@ class FeaturedFacets(FacetsWithEntryPoint):
 
     DETERMINISTIC = object()
 
-    def __init__(self, minimum_featured_quality, uses_customlists=False,
```
leonardr: `uses_customlists` is gone because `FeaturedFacets` no longer needs to know whether a `WorkList` uses custom lists. The Elasticsearch scoring function will work the same way whether or not a work is on any lists, and the code to calculate the 'quality tier' field through SQL has been removed.

lane.py
```diff
@@ -1114,6 +1075,15 @@ def initialize(self, library, display_name=None, genres=None,
         :param media: Only Works in one of these media will be included
         in lists.
 
+        :param fiction: Only Works with this fiction status will be included
+        in lists.
+
+        :param target_age: Only Works targeted at readers in this age range
+        will be included in lists.
```
leonardr: Previously you couldn't create a `WorkList` with a fiction or target age restriction -- not for any particular reason, it's just that only `Lane` needed that functionality. But now all the database-specific code goes into `DatabaseBackedWorkList`, and I wanted all the tests of that code to go into a class that didn't rely on creating `Lane`s at the crucial moment.

It turns out we still need to create a few `Lane`s in the test, because there's one feature that `Lane` has that `WorkList` is hard-coded not to have: a parent and the ability to inherit restrictions from the parent. But we can deal with that later.

The `license_datasource` restriction is brand new.

lane.py
```diff
@@ -1645,335 +1476,32 @@ def works_for_hits(self, _db, hits, work_model=mw):
 
         b = time.time()
         logging.info(
-            u"Obtained %sx%d in %.2fsec", work_model.__name__, len(results), b-a
+            u"Obtained %sxWork in %.2fsec", len(results), b-a
         )
         return results
 
-    def apply_filters(self, _db, qu, facets, pagination, featured=False):
```
leonardr: All this code was removed from `WorkList` and quarantined in the new `DatabaseBackedWorkList` class.

lane.py
```diff
@@ -2138,99 +1664,325 @@ def _featured_works_with_lanes(
...
+class DatabaseBackedWorkList(WorkList):
+    """A WorkList that can get its works from the database in addition to
+    (or possibly instead of) the search index.
+
+    Even when works _are_ obtained through the search index, a
+    DatabaseBackedWorkList is then created to look up the Work objects
+    for use in an OPDS feed.
+    """
+
+    def works_from_database(self, _db, facets=None, pagination=None, **kwargs):
```
leonardr: Here's the code that I removed from `WorkList`. Since it no longer uses the materialized view, it should only be used when performance is not important or when there's no way to get what you want using the normal `works()` method.

lane.py
```diff
@@ -2138,99 +1664,325 @@ def _featured_works_with_lanes(
...
+        if self.languages:
+            clauses.append(Edition.language.in_(self.languages))
+        if self.media:
+            clauses.append(Edition.medium.in_(self.media))
+        if self.fiction is not None:
+            clauses.append(Work.fiction==self.fiction)
+        if self.license_datasource_id:
+            clauses.append(
+                LicensePool.data_source_id==self.license_datasource_id
+            )
```
leonardr: The license_datasource_id code is new. The fiction and inherit_parent_restrictions code was merged in from this method's now-removed implementation in `Lane`.

lane.py
```diff
@@ -2138,99 +1664,325 @@ def _featured_works_with_lanes(
...
+            # This lane does not require that books be on any particular
+            # CustomList.
+            return qu, []
+
+        # We will be joining against CustomListEntry at least
+        # once. For a lane derived from the intersection of two or
+        # more custom lists, we may be joining CustomListEntry
+        # multiple times. To avoid confusion, we make a new alias for
+        # the table every time.
+        a_entry = aliased(CustomListEntry)
```
leonardr: This code (and the genre code) is simplified since we're no longer starting with the materialized view. We now have to join against `CustomListEntry` or `WorkGenre` every time, since we don't start off with a 'free' join created when the materialized view was created. This is less efficient, but again, we almost never need to use this code now.

lane.py
```diff
@@ -2591,13 +2343,17 @@ def uses_customlists(self):
...
         # Do the estimate for every known entry point.
         by_entrypoint = dict()
         for entrypoint in EntryPoint.ENTRY_POINTS:
-            qu = entrypoint.apply(query)
+            facets = DatabaseBackedFacets(
+                library, FacetConstants.COLLECTION_FULL,
+                FacetConstants.AVAILABLE_ALL,
+                order=FacetConstants.ORDER_WORK_ID, entrypoint=entrypoint
+            )
+            qu = self.works_from_database(_db, facets)
```
leonardr: Again, I don't think it would be too difficult to use the search index to count the size of a lane under different entry points.

lane.py
```diff
@@ -2868,80 +2591,6 @@ def search(self, _db, query_string, search_client, pagination=None,
         return m(_db, query_string, search_client, pagination,
                  facets=facets)
 
-    def bibliographic_filter_clause(self, _db, qu, featured, outer_join=False):
```
leonardr: This whole thing was merged into `DatabaseBackedWorkLane.bibliographic_filter_clauses` (it returns multiple clauses rather than one joined clause because that's easier to test).

tests/test_lane.py
```diff
@@ -827,230 +884,121 @@ def facetify(collection=Facets.COLLECTION_FULL,
...
-        )
-        high_quality_low_random.quality = 1
-        high_quality_low_random.random = 0
-
-        high_quality_high_random = self._work(
-            title="High quality, high random", with_license_pool=True
-        )
-        high_quality_high_random.quality = 0.7
-        high_quality_high_random.random = 1
+    def test_scoring_functions(self):
```
leonardr: This wasn't previously tested in detail -- only the results of using it to generate grouped feeds.

tests/test_lane.py
```diff
@@ -1948,295 +1731,736 @@ def __contains__(self, k):
...
                 qu = qu.filter(and_(*clauses))
-            return [x.works_id for x in qu.all()]
-
-        eq_([gutenberg_adult.id], for_audiences(Classifier.AUDIENCE_ADULT))
+            return qu.all()
 
-        # The Gutenberg "children's" book is filtered out because it we have
-        # no guarantee it is actually suitable for children.
-        eq_([non_gutenberg_children.id],
-            for_audiences(Classifier.AUDIENCE_CHILDREN))
```
leonardr: This restriction has been removed. I've removed nearly all the Project Gutenberg books from the NYPL collection, so inappropriate "children's" books shouldn't be an issue anymore.

tests/test_lane.py
```diff
@@ -2249,67 +2473,52 @@ def test_customlist_filter_clauses(self):
...
 
-        # It's possible to restrict a WorkList so that only works that
-        # are _featured_ on a list show up. The work isn't featured,
-        # so it doesn't show up.
-        eq_([], results(must_be_featured=True))
-
-        # Now it's featured, and it shows up.
-        gutenberg_list_entry.featured = True
-        eq_([work.id], results(must_be_featured=True))
-
```
leonardr: Again, this feature was removed because it was only used in the context of generating a grouped feed for a Lane -- something that's now handled through the search index.

tests/test_lane.py
```diff
@@ -2358,186 +2567,22 @@ def results(wl=works_on_gutenberg_lists, must_be_featured=False):
...
         # If we remove `work` from either list, the combined query
-        # matches nothing. This works even though the materialized
-        # view has not been refreshed.
+        # matches nothing.
         for l in [gutenberg_list, gutenberg_list_2]:
             l.remove_entry(work)
             eq_([], _run(both_lists_qu, both_lists_clauses))
             l.add_entry(work)
 
-    def test_random_sample(self):
```
leonardr: Random sampling was only used to generate grouped feeds, which now happens through the search index.

tests/test_lane.py
```diff
@@ -3090,241 +3135,6 @@ def search(self, *args, **kwargs):
         Lane.search_target = old_lane_search_target
         WorkList.search = old_wl_search
 
-    def test_bibliographic_filter_clause(self):
```
leonardr: These tests were merged into the corresponding test in `DatabaseBackedWorkList`, since `Lane` no longer has any database code of its own.

tests/test_lane.py
```diff
@@ -3770,61 +3571,6 @@ def works(self, _db, facets, pagination, *args, **kwargs):
         eq_(facets, called_with_facets)
         eq_(pagination, pagination)
 
-    def test_featured_window(self):
```
leonardr: More code used only to generate grouped feeds from the database.

opds.py
```diff
@@ -798,8 +799,10 @@ def make_link(ep):
     def from_query(cls, query, _db, feed_name, url, pagination, url_fn, annotator):
         """Build  a feed representing one page of a given list. Currently used for
         creating an OPDS feed for a custom list and not cached.
+
+        # TODO: Can this be removed?
```
aslagle: This is used for the admin interface.

PR #1078, opened 2019-06-27
Allow WorkLists to modify an incoming FeaturedFacets object to a faceting object that works for them

leonardr:
This is a support branch for https://jira.nypl.org/browse/SIMPLY-1981. It changes the way grouped feeds are generated.

For normal Lanes, the behavior is unaffected -- a FeaturedFacets object means that the ElasticSearch scoring function is changed to make a random sample of high-quality works show up at the top (rather than e.g. works showing up in alphabetical order).

For other types of WorkLists, FeaturedFacets isn't very useful -- it just says "show some featured stuff". This branch introduces the `WorkList.overview_facets` method, which takes a FeaturedFacets object and returns some other object appropriate for _that WorkList_ to use when "showing some featured stuff".

This branch also brings in some generally useful code from circulation -- a faceting object that works just like `Facets` except it changes the default sort order, possibly to a field not normally enabled in `Facets`.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -320,8 +320,10 @@ class Facets(FacetsWithEntryPoint):
     feeds that list all the works in some WorkList.
     """
     @classmethod
-    def default(cls, library):
-        return cls(library, collection=None, availability=None, order=None)
+    def default(cls, library, collection=None, availability=None, order=None,
+                entrypoint=None):
+        return cls(library, collection=collection, availability=availability,
+                   order=availability, entrypoint=entrypoint)
 
     @classmethod
     def available_facets(cls, config, facet_group_name):
--- a/lane.py
+++ b/lane.py
@@ -1036,6 +1036,11 @@ class WorkList(object):
     # default. Most WorkList subclasses will override this.
     MAX_CACHE_AGE = 14*24*60*60
 
+    # In your subclass, set MAX_CACHE_AGE to this value to guarantee
+    # that cached feeds never expire -- they must be explicitly
+    # regenerated.
+    CACHE_FOREVER = 'forever'
+
     # By default, a WorkList is always visible.
     visible = True
 
--- a/lane.py
+++ b/lane.py
@@ -1413,8 +1413,9 @@ class WorkList(object):
                 continue
 
             if isinstance(child, Lane):
-                # Children that turn out to be Lanes go into relevant_lanes.
-                # Their Works will all be filled in with a single query.
+                # Children that turn out to be Lanes go into
+                # relevant_lanes. Their Works will be obtained from
+                # the search index.
                 relevant_lanes.append(child)
             # Both Lanes and WorkLists go into relevant_children.
             # This controls the yield order for Works.
--- a/lane.py
+++ b/lane.py
@@ -323,7 +323,7 @@ class Facets(FacetsWithEntryPoint):
     def default(cls, library, collection=None, availability=None, order=None,
                 entrypoint=None):
         return cls(library, collection=collection, availability=availability,
-                   order=availability, entrypoint=entrypoint)
+                   order=order, entrypoint=entrypoint)
 
     @classmethod
     def available_facets(cls, config, facet_group_name):
--- a/lane.py
+++ b/lane.py
@@ -1032,8 +1032,8 @@ class WorkList(object):
     By default, these Work objects come from a search index.
     """
 
-    # Unless a sitewide setting intervenes, the set of Works in a
-    # WorkList is cacheable for two weeks by default.
+    # The set of Works in a WorkList is cacheable for two weeks by
+    # default. Most WorkList subclasses will override this.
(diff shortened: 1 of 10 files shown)


lane.py
```diff
@@ -320,8 +320,10 @@ class Facets(FacetsWithEntryPoint):
     feeds that list all the works in some WorkList.
     """
     @classmethod
-    def default(cls, library):
-        return cls(library, collection=None, availability=None, order=None)
+    def default(cls, library, collection=None, availability=None, order=None,
+                entrypoint=None):
+        return cls(library, collection=collection, availability=availability,
+                   order=availability, entrypoint=entrypoint)
```
leonardr: At this point we could probably just change the constructor to make these arguments optional.

model/cachedfeed.py
```diff
@@ -78,14 +80,9 @@ def fetch(cls, _db, lane, type, facets, pagination, annotator,
...
-                max_age = AcquisitionFeed.grouped_max_age(_db)
-            elif type == cls.PAGE_TYPE:
-                max_age = AcquisitionFeed.nongrouped_max_age(_db)
-            elif hasattr(lane, 'MAX_CACHE_AGE'):
-                max_age = lane.MAX_CACHE_AGE
-            else:
-                max_age = 0
+            # Figure out the default max age for a feed of this type
+            # for this lane.
+            max_age = cls.calculate_max_age(_db, lane, type)
```
leonardr: I was getting test failures in circulation because of this code, and I think this code is wrong in a world where there are lots of custom WorkLists which all define MAX_CACHE_AGE. But it didn't have good test coverage. I moved it into a separate method, implemented the behavior I think is right (although a lot of this behavior will probably never happen in real life), and tested it.

model/cachedfeed.py
```diff
@@ -66,6 +66,8 @@ class CachedFeed(Base):
     GROUPS_TYPE = u'groups'
     PAGE_TYPE = u'page'
     NAVIGATION_TYPE = u'navigation'
+    CRAWLABLE_TYPE = u'crawlable'
+    RELATED_TYPE = u'related'
```
leonardr: Adding some constants here for feed types used in circulation.

model/cachedfeed.py
```diff
@@ -172,6 +169,45 @@ def fetch(cls, _db, lane, type, facets, pagination, annotator,
...
+        """Helper method to calculate the proper cache time for
+        a WorkList.
+        """
+        from ..opds import AcquisitionFeed
+        from ..lane import Lane
+        # The cache time might come from a site-wide database setting
+        # or from the class definition.
+        database_default = None
+        if type == cls.GROUPS_TYPE:
+            database_default = AcquisitionFeed.grouped_max_age(_db)
```
aslagle: What's the point of keeping a setting for this if we're never going to use it?

aslagle: I guess this is used for WorkLists that aren't lanes, but one of those is the top-level WorkList for a library, which should also be cached forever.

aslagle: And looks like that has a class setting of 2 weeks.

aslagle: Since WorkList has a class setting, this database default will never be used unless a subclass of WorkList resets it to None.

leonardr: I think we probably should get rid of the grouped_max_age setting.

I also think we should either get rid of ungrouped_max_age, or set Lane.MAX_CACHE_AGE to None so that the setting will take effect more often.

Should we do this now?

aslagle: I'd say so, I kind of doubt anyone uses these settings unless it's for development or testing.

tests/test_lane.py
```diff
@@ -1446,6 +1446,15 @@ def test_initialize(self):
...
 
+    def test_initialize_without_library(self):
+        # It's possible to initialize a WorkList with no Library.
+        worklist = WorkList()
+        worklist.initialize(None)
+
+        # No restriction is placed on the collection IDs of the
+        # Works in this list.
+        eq_(None, worklist.collection_ids)
+
```
leonardr: This fixes a problem I saw in circulation: a CrawlableCollectionLane was initialized with a collection ID instead of a library. That worked fine, but then a SpecificWorkList was initialized based on the (nonexistent) library of the CrawlableCollectionLane. The behavior should have been "get these works regardless of which collections they're in", but it was "get these works but only if they're in the null set of collections", which was matching nothing.

model/cachedfeed.py
leonardr: OK, I took the whole thing out.

There's more we could do -- theoretically you can always derive `type` from the WorkList + the Facets, but I don't want to add that complexity in this branch.

aslagle: I think they're still in config.py. Otherwise, looks good!

PR #1085, opened 2019-07-09
Body refactoring

leonardr:
This is hopefully the last branch before I merge elasticsearch-paginated-lists into master. It makes more miscellaneous changes to external_search.py to make the search system easier to explain.

The main changes:

* Instead of a bunch of generic query objects, we now import specific classes like `Term` and `DisMax` from `elasticsearch_dsl.query`. This makes it easier to talk about pieces of queries.
* I replaced the `icu_collation_property_hook` method with `sort_author_keyword_property_hook`. The `icu_collation_property_hook` was designed for an `icu_collation_property` that also needs a normalizer, but the only such field is `sort_author`, and we were adding a bunch of stuff onto `sort_author` after defining it as a custom type. It's simpler to define a custom data type that's just for `sort_author`, and let `sort_title` be a standard `icu_collation_property`.
* I moved the code that makes scoring functions for "featurability" out of lane.py and into `Filter`. Conceptually it belongs in `FeaturedFacets`, but this keeps all of the Elasticsearch code in `external_search.py`, which I think makes it easier to wrap your head around the whole system. Instead of doing this, I could move `FeaturedFacets` itself into `external_search.py`.

### DIFF
--- a/tests/test_external_search.py
+++ b/tests/test_external_search.py
@@ -1356,9 +1356,9 @@ class TestAuthorFilter(EndToEndSearchTest):
             lc="n2013008575"
         )
         self.display_name = Contributor(
-            sort_name=Edition.UNKNOWN_AUTHOR, display_name='Ann Leckie'
+            sort_name=Edition.UNKNOWN_AUTHOR, display_name='ann leckie'
         )
-        self.sort_name = Contributor(sort_name='Leckie, Ann')
+        self.sort_name = Contributor(sort_name='LECKIE, ANN')
         self.viaf = Contributor(
             sort_name=Edition.UNKNOWN_AUTHOR, viaf="73520345"
         )
--- a/tests/test_external_search.py
+++ b/tests/test_external_search.py
@@ -1754,8 +1754,7 @@ class TestFeaturedFacets(EndToEndSearchTest):
             "Works permuted by a random seed",
             worklist, random_facets,
             [self.hq_available_2, self.hq_available,
-             self.not_featured_on_list,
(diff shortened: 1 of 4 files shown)


tests/test_external_search.py
```diff
@@ -1352,9 +1356,9 @@ def populate_works(self):
             lc="n2013008575"
         )
         self.display_name = Contributor(
-            sort_name=Edition.UNKNOWN_AUTHOR, display_name='Ann Leckie'
+            sort_name=Edition.UNKNOWN_AUTHOR, display_name='ann leckie'
         )
-        self.sort_name = Contributor(sort_name='Leckie, Ann')
+        self.sort_name = Contributor(sort_name='LECKIE, ANN')
```
leonardr: I was concerned that without a `lowercase` filter the sort_author_keyword type wouldn't treat these as the same person, so I added a test for it, but it appears to work as is.

tests/test_external_search.py
```diff
@@ -1684,12 +1749,13 @@ def assert_featured(description, worklist, facets, expect):
         #
         # The random element is relatively small, so it mainly acts
         # to rearrange works whose scores were similar before.
-        random_facets = FeaturedFacets(1, random_seed=41)
+        random_facets = FeaturedFacets(1, random_seed=43)
```
leonardr: I had to change the seed to get the same results as before. This doesn't make sense to me. The seed goes into Elasticsearch, and I don't know what happens in there.

aslagle: Weird!

PR #1087, opened 2019-07-10
Add a script to update lane sizes

leonardr:
I also added a test for the very similar UpdateCustomListSizeScript, and fixed an apparent bug in its superclass.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -817,7 +817,7 @@ class CustomListSweeperScript(LibraryInputScript):
     def process_library(self, library):
         lists = self._db.query(CustomList).filter(CustomList.library_id==library.id)
         for l in lists:
-            self.process_custom_list(self, l)
+            self.process_custom_list(l)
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -817,7 +817,7 @@ class CustomListSweeperScript(LibraryInputScript):
     def process_library(self, library):
         lists = self._db.query(CustomList).filter(CustomList.library_id==library.id)
         for l in lists:
-            self.process_custom_list(self, l)
+            self.process_custom_list(l)
```
leonardr: Here's the bug -- I don't think this script is working now.

aslagle: Checked the open bookshelf logs and this has in fact been failing - thanks!

This is a performance optimization for the admin interface which currently only shows lists associated with a library anyway. It's not used anywhere else.

aslagle (approved):

PR #1088, opened 2019-07-11
Fix self-test and index rebuild issues

leonardr:
This branch fixes real-world bugs I found while finishing https://jira.nypl.org/browse/SIMPLY-2107 and https://jira.nypl.org/browse/SIMPLY-2109, plus real-world bugs I found in the Elasticsearch self-test code (which is heavily mocked in our unit tests).

I've tested this on the circulation branch I have in progress by running `bin/repair/search_index`, `bin/update_lane_size`, and by running the Elasticsearch self-tests.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -243,6 +243,9 @@ class BaseCoverageProvider(object):
             # at the start of the database table.
             original_finish = progress.finish = None
             progress.offset = 0
+
+            # Call run_once() until we get an exception or
+            # progress.finish is set.
             while not progress.is_complete:
                 try:
                     new_progress = self.run_once(
--- a/coverage.py
+++ b/coverage.py
@@ -241,7 +241,7 @@ class BaseCoverageProvider(object):
             #
             # Also set the offset to zero to ensure that we always start
             # at the start of the database table.
-            progress.finish = None
+            original_finish = progress.finish = None
             progress.offset = 0
 
             # Call run_once() until we get an exception or
@@ -274,8 +274,8 @@ class BaseCoverageProvider(object):
                 # value for progress.finish to stand. It might
                 # incorrectly make progress.is_complete appear to be
                 # True, making us exit the loop before we mean to.
-                if progress is not None:
-                    progress.finish = old_finish
+                if (progress is not None and not progress.exception):
+                    progress.finish = original_finish
         return progress
 
     @property
--- a/coverage.py
+++ b/coverage.py
@@ -243,39 +243,46 @@ class BaseCoverageProvider(object):
             # at the start of the database table.
             original_finish = progress.finish = None
             progress.offset = 0
+            progress = self._run_until_complete(
+                progress, count_as_covered=covered_statuses
+            )
+        # TODO: We should be able to return a list of progress objects,
+        # not just one.
+        return progress
 
-            # Call run_once() until we get an exception or
-            # progress.finish is set.
-            while not progress.is_complete:
-                try:
-                    new_progress = self.run_once(
-                        progress, count_as_covered=covered_statuses
-                    )
-                    # run_once can either return a new
-                    # CoverageProviderProgress object, or modify
-                    # in-place the one it was passed.
-                    if new_progress is not None:
-                        progress = new_progress
-                        original_finish = new_progress.finish
-                except Exception, e:
-                    logging.error(
-                        "CoverageProvider %s raised uncaught exception.",
-                        self.service_name, exc_info=e
-                    )
-                    progress.exception=traceback.format_exc()
+    def _run_until_complete(self, progress, **run_once_kwargs):
+        """Call run_once() until we get an exception or run_once() sets
+        progress.finish.
+        """
+        while not progress.is_complete:
+            try:
+                new_progress = self.run_once(progress, **run_once_kwargs)
+                # run_once can either return a new
+                # CoverageProviderProgress object, or modify
+                # in-place the one it was passed.
+                if new_progress is not None:
+                    progress = new_progress
+            except Exception, e:
+                logging.error(
+                    "CoverageProvider %s raised uncaught exception.",
+                    self.service_name, exc_info=e
+                )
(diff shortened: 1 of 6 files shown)


external_search.py
```diff
@@ -326,6 +331,7 @@ def setup_index(self, new_index=None, **index_settings):
         """
         index_name = new_index or self.works_index
         if self.indices.exists(index_name):
+            self.log.info("Deleting index %s", index_name)
```
leonardr: This way someone who runs bin/repair/search_index hears about the index being deleted.

external_search.py
```diff
@@ -602,13 +608,13 @@ def _search():
         def _works():
             return self.query_works(
                 self.test_search_term, filter=None, pagination=None,
-                debug=False
+                debug=True
```
leonardr: Without this, we don't have .sort_title and .sort_author to display when we render the results.

external_search.py
```diff
@@ -602,13 +608,13 @@ def _search():
...
             )
 
         # The self-tests:
 
         def _search_for_term():
-            titles = [("%s (%s)" %(x.title, x.author)) for x in _works()]
+            titles = [("%s (%s)" %(x.sort_title, x.sort_author)) for x in _works()]
```
leonardr: Those fields aren't available anymore.

external_search.py
```diff
@@ -640,21 +646,16 @@ def _get_raw_results():
...
             _count_docs
         )
 
         def _total_count():
-            # The mock methods used in testing return a list, so we have to call len() rather than count().
-            if in_testing:
-                return str(len(self.search))
-
-            return str(self.search.count())
+            return str(self.count_works(None))
```
leonardr: Here we don't need to treat the mocked method differently, because there's now a special method to count all the works in the index, and we mock that method.

external_search.py
```diff
@@ -2663,57 +2664,12 @@ def __contains__(self, k):
...
         return {
-            "title": self.title,
-            "author": self.author,
+            "title": self.sort_title,
+            "author": self.sort_author,
             "id": self.meta["id"],
             "meta": self.meta,
         }
 
-class SearchIndexMonitor(WorkSweepMonitor):
```
leonardr: Since bin/repair/search_index now deletes all the search coverage records and then recreates coverage, there's no need for `SearchIndexMonitor`, which was a less reliable way of doing the same thing.

scripts.py
```diff
@@ -345,22 +345,6 @@ def monitors(self, **kwargs):
...
 
 
-class UpdateSearchIndexScript(RunMonitorScript):
-
-    def __init__(self):
-        parser = argparse.ArgumentParser()
-        parser.add_argument(
-            '--works-index',
-            help='The ElasticSearch index to update, if other than the default.'
-        )
```
leonardr: I don't think we need the `--works-index` argument. If you use a different index for works, you'll configure it as part of your search integration, and we'll pick it up when we instantiate `ExternalSearchIndex`.

scripts.py
```diff
@@ -391,6 +381,7 @@ def do_run(self):
 
                 self.log.debug("Completed %s", provider.service_name)
                 providers.remove(provider)
+        return progress
```
leonardr: I wrote a test that needed to see what happened inside `do_run`, and since it can potentially do multiple things it made the most sense to return a list of CoverageProviderProgress objects.

scripts.py
```diff
@@ -872,19 +863,19 @@ def __init__(self, provider, _db=None, cmd_args=None, *provider_args, **provider
...
-            if parsed_args.identifiers:
-                self.identifiers = parsed_args.identifiers
-            else:
-                self.identifiers = []
+        if parsed_args.identifiers:
+            self.identifiers = parsed_args.identifiers
+        else:
+            self.identifiers = []
 
+        if callable(provider):
```
leonardr: This is totally unrelated -- I found it because I was mistakenly treating the search coverage provider as a CollectionCoverageProvider instead of a WorkCoverageProvider -- but I think it is a real bug fix. Previously, if you passed in an already-instantiated CollectionCoverageProvider object, `identifiers` and `identifier_types` wasn't being set, and you'd get a crash.

scripts.py
```diff
@@ -3148,7 +3139,15 @@ def add_line(id, name, protocol, metadata_identifier):
 
 
 class UpdateLaneSizeScript(LaneSweeperScript):
+
+    def should_process_lane(self, lane):
+        """We don't want to process generic WorkLists -- there's nowhere
+        to store the data.
+        """
+        return isinstance(lane, Lane)
+
```
leonardr: Without this, `bin/update_lane_size` starts off trying to process the top-level WorkList for a library, and it crashes immediately.

coverage.py
```diff
@@ -262,11 +262,25 @@ def run_once_and_update_timestamp(self):
...
                 self.finalize_timestampdata(progress)
+
+                # That wrote a value for progress.finish to the
+                # database, which is fine, but we don't necessarily
+                # want that value for progress.finish to stand. It
+                # might incorrectly make progress.is_complete appear
+                # to be True, making us exit the loop before we mean
+                # to.
+                if not progress.exception:
+                    progress.finish = original_finish
```
leonardr: I was really surprised by this bug. It means that no CoverageProvider will handle more than two batches (one batch for each value in `covered_status_lists`) on any given run. The first time it runs, `finalize_timestampdata` sets `progress.finish` for purposes of writing to the database. The loop thinks `finish` was set by `run_once`, and stops looking.

The only CoverageProvider where I would have noticed this problem is the search index one, and that one used a Monitor (removed in this branch), so it wouldn't have had this problem. It hasn't been a huge deal because our coverage providers run pretty frequently, but it does mean we've been getting coverage slower than we could have been.

scripts.py
aslagle: I think the original use case was to build a new index while still using the old index during an update.

PR #1089, opened 2019-07-12
Change 'genres' to a real nested document

leonardr:
This fixes https://jira.nypl.org/browse/SIMPLY-2119.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -1415,7 +1415,14 @@ class Query(SearchBase):
     def _match(cls, field, query_string):
         """A clause that matches the query string against a specific field in the search document.
         """
-        return Match(**{field: query_string})
+        match_query = Match(**{field: query_string})
+        if '.' in field:
+            # This is a query against a field from a subdocument. We
+            # can't run it against the top-level document; it has to
+            # be run in the context of its subdocument.
+            subdocument = field.split('.', 1)[0]
+            match_query = Nested(path=subdocument, query=match_query)
+        return match_query
 
     @classmethod
     def _match_phrase(cls, field, query_string):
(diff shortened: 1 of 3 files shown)


external_search.py
```diff
@@ -1408,7 +1415,14 @@ def fuzzy_string_query(cls, query_string):
...
         """A clause that matches the query string against a specific field in the search document.
         """
-        return Match(**{field: query_string})
+        match_query = Match(**{field: query_string})
+        if '.' in field:
+            # This is a query against a field from a subdocument. We
+            # can't run it against the top-level document; it has to
+            # be run in the context of its subdocument.
+            subdocument = field.split('.', 1)[0]
+            match_query = Nested(path=subdocument, query=match_query)
```
leonardr: At this point we can't refer to "foo.bar" unless "foo" is defined as a subdocument, but "genres" was the last subdocument we had that wasn't formally defined as one.

external_search.py
```diff
@@ -1922,7 +1936,10 @@ def build(self, _chain_filters=None):
             f = chain(f, self.target_age_filter)
 
         for genre_ids in self.genre_restriction_sets:
-            f = chain(f, Terms(**{'genres.term' : filter_ids(genre_ids)}))
+            ids = filter_ids(genre_ids)
+            nested_filters['genres'].append(
+                Terms(**{'genres.term' : filter_ids(genre_ids)})
+            )
```
leonardr: A filter on genres.term now needs to be a nested filter.

external_search.py
aslagle: That seems like it should have let you remove code.... why wasn't this needed for the other subdocuments?

leonardr: I think this is the only time we need to run a _query_ against a field that lives in a subdocument. All the other times we reference a subdocument it's because we're applying a filter.

It's possible that this will fall apart the second time we need to run a query against a subdocument field, and we'll have to go to something analogous to `nested_queries`.

leonardr: Yesterday I presented on the Elasticsearch code to Adriana, Edwin and Robert, and I came to the conclusion that this should, in fact, become a nested filter. The hypothesis we're testing is that we'll get better results if we restrict the search results to (let's say) "Young Adult" and then run a search for `divorce`, than if we just tell Elasticsearch to handle `young adult divorce`.

So that's one of the things we'll be looking into when it comes time to improve search results.

PR #1090, opened 2019-07-13
Extract add time for each title from Overdrive JSON data

leonardr:
This is a support branch for https://jira.nypl.org/browse/SIMPLY-2121. It surfaces one extra piece of information from Overdrive -- the time at which a book was added to Overdrive's collection -- which is necessary for the implementation of `NewTitlesOverdriveCollectionMonitor.should_stop`.

### DIFF
--- a/overdrive.py
+++ b/overdrive.py
@@ -355,8 +355,11 @@ class OverdriveAPI(object):
     def _get_book_list_page(self, link, rel_to_follow='next'):
         """Process a page of inventory whose circulation we need to check.
 
-        Returns a list of (title, id, availability_link) 3-tuples,
-        plus a link to the next page of results.
+        Returns a 2-tuple: (availability_info, next_link).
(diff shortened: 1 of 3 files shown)


overdrive.py
```diff
@@ -355,8 +355,11 @@ def _all_products_link(self):
     def _get_book_list_page(self, link, rel_to_follow='next'):
         """Process a page of inventory whose circulation we need to check.
 
-        Returns a list of (title, id, availability_link) 3-tuples,
-        plus a link to the next page of results.
+        Returns a 2-tuple: (availability_info, next_link).
```
EdwinGuzman: The documentation for the code wasn't right before, right?

leonardr: Yes, the old documentation thought `availability_info` was a list of 3-tuples, and it's actually a list of dictionaries.

PR #1091, opened 2019-07-16
Don't just index the subjects associated with a work's equivalent identifiers -- index the identifiers themselves

leonardr:
This is the core part of https://jira.nypl.org/browse/SIMPLY-2110.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -41,6 +41,7 @@ from classifier import (
     AgeClassifier,
 )
 from facets import FacetConstants
+from metadata_layer import IdentifierData
 from model import (
     numericrange_to_tuple,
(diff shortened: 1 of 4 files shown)


external_search.py
```diff
@@ -1953,7 +1968,7 @@ def build(self, _chain_filters=None):
             # copies should be displayed.
             available = Term(**{'licensepools.available' : True})
             nested_filters['licensepools'].append(
-                Bool(should=[open_access, available])
+                Bool(should=[open_access, available], minimum_should_match=1)
```
leonardr: This is the default behavior but I like to make it explicit.

PR #1093, opened 2019-07-16
Reindex a work whenever its last_update_time changes

leonardr:
This ensures that we reindex whenever LicensePool.update_availability is called. 

Ticket: https://jira.nypl.org/browse/SIMPLY-2122

### DIFF
--- a/tests/models/test_listeners.py
+++ b/tests/models/test_listeners.py
@@ -15,6 +15,7 @@ from ...model import (
     create,
     site_configuration_has_changed,
     Timestamp,
+    WorkCoverageRecord,
 )
 
 class TestSiteConfigurationHasChanged(DatabaseTest):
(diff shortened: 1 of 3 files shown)


tests/models/test_listeners.py
```diff
@@ -181,3 +182,56 @@ def test_configuration_relevant_collection_change_updates_configuration(self):
                facets='', library=library)
         self._db.commit()
         self.mock.assert_was_not_called()
+
+class TestWorkReindexing(DatabaseTest):
+    """Test the circumstances under which a database change
+    requires that a Work's entry in the search index be recreated.
+    """
+
+    def _assert_work_needs_update(self, work):
```
leonardr: Most of these tests are of preexisting code that didn't previously have tests.

PR #1095, opened 2019-07-24
Overhaul the system that builds hypotheses to handle search requests

leonardr:
This branch overhauls the system that builds the dis_max query to handle search requests. This dramatically improves the quality of search results and simplifies the hypotheses that were going into Elasticsearch -- previously this was scattershot and based on trial and error rather than a coherent plan.

The downside is reduced performance. We're testing many more different hypotheses, including some that are quite complicated. Here's a timing report I made using the `test_search.py` integration test in [my circulation PR](https://github.com/NYPL-Simplified/circulation/pull/1283) (which will be ready tomorrow), which has an integration test suite that runs about 280 realistic searches against a real library collection.

The report for 2.x:

```
 mean 0.152685121107
 median 0.141
 std 0.0933304273608
 Most expensive:
   (1.139, 'wolfgang burger')
   (0.592, 'black age 3-5')
   (0.437, 'rguhriregiuh43pn5rtsadpfnsadfausdfhaspdiufnhwe42uhdsaipfh')
   (0.409, 'law of the mountain man william johnstone')
   (0.395, 'seventh grade science')
   (0.369, 'books about texas like the fair')
   (0.362, 'rguhriregiuh 43pn5rts adpfnsadfaus dfhaspdiufnhwe4 2uhdsaipfh')
   (0.345, 'the gurnsey literary and potato peel society')
   (0.327, 'isaac asimov foundation book 1')
   (0.324, 'panda age 3-5')
```

The report for this branch:

```
 mean 0.21395862069
 median 0.184
 std 0.122345743883
 Most expensive:
   (1.146, 'harlequin historical')
   (0.845, 'law of the mountain man william johnstone')
   (0.789, 'the gurnsey literary and potato peel society')
   (0.646, 'rguhriregiuh43pn5rtsadpfnsadfausdfhaspdiufnhwe42uhdsaipfh')
   (0.616, 'agatha christie mystery')
   (0.61, 'rguhriregiuh 43pn5rts adpfnsadfaus dfhaspdiufnhwe4 2uhdsaipfh')
   (0.595, 'modern romance aziz ansari')
   (0.531, 'the hunger games mockingjay')
   (0.493, 'percy jackson graphic novel')
   (0.474, 'iain banks science fiction')
```

So the the "search" part is about 40% slower. It used to be much worse but I was able to improve it quite a bit by reducing the complexity of our author match queries, at little cost to search result quality.


external_search.py
```diff
@@ -996,7 +1051,8 @@ def __init__(self):
         fields_by_type = {
             "basic_text": ['summary'],
             'filterable_text': [
-                'title', 'subtitle', 'series', 'classifications.term'
+                'title', 'subtitle', 'series', 'classifications.term',
+                'author',
```
leonardr: We index `author` even though we have all the `contributors`, because it dramatically simplifies the handling of queries like  `nalini singh archangel` or `fallen baldacci`. In those cases we want to do a cross-field search including `title` and `contributors.display_name`. I wasn't able to get cross-field search working between a field in the main document and a field in the subdocument, but we have `author` already available as a field in the main document, and searching on that is a lot faster than going into the subdocuments anyway.

external_search.py
```diff
@@ -1103,85 +1159,6 @@ def _match_range(cls, field, operation, value):
...
-        'subtitle^3',
-        'summary^2',
-        "classifications.term^2",
-
-        # These fields only use the standard analyzer and are closer to the
-        # original text.
-        'author^6',
-        'publisher',
-        'imprint'
-    ]
```
leonardr: We no longer rely on Lucene syntax to weight fields differently, and we no longer tell Elasticsearch to do blanket searches across large numbers of fields.

external_search.py
```diff
@@ -1103,85 +1159,6 @@ def _match_range(cls, field, operation, value):
...
-        "biology", "ecology", "zoology", "geology",
-
-        "joke", "jokes" # "jake"
-
-        "cat", "cats",
-        "car", "cars",
-        "war", "wars",
-
-        "away", "stay",
-    ]
```
leonardr: We now "spell-check" the search result, and if it looks like everything is spelled correctly the fuzzy hypotheses are given a much lower weight. So there's no need to check for specific words.

I added integration tests of some of these fuzzy confounders to the circulation branch, to verify that any future solutions don't reintroduce this problem.

leonardr:
The way I see it is, the Python work is all business logic. People are more likely to do an author search than a subject search, stopwords could be important in series name but probably not in publisher name, searches are more likely to split title/subtitle than title/subject, human names are really weird, etc. The performance hit comes from Elasticsearch -- we're asking it to do 5-10 times the work we were asking before and it's pretty impressive that it only takes 40% longer.

Elasticsearch has a DSL (the Lucene query parser syntax) which we were using to change the relative importance of different fields. We could have implemented some of this business logic within that DSL, but it's more readable (and much more powerful) to do it in Python.

The only thing that I could see easily moving from Python to Elasticsearch is the bit where we convert "ursula k leguin" into "leguin, ursula k" in case we don't have a display name for a contributor. Even then, that work would probably be done in `Work.to_search_document`, so it would just move from one bit of Python to another.

PR #1100, opened 2019-08-02
Changes necessary to get core tests working after Python 3 converstion

leonardr:


s3.py
```diff
@@ -237,7 +237,7 @@ def key_join(self, key):
             if isinstance(part, str):
                 part = part.encode("utf-8")
             else:
-                part = str(part)
+                part = bytes(part)
```
EdwinGuzman: @leonardr come back to this later. What if there is a list containing a number.

PR #1101, opened 2019-08-09
Weight keyword matches for publisher and imprint much lower than keyw…

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-2203.
Most of the tests are integration tests, which I've added to https://github.com/NYPL-Simplified/circulation/pull/1296

This solution isn't ideal, but without intelligence about whether a string is likely to be a publisher search, I don't see a better option.

### DIFF
--- a/tests/test_external_search.py
+++ b/tests/test_external_search.py
@@ -418,7 +418,7 @@ class TestExternalSearchWithWorks(EndToEndSearchTest):
         self.subtitle_match.presentation_edition.subtitle = "Match"
 
         self.summary_match = _work(title="SummaryM")
-        self.summary_match.summary_text = "It's a Match!"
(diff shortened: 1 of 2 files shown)


tests/test_external_search.py
```diff
@@ -595,12 +595,12 @@ def test_query_works(self):
...
         expect(self.moby_dick, "gutenberg")
 
-        # Title > subtitle > publisher > word found in summary
+        # Title > subtitle > word found in summary > publisher
         order = [
             self.title_match,
             self.subtitle_match,
-            self.publisher_match,
             self.summary_match,
+            self.publisher_match,
```
leonardr: This happens even though the summary match isn't a keyword match. In real usage, though, I do see publisher matches frequently showing up above summary matches -- they're just a lot closer together.

PR #1102, opened 2019-08-12
Prepare Python 2 code for conversion to Python 3

leonardr:
This branch rewrites a bunch of code that can't be automatically converted to Python 3, so that it _can_ be automatically converted to Python 3. Ultimately, this conversion will stand or fall on the strength of our test coverage, but making all the changes we can make beforehand will make it easier to evaluate the eventual conversion.

Most of these changes have to do with the distinction between bytestrings and Unicode strings. Our current code is ambiguous about whether a 'string' is bytes or Unicode, and `2to3` either guesses wrong or doesn't touch code that ought to be converted.

In general, we would like everything to be Unicode, but we have to deal with bytestrings in the following places:

* We read bytestrings from files
* We read bytestrings from HTTP response bodies
* The `Representation.content` database field is a bytestring
* The `Work.marc_record` database field is a bytestring (see below)
* base64 encoding or decoding operates on a bytestring and produces a bytestring
* Encryption or decryption operates on a bytestring and produces a bytestring
* `os.urandom` returns a bytestring

The base64 thing, in particular, was a huge pain. I ended up defining a helper class `UnicodeAwareBase64`, an instance of which acts like a drop-in replacement for the standard `base64` module. This lets us base64-encode Unicode strings and get Unicode strings back. Strings are transparently encoded as UTF-8 (if necessary), then base64-encoded and then decoded to Unicode.

Other changes we had to anticipate:

* In Python 3, you can't check whether `None`  is greater or less than a non-`None` value. This means you can't just call `sorted()` on  a list of values if `None` is in the list. You can't just call `sorted()` on list of dictionaries if any of the keys or values are `None`.
* In Python 3, `Exception.message` is gone. Instead, you convert the exception to Unicode with `str(exception)`. In Python 3 this means `unicode(exception)`.
* Functions like `map` and `reduce` have been moved to `functools` and their use is discouraged.
* The `md5` module is gone and `hashlib.md5` works slightly differently.
* The scoping rules have changed. In particular, you can't refer to class variables within a list comprehension that's running later in the class definition. Variables defined in an `except` block are also local.
* We had some files that mixed spaces and tabs, which isn't allowed in Python 3.
* In Python 2, `__repr__` must return a bytestring. In Python 3, `__repr__` must return a Unicode string. I defined a helper function for this which we can remove after the converstion.

I couldn't do anything about these changes within Python 2, but I added `TODO PYTHON3` notices around the points where we'll have to change:

* In Python 2, all output (to standard output or a logger) must be encoded. In Python 3, all such output must be Unicode.
* In Python 3, dictionaries are sorted by insertion order. It's not necessary to sort dictionaries in most tests, which is good because most of our dictionaries have a `None` somewhere in them and can't be sorted normally. We can use `collections.OrderedDict` to get the Python 3 behavior in Python 2, but sometimes it's not possible (because the dictionary is created by another library such as lxml) or it's just not worth the trouble.
* Our current implementation of `S3Uploader.key_join` returns a bytestring. In Python 3 it should return a Unicode string instead, since it's going to be mixed with Unicode URL fragments.

Regarding `Work.marc_record`, I don't know what we need in terms of a database migration. It looks like changing the sqlalchemy column to `Binary` works fine on new databases (where marc_record will be created as a binary field) and on old databases (where marc_record will be a string field). Since MARC records aren't huge, there shouldn't be a performance problem to letting `Work.marc_record` remain a string field, and it would avoid a potentially expensive migration script. 

But, it's ugly, and once we actually migrate to Python 3, the sqlalchemy behavior might change in a way that makes a migration script necessary. I'd like your thoughts on this.


scripts.py
```diff
@@ -2497,16 +2500,30 @@ def update_timestamps(self, migration_file):
...
+                return
+            # The previous script has a higher counter
+            if (self.overall_timestamp.counter is not None and
+                self.overall_timestamp.counter > counter):
+                return
+
+        self.overall_timestamp.update(
+            self._db, finish=last_run_date,
+            counter=counter, migration_name=migration_file
+        )
```
leonardr: This is an especially bad case of "can't compare None against a non-None value". We had to unroll the entire algorithm.

config.py
```diff
@@ -509,7 +513,7 @@ def site_configuration_last_update(cls, _db, known_value=None,
             from model import ConfigurationSetting
             timeout = ConfigurationSetting.sitewide(
                 _db, cls.SITE_CONFIGURATION_TIMEOUT
-            ).value
+            ).int_value
```
leonardr: This is a bug in our original code that was exposed by the Python 3 conversion.

opds.py
```diff
@@ -1806,10 +1805,8 @@ def group_uri(self, work, license_pool, identifier):
         lanes = self.lanes_by_work.get(work, None)
 
         if lanes:
-            lane_name = lanes[0]['lane'].display_name
-            additional_lanes = lanes[1:]
-            if additional_lanes:
-                self.lanes_by_work[work] = additional_lanes
+            lane_dic = lanes.pop(0)
+            lane_name = lane_dic['lane'].display_name
```
leonardr: I don't remember what this change was, so I'm a little suspicious of it, but it's another bug that the Python 3 conversion caught. I remember it was causing the same work to show up multiple times in a list. And it doesn't seem like it belongs in a method that generates a URI.

scripts.py
```diff
@@ -2156,7 +2157,7 @@ def sort_migrations(self, migrations):
         one.
         """
 
-        def compare_migrations(first, second):
+        def compare_migrations(first):
```
leonardr: In Python 3 you can't sort with a comparison method that takes two arguments and says which one is bigger. You have to take one argument and generate a sort key.

testing.py
```diff
@@ -1372,10 +1382,12 @@ def json(self):
         content = self.content
         # The queued content might be a JSON string or it might
         # just be the object you'd get from loading a JSON string.
-        if isinstance(content, basestring):
+        if isinstance(content, (unicode, bytes)):
```
leonardr: `2to3` replaces 'basestring' with 'str', which means Unicode strings as opposed to bytestrings. This is how we say 'both types of strings' in a way that will convert over.

util/__init__.py
```diff
@@ -1140,6 +1140,7 @@ def parse(cls, amount):
         currency = cls.DEFAULT_CURRENCY
         if not amount:
             amount = '0'
+        amount = unicode(amount)
```
leonardr: Not sure about this one, but I recall circulation code was sometimes passing a string into this method and sometimes a number.

config.py
```diff
@@ -247,7 +251,7 @@ class Configuration(object):
             "category": "Lanes & Filters",
         },
     ] + [
-        { "key": ENABLED_FACETS_KEY_PREFIX + group,
+        { "key": ConfigurationConstants.ENABLED_FACETS_KEY_PREFIX + group,
```
aslagle: This is pretty annoying but I don't see a better option.

model/constants.py
```diff
@@ -286,5 +286,16 @@ class MediaTypes(object):
         ".htm" : TEXT_HTML_MEDIA_TYPE,
         ".jpeg" : JPEG_MEDIA_TYPE,
     }
+    # PYTHON3 TODO: This happens to work in Python 2, but in Python 3,
+    # items come out of MEDIA_TYPE_FOR_EXTENSION in the order they
+    # were inserted, so we need to take steps to ensure
+    # SCORM_MEDIA_TYPE doesn't claim '.zip'.
```
aslagle: Are you sure this always works in Python 2?

model/customlist.py
```diff
@@ -61,9 +61,18 @@ class CustomList(Base):
     # interface for managing this.
 
     def __repr__(self):
+        # TODO PYTHON3 repr is unicode string
         return (u'<Custom List name="%s" foreign_identifier="%s" [%d entries]>' % (
             self.name, self.foreign_identifier, len(self.entries))).encode('utf8')
 
+    def __lt__(self, other):
```
aslagle: Why only this method?

leonardr: This one makes `sorted` work if you run it on a homogenous list of `CustomList` objects.

model/constants.py
leonardr: I'm not, but it's pretty stable -- otherwise I feel like we'd see intermittent test failures.
Python 2.7 has a class called `OrderedDict` which we can use here to get Python 3 behavior in Python 2. I can use that here.

model/customlist.py
aslagle: Can we rely on `sorted` only using this operator? It seems like an internal Python implementation detail. 

I also found https://docs.python.org/3.3/library/functools.html#functools.total_ordering when I was trying to understand this.

model/integrationclient.py
```diff
@@ -20,6 +20,7 @@
 from sqlalchemy.orm import (
     relationship,
 )
+from ..util.string_helpers import random_string
```
aslagle: You didn't use this.

model/resource.py
```diff
@@ -759,10 +760,11 @@ def get(cls, _db, url, do_get=None, extra_request_headers=None,
             media_type = cls._best_media_type(url, headers, presumed_media_type)
             if isinstance(content, unicode):
                 content = content.encode("utf8")
-        except Exception, fetch_exception:
+        except Exception as e:
```
aslagle: Why not write `Exception as fetch_exception`?

opds.py
aslagle: Looks like this only used for tests anyway.

model/resource.py
leonardr: The variable name shadows any variable with the same name in the enclosing code block, and when the except block ends, `fetch_exception` actually goes out of scope. (I wasn't expecting this; I thought at most it would go back to its previous value).

Here's a simple illustration of the difference between Python 2 and Python 3 behavior:

```
error = None
print("Error starts out", error)
try:
    a = int("abc")
except Exception as error:
    print("I just caught", error)
print("Error is now", error)
```

aslagle:
For MARC records, I'd be worried about having some people with binary and others with strings on the same version of the CM. I'd probably do the migration, or try to avoid changing the sqlalchemy column.

leonardr:
OK, tomorrow I'll look at `works.marc_record` in more detail. The SQLAlchemy column can't stay `Unicode`, but maybe `String` will work.

tests/test_metadata.py
```diff
@@ -1226,28 +1226,51 @@ def assert_registered(full):
 
     def test_apply_identifier_equivalency(self):
 
-        # Set up primary identifier with matching & new IdentifierData objects
+        # Set up an Edition.
```
leonardr: The behavior of this test changed when I first implemented `Identifier.__eq__`. I didn't have the type check on `other`, so an Identifier and an IdentifierData could have Python equality if they represented the same conceptual identifier. This meant that if you mentioned an `Identifier` and an `IdentifierData` in the `Metadata` constructor, you might end up with only one or the other in `Metadata.identifiers`.

I think a more expansive implementation if `Identifier.__eq__` might be a good idea, but it's way beyond the scope of this project. But by the time I'd figured out what was going on, I'd expanded this test and made it easier to read, so I'm leaving that bit in place.

model/work.py
```diff
@@ -218,7 +218,7 @@ class Work(Base):
     # A precalculated MARC record containing metadata about this
     # work that would be relevant to display in a library's public
     # catalog.
-    marc_record = Column(Binary, default=None)
+    marc_record = Column(String, default=None)
```
leonardr: I did a preliminary test of this code when converted to Python 3 and it seems to work -- data is read from a `String` field as a bytestring. A `Unicode` field is stored as a bytestring in the database, but decoded as UTF-8 on the way into Python code.

testing.py
leonardr: According to the porting guide, `2to3` handles this in a more sensible way, but it wasn't working here for some reason.

model/customlist.py
leonardr: That's a good find, I made `Identifier` and `CustomList` into @total_orderings.

util/string_helpers.py
```diff
@@ -64,3 +65,22 @@ def random_string(size):
     :return: A Unicode string.
     """
     return binascii.hexlify(os.urandom(size)).decode("utf8")
+
+
+def native_string(x):
```
leonardr: [The porting guide](https://portingguide.readthedocs.io/en/latest/strings.html#the-native-string) talks about the "native string" type, a.k.a. `str`, which represents Unicode data in Python 3 and byte data in Python 2. Usually we can avoid this data type by being explicit, but the `__repr__` method needs to return this type, which led to a lot of PYTHON3 TODOs.

But thinking about it that way, I figured I could get rid of a lot of those TODOs with a helper function that turns... some kind of string... into... the native string type for the current Python version.

log.py
```diff
@@ -43,18 +44,20 @@ def format(self, record):
...
     def format(self, record):
         try:
-            data = super(UTF8Formatter, self).format(record)
+            data = super(StringFormatter, self).format(record)
         except Exception, e:
-            data = super(UTF8Formatter, self).format(record)
-        if isinstance(data, unicode):
-            data = data.encode("utf8")
-        return data
+            data = super(StringFormatter, self).format(record)
```
aslagle: This isn't related to your change, but I'm not sure what this try-except could be for. It's catching an exception and then doing the exact same thing.

leonardr: That's true. There's also a place in JSONFormatter where we catch an exception and immediately re-raise it.

PR #1104, opened 2019-08-13
Make MockRequestsResponse more realistic

leonardr:
This branch adds some features to the mock Requests code so that we can mock Requests as used by the zeep SOAP client.

### DIFF
--- a/testing.py
+++ b/testing.py
@@ -1357,6 +1357,10 @@ class DummyHTTPClient(object):
         self.requests.append(url)
         return self.responses.pop(0)
 
+    def do_post(self, url, data, *wargs, **kwargs):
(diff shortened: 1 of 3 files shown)


testing.py
```diff
@@ -1391,3 +1396,9 @@ def text(self):
         if isinstance(self.content, bytes):
             return self.content.decode("utf8")
         return self.content
+
+    def raise_for_status(self):
```
EdwinGuzman: So this function is needed to make the mock request for realistic?

leonardr: Exactly. sudz thinks it has a real Requests response and calls this method on it, but it actually has a mocked response.

PR #1105, opened 2019-08-16
Improve Collection.for_metadata_identifier

leonardr:
This branch makes some improvements to the code that turns a metadata identifier into a Collection.
In particular, invalid or missing metadata identifiers result in distinct ValueErrors, not random base64 errors.

This is core work for https://github.com/NYPL-Simplified/metadata_wrangler/pull/199. Only the metadata wrangler uses Collection.for_metadata_identifier, so this is effectively metadata wrangler code.

### DIFF
--- a/model/collection.py
+++ b/model/collection.py
@@ -468,8 +468,11 @@ class Collection(Base, HasFullTableCache):
         this Collection on the metadata wrangler. This identifier is
         composed of the Collection protocol and account identifier.
 
-        In the metadata wrangler, this identifier is used as the unique
-        name of the collection.
+        A circulation manager provides a Collection's metadata
+        identifier as part of collection registration. The metadata
+        wrangler creates a corresponding Collection on its side,
+        _named after_ the metadata identifier -- regardless of the name
+        of that collection on the circulation manager side.
         """
         account_id = self.unique_account_id
         if self.protocol == ExternalIntegration.OPDS_IMPORT:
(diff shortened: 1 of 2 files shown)


model/collection.py
```diff
@@ -486,35 +486,57 @@ def metadata_identifier(self):
...
+            collection.create_external_integration(protocol)
+
+        if protocol == ExternalIntegration.OPDS_IMPORT:
+            # For OPDS Import collections only, we store the URL to
+            # the OPDS feed (the "account ID") and the data source.
+            collection.external_account_id = account_id
+            if data_source and not isinstance(data_source, DataSource):
+                data_source = DataSource.lookup(
+                    _db, data_source, autocreate=True
+                )
```
leonardr: Previously you had to look up a `DataSource` before calling this method. This was a bad idea because the lookup process might involve _creating_ a new data source, and then you might call this method on an invalid metadata identifier, causing an error. Now you've created a new data source for no reason. So I made it possible to pass a data source _name_ into this method before -- it'll only be created if the metadata identifier is valid.

tests/models/test_collection.py
```diff
@@ -441,36 +441,75 @@ def build_expected(protocol, unique_id):
...
+            self._db, collection.metadata_identifier, data_source=data_source
         )
         eq_(False, is_new)
         eq_(mirror_collection, result)
         # The external_account_id and data_source have been set now.
         eq_(collection.external_account_id, mirror_collection.external_account_id)
+
+        # A new DataSource object has been created.
+        source = new_data_source()
+        eq_("New data source", source.name)
```
EdwinGuzman: So here it is created because the metadata identifier is valid?

leonardr: Yes, that's right.

PR #1112, opened 2019-09-06
Add 'library' and 'location' to CirculationEvent

leonardr:
This is support work for https://jira.nypl.org/browse/SIMPLY-2185. It adds two fields to CirculationEvent -- `library_id` (a foreign key reference to Library.id) and `location` (a string that may only have meaning within a particular library system). It removes one field, `foreign_patron_identifier`, which hasn't been used for several years and could cause privacy problems, especially when combined with `location`.

Previously we had a unique constraint on (license_pool_id, event, start, foreign_patron_identifier). This constraint didn't actually constrain anything, because the constraint doesn't apply when one of the fields is null, and foreign_patron_identifier is always null. I redid this as two different unique indexes -- one on (license_pool_id, library_id, event, start) and one on (license_pool_id, event, start) for cases where library_id is null.

However, I don't think this is going to help in real situations, because in almost all situations we don't know the start date of the event, so we use the current time. AFAIK the only events for which we know the start date are events from the Bibliotheca API. It might be better for performance to remove these indexes.

### DIFF
--- a/model/circulationevent.py
+++ b/model/circulationevent.py
@@ -16,7 +16,7 @@ from sqlalchemy import (
     Index,
     Integer,
     String,
-    UniqueConstraint,
+    Unicode,
 )
 
 class CirculationEvent(Base):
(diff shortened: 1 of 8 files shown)


model/circulationevent.py
```diff
@@ -89,7 +131,10 @@ class CirculationEvent(Base):
 
     @classmethod
     def log(cls, _db, license_pool, event_name, old_value, new_value,
-            start=None, end=None, foreign_patron_id=None):
+            start=None, end=None, library=None, location=None):
+        """Log a CirculationEvent to the database, assuming it
+        hasn't already been recorded.
+        """
```
leonardr: This method was missing a test so I added one to make sure there's no regression here (apart from the removal of foreign_patron_id).

aslagle:
We do know the start date for events coming through the circulation manager, which are the most useful events anyway.

leonardr:
That's true, but those events are initiated immediately after they happen, so it's basically like the start date being the current time. Duplicates are only a potential problem when we're going over an event log describing things that happened in the past. And even then, duplicates will be caught so long as we use CirculationEvent.log instead of trying to make our own events.

I can commit this branch as-is and potentially remove the unique indexes later.

leonardr:
Looking at the circulation code, I think the change I made to the LocalAnalyticsProvider completes the fix to 2185. We already pass the library in to the analytics provider in CirculationAPI and AnalyticsController.

tests/models/test_circulationevent.py
```diff
@@ -106,3 +110,113 @@ def test_new_title(self):
...
+            library=library, old_value=old_value, new_value=new_value,
+            start=start, end=end, location=location
+        )
+        eq_(True, is_new)
+        eq_(pool, event.license_pool)
+        eq_(library, event.library)
+        eq_(-2, event.delta)  # calculated from old_value and new_value
+        eq_(start, event.start)
+        eq_(end, event.end)
+        eq_(location, location)
```
EdwinGuzman: Should these be `eq_(location, event.location)`?

leonardr: Yep.

PR #1113, opened 2019-09-06
Added a monitor that scrubs .location from old CirculationEvents.

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-2272.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -18,6 +18,7 @@ from model import (
     get_one,
     get_one_or_create,
     CachedFeed,
+    CirculationEvent,
     Collection,
     CollectionMissing,
     CoverageRecord,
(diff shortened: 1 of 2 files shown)


monitor.py
```diff
@@ -849,3 +850,37 @@ def delete(self, collection):
...
+    MAX_AGE = datetime.timedelta(days=365)
+
+    def run_once(self, *args, **kwargs):
+        rows_scrubbed = 0
+        qu = self.query()
+        self.log.info("Scrubbing %d row(s)", qu.count())
+        for i in qu:
+            self.scrub(i)
+            rows_scrubbed += 1
+        return TimestampData(achievements="Items scrubbed: %d" % rows_scrubbed)
```
leonardr: Most reapers don't implement run_once(), and this one only does so that we don't issue misleading messages about "Items deleted" or have a misleadingly-named delete() method.

PR #1117, opened 2019-09-13
Main collection is everything

leonardr:
This branch implements https://jira.nypl.org/browse/SIMPLY-2286. It removes one facet -- "main collection" from the "collection" facet group, leaving only "everything" and "popular books".

In addition to a migration script (which is a little complicated) I changed the one piece of the code where an unrecognized facet or facet group could cause a crash: `AcquisitionFeed.facet_links`. This is the code that looks up the human-readable names of facets or facet groups. If there's no human-readable name for a facet or facet group, the code used to crash. Now it just ignores that facet or facet group.

### DIFF
--- a/migration/20190916-remove-main-collection-facet.sql
+++ b/migration/20190916-remove-main-collection-facet.sql
@@ -8,7 +8,7 @@
 update configurationsettings set value = replace(value, '"main",', '') where key='facets_enabled_collection';
 
 -- Remove ', "main"' for cases where "main" is the last item in the JSON list.
-update configurationsettings set value = replace(value, ', "main"', '') where key='facets_enabled_collection';
+update configurationsettings set value = replace(value, ', "main"]', ']') where key='facets_enabled_collection';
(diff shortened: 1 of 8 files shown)


migration/20190916-remove-main-collection-facet.sql
```diff
@@ -0,0 +1,14 @@
...
+-- group.
+
+-- This migration script removes it from the list of enabled facets,
+-- if it's present, and changes the default if 'main' is the default.
+
+-- Remove '"main",' for cases where "main" is not the last item in the JSON list.
+update configurationsettings set value = replace(value, '"main",', '') where key='facets_enabled_collection';
+
+-- Remove ', "main"' for cases where "main" is the last item in the JSON list.
+update configurationsettings set value = replace(value, ', "main"]', ']') where key='facets_enabled_collection';
```
aslagle: This won't work if main is somehow the only enabled facet.

PR #1120, opened 2019-09-24
Updating the mirror uploader to not be sitewide

EdwinGuzman:
Resolves [SIMPLY-2296](https://jira.nypl.org/browse/SIMPLY-2296).

### DIFF
--- a/marc.py
+++ b/marc.py
@@ -521,15 +521,7 @@ class MARCExporter(object):
         },
     ]
 
-    SETTINGS = [
-        { "key": STORAGE_PROTOCOL,
-          "label": _("Storage service protocol"),
-          "description": _("Storage protocol to use for uploading generated MARC files. The service must already be configured under 'Storage Services'."),
-          "type": "select",
-          "options": [{ "key": ExternalIntegration.S3, "label": ExternalIntegration.S3 }],
-          "default": ExternalIntegration.S3,
-        },
-    ]
+    SETTING_DESCRIPTION = "Storage protocol to use for uploading generated MARC files. The service must already be configured under 'Storage Services'."
 
     @classmethod
     def from_config(cls, library):
@@ -597,7 +589,7 @@ class MARCExporter(object):
 
     def records(self, lane, annotator, start_time=None, force_refresh=False,
                 mirror=None, search_engine=None, query_batch_size=500,
-                upload_batch_size=7500
+                upload_batch_size=7500, storage_name=None
     ):
         """
         Create and export a MARC file for the books in a lane.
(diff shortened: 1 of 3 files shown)


mirror.py
```diff
@@ -44,38 +44,26 @@ def sitewide_integration(cls, _db):
...
-                    _db = Session.object_session(collection)
-                    return cls.sitewide(_db)
-                except CannotLoadConfiguration, e:
-                    return None
-            else:
+            try:
+                from model import Session
+                _db = Session.object_session(collection)
+                #todo get name of storage
+                return cls.mirror(_db)
```
EdwinGuzman: Here I'm not sure how to get the name of the storage integration because the collection alone is not enough to get the right implementation.

marc.py
```diff
@@ -618,9 +610,10 @@ def records(self, lane, annotator, start_time=None, force_refresh=False,
         # and Representation, but don't actually mirror it.
         if not mirror:
             storage_protocol = self.integration.setting(self.STORAGE_PROTOCOL).value
-            mirror = MirrorUploader.sitewide(self._db)
+            # todo mirror
+            mirror = MirrorUploader.mirror(self._db, storage_name)
```
EdwinGuzman: The only place this function `records` is called is in scripts.py `CacheMARCFiles` but I'm not sure how to get the name of the collection or storage name to pass it to `.records()`.

marc.py
```diff
@@ -521,15 +521,7 @@ class MARCExporter(object):
...
 
-    SETTINGS = [
-        { "key": STORAGE_PROTOCOL,
-          "label": _("Storage service protocol"),
-          "description": _("Storage protocol to use for uploading generated MARC files. The service must already be configured under 'Storage Services'."),
-          "type": "select",
-          "options": [{ "key": ExternalIntegration.S3, "label": ExternalIntegration.S3 }],
-          "default": ExternalIntegration.S3,
-        },
-    ]
```
leonardr: This used to be a drop-down list where you chose which protocol to use. It needs to become a drop-down list where you choose which _integration_ to use. If you have seven different storage integrations, there need to be 7 options in this list.

This is the big challenge and I think this is why we didn't make this change earlier -- we don't have any other situation where you have to connect an ExternalIntegration to _another ExternalIntegration_. The options for this configuration setting need to be gathered at runtime, by searching for all ExternalIntegrations of the STORAGE type.

The value of this configuration setting would presumably be the database ID of the storage integration you chose. If you set this value and then delete the storage integration, I'd act as if the value was not set -- this would probably mean you can't export any MARC files until you go into the admin interface and fix it.

marc.py
leonardr: If you have a configuration setting called STORAGE_INTEGRATION_ID (or whatever) you can load the ExternalIntegration by its database ID and then pass the ExternalIntegration into MirrorUplaoder.implementation().

mirror.py
leonardr: Collection.mirror_integration will get you the ExternalIntegration that should be used to mirror stuff for a given collection. Then you can call MirrorUploader.implementation() on that ExternalIntegration.

marc.py
leonardr: I can't see very far into the future, but I can foresee a scenario where a Collection needs to designate one storage integration (an S3 integration) as the place to store its cover images, and a second storage integration (an LCP integration) as the place to store its books. So at some point we may need to have a separate database table that explains which storage integration is used as the mirror for which Collection/ExternalIntegration for which purpose.

EdwinGuzman: Does the setting creation need to be done here? I did the integration dropdown in `catalog_services` https://github.com/NYPL-Simplified/circulation/pull/1327/files#diff-cafab51e9ce4ad6b89864e1f061e3beeR23 but I did encounter of the issue of not being able to get the external integration from within this MARCExporter class.

PR #1121, opened 2019-09-25
ExternalIntegrationLinks table

EdwinGuzman:
First part to resolve [SIMPLY-2296](https://jira.nypl.org/browse/SIMPLY-2296). Updates are mostly in mirror.py to remove the sitewide configuration and allow for multiple mirrors for different purposes.


s3.py
```diff
@@ -93,9 +92,29 @@ class S3Uploader(MirrorUploader):
...
 
+    PURPOSE = 'Purpose'
+    COVERS = 'Covers'
+    BOOKS = 'Books'
+    MARC = 'MARC'
+
     SETTINGS = [
         { "key": ExternalIntegration.USERNAME, "label": _("Access Key"), "required": True },
         { "key": ExternalIntegration.PASSWORD, "label": _("Secret Key"), "required": True },
+        { "key": PURPOSE, "label": _("Purpose"), "required": True,
```
EdwinGuzman: If I understand this correctly, under "Storage" a purpose for that storage integration is assigned.
Then, in the External Catalogs, you'd only see storages that should be used for MARC files. Similar for collections for covers and books storages.

The externalintegrationlink table would have something where the MARC exporter integration has a storage integration as its linked integration and the purpose is "MARC"?

marc.py
```diff
@@ -521,15 +521,7 @@ class MARCExporter(object):
...
-    SETTINGS = [
-        { "key": STORAGE_PROTOCOL,
-          "label": _("Storage service protocol"),
-          "description": _("Storage protocol to use for uploading generated MARC files. The service must already be configured under 'Storage Services'."),
-          "type": "select",
-          "options": [{ "key": ExternalIntegration.S3, "label": ExternalIntegration.S3 }],
-          "default": ExternalIntegration.S3,
-        },
-    ]
+    SETTING_DESCRIPTION = "Storage protocol to use for uploading generated MARC files. The service must already be configured under 'Storage Services'."
```
EdwinGuzman: Should the list of s3 integrations for MARC files be generated here? I'm currently setting it up but in circulation.

leonardr: You could have a method `ExternalIntegration.storage_for(self, _db, artifact_type)`which found all of the storage integrations that were suitable for storing certain things. "Suitability" would be hard-coded -- we know that an S3 integration can be used for anything so long as you've defined an appropriate bucket for that purpose, and we know that an LCP integration can only be used to store ebooks. I think this code could definitely go into core.

marc.py
```diff
@@ -618,9 +611,9 @@ def records(self, lane, annotator, start_time=None, force_refresh=False,
         # and Representation, but don't actually mirror it.
         if not mirror:
             storage_protocol = self.integration.setting(self.STORAGE_PROTOCOL).value
-            mirror = MirrorUploader.sitewide(self._db)
+            mirror = MirrorUploader.implementation(integration)
```
leonardr: It looks like `integration` is required, and if so, you shouldn't set a default of `None` in the method signature.

model/configuration.py
```diff
@@ -33,6 +33,26 @@
 from sqlalchemy.sql import select
 from sqlalchemy.sql.functions import func
 
+class ExternalIntegrationLink(Base, HasFullTableCache):
+
+    __tablename__ = 'externalintegrationslinks'
+
+    COVERS = "covers"
+    MARC = "MARC"
+    BOOKS = "books"
```
leonardr: These are the purposes, right?

model/configuration.py
```diff
@@ -33,6 +33,26 @@
...
+        Integer, ForeignKey('externalintegrations.id'), index=True
+    )
+    library_id = Column(
+        Integer, ForeignKey('libraries.id'), index=True
+    )
+    other_integration_id = Column(
+        Integer, ForeignKey('externalintegrations.id'), index=True
+    )
+    purpose = Column(Unicode)
+
```
leonardr: I think index `purpose` as well, just to be safe.

migration/20190924-add-externalintegrationslink-table.sql
```diff
@@ -0,0 +1,15 @@
...
+                external_integration_id INTEGER REFERENCES externalintegrations(id),
+                library_id INTEGER REFERENCES libraries(id),
+                other_integration_id INTEGER REFERENCES externalintegrations(id),
+                purpose VARCHAR
+            );
+        EXCEPTION
+            WHEN duplicate_table THEN RAISE NOTICE 'Warning: externalintegrationslinks already exists.';
+        END;
+    END;
+$$;
```
leonardr: I tried this out on my laptop and it looks like this doesn't create the indexes. However, when you're creating a brand new database table you don't need a migration script at all. The database table will be created automatically the first time SQLAlchemy is initialized.

opds_import.py
```diff
@@ -460,7 +461,9 @@ def __init__(self, _db, collection, data_source_name=None,
             # If this Collection is configured to mirror the assets it
             # discovers, this will create a MirrorUploader for that
             # Collection. Otherwise, this will return None.
-            mirror = MirrorUploader.for_collection(collection)
+            mirror = MirrorUploader.for_collection(
+                collection, ExternalIntegrationLink.COVERS
+            )
```
leonardr: This is a great payoff -- it's totally clear what's happening here.

However, it's also clear that `mirror` now is just for the book covers. We're going to need a separate `MirrorUploader` to store any ebooks we find. In this code, instead of self.mirror we should be using self.cover_mirror or self.book_mirror.

scripts.py
```diff
@@ -1890,7 +1891,9 @@ def collections_with_uploader(self, collections):
             to use for that Collection.
         """
         for collection in collections:
-            uploader = MirrorUploader.for_collection(collection)
+            uploader = MirrorUploader.for_collection(
+                collection, ExternalIntegrationLink.COVERS
+            )
```
leonardr: Again, we now have different MirrorUploaders for different types of items. So  `ReplacementPolicy` needs to take two mirrors -- one for every type of thing that might be mirrored. It might be better to make `ReplacementPolicy` take a dictionary that maps "purpose" to an appropriate MirrorUploader.

model/configuration.py
EdwinGuzman: Yes.

migration/20190924-add-externalintegrationslink-table.sql
EdwinGuzman: I see. So then this isn't really needed at all?

leonardr: Yeah, you can lose it.

opds_import.py
leonardr: Or we should have a dictionary that maps a purpose to a MirrorUploader designed for that purpose.

marc.py
EdwinGuzman: That makes sense but I'm giving the type of bucket each integration is used for when it is added to a collection or catalog service (it's "purpose"). So here, it wouldn't be able to know that some buckets are specifically for, say, MARC files. In circulation, I was listing all the available buckets regardless of how they could be used.

leonardr: That's fine -- I'm only suggesting you be able to grab the buckets that _could_ be used for MARC files because it's technically possible.

scripts.py
```diff
@@ -1891,11 +1891,15 @@ def collections_with_uploader(self, collections):
...
+            covers = MirrorUploader.for_collection(
                 collection, ExternalIntegrationLink.COVERS
             )
-            if uploader:
-                policy = self.replacement_policy(uploader)
+            books = MirrorUploader.for_collection(
+                collection, ExternalIntegrationLink.BOOKS
+            )
+            mirror = dict(covers=covers, books=books)
+            if mirror["covers"] or mirror["books"]:
```
leonardr: I'm pretty sure this is stylistic, but i would write `if covers or books:` and build the dictionary inside -- otherwise you're making a dictionary you're not using.

marc.py
```diff
@@ -521,15 +521,19 @@ class MARCExporter(object):
...
+        "label": _("MARC Mirror"),
+        "description": _("Storage protocol to use for uploading generated MARC files. The service must already be configured under 'Storage Services'."),
+        "type": "select",
+        "options" : [
+            dict(
+                key=NO_MIRROR_INTEGRATION,
+                label=_("None - Do not mirror MARC files")
+            )
+        ]
+    }
```
leonardr: I'm at the very beginning of this diff so maybe this will make sense later, but how do the mirror integrations get added to `options`?

I would expect the config here to look more like:

```
"type": "storage_integration",
"storage_purpose": "MARC",
```

marc.py
```diff
@@ -548,6 +552,19 @@ def __init__(self, _db, library, integration):
...
+        integrations = ExternalIntegration.get_integrations_of_goal(
+            _db, ExternalIntegration.STORAGE_GOAL
+        )
+        for integration in integrations:
+            cls.SETTING['options'].append(
+                dict(key=str(integration.id), label=integration.name)
+            )
+        
+        return cls.SETTING
+
```
leonardr: I see, this method gets called at some point and it adds all the existing storage integrations to the list.

This is fine but--again, just thinking out loud as I go through the diff--it seems like you're going to need to have similar code for  setting up the "cover" and "book" ExternalIntegrationLinks for a collection.

This can wait, but the other issue is that not all ExternalIntegrations with STORAGE_GOAL can actually store MARC records. An S3 integration needs to have a MARC bucket defined, and an LCP integration won't be able to store MARC records at all.

marc.py
leonardr: This is looking good, but `integration` is too vague a term. How about `mirror_integration`?

metadata_layer.py
```diff
@@ -721,7 +721,7 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
             )
             return
 
-        mirror = policy.mirror
+        mirror = policy.mirror[mirror_type]
```
leonardr: You need to handle the case where `mirror_type` doesn't show up in the dictionary.

metadata_layer.py
```diff
@@ -1176,9 +1176,11 @@ def apply(self, _db, collection, replace=None):
...
                 link_obj = link_objects[link]
                 if replace.mirror:
-                    # We need to mirror this resource. If it's an image, a
-                    # thumbnail may be provided as a side effect.
-                    self.mirror_link(pool, data_source, link, link_obj, replace)
+                    for mirror_type in replace.mirror.keys():
+                        if replace.mirror[mirror_type]:
+                            # We need to mirror this resource. If it's an image, a
+                            # thumbnail may be provided as a side effect.
+                            self.mirror_link(pool, data_source, link, link_obj, replace, mirror_type)
```
leonardr: I don't understand this piece. I feel like this is the point where you need to decide which of the mirrors in replace.mirror to use for this particular link -- so you do need new code here and looking at replace.mirror is appropriate. But I would expect it to have something to do with `link.rel` -- images go to the image mirror, books go to the book mirror. Maybe there needs to be a `ReplacementPolicy.mirror_for(LinkData)` method to sort this out.

I think the code you have is going to tell _every_ mirror to mirror every link.

metadata_layer.py
```diff
@@ -1909,9 +1911,11 @@ def _key(classification):
...
 
             if replace.mirror:
-                # We need to mirror this resource. If it's an image, a
-                # thumbnail may be provided as a side effect.
-                self.mirror_link(edition, data_source, link, link_obj, replace)
+                for mirror_type in replace.mirror.keys():
+                    if replace.mirror[mirror_type]:
+                        # We need to mirror this resource. If it's an image, a
+                        # thumbnail may be provided as a side effect.
+                        self.mirror_link(edition, data_source, link, link_obj, replace, mirror_type)
```
leonardr: Again, I think every mirror is going to mirror every link. And this looks like duplicate code (though it was duplicate code before you changed it). Maybe the right thing to do is to put all the code in `mirror_link`, and make `mirror_link` figure out whether the link should actually be mirrored, and if so, which is the right mirror to use.

mirror.py
```diff
@@ -18,72 +17,64 @@ class MirrorUploader(object):
     IMPLEMENTATION_REGISTRY = {}
 
     @classmethod
-    def sitewide(cls, _db):
-        """Create a MirrorUploader from a sitewide configuration.
+    def mirror(cls, _db, storage_name=None, integration=None):
+        """Create a MirrorUploader from an integration or storage name.
+
+        :param storage_name: The name of the S3 storage integration.
+        :param integration: The S3 external integration.
```
leonardr: It's good to have a docstring here but we're doing this to eventually support MirrorUploaders other than S3, so let's not start with the assumption that this is an S3 integration specifically.

model/configuration.py
```diff
@@ -236,6 +257,45 @@ def cache_key(self):
         # bit and by_cache_key() not as much.
         return self.id
 
+    @classmethod
+    def get_integrations_of_goal(cls, _db, goal):
```
leonardr: For consistency with other methods this should be called `get_integrations_for_goal`.

model/configuration.py
```diff
@@ -236,6 +257,45 @@ def cache_key(self):
...
+        """Return all external integrations by goal type.
+        """
+        integrations = _db.query(cls).filter(
+            cls.goal==goal
+        ).order_by(
+            cls.name
+        ).all()
```
leonardr: You don't need to call `all()` here -- the query wil lbe run when the results are needed. This will allow someone to call `ExternalIntegration.for_goal` and then tack on any extra query conditions they need.

model/configuration.py
```diff
@@ -236,6 +257,45 @@ def cache_key(self):
...
+            ExternalIntegrationLink,
+            ExternalIntegrationLink.other_integration_id==cls.id
+        ).filter(
+            ExternalIntegrationLink.external_integration_id==collection.external_integration_id,
+            ExternalIntegrationLink.purpose==purpose
+        )
+        integrations = qu.all()
+        if not integrations:
+            raise CannotLoadConfiguration(
+                "No storage integration for purpose %s is configured." % purpose
```
leonardr: This error message should also mention the collection name.

opds_import.py
```diff
@@ -402,9 +403,10 @@ class OPDSImporter(object):
     SUCCESS_STATUS_CODES = None
 
     def __init__(self, _db, collection, data_source_name=None,
-                 identifier_mapping=None, mirror=None, http_get=None,
+                 identifier_mapping=None, http_get=None,
                  metadata_client=None, content_modifier=None,
-                 map_from_collection=None,
+                 map_from_collection=None, covers_mirror=None,
+                 books_mirror=None,
```
leonardr: In ReplacementPolicy you have a dictionary of mirrors keyed by purpose, and I think it's best to do the same here. Not only will this make everything consistent, this will let us add other kinds of mirrors (such as for book previews) without changing the method signature.

opds_import.py
```diff
@@ -456,12 +461,20 @@ def __init__(self, _db, collection, data_source_name=None,
...
+                # Collection for its purpose. Otherwise, this will return None.
+                covers_mirror = MirrorUploader.for_collection(
+                    collection, ExternalIntegrationLink.COVERS
+                )
+            if not books_mirror:
+                books_mirror = MirrorUploader.for_collection(
+                    collection, ExternalIntegrationLink.BOOKS
+                )
+
+        self.mirror = dict(covers=covers_mirror, books=books_mirror)
```
leonardr: Yeah, you end up with a dictionary, so take a dictionary as input and make your own if the input is `None`.

s3.py
```diff
@@ -79,6 +78,7 @@ class S3Uploader(MirrorUploader):
     S3_BASE = "https://%s/" % S3_HOSTNAME
 
     BOOK_COVERS_BUCKET_KEY = u'book_covers_bucket'
+    BOOK_BUCKET_KEY = u'book__bucket'
```
leonardr: I've been saying "book bucket" when I actually need the "open_access_content_bucket". We shouldn't need a separate "book bucket" because we'll probably never store non-open-access content on S3.

We do need a 'books' _purpose_ (which indicates that an S3 integration can be used to store open-access books or an LCP integration can be used to store any kind of books), but that's different and you already have that.

marc.py
```diff
@@ -555,10 +555,10 @@ def __init__(self, _db, library, integration):
         
     @classmethod
     def get_storage_settings(cls, _db):
-        integrations = ExternalIntegration.get_integrations_of_goal(
+        integrations = ExternalIntegration.for_goal(
             _db, ExternalIntegration.STORAGE_GOAL
         )
-        for integration in integrations:
+        for integration in integrations.all():
```
leonardr: This works but you don't need to call `all`. Iterating over the query object will run the query.

metadata_layer.py
EdwinGuzman: So would it make sense to assume that the link should be mirrored for cover images if the `rel` is in `MIRRORED = [OPEN_ACCESS_DOWNLOAD, IMAGE, THUMBNAIL_IMAGE]`? Otherwise, it is a book mirror. In this specific case, there is a check to see if `rel` is in `CIRCULATION_ALLOWED` so that limits what type it is and I guess in this case, it is a book mirror?

leonardr: Yes, it sounds like in this case it's a book mirror. For `[IMAGE, THUMBNAIL_IMAGE]` it would be a cover mirror.

metadata_layer.py
```diff
@@ -148,6 +149,15 @@ def append_only(self, **args):
...
 
+    def mirror_type_for(self, link):
+        """Returns the type of mirror that should be used for the link.
+        """
+        if link.rel in [Hyperlink.IMAGE, Hyperlink.THUMBNAIL_IMAGE]:
+            return ExternalIntegrationLink.COVERS
+        elif link.rel in Hyperlink.CIRCULATION_ALLOWED:
+            return ExternalIntegrationLink.BOOKS
+        return None
+
```
leonardr: This doesn't use `self` at all. As is, this would be a better fit in `LinkData`. It would make sense as part of `ReplacementPolicy` if it looked in `self.mirrors` and returned the actual mirror object to use.

Normally I don't think this is a big deal but `metadata_layer.py`, like the model, is a semi-public API that's used by a lot of different code so I'm more concerned with the API design.

metadata_layer.py
```diff
@@ -1183,11 +1193,9 @@ def apply(self, _db, collection, replace=None):
...
                 link_obj = link_objects[link]
                 if replace.mirror:
-                    for mirror_type in replace.mirror.keys():
-                        if replace.mirror[mirror_type]:
-                            # We need to mirror this resource. If it's an image, a
-                            # thumbnail may be provided as a side effect.
-                            self.mirror_link(pool, data_source, link, link_obj, replace, mirror_type)
+                    mirror_type = replace.mirror_type_for(link)
+                    # We need to mirror this resource.
+                    self.mirror_link(pool, data_source, link, link_obj, replace, mirror_type)
```
leonardr: However you resolve the `mirror_type_for` issue, it's now possible to put _all_ of this logic inside `mirror_link`. You're already passing the LinkData and the ReplacementPolicy into that method, and that gives it all the information necessary to figure out whether, how and where to do the mirroring.

leonardr (commented):
This is looking good. I think you can make the metadata-layer "mirror this link" code easier to use and then we'll be ready.

leonardr (approved):
:sunglasses:

model/configuration.py
```diff
@@ -54,6 +56,30 @@ class ExternalIntegrationLink(Base, HasFullTableCache):
     )
     purpose = Column(Unicode, index=True)
 
+    @classmethod
+    def _get_settings(cls):
```
EdwinGuzman: Is this what you meant for https://github.com/NYPL-Simplified/circulation/pull/1333/files#r333120505 ?

leonardr: Yes, but `_get_settings` is a generic name and this doesn't need to be a function. You can make a list of COLLECTION_MIRROR_SETTINGS and then use it from controller code.

leonardr:
This is looking good overall. We're missing two things:

1. Migration script that drops collection.mirror_integration_id and (if it's set) replaces it with two ExternalIntegrationLinks.
2. Make it possible to set the data source for a "Manual import" integration from the admin interface.

#2 can wait for later but I would like to see #1 in the core branch before merging.

model/configuration.py
```diff
@@ -259,11 +259,11 @@ class ExternalIntegration(Base, HasFullTableCache):
...
-    # An ExternalIntegration may be used by many Collections
-    # to mirror book covers or other files.
-    mirror_for = relationship(
-        "Collection", backref="mirror_integration",
-        foreign_keys='Collection.mirror_integration_id',
+    links = relationship(
+        "ExternalIntegrationLink",
+        backref="other_integration",
+        foreign_keys="ExternalIntegrationLink.other_integration_id",
+        cascade="all, delete-orphan"
```
EdwinGuzman: @leonardr this worked great, thanks!

leonardr: Great. Hold off merging this until the circulation branch is ready to go, but I think this is done once you add one more test.

tests/models/test_configuration.py
```diff
@@ -300,6 +300,53 @@ def test_collection_mirror_settings(self):
...
+            integration=collection.external_integration,
+            other_integration=storage2, purpose="books"
+        )
+
+        qu = self._db.query(ExternalIntegrationLink)
+        external_integration_links = qu.all()
+
+        eq_(len(external_integration_links), 2)
+        eq_(external_integration_links[0].other_integration_id, storage1.id)
+        eq_(external_integration_links[1].other_integration_id, storage2.id)
```
leonardr: In general you can't rely on objects coming out of a database query in a specific order. The simplest way to enforce this is to add `.order_by(ExternalIntegrationLink.other_integration_id)` to the `qu`. You can also do this by comparing two `set`s.

leonardr:
I'd like @aslagle to double-check the migration script (https://github.com/NYPL-Simplified/server_core/pull/1121/files#diff-d9b951efad2ec7b38b0ad144e4df994e), both for its functionality and for a problem I think I see in it:

This migration script uses a table that is only created when the SQLAlchemy model is loaded. DatabaseMigrationScript tries really hard not to load the SQLAlchemy model. I know I said earlier that you didn't need to create the table inside the migration script, and that's generally true, because generally the new table starts out empty.

This situation is unusual because we need the table to start out with a couple of rows. The only time we can add these rows is during a migration script, and (by design) the table won't get added until after the migration script runs. So I think you need to bring back the SQL you had earlier that creates the database table with all of its indexes and foreign keys. Running `pg_dump --schema-only --schema=externalintegrationlinks` on a working database should give you the SQL to use.

leonardr (changes requested):
Un-approving because of concerns about the existing migration script.

aslagle:
Yeah, I agree. The table needs to be created as part of the migration in this case. Also, seems like the purpose should be more specific, like "mirror_books" and "mirror_covers".

leonardr:
I'm okay with adding "mirror_" or "storage_" to the purpose of links used for storage.

migration/20191010-update-external-integration-mirrors.sql
```diff
@@ -11,26 +11,26 @@ DO $$
         EXCEPTION
             WHEN duplicate_table THEN RAISE NOTICE 'Warning: externalintegrationslinks already exists.';
         END;
-    END;
-$$; 
-
-ALTER TABLE ONLY externalintegrationslinks ADD CONSTRAINT externalintegrationslinks_pkey PRIMARY KEY (id);
+    
+        ALTER TABLE ONLY externalintegrationslinks ADD CONSTRAINT externalintegrationslinks_pkey PRIMARY KEY (id);
```
leonardr: Still not quite right -- this is inside the first BEGIN statement. I ran it on my dev database and got the warning as expected, but then an error, "multiple primary keys for table "externalintegrationslinks" are not allowed", which would have caused the script to fail.

leonardr (approved):
Looks good now!

PR #1122, opened 2019-09-30
Fix classification weights

leonardr:
This branch fixes the part of https://jira.nypl.org/browse/SIMPLY-2320 that can be fixed in core. The circulation PR is https://github.com/NYPL-Simplified/circulation/pull/1332.

I decided to stop deriving implicit "Adult" audience information from LCC and Dewey classifications, even though I don't know how frequently a children's book is classified under a non-juvenile LCC or Dewey classification. It has definitely happened and we don't need to scrape for hints that a book's classification is "Adult". That's the default and that's where we'll go in the absence of information.

I'm pretty confident that the migration script works but as I say in the script, it's  simpler and faster to delete _all_ of the "classify" WorkCoverageRecords. Almost every work on a normal site will need to be reclassified.

Since we're reclassifying everything, I'm very interested in hearing about other problems of this sort, so that we don't have to do this again for a while.

### DIFF
--- a/migration/20190926-reweight-classifications.sql
+++ b/migration/20190926-reweight-classifications.sql
@@ -6,8 +6,11 @@
 -- we see from other commercial distributors.
 update classifications set weight=100 where id in (select c.id from classifications c join datasources ds on ds.id=c.data_source_id where ds.name='Axis 360' and weight=1);
 
--- Weight any Bibliotheca BISAC classifiers similarly.
-update classifications set weight=100 where s.id in (select c.id from classifications c join subjects s on s.id=c.subject_id join datasources ds on ds.id=c.data_source_id where ds.name='Bibliotheca' and s.type='BISAC' and c.weight=15);
+-- Weight any Bibliotheca BISAC classifications similarly.
+update classifications set weight=100 where id in (select c.id from classifications c join subjects s on s.id=c.subject_id join datasources ds on ds.id=c.data_source_id where ds.name='Bibliotheca' and s.type='BISAC' and c.weight=15);
+
+-- Same for ENKI tags.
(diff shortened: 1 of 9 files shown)


leonardr:
@aslagle what's your take on deleting all of the "classify" WorkCoverageRecords?

aslagle:
I think it's probably fine - the extra work gets done by a script so the only problem I can think of is that it might delay fixing the classifications that are actually wrong if it's reprocessing a lot of others first.

migration/20190926-reweight-classifications.sql
```diff
@@ -0,0 +1,71 @@
...
+   -- violated.
+   OR (s.type in ('DDC', 'LCC') and s.audience='Adult')
+  )
+ )
+
+);
+
+-- Now that we've deleted the appropriate coverage records (which
+-- required the old system to be in place), change the weights of
+-- certain classifications to reflect the new system.
```
EdwinGuzman: So, updating the weights here means that we believe that these values are more correct now than they were before?

leonardr: Yes, I think the new values will give better classification results than the old values, mainly because we're trusting distributors more.

leonardr:
I'd like another quick look at this branch and https://github.com/NYPL-Simplified/circulation/pull/1332. I normalized the value for Enki tags and all RBdigital classifications to the TRUSTED_DISTRIBUTOR_WEIGHT. Enki tags were weighted at 1 (basically ignored) and RBdigital classifications went from 100 to 500 (!), which was probably find but I think it's better to normalize the numbers.

There are unanswered questions around Odilo, NoveList, and ONIX imports, but we can address those later if necessary.

PR #1123, opened 2019-10-01
Implement all searches using the multi-search functionality

leonardr:
This is a precursor branch for https://jira.nypl.org/browse/SIMPLY-2015. It exposes the "multi-search" feature of Elasticsearch, which makes it possible to run several queries simultaneously and get back a list of lists of results.

In this branch, _all_ searches use the multi-search functionality, even when there's only one underlying query. This makes sure all searches go through the same code path, and so all our expensive tests where we search for "moby dick" or whatever become a test of the multi-search functionality. Of course I also added tests of multiple searches at once. I didn't see any performance difference between running a single query and running a multi-query that contained only one query.

The multi-search functionality is not yet exploited outside the unit tests -- I'll get to that in the next branch.

I had to change the behavior of `Filter.match_nothing` so that we could actually generate an Elasticsearch filter that matches nothing. `query_works()` implements an optimization where if `Filter.match_nothing` is True it doesn't even run the query. But `query_works_multi` needs to handle the case where it was given one query with `Filter.match_nothing=True` and one query that will match something. Since we're already sending an HTTP request out to Elasticsearch, the simplest thing to do is to build a real query that does nothing. (I don't think this will happen in real life.)

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -466,9 +466,6 @@ class ExternalSearchIndex(HasSelfTests):
             script fields calculated by ElasticSearch during the
             search process.
         """
-        if not self.works_alias:
-            return []
-
         if isinstance(filter, Filter) and filter.match_nothing is True:
             # We already know this search should match nothing.  We
             # don't even need to perform the search.
@@ -476,9 +473,7 @@ class ExternalSearchIndex(HasSelfTests):
 
         pagination = pagination or Pagination.default()
         query_data = (query_string, filter, pagination)
-        [result] = [
-            x for x in self.query_works_multi([query_data], debug)
-        ]
+        [result] = list(self.query_works_multi([query_data], debug))
         return result
 
     def query_works_multi(self, queries, debug=False):
@@ -487,10 +482,19 @@ class ExternalSearchIndex(HasSelfTests):
 
         :param queries: A list of (query string, Filter, Pagination) 3-tuples.
 
-        :yield: A sequence of lists, one per item in `filters`,
+        :yield: A sequence of lists, one per item in `queries`,
             each containing the search results from that
             (query string, Filter, Pagination) 3-tuple.
         """
+        # If the works alias is not set, all queries return empty.
+        #
+        # TODO: Maybe an unset works_alias should raise
+        # CannotLoadConfiguration in the constructor. Then we wouldn't
+        # have to worry about this.
+        if not self.works_alias:
+            for q in queries:
+                yield []
+
         multi = MultiSearch(using=self.__client)
         for (query_string, filter, pagination) in queries:
             search = self.create_search_doc(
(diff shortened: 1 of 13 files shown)


external_search.py
```diff
@@ -458,58 +462,93 @@ def query_works(self, query_string, filter=None, pagination=None,
...
         #
+        # TODO: Maybe an unset works_alias should raise
+        # CannotLoadConfiguration in the constructor. Then we wouldn't
+        # have to worry about this.
+        if not self.works_alias:
+            for q in queries:
+                yield []
+
+        # Create a MultiSearch. 
+        multi = MultiSearch(using=self.__client)
```
EdwinGuzman: `__client` is the ElasticSearch client, right?

external_search.py
```diff
@@ -458,58 +462,93 @@ def query_works(self, query_string, filter=None, pagination=None,
...
+            if function_scores:
+                function_score = FunctionScore(
+                    query=dict(match_all=dict()),
+                    functions=function_scores,
+                    score_mode="sum"
+                )
+                search = search.query(function_score)
+            multi = multi.add(search)
+
+        a = time.time()
```
EdwinGuzman: Are `a` and `b` used to compare time?

external_search.py
```diff
@@ -458,58 +462,93 @@ def query_works(self, query_string, filter=None, pagination=None,
...
-        # it set up to generate a link to the next page.
-        pagination.page_loaded(results)
-
-        return results
+        # Process each item in the resultset according to the
+        # Pagination object that was used to create it.
+        for i, results in enumerate(resultset):
+            # Use the Pagination object to slice up the results if
+            # necessary.
+            results = resultset[i]
```
EdwinGuzman: Why use `resultset[i]` when the loop has `i, results` above?

external_search.py
leonardr: Yes, it's the Elasticsearch object from elasticsearch-dsl.

external_search.py
leonardr: Yes.

external_search.py
leonardr: Good catch.

external_search.py
```diff
@@ -458,58 +462,92 @@ def query_works(self, query_string, filter=None, pagination=None,
...
+        if isinstance(filter, Filter) and filter.match_nothing is True:
+            # We already know this search should match nothing.  We
+            # don't even need to perform the search.
             return []
 
-        if not pagination:
-            pagination = Pagination.default()
+        pagination = pagination or Pagination.default()
+        query_data = (query_string, filter, pagination)
+        [result] = list(self.query_works_multi([query_data], debug))
```
EdwinGuzman: Does calling `list` on the method iterate through the generator to get all the values?

leonardr: Yeah, without the list()... we'd probably be okay here because `[result]` implicitly iterates over the generator.

PR #1124, opened 2019-10-03
Generate grouped feeds with a single multisearch query

leonardr:
This branch completes https://jira.nypl.org/browse/SIMPLY-2015. It changes the query used to generated grouped OPDS feeds so that it uses a single Elasticsearch query and a single database query to get all the books that will be used in the feed, in the proper order.

As I mentioned in the ticket, the speedup is 20-30% -- nothing to sneeze at, but less than I was hoping for.

Rather than calling `lane.works()` on every `Lane` it's given, `WorkList._featured_works_for_lanes()` now builds a query for each `Lane` and runs all the queries using a single `ExternalSearchIndex.query_works_multi()` call.

The result is a list of lists of Hit objects. These are transformed into Work objects using a single call to `WorkList.works_for_resultsets`. This new method contains code from the old `works_for_hits` method (which turned a single list of Hits into a list of WorkList). It now goes through a list of lists, building a list of work IDs, makes a single database query to fetch all the corresponding Works, and then builds a list of list of Works corresponding to the list of list of Hits.

The old `works_for_hits` method now wraps its argument in an extra list, calls `works_for_resultsets`, and unwraps the result. This is analogous to how my previous branch changed `query_works` to wrap its single query in a list, call `query_works_multi`, and unwrap the results.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1484,7 +1484,7 @@ class WorkList(object):
             script fields), WorkSearchResult objects.
         """
 
-        [results] = self.works_for_resultsets(self, _db, [hits])
+        [results] = self.works_for_resultsets(_db, [hits])
         return results
 
     def works_for_resultsets(self, _db, resultsets):
@@ -1516,7 +1516,8 @@ class WorkList(object):
         # DatabaseBackedWorkList that fetches those specific Works
         # while applying the general availability filters.
         #
-        # TODO: There's a lot of room for improvement here.
+        # TODO: There's a lot of room for improvement here, but
+        # performance isn't a big concern -- it's just ugly.
         wl = SpecificWorkList(work_ids)
         wl.initialize(self.get_library(_db))
         qu = wl.works_from_database(_db)
@@ -1732,16 +1733,6 @@ class WorkList(object):
         # need to pick out unusable WorkLists, make a recursive call
         # to WorkList.works() for each one, and fold them into the
         # results we got from the big multi-query.
-
-        searchable = []
-        not_searchable = []
-        for lane in lanes:
-            if isinstance(lane, Lane):
-                l = searchable
-            else:
-                l = not_searchable
-            l.append(lane)
-
         queries = []
         for lane in lanes:
             overview_facets = lane.overview_facets(_db, facets)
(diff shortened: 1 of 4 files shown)


lane.py
```diff
@@ -1475,61 +1475,87 @@ def modify_search_filter_hook(self, filter):
...
+                work_ids.add(result.work_id)
+                if has_script_fields is None:
+                    # We don't know whether any script fields were
+                    # included, and now we're in a position to find
+                    # out.
+                    has_script_fields = (
+                        any(
+                            x in result for x in Filter.KNOWN_SCRIPT_FIELDS
+                        )
+                    )
```
leonardr: This is instead of looking at the first result (if there is a first result) after iterating over the list. There's no reason we can't do it _while_ we're iterating over the list.

EdwinGuzman: What kind of data is in the script fields? It makes sense to do this here but why only check the first result?

leonardr: The only script field we have right now is "simplified.work_last_update", which calculates the last-update time to use when including a book in an OPDS feed.

We don't actually use the script data here, we just need to check whether the script field is present so we can know to return `WorkSearchResult`s instead of `Work`s. (If we just return `Work`s then the script data is lost.) 

The script data is either present on all the results, or not present at all, so checking the first result is enough to answer that question.

lane.py
```diff
@@ -1707,12 +1730,28 @@ def _featured_works_with_lanes(
...
+            overview_facets = lane.overview_facets(_db, facets)
+            from external_search import Filter
+            filter = Filter.from_worklist(_db, lane, overview_facets)
+            queries.append((None, filter, pagination))
+        resultsets = list(search_engine.query_works_multi(queries))
+        works = self.works_for_resultsets(_db, resultsets)
+
+        for i, lane in enumerate(lanes):
+            results = works[i]
+            for work in results:
```
EdwinGuzman: So this will return a work and its lane for every work that was found?

leonardr: Yes, that's right. I don't think we need the 2-tuples anymore -- we could just yield `(lane, results)` -- but I'd rather make that change in a separate branch.

PR #1125, opened 2019-10-07
Optimize feed generation

leonardr:
This branch makes a couple small changes to stop certain bits of expensive code from running while we're generating OPDS feeds. I've got a circulation branch that adopts this code and adds one more optimization.

From most to least effective, the changes I made are:

* Instantiating a single `SpellChecker` rather than making a new one every time a `Query` is initialized.
* Stop instantiating a whole other `Query` object on every search that never gets used.
* Preload work.license_pools.presentation_edition.primary_identifier to avoid an extra database request to find the primary identifier of the active license pool for each work. I'm not 100% sure about this one because it makes the database query heavier, but I'm pretty sure we do need it every time we make an OPDS feed.
* Generating and running the fuzzy hypotheses for a query that has no query string and so can't have any misspellings. This one isn't very bad at all -- presumably Elasticsearch ignores those hypotheses -- but it makes the code more readable.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -419,7 +419,6 @@ class ExternalSearchIndex(HasSelfTests):
                           debug):
 
         query = Query(query_string, filter)
-        query_without_filter = Query(query_string)
         search = query.build(self.search, pagination)
         if debug:
             search = search.extra(explain=True)
--- a/external_search.py
+++ b/external_search.py
@@ -1468,8 +1468,8 @@ class Query(SearchBase):
                 self.fuzzy_coefficient = 0.5
         else:
             # Since this query does not contain any words, there is no
-            # risk that a word might misspell. Do not create or run
-            # the 'fuzzy' hypotheses at all.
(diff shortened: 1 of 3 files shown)


lane.py
```diff
@@ -1869,6 +1869,8 @@ def _modify_loading(cls, qu):
             joinedload(license_pool_name, "delivery_mechanisms"),
             joinedload(license_pool_name, "delivery_mechanisms", "delivery_mechanism"),
 
+            joinedload(license_pool_name, "identifier"),
```
leonardr: This lets us access active_license_pool.identifier from OPDS generation code without making a separate database query -- the data is preloaded as part of the original lookup.

external_search.py
```diff
@@ -419,7 +419,6 @@ def create_search_doc(self, query_string, filter, pagination,
                           debug):
 
         query = Query(query_string, filter)
-        query_without_filter = Query(query_string)
```
leonardr: This line didn't need to be there at all.

PR #1126, opened 2019-10-09
Add a reaper for works with no license pools

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2061.

Amy, I'm asking you to do this review as a way of giving one more piece of thought to whether there might be bad side effects to removing Works with no LicensePools. I'm almost positive it'll be fine, both from thinking about it and because a running circ manager recently had to manually delete a ton of Works with no LicensePools.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -830,6 +830,20 @@ class PatronRecordReaper(ReaperMonitor):
 ReaperMonitor.REGISTRY.append(PatronRecordReaper)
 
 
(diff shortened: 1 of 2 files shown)


aslagle:
The only problem I can think of is if a work ended up with an orphaned search index document. Deleting a collection handles that, but if the license pools were deleted manually or some other way it might be possible.

PR #1129, opened 2019-10-22
Add uniqueness constraints on Credentials

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2353 by implementing uniqueness constraints on the `credentials` table. Requests that previously would have triggered a race condition will now fail _without_ screwing up the database.

I've tested this with a fresh database and with an old database migrated using the migration script.

### DIFF
--- a/model/credential.py
+++ b/model/credential.py
@@ -42,9 +42,43 @@ class Credential(Base):
     )
 
     __table_args__ = (
-        UniqueConstraint('data_source_id', 'patron_id', 'collection_id', 'type'),
-    )
 
+        # Unique indexes to prevent the creation of redundant credentials.
+
+        # If both patron_id and collection_id are null, then (data_source_id,
+        # type, credential) must be unique.
+        Index(
+            "ix_credentials_data_source_id_type_token",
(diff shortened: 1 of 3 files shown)


leonardr:
I do not get why these tests are failing; they work locally and on Travis they seem to fail differently every time.

leonardr:
OK, I figured it out. I hadn't actually tested on a fresh database; I was just testing on my migrated database so bugs in my "initialize the schema from scratch code" weren't being hit.

model/credential.py
```diff
@@ -42,9 +43,43 @@ class Credential(Base):
...
+            data_source_id, type, patron_id,
+            unique=True, postgresql_where=(collection_id==None)
+        ),
+
+        # If neither collection_id nor patron_id is null, then
+        # (data_source, type, collection_id, patron_id)
+        # must be unique.
+        Index(
+            "ix_credentials_data_source_id_type_collection_id_patron_id",
+            data_source_id, type, collection_id, patron_id,
```
aslagle: This has patron id and collection id in the opposite order from the migration.

PR #1130, opened 2019-10-22
Create a default Local Analytics service if one already does not exist.

EdwinGuzman:
Resolves [SIMPLY-2346](https://jira.nypl.org/browse/SIMPLY-2346). Adding a function to LocalAnalyticsProvider to get or create a local analytics integration. It'll be used when the app starts so there will be a default service if none exists.

### DIFF
--- a/tests/test_analytics.py
+++ b/tests/test_analytics.py
@@ -188,8 +188,6 @@ class TestAnalytics(DatabaseTest):
         assert isinstance(local_analytics, ExternalIntegration)
         eq_(local_analytics.name, LocalAnalyticsProvider.NAME)
 
-        self._db.delete(local_analytics)
-
         # When an analytics provider is initialized, retrieving a
         # local analytics service should return the same one.
         local_analytics = LocalAnalyticsProvider.initialize(self._db)
--- a/tests/test_analytics.py
+++ b/tests/test_analytics.py
@@ -15,6 +15,7 @@ from ..model import (
     ExternalIntegration,
     Library,
     create,
+    get_one
 )
 import json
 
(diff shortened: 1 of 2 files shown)


leonardr (commented):
This makes sense but I'd like @aslagle to weigh in it, because once this branch lands there is no way to _do without_ a local analytics service. If you delete it, it'll just be recreated. We can't distinguish between "no analytics service because it was deleted" and "no analytics service because the default code hasn't run yet."

I believe Amy thinks that there are cases where you just don't want to do local analytics. So it might be best to add a configuration option to the local analytics provider that lets you _disable_ gathering of analytics, rather than deleting the "integration" itself.

leonardr:
Amy, to summarize SIMPLY-2346, it looks like people who run circulation managers are, in general, expecting the "Export CSV" feature to be enabled by default, but that feature requires a local analytics integration.

aslagle:
This sounds like more of a problem with training and marketing, but I don't know of any specific cases where someone wants to disable it. I could imagine wanting to save the disk space if you don't need the data, or wanting to not save it for even better patron privacy. I'm more concerned that we are promoting this feature to libraries when it's not actually that useful yet.

leonardr:
I think 3.0.2 makes the feature pretty useful by associating events with libraries and optionally neighborhoods.

We have a clear path forward if someone does want to disable this -- add a flag that lets you turn off gathering this info without deleting the integration, so I'm okay with merging this.

tests/test_analytics.py
```diff
@@ -168,3 +169,36 @@ def test_collect_event(self):
...
+        eq_(None, local_analytics)
+
+        # So when the Local Analytics provider is initialized, it will
+        # create one with the default name of "Local Analytics".
+        local_analytics = LocalAnalyticsProvider.initialize(self._db)
+
+        assert isinstance(local_analytics, ExternalIntegration)
+        eq_(local_analytics.name, LocalAnalyticsProvider.NAME)
+
+        self._db.delete(local_analytics)
```
leonardr: I don't think this line does anything (possibly because the delete isn't flushed to the database), because if local_analytics was really deleted, a new one would be created and `local_analytics_2` would have a different ID. I'd just take this line out.

leonardr (approved):

PR #1134, opened 2019-11-06
Update the list of media types fulfillable by the default client

leonardr:
This branch is one of a couple I'm writing to address https://jira.nypl.org/browse/SIMPLY-2370.

The circulation manager keeps a list of media type/DRM scheme combos that SimplyE can render. Since SimplyE gains capabilities over time, it's possible for a circulation manager to have an old copy of this list, and for this reason we haven't used it for very much and haven't put much work into keeping it up to date.

However, SIMPLY-2370 presents a good argument for keeping it up to date, so this branch does the work of bringing it up to date.

### DIFF
--- a/model/licensing.py
+++ b/model/licensing.py
@@ -1421,13 +1421,30 @@ class DeliveryMechanism(Base, HasFullTableCache):
 
     # These are the media type/DRM scheme combos known to be supported
     # by the default Library Simplified client.
+    #
+    # This is primarily used when deciding which books can be imported
+    # from an OPDS For Distributors collection.
     default_client_can_fulfill_lookup = set([
+        # EPUB books
         (MediaTypes.EPUB_MEDIA_TYPE, NO_DRM),
         (MediaTypes.EPUB_MEDIA_TYPE, ADOBE_DRM),
-        (MediaTypes.EPUB_MEDIA_TYPE, BEARER_TOKEN),
+
+        # PDF books
+        (MediaTypes.PDF_MEDIA_TYPE, NO_DRM),
+
+        # Various audiobook formats
         (None, FINDAWAY_DRM),
-        (MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE, None),
+        (MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE, NO_DRM),
     ])
+
+    # If the default client supports a given media type with no DRM,
+    # we can infer that the client _also_ supports that media type via
+    # bearer token exchange.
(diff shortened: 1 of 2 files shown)


model/licensing.py
```diff
@@ -1421,14 +1421,31 @@ class DeliveryMechanism(Base, HasFullTableCache):
...
 
+    # If the default client supports a given media type with no DRM,
+    # we can infer that the client _also_ supports that media type via
+    # bearer token exchange.
+    for media_type, drm in list(default_client_can_fulfill_lookup):
+        if media_type is not None and drm == NO_DRM:
+            default_client_can_fulfill_lookup.add(
+                (media_type, BEARER_TOKEN)
+            )
+
```
leonardr: This is an assumption on my part which needs to be verified by @kyles-ep and @io7m.

tests/models/test_licensing.py
```diff
@@ -76,15 +77,35 @@ def test_content_type_media_type(self):
...
+            # database was initialized.
+            mechanism, is_new = DeliveryMechanism.lookup(self._db, media, drm)
+            eq_(False, is_new)
+            eq_(True, mechanism.default_client_can_fulfill)
+
+        # It's possible to create new DeliveryMechanisms at runtime,
+        # but their .default_client_can_fulfill will be False.
         mechanism, is_new = DeliveryMechanism.lookup(
-            self._db, Representation.EPUB_MEDIA_TYPE,
+            self._db, MediaTypes.EPUB_MEDIA_TYPE,
```
leonardr: Beyond this point I'm just changing `Representation` to `MediaTypes`, which is where these constants are actually stored.

model/licensing.py
leonardr: Update: tentatively confirmed by both. Let's go forward and if there are problems we can fix them on the SimplyE side.

model/licensing.py
```diff
@@ -1421,14 +1421,31 @@ class DeliveryMechanism(Base, HasFullTableCache):
...
         (MediaTypes.EPUB_MEDIA_TYPE, ADOBE_DRM),
-        (MediaTypes.EPUB_MEDIA_TYPE, BEARER_TOKEN),
+
+        # PDF books
+        (MediaTypes.PDF_MEDIA_TYPE, NO_DRM),
+
+        # Various audiobook formats
         (None, FINDAWAY_DRM),
-        (MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE, None),
+        (MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE, NO_DRM),
```
leonardr: NO_DRM is None, this is just to improve readability.

PR #1135, opened 2019-11-14
Incorporate pagination instructions into the Elasticsearch query itself.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2409 by consistently moving the point at which lists are paginated from outside `query_works_multi` to inside `query_works_multi`. In some cases, we were doing pagination twice -- once in the Elasticsearch server and once in Python code afterwards -- which led to empty result sets.

Imagine two lists of 50 books each. We want to issue a single Elasticsearch query that gets items 1-10 in list A, and items 31-40 in list B. That part was working just fine -- `query_works_multi` would return two lists of ten books each.

But then we ran some _additional_ code that selected items [0:10] from list A and items [30:40] from list B. List A is fine -- there are ten items in the list and we want the first ten. List B would end up empty: we're asking for items 31-40 but there are only ten items in the list (representing items 31-40 in the underlying resultset). 

Our existing tests didn't catch this problem because we didn't have a test against a real search index for regular `Pagination` -- only for `SortKeyPagination`.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -532,20 +532,14 @@ class ExternalSearchIndex(HasSelfTests):
                         result.meta['score'] or 0, result.meta['shard']
                     )
 
-        # Process each item in the resultset according to the
-        # Pagination object that was used to create it.
         for i, results in enumerate(resultset):
-            # Use the Pagination object to slice up the results if
-            # necessary.
-            query_string, filter, pagination = queries[i]
-            start = pagination.offset
-            stop = start + pagination.size
-            page = results[start:stop]
-
             # Tell the Pagination object about the page that was just
             # 'loaded' so that Pagination.next_page will work.
-            pagination.page_loaded(page)
-            yield page
+            #
+            # The pagination itself happened inside the Elasticsearch
+            # server when the query ran.
+            pagination.page_loaded(results)
+            yield results
 
     def count_works(self, filter):
         """Instead of retrieving works that match `filter`, count the total."""
@@ -1478,6 +1472,9 @@ class Query(SearchBase):
         :param elasticsearch: An Elasticsearch-DSL Search object. This
             object is ready to run a search against an Elasticsearch server,
             but it doesn't represent any particular Elasticsearch query.
+
+        :param pagination: A Pagination object indicating a slice of
+            results to pull from the search index.
 
         :return: An Elasticsearch-DSL Search object that's prepared
             to run this specific query.
@@ -2990,7 +2987,7 @@ class SortKeyPagination(Pagination):
             search = search.update_from_dict(
                 dict(search_after=self.last_item_on_previous_page)
             )
-        return search
+        return super(SortKeyPagination, self).modify_search_query(search)
 
     @property
     def previous_page(self):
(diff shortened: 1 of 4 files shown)


external_search.py
```diff
@@ -532,20 +532,14 @@ def query_works_multi(self, queries, debug=False):
...
-        # Process each item in the resultset according to the
-        # Pagination object that was used to create it.
         for i, results in enumerate(resultset):
-            # Use the Pagination object to slice up the results if
-            # necessary.
-            query_string, filter, pagination = queries[i]
-            start = pagination.offset
-            stop = start + pagination.size
-            page = results[start:stop]
-
```
leonardr: This is the code that caused the bug.

external_search.py
```diff
@@ -2990,7 +2987,7 @@ def modify_search_query(self, search):
             search = search.update_from_dict(
                 dict(search_after=self.last_item_on_previous_page)
             )
-        return search
+        return super(SortKeyPagination, self).modify_search_query(search)
```
leonardr: To be safe, `SortKeyPagination` also needs to run regular pagination, just in case the server returned more results than we asked for.

lane.py
```diff
@@ -997,9 +997,12 @@ def modify_database_query(self, _db, qu):
         return qu.offset(self.offset).limit(self.size)
 
     def modify_search_query(self, search):
-        # Do nothing -- all necessary pagination information is kept in
-        # offset and size, which external_search knows how to apply.
-        return search
```
leonardr: external_search applying this information is what caused the bug. We need to handle pagination _before_ the query goes to the Elasticsearch server, not afterwards.

tests/test_external_search.py
```diff
@@ -1321,35 +1321,42 @@ def assert_order(sort_field, order, **filter_kwargs):
             facets.order_ascending = False
             expect(list(reversed(order)), None, Filter(facets=facets, **filter_kwargs))
 
-            # Get each item in the list as a separate page. This proves
-            # that pagination based on SortKeyPagination works for this
-            # sort order.
+            # Get each item in the list as a separate page. This
+            # proves that pagination works for this sort order for
+            # both Pagination and SortKeyPagination.
```
leonardr: This is the test that would have caught the problem earlier.

tests/test_external_search.py
```diff
@@ -4120,9 +4127,14 @@ def test_unimplemented_features(self):
...
-            called_with = "not called"
+            update_from_dict_called_with = "not called"
+            getitem_called_with = "not called"
             def update_from_dict(self, dict):
-                self.called_with = dict
+                self.update_from_dict_called_with = dict
+                return self
+
+            def __getitem__(self, slice):
+                self.getitem_called_with = slice
```
leonardr: It's now necessary to verify that our mocked-up SearchKeyPagination object gets sliced, just like a regular Pagination object would.

external_search.py
EdwinGuzman: So, this will return `self.size` amount of results since `offset` is always 0?

leonardr: yep.

PR #1136, opened 2019-11-14
Copied the slice_timespan method in from circulation and added a test for it.

leonardr:
A tiny support branch for https://jira.nypl.org/browse/SIMPLY-1221. The Bibliotheca monitor has a helper method that slices a span of time into chunks, so that you can make API requests for small periods of time that won't overwhelm the API. 1221 is about bringing this same functionality to the Enki monitor, so I moved the helper method to `TimelineMonitor`, the common superclass of the Bibliotheca and Enki monitors.

The code is the same as what's currently in circulation/api/bibliotheca.py, except I made it a class method. The test is new; this previously didn't have a standalone test.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -286,6 +286,27 @@ class TimelineMonitor(Monitor):
         """
         raise NotImplementedError()
 
+    @classmethod
+    def slice_timespan(cls, start, cutoff, increment):
+        """Slice a span of time into segements no large than [increment].
+
+        This lets you divide up a task like "gather the entire
+        circulation history for a collection" into chunks of one day.
+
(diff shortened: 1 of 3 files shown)


monitor.py
```diff
@@ -286,6 +286,27 @@ def catch_up_from(self, start, cutoff, progress):
         """
         raise NotImplementedError()
 
+    @classmethod
+    def slice_timespan(cls, start, cutoff, increment):
+        """Slice a span of time into segements no large than [increment].
```
EdwinGuzman: Small typo for 'segments'.

PR #1138, opened 2019-11-19
Add IntegrationClient.enabled

leonardr:
This branch is the core portion of https://github.com/NYPL-Simplified/metadata_wrangler/pull/202, for https://jira.nypl.org/browse/SIMPLY-2418. It adds a new database field to `integrationclients` so we can keep track of which clients are enabled.

### DIFF
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -98,14 +98,14 @@ class ReplacementPolicy(object):
         )
 
     @classmethod
-    def from_license_source(self, _db, **args):
+    def from_license_source(cls, _db, **args):
         """When gathering data from the license source, overwrite all old data
         from this source with new data from the same source. Also
         overwrite an old rights status with an updated status and update
         the list of available formats. Log availability changes to the
         configured analytics services.
         """
-        return ReplacementPolicy(
+        return cls(
             identifiers=True,
             subjects=True,
             contributions=True,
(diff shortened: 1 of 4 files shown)


model/__init__.py
```diff
@@ -506,6 +506,7 @@ def production_session(initialize_data=True):
 from configuration import (
     ConfigurationSetting,
     ExternalIntegration,
+    ExternalIntegrationLink,
```
EdwinGuzman: Why does this need to be imported if it's not being used in the file?

metadata_layer.py
```diff
@@ -98,14 +98,14 @@ def __init__(
         )
 
     @classmethod
-    def from_license_source(self, _db, **args):
+    def from_license_source(cls, _db, **args):
```
EdwinGuzman: Can these methods still work if they are defined as class methods but take in `self` as the first param?

model/__init__.py
leonardr: These import statements are what let us do stuff like `from core.model import ExternalIntegration` everywhere instead of having to remember `from core.model.configuration import ExternalIntegration`. This functionality wasn't added when the `ExternalIntegrationLink` was created and I didn't notice until I actually started writing code that used `ExternalIntegrationLink`.

metadata_layer.py
leonardr: Yeah, the variable name is purely informative. But the previous code was wrong from an understandability viewpoint, and code like `self()` looks really off.

PR #1139, opened 2019-11-21
Make MetadataWranglerOPDSLookup implement HasSelfTests

leonardr:
This branch implements https://jira.nypl.org/browse/SIMPLY-2421, the first stage of https://jira.nypl.org/browse/SIMPLY-2230.

Calling run_self_tests will check every appropriate collection on the circulation manager against the corresponding collection on the metadata wrangler. Two tests are run for each collection:

1. How many metadata updates were there for items in this collection in the past 24 hours?
2. Is the metadata wrangler waiting for any information (e.g. book covers) on this collection that the circulation manager could provide but hasn't?

In each case, the test is run by making a request to the metadata wrangler that ought to return an OPDS feed in response. The feed is parsed and the list of titles is used as the `.result` of the `SelfTestResult` object.

If the metadata wrangler returns 5xx in response to the request (this is pretty common, especially for #2, which hasn't worked properly for a while), that is useful information that goes into the `.result`.

As part of this work, I changed `HasSelfTests` to accept a `.collection`. This lets us implement `MetadataWranglerOPDSLookup._run_self_tests` to yield a stream of SelfTestResults objects, the same as all other implementations, without losing track of which "past 24 hours" result came from which collection.

I noticed that even when everything is working, the self-test process can take a long time if there are a lot of collections. Not sure what we can do about this -- it's slow because there are a lot of potentially slow HTTP requests happening.

The tests here have a lot of "test method A by verifying that it calls method B, test method B by verifying that it calls method C". So you can visualize it, here's how the methods line up:

```
HasSelfTest.run_self_tests
 MetadataWranglerOPDSLookup._run_self_tests
  MetadataWranglerOPDSLookup._run_collection_self_tests (x number of collections)
   MetadataWranglerOPDSLookup._feed_self_test (2x)
    MetadataWranglerOPDSLookup._annotate_feed_response
```

### DIFF
--- a/opds_import.py
+++ b/opds_import.py
@@ -198,8 +198,8 @@ class MetadataWranglerOPDSLookup(SimplifiedOPDSLookup, HasSelfTests):
                 continue
 
             lookup = lookup_class.from_config(_db, c)
-            results[c] = list(lookup._run_collection_self_tests())
-        return results
+            for i in lookup._run_collection_self_tests():
+                yield i
 
     def _run_collection_self_tests(self):
         """Run the self-test suite on the Collection associated with this
@@ -232,12 +232,12 @@ class MetadataWranglerOPDSLookup(SimplifiedOPDSLookup, HasSelfTests):
         ):
             yield self._feed_self_test(title, m, *args)
 
-    @classmethod
-    def _feed_self_test(cls, name, method, *args):
+    def _feed_self_test(self, name, method, *args):
         """Retrieve a feed from the metadata wrangler and 
         turn it into a SelfTestResult.
         """
         result = SelfTestResult(name)
+        result.collection = self.collection
 
         # If the server returns a 500 error we don't want to raise an
         # exception -- we want to record it as part of the test
@@ -245,9 +245,7 @@ class MetadataWranglerOPDSLookup(SimplifiedOPDSLookup, HasSelfTests):
         kwargs = dict(allowed_response_codes=['%sxx' % f for f in range(1,6)])
 
         response = method(*args, **kwargs)
-        result.result = "\n".join(
-            cls._summarize_feed_response(response)
-        )
+        result.result = self._summarize_feed_response(response)
 
         # Once we process the OPDS feed we can call this a success.
         result.success = True
--- a/opds_import.py
+++ b/opds_import.py
@@ -219,7 +219,6 @@ class MetadataWranglerOPDSLookup(SimplifiedOPDSLookup, HasSelfTests):
 
         # Check various endpoints that yield OPDS feeds.
         one_day_ago = datetime.datetime.utcnow() - datetime.timedelta(days=1)
-        prefix = '"%s" - ' % self.collection.name
         for title, m, args in (
             (
                 "Metadata updates in last 24 hours", 
@@ -245,17 +244,16 @@ class MetadataWranglerOPDSLookup(SimplifiedOPDSLookup, HasSelfTests):
         kwargs = dict(allowed_response_codes=['%sxx' % f for f in range(1,6)])
 
         response = method(*args, **kwargs)
-        result.result = self._summarize_feed_response(response)
+        self._annotate_feed_response(result, response)
 
-        # Once we process the OPDS feed we can call this a success.
-        result.success = True
+        # We're all done.
         result.end = datetime.datetime.utcnow()
         return result
 
     @classmethod
-    def _summarize_feed_response(cls, response):
-        """Parse an OPDS feed and return some information
-        about it:
+    def _annotate_feed_response(cls, result, response):
+        """Parse an OPDS feed and annotate a SelfTestResult with some
+        information about it:
 
         * How the feed was requested.
         * What the response code was.
@@ -263,8 +261,8 @@ class MetadataWranglerOPDSLookup(SimplifiedOPDSLookup, HasSelfTests):
         * The title of each item on the page, if any.
         * The total number of items in the feed, if available.
 
+        :param result: A SelfTestResult object.
         :param response: A requests Response object.
-        :return: A list of strings summarizing the feed.
         """
         lines = []
         lines.append("Request URL: %s" % response.url)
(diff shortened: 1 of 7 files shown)


opds_import.py
```diff
@@ -173,6 +176,117 @@ def from_config(cls, _db, collection=None):
...
+        # its _run_collection_self_tests method.
+        for c in _db.query(Collection):
+            try:
+                metadata_identifier = c.metadata_identifier
+            except ValueError, e:
+                continue
+
+            lookup = lookup_class.from_config(_db, c)
+            for i in lookup._run_collection_self_tests():
+                yield i
```
leonardr: This is a little awkward, because we have to instantiate a bunch of `MetadataWranglerOPDSLookup` objects and call `_run_collection_self_tests` on each one. The advantage is that this preserves the behavior of the top-level method, `run_self_tests` --  instantiate a single object, call `_run_self_tests` on it, and expect to get a stream of `HasSelfTest` objects back.

opds_import.py
```diff
@@ -250,14 +363,14 @@ def add_with_metadata(self, feed):
         add_with_metadata_url = self.get_collection_url(self.ADD_WITH_METADATA_ENDPOINT)
         return self._post(add_with_metadata_url, unicode(feed))
 
-    def metadata_needed(self):
+    def metadata_needed(self, **kwargs):
```
leonardr: These methods need to start taking **kwargs because they're being called with the `allowed_response_codes` keyword argument.

testing.py
```diff
@@ -1426,16 +1426,31 @@ def do_post(self, url, data, *wargs, **kwargs):
...
+class MockRequestsRequest(object):
+    """A mock object that simulates an HTTP request from the
+    `requests` library.
+    """
+    def __init__(self, url, method="GET", headers=None):
+        self.url = url
+        self.method = method
+        self.headers = headers or dict()
+
+
```
leonardr: This is necessary to test `_annotate_feed_response`, which looks at the Authorization header that was sent as part of the interaction with the metadata wrangler. (This, in turn, makes it easy for someone looking at a test result to duplicate the HTTP request that caused it.)

opds_import.py
```diff
@@ -173,6 +176,116 @@ def from_config(cls, _db, collection=None):
...
+        lines.append("Request URL: %s" % response.url)
+        lines.append(
+            "Request authorization: %s" %
+            response.request.headers.get('Authorization')
+        )
+        lines.append("Status code: %d" % response.status_code)
+        result.success = response.status_code == 200
+        if result.success:
+            feed = feedparser.parse(response.content)
+            total_results = feed['feed'].get('opensearch_totalresults')
```
leonardr: This feature doesn't exist yet on the metadata wrangler side -- that's https://jira.nypl.org/browse/SIMPLY-2422.

opds_import.py
```diff
@@ -173,6 +176,116 @@ def from_config(cls, _db, collection=None):
...
+        * The total number of items in the feed, if available.
+
+        :param result: A SelfTestResult object.
+        :param response: A requests Response object.
+        """
+        lines = []
+        lines.append("Request URL: %s" % response.url)
+        lines.append(
+            "Request authorization: %s" %
+            response.request.headers.get('Authorization')
```
leonardr: See the comment below about `MockRequestsRequest`.

tests/files/opds/metadata_wrangler_overdrive.opds
```diff
@@ -1,9 +1,10 @@
-<feed xmlns:simplified="http://librarysimplified.org/terms/" xmlns:app="http://www.w3.org/2007/app" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:opds="http://opds-spec.org/2010/catalog" xmlns:schema="http://schema.org/" xmlns="http://www.w3.org/2005/Atom" xmlns:bibframe="http://bibframe.org/vocab/">
+<feed xmlns:simplified="http://librarysimplified.org/terms/" xmlns:app="http://www.w3.org/2007/app" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:opds="http://opds-spec.org/2010/catalog" xmlns:schema="http://schema.org/" xmlns="http://www.w3.org/2005/Atom" xmlns:bibframe="http://bibframe.org/vocab/" xmlns:opensearch="http://a9.com/-/spec/opensearch/1.1/">
   <id>http://localhost:5000/</id>
   <title>Open-Access Content</title>
   <updated>2015-01-02T16:56:40Z</updated>
   <link href="http://localhost:5000/"/>
   <link href="http://localhost:5000/" rel="self"/>
+  <opensearch:totalResults>201</opensearch:totalResults>
```
leonardr: This verifies that our side of the code will work once https://jira.nypl.org/browse/SIMPLY-2422 is implemented on the metadata wrangler side.

tests/test_opds_import.py
```diff
@@ -89,7 +94,16 @@ def update_work_for_edition(self, edition, *args, **kwargs):
             raise Exception("Utter work failure!")
 
 
-class TestMetadataWranglerOPDSLookup(DatabaseTest):
+class OPDSTest(DatabaseTest):
+    """A unit test that knows how to find OPDS files for use in tests."""
+
+    def sample_opds(self, filename):
```
leonardr: Later on in this file there's a lot of duplicative code for reading sample OPDS files from disk. Bringing that code up here to a superclass lets me simplify the old code and write `test__annotate_feed_response` in a simple way.

util/opds_writer.py
```diff
@@ -47,7 +48,8 @@ class AtomFeed(object):
         'schema' : SCHEMA_NS,
         'simplified' : SIMPLIFIED_NS,
         'bibframe' : BIBFRAME_NS,
-        'bib': BIB_SCHEMA_NS
+        'bib': BIB_SCHEMA_NS,
+        'opensearch': OPENSEARCH_NS,
```
leonardr: This is a tiny core change for https://jira.nypl.org/browse/SIMPLY-2422 so I don't have to do a whole other core branch. It's just making sure we can use tag names from the OpenSearch namespace in our OPDS documents.

PR #1140, opened 2019-12-02
Make works_updated_since more like our other complex queries, and improve performance

leonardr:
This branch is the core part of the fix to https://jira.nypl.org/browse/SIMPLY-2427.

Previously, works_updated_since was returning 3-tuples of (Work, LicensePool, Identifier) objects, which isn't generally how we do this kind of query. I changed it to return LicensePool objects with the Work and Identifier pre-loaded. This let me apply the "Work and Identifier objects are pre-loaded" optimization using `contains_eager`, dramatically improving performance. Previously we were using `joinedload`, which I think was incorrect -- it makes the database do _more_ work.

### DIFF
--- a/model/collection.py
+++ b/model/collection.py
@@ -627,13 +627,14 @@ class Collection(Base, HasFullTableCache):
         return query
 
     def works_updated_since(self, _db, timestamp):
-        """Finds all works in a collection's catalog that have been updated
-           since the timestamp. Used in the metadata wrangler.
+        """Finds all LicensePools in a collection's catalog whose Works' OPDS
+        entries have been updated since the timestamp. Used by the
+        metadata wrangler.
 
-           :return: a Query that yields (Work, LicensePool,
-              Identifier) 3-tuples. This gives caller all the
-              information necessary to create full OPDS entries for
-              the works.
+           :return: a Query that yields LicensePools. The Work and
+              Identifier associated with each LicensePool have been
+              pre-loaded, giving the caller all the information
+              necessary to create full OPDS entries for the works.
         """
         opds_operation = WorkCoverageRecord.GENERATE_OPDS_OPERATION
         qu = _db.query(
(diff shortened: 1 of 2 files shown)


model/collection.py
```diff
@@ -624,31 +625,40 @@ def unresolved_catalog(self, _db, data_source_name, operation):
...
+        ).join(
+            LicensePool.work,
+        ).join(
+            LicensePool.identifier,
         ).join(
             Work.coverage_records,
         ).join(
-            Identifier.collections,
+            CollectionIdentifier,
+            Identifier.id==CollectionIdentifier.identifier_id
```
EdwinGuzman: Is there a difference from doing this as a join here instead of a filter which was removed below?

model/collection.py
```diff
@@ -624,31 +625,40 @@ def unresolved_catalog(self, _db, data_source_name, operation):
...
+           necessary to create full OPDS entries for the works.
         """
         opds_operation = WorkCoverageRecord.GENERATE_OPDS_OPERATION
         qu = _db.query(
-            Work, LicensePool, Identifier
+            LicensePool
+        ).join(
+            LicensePool.work,
+        ).join(
+            LicensePool.identifier,
```
EdwinGuzman: You mentioned that there are two joins against the licensepools table but there should only be one. Are these different since they are joining against certain properties from LicensePool?

leonardr:
There's a SQL difference for sure: it's the difference between

```
select * from a join b on a.id=b.id
```

and

```
select * from a, b where a.id==b.id
```

Does it make a difference from a performance perspective? Theoretically it could -- JOIN is executed before WHERE, so you select a bunch of rows that are immediately eliminated -- but I believe Postgres is able to optimize this kind of query.

leonardr:
Yes, the joins against `LicensePool.work` and `LicensePool.identifier` aren't two joins against `licensepools`. They're a join between `licensepools` and `works`, plus a join between the same `licensepools` table and `identifiers`.

PR #1141, opened 2019-12-04
Updating the URNLookupController to URNLookupHandler.

leonardr:
This is a collaboration between @EdwinGuzman and myself to fix https://jira.nypl.org/browse/SIMPLY-2432.

All code in `URNLookupController` that keeps state is split out into a new class, `URNLookupHandler`. The controller code performs pre-request processing, instantiates a `URNLookupHandler` (a new one on every request), and uses the state stored in _that_ object to generate an OPDS feed.

The behavior is the same as before, but we've added test cases for a couple small bits of untested code, and a test that verifies that SIMPLY-2432 has been fixed.

This branch requires minor changes to the circulation branch and medium-sized refactoring in the metadata branch (which is where the main problem is happening). Those PRs are forthcoming.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -266,8 +266,15 @@ class HeartbeatController(object):
 
 
 class URNLookupController(object):
+    """A controller for looking up OPDS entries for specific books,
+    identified in terms of their Identifier URNs.
+    """
 
     def __init__(self, _db):
+        """Constructor.
+
+        :param _db: A database connection.
+        """
         self._db = _db
 
     def work_lookup(self, annotator, route_name='lookup', **process_urn_kwargs):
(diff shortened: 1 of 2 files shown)


app_server.py
```diff
@@ -266,55 +266,88 @@ def heartbeat(self, conf_class=None):
...
+        handler = URNLookupHandler(self._db)
+        handler.process_urns(urns, **process_urn_kwargs)
+        return handler
+
     def permalink(self, urn, annotator, route_name='work'):
-        """Look up a single identifier and generate an OPDS feed."""
+        """Look up a single identifier and generate an OPDS feed.
+
+        TODO: This method is tested, but it might not be used and it
+        may be possible to remove it.
```
leonardr: I don't think now is the time to remove this, but I've verified that we could remove it. I think it was used in the content server, but the circulation manager has its own `permalink` implementation and the metadata wrangler doesn't use this.

leonardr:
Here's the circulation PR, reviewed by Edwin: https://github.com/NYPL-Simplified/circulation/pull/1363. It wasn't really necessary -- I thought a test was missing but the test class had the wrong name.

PR #1142, opened 2019-12-10
Not raising a ValueError if a language code does not have any name.

EdwinGuzman:
Resolves [SIMPLY-458](https://jira.nypl.org/browse/SIMPLY-458). Instead of raising a ValueError for a language code that has no name, just skip over it.

### DIFF
--- a/util/__init__.py
+++ b/util/__init__.py
@@ -649,6 +649,7 @@ zza|||Zaza; Dimili; Dimli; Kirdki; Kirmanjki; Zazaki|zaza; dimili; dimli; kirdki
 
     @classmethod
     def name_for_languageset(cls, languages):
+        """Attempt to find any name for a language code."""
         if isinstance(languages, basestring):
             languages = languages.split(",")
         all_names = []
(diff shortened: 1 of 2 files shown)


util/__init__.py
```diff
@@ -661,9 +662,8 @@ def name_for_languageset(cls, languages):
                 all_names.append(native_names[0])
             else:
                 names = cls.english_names.get(normalized, [])
-                if not names:
-                    raise ValueError("No native or English name for %s" % l)
```
leonardr: Since this is in utility code and we don't know how it could be used (or even how it is already used in different parts of the code), I'd like to keep the `ValueError`. We just don't know how big a problem it is that the language didn't match, or five languages were passed in and only three of them matched.

In your circulation branch, it's not a big problem, so catching the `ValueError` and deciding not to create the lane is an appropriate response.

@aslagle Do you have an opinion?

aslagle: That sounds fine. I think these errors should at least be logged somewhere since it may be a problem with a book's metadata.

EdwinGuzman: Yes, I can make that change. If `name_for_languageset` returns a `ValueError`, then no lanes will be created at all since that is the value returned instead of any found language names, no?

PR #1143, opened 2019-12-12
Closer summaries are better

leonardr:
This branch breaks into two stages the procedure for choosing a summary for a work. First, summaries associated with an Identifier that is associated with one of the Work's LicensePools are considered. Second, summaries associated with Identifiers equivalent to those original Identifiers are considered. If nothing works, the Work's summary is cleared out.

This doesn't resolve https://jira.nypl.org/browse/SIMPLY-1696 as described, but it solves the most common real-world problem (which I think is exemplified by https://jira.nypl.org/browse/SIMPLY-1891). It's possible for a nice-looking (to a computer) summary to be chosen for a work, even if it has obvious (to a human) problems such as being in a different language. By checking descriptions close to the work first, we ensure that the two most common sources of good descriptions -- library staff and ebook vendors -- are always considered first.

We've had cases where the ebook vendor doesn't provide a description and we have to fill it in from elsewhere (in which case the second stage of this algorithm will run), but all commercial vendors currently provide decent descriptions, and the only case where the book vendor provides _bad_ descriptions is Project Gutenberg. That case can trip up this new algorithm, but we don't have Gutenberg titles in our system anymore, so I don't think it's a big deal.

### DIFF
--- a/model/work.py
+++ b/model/work.py
@@ -737,34 +737,33 @@ class Work(Base):
         )
         return q
 
-    def all_identifier_ids(self, policy=None, unified=True):
-        """Return all Identifier IDs associated with this Work.
-
-        :param policy: A `PresentationCalculationPolicy`.
-        :return: If unified=False, a 2-tuple (`direct_ids`, `all_ids`).
-           `direct_ids` is a list of Identifier IDs associated with
-             the Work's LicensePools.
-           `all_ids` is a set containing those Identifier IDs, plus
-             IDs of any Identifiers indirectly associated with them (
-             as per the rules set down in `policy`).
-           If unified=True, only `all_ids` is returned.
+    @property
+    def _direct_identifier_ids(self):
+        """Return all Identifier IDs associated with one of this
+        Work's LicensePools.
         """
-        _db = Session.object_session(self)
-        primary_identifier_ids = [
+        return [
             lp.identifier.id for lp in self.license_pools
             if lp.identifier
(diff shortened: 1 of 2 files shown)


model/work.py
```diff
@@ -737,7 +737,18 @@ def all_editions(self, policy=None):
...
         return q
 
-    def all_identifier_ids(self, policy=None):
+    def all_identifier_ids(self, policy=None, unified=True):
+        """Return all Identifier IDs associated with this Work.
+
+        :param policy: A `PresentationCalculationPolicy`.
+        :return: If unified=False, a 2-tuple (`direct_ids`, `all_ids`).
+           `direct_ids` is a list of Identifier IDs associated with
+             the Work's LicensePools.
```
leonardr: The process for deriving `direct_ids` is _just_ complicated enough that I decided to change the return value of this method so I could reuse it rather than copy-and-pasting the code (if it was less complicated) or refactoring it (if it was more complicated). I'm happy to do either instead.

For clarity, here's the code:

```
  primary_identifier_ids = [
            lp.identifier.id for lp in self.license_pools
            if lp.identifier
=]
```

aslagle: I'd have copied it just to make this method easier to understand, but it's fine either way. It's a little weird that one of the return values is a list and one is a set though.

leonardr: OK, I'll copy it to simplify the API.

PR #1144, opened 2019-12-12
Metadata self-tests

adriana-alter:


selftest.py
```diff
@@ -224,6 +225,11 @@ def external_integration(self, _db):
         By default, there is no way to get from an object to its
         ExternalIntegration, and self-test status will not be stored.
         """
+        import logging
+        logger = logging.getLogger()
```
leonardr: You should be able to import logging at the top of the file.

Also, the getLogger() call should name a specific logger so that it shows up as a message from the self-test framework rather than the Python process as a whole. The fancy way to do this would be to require that NAME be defined on every single `HasSelfTests` class, and define a `log` property that automatically looks up a logger based on NAME -- see `BaseCoverageProvider.log` for an example.

But since we're only doing one log message and it should only show up when there's code we forgot to write, I think it's fine to do `logging.getLogger("Self-test system")`.

PR #1145, opened 2019-12-13
Added a bunch of unit tests and changed around the sort name processor.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-378 without making things worse, by changing the way display names are converted to sort names. I added a number of unit tests, some of which reveal problems with the way we handle certain names, but equally bad problems existed before, so this is an improvement on average.

### DIFF
--- a/tests/test_personal_names.py
+++ b/tests/test_personal_names.py
@@ -44,56 +44,65 @@ class TestNameConversions(DatabaseTest):
         # Make sure the sort name algorithm processes the messy reality of contributor
         # names in a way we expect.
 
+        m = display_name_to_sort_name
+
         # no input means don't do anything
-        sort_name = display_name_to_sort_name(None)
+        sort_name = m(None)
         eq_(None, sort_name)
 
-        # already sort-ready input means don't do anything
-        sort_name = display_name_to_sort_name(u"Bitshifter, Bob")
-        eq_(u"Bitshifter, Bob", sort_name)
+        def unchanged(x):
+            # Verify that the input is already a sort name -- either
+            # because it's in "Family, Given" format or for some other
+            # reason.
+            eq_(x, m(x))
+        unchanged(u"Bitshifter, Bob")
+        unchanged(u"Prince")
+        unchanged(u"Pope Francis")
+        unchanged(u"Lord Byron")
+        unchanged(u"Heliodorus (bp. of Tricca.)")
+        unchanged(u"谢新源 (Xie Xinyuan)")
 
-        sort_name = display_name_to_sort_name(u"Prince")
-        eq_(u"Prince", sort_name)
-
-        sort_name = display_name_to_sort_name(u"Pope Francis")
-        eq_(u"Pope, Francis", sort_name)
-
-        sort_name = display_name_to_sort_name(u"Bob Bitshifter")
+        sort_name = m(u"Bob Bitshifter")
         eq_(u"Bitshifter, Bob", sort_name)
 
         # foreign characters don't confuse the algorithm
-        sort_name = display_name_to_sort_name(u"Боб Битшифтер")
+        sort_name = m(u"Боб Битшифтер")
         eq_(u"Битшифтер, Боб", sort_name)
 
-        sort_name = display_name_to_sort_name(u"Bob Bitshifter, Jr.")
+        sort_name = m(u"Bob Bitshifter, Jr.")
         eq_(u"Bitshifter, Bob Jr.", sort_name)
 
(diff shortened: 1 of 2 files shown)


tests/test_personal_names.py
```diff
@@ -44,56 +44,65 @@ def test_display_name_to_sort_name(self):
...
+            # Verify that the input is already a sort name -- either
+            # because it's in "Family, Given" format or for some other
+            # reason.
+            eq_(x, m(x))
+        unchanged(u"Bitshifter, Bob")
+        unchanged(u"Prince")
+        unchanged(u"Pope Francis")
+        unchanged(u"Lord Byron")
+        unchanged(u"Heliodorus (bp. of Tricca.)")
+        unchanged(u"谢新源 (Xie Xinyuan)")
```
leonardr: These four names were treated incorrectly by the old code -- the sort name contained an extra comma when it should have been left alone.

tests/test_personal_names.py
```diff
@@ -44,56 +44,65 @@ def test_display_name_to_sort_name(self):
...
+        # corporate names are unchanged
+        unchanged(u"Church of Jesus Christ of Latter-day Saints")
+        unchanged(u"(C) 2006 Vanguard")
 
+        # NOTE: These results are not the best.
+        eq_("Pope XVI, Benedict", m(u"Pope Benedict XVI"))
+        eq_("Lord Alfred, Tennyson", m(u"Alfred, Lord Tennyson"))
+        eq_("Lord Lennox, William", m("Lord William Lennox"))
+        sort_name = m(u"Bob, The Grand Duke of Awesomeness")
+        eq_(u"The Grand Bob, Duke of Awesomeness", sort_name)
```
leonardr: In these, we still get a wrong answer, but the wrong answer is slightly different. 
The only one here that's not new is "Bob, The Grand Duke of Awesomeness", which was previously normalized to "Bob, Duke of Awesomeness The Grand".

The difference is that "The Grand" is now treated as a title, and titles like "Pope" and "Lord" and "The Grand" now go at the front of someone's sort name.

tests/test_personal_names.py
```diff
@@ -44,56 +44,63 @@ def test_display_name_to_sort_name(self):
...
-        # corporate name is unchanged
-        sort_name = display_name_to_sort_name(u"Church of Jesus Christ of Latter-day Saints")
-        eq_(u"Church of Jesus Christ of Latter-day Saints", sort_name)
+        # corporate names are unchanged
+        unchanged(u"Church of Jesus Christ of Latter-day Saints")
+        unchanged(u"(C) 2006 Vanguard")
 
+        # NOTE: These results are not the best.
+        eq_("XVI, Pope Benedict", m(u"Pope Benedict XVI"))
+        eq_("Byron, Lord", m(u"Lord Byron"))
```
leonardr: These results were bad before, though the way in which they were bad was slightly different.

tests/test_personal_names.py
```diff
@@ -44,56 +44,63 @@ def test_display_name_to_sort_name(self):
...
+            # Verify that the input is already a sort name -- either
+            # because it's in "Family, Given" format or for some other
+            # reason.
+            eq_(x, m(x))
+        unchanged(u"Bitshifter, Bob")
+        unchanged(u"Prince")
+        unchanged(u"Pope Francis")
+        unchanged(u"Heliodorus (bp. of Tricca.)")
+        unchanged(u"谢新源 (Xie Xinyuan)")
+        unchanged(u"Alfred, Lord Tennyson")
```
leonardr: These results were bad before and now they're good.

util/personal_names.py
```diff
@@ -149,24 +149,34 @@ def display_name_to_sort_name(display_name):
...
+    # name.
+    base_name = u' '.join([name.title, name.first, name.middle, name.suffix]).strip()
     if not name.last:
         # Examples: 'Pope Francis', 'Prince'.
-        sort_name = u' '.join([name.first, name.middle, name.suffix, name.nickname])
-        if name.title:
-            sort_name = u''.join([name.title, ", ", sort_name])
+        sort_name = base_name
     else:
-        sort_name = u' '.join([name.first, name.middle, name.suffix, name.nickname, name.title])
```
EdwinGuzman: So now the title always goes first in the base name and not last?

leonardr: That's right. I made that decision based on https://nameparser.readthedocs.io/en/latest/, which lists three different ways of writing a name with Title always coming before the given name.

When I think of "title" I think of stuff like "5th Earl of Combsbury" which goes last, but by "title" nameparser means stuff like "Pope" or "Dr."

PR #1147, opened 2020-01-03
Add new audiences

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-538 by introducing two new audiences, "All Ages" and "Research".

An "All Ages" book will show up in Adult and Young Adult queries. It will show up in "Children" queries so long as the query includes children eight years old or older. (This is my guesstimate as to when children are expected to have the appropriate reading fluency -- it doesn't make sense to give Sherlock Holmes to three-year-olds. If this turns out to be inaccurate, we can change the cutoff age in the code and the new age will take effect immediately with no database changes)

A "Research" book will be excluded from all queries unless "Research" is explicitly listed as an audience.

Right now the only way for a book to be classified as 'All Ages' or 'Research' is for it to have a Classification associated with a schema:audience Subject with the name "All Ages" or "Research". This can happen through the admin interface or through the metadata wrangler. This is because there's no standard way of conveying this audience information in BISAC, DDC, etc.

### DIFF
--- a/classifier/__init__.py
+++ b/classifier/__init__.py
@@ -1294,6 +1294,8 @@ class WorkClassifier(object):
         ya_weight = w.get(Classifier.AUDIENCE_YOUNG_ADULT, 0)
         adult_weight = w.get(Classifier.AUDIENCE_ADULT, 0)
         adults_only_weight = w.get(Classifier.AUDIENCE_ADULTS_ONLY, 0)
+        all_ages_weight = w.get(Classifier.AUDIENCE_ALL_AGES, 0)
+        research_weight = w.get(Classifier.AUDIENCE_RESEARCH, 0)
 
         total_adult_weight = adult_weight + adults_only_weight
         total_weight = sum(w.values())
@@ -1323,6 +1325,10 @@ class WorkClassifier(object):
             audience = Classifier.AUDIENCE_YOUNG_ADULT
         elif total_adult_weight > 0:
             audience = Classifier.AUDIENCE_ADULT
+        elif all_ages_weight > 0:
+            audience = Classifier.AUDIENCE_ALL_AGES
+        elif research_weight > 0:
+            audience = Classifier.AUDIENCE_RESEARCH
 
         # If the 'adults only' weight is more than 1/4 of the total adult
         # weight, classify as 'adults only' to be safe.
--- a/classifier/__init__.py
+++ b/classifier/__init__.py
@@ -250,7 +250,7 @@ class Classifier(object):
             # likely the data is simply missing.
             return None
         if not lower:
-            if upper >= cls.AUDIENCE_ADULT:
+            if upper >= cls.AUDIENCE_ADULT_AGE_CUTOFF:
                 # e.g. "up to 20 years", though this doesn't
                 # really make sense.
                 #
--- a/classifier/__init__.py
+++ b/classifier/__init__.py
@@ -1314,7 +1314,11 @@ class WorkClassifier(object):
         # If the 'children' weight passes the threshold on its own
         # we go with 'children'.
         total_juvenile_weight = children_weight + ya_weight
-        if children_weight > threshold and children_weight > ya_weight:
+        if (research_weight > (total_adult_weight + all_ages_weight) and
+            research_weight > (total_juvenile_weight + all_ages_weight) and
+            research_weight > threshold):
+            audience = Classifier.AUDIENCE_RESEARCH
+        elif children_weight > threshold and children_weight > ya_weight:
             audience = Classifier.AUDIENCE_CHILDREN
         elif ya_weight > threshold:
             audience = Classifier.AUDIENCE_YOUNG_ADULT
@@ -1323,9 +1327,6 @@ class WorkClassifier(object):
             # combined they do pass the threshold. Go with
             # 'Young Adult' to be safe.
             audience = Classifier.AUDIENCE_YOUNG_ADULT
-        elif (research_weight > (total_adult_weight + all_ages_weight) and
-            research_weight > (total_juvenile_weight + all_ages_weight)):
-            audience = Classifier.AUDIENCE_RESEARCH
         elif (all_ages_weight > total_adult_weight and
             all_ages_weight > total_juvenile_weight):
             audience = Classifier.AUDIENCE_ALL_AGES
--- a/classifier/__init__.py
+++ b/classifier/__init__.py
@@ -1318,6 +1318,9 @@ class WorkClassifier(object):
             research_weight > (total_juvenile_weight + all_ages_weight) and
             research_weight > threshold):
             audience = Classifier.AUDIENCE_RESEARCH
+        elif (all_ages_weight > total_adult_weight and
+            all_ages_weight > total_juvenile_weight):
+            audience = Classifier.AUDIENCE_ALL_AGES
         elif children_weight > threshold and children_weight > ya_weight:
             audience = Classifier.AUDIENCE_CHILDREN
         elif ya_weight > threshold:
@@ -1327,9 +1330,6 @@ class WorkClassifier(object):
             # combined they do pass the threshold. Go with
             # 'Young Adult' to be safe.
             audience = Classifier.AUDIENCE_YOUNG_ADULT
-        elif (all_ages_weight > total_adult_weight and
-            all_ages_weight > total_juvenile_weight):
-            audience = Classifier.AUDIENCE_ALL_AGES
         elif total_adult_weight > 0:
             audience = Classifier.AUDIENCE_ADULT
 
--- a/classifier/__init__.py
+++ b/classifier/__init__.py
@@ -1323,12 +1323,14 @@ class WorkClassifier(object):
             # combined they do pass the threshold. Go with
             # 'Young Adult' to be safe.
             audience = Classifier.AUDIENCE_YOUNG_ADULT
+        elif (research_weight > (total_adult_weight + all_ages_weight) and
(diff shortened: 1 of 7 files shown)


classifier/__init__.py
```diff
@@ -255,6 +270,12 @@ def default_audience_for_target_age(cls, range):
             return cls.AUDIENCE_ADULT
         elif lower >= cls.YOUNG_ADULT_AGE_CUTOFF:
             return cls.AUDIENCE_YOUNG_ADULT
+        elif lower <= cls.ALL_AGES_AGE_CUTOFF and (
+            upper is not None and upper >= cls.ADULT_AGE_CUTOFF
+        ):
+            # e.g. "for children ages 7-77". The 'all ages' reading
+            # is here the most plausible.
+            return cls.AUDIENCE_ALL_AGES
```
leonardr: add to test_default_audience_for_target_age

classifier/__init__.py
```diff
@@ -876,6 +905,8 @@ def audience(cls, identifier, name):
             return cls.AUDIENCE_ADULT
         elif identifier == 'adults only':
             return cls.AUDIENCE_ADULTS_ONLY
+        elif identifier == 'all ages':
+            return cls.AUDIENCE_ALL_AGES
```
leonardr: FreeformAudienceClassifier doesn't seem to have its own test suite. It's possible it got lost when Adrianna and I refactored the classifier code. We can make a new one based on test_age.py.

classifier/__init__.py
```diff
@@ -1068,7 +1102,11 @@ def add(self, classification):
                     self.audience_weights[Classifier.AUDIENCE_YOUNG_ADULT] += (weight * 0.6)
                     self.audience_weights[Classifier.AUDIENCE_CHILDREN] += (weight * 0.4)
                     for audience in Classifier.AUDIENCES_ADULT:
-                        self.audience_weights[audience] -= weight * 0.5
+                        if audience != Classifier.AUDIENCE_ALL_AGES:
+                            # 'All Ages' is considered an adult audience,
+                            # but a generic 'juvenile' classification
+                            # is not evidence against it.
+                            self.audience_weights[audience] -= weight * 0.5
```
leonardr: This is tested by our change to test_juvenile_classification_is_split_between_children_and_ya.

external_search.py
```diff
@@ -2331,6 +2332,53 @@ def __init__(self, collections=None, media=None, languages=None,
         else:
             self.scoring_functions = []
 
+    @property
+    def audiences(self):
+        """Return the appropriate audiences for this query.
```
leonardr: Filter.audiences was previously a static property. We need a `TestFilter.test_audiences` that sets `filter._audiences` to a bunch of different values and then verifies that `filter.audiences` gives the right result.

external_search.py
```diff
@@ -2331,6 +2332,53 @@ def __init__(self, collections=None, media=None, languages=None,
...
+
+        This will be whatever audiences were provided, but it will
+        probably also include the 'All Ages' audience, and it may exclude
+        the 'Research' audience.
```
leonardr: Also this bit of the comment is out of date; we no longer exclude 'Research' here.

external_search.py
```diff
@@ -2395,6 +2443,8 @@ def build(self, _chain_filters=None):
 
         if self.audiences:
             f = chain(f, Terms(audience=scrub_list(self.audiences)))
+        else:
+            f = chain(f, Bool(must_not=[Term(audience=Classifier.AUDIENCE_RESEARCH)]))
```
leonardr: I think we've got this covered.

tests/test_external_search.py
```diff
@@ -90,6 +90,7 @@
     EndToEndSearchTest,
 )
 
+RESEARCH = Term(audience=Classifier.AUDIENCE_RESEARCH)
```
leonardr: In TestExternalSearchWithWorks add a really boring research book about Alice in Wonderland. Around line 825, add a test that this book only shows up on a search for "alice" if there is an audience that explicitly includes AUDIENCE_RESEARCH. In particular, when there is no audience filter, the book shouldn't show up.

tests/test_external_search.py
```diff
@@ -538,7 +539,8 @@ def populate_works(self):
 
         self.sherlock = _work(
             title="The Adventures of Sherlock Holmes",
-            with_open_access_download=True
+            with_open_access_download=True,
+            audience=Classifier.AUDIENCE_ADULT
```
leonardr: You can take this out; it turns out the default is ADULT.

tests/test_external_search.py
EdwinGuzman: Also, need an "all ages" book.

classifier/__init__.py
```diff
@@ -886,7 +919,10 @@ def target_age(cls, identifier, name):
             return cls.range_tuple(9, 12)
         if identifier == 'early adolescents':
             return cls.range_tuple(13, 15)
-
+        if identifier == 'all ages':
+            return cls.range_tuple(
+                cls.ALL_AGES_AGE_CUTOFF, cls.ADULT_AGE_CUTOFF
```
aslagle: this doesn't seem like it should have an upper bound

classifier/__init__.py
```diff
@@ -1323,6 +1325,10 @@ def audience(self, genres=[], default_audience=None):
             audience = Classifier.AUDIENCE_YOUNG_ADULT
         elif total_adult_weight > 0:
             audience = Classifier.AUDIENCE_ADULT
+        elif all_ages_weight > 0:
+            audience = Classifier.AUDIENCE_ALL_AGES
+        elif research_weight > 0:
+            audience = Classifier.AUDIENCE_RESEARCH
```
EdwinGuzman: @leonardr I had to add these in order for the work to successfully save the value that was being saved in the admin. I'm actually not too sure if the logic is correct.

model/classification.py
```diff
@@ -118,7 +118,8 @@ class Subject(Base):
     # the book's audience.
     audience = Column(
         Enum("Adult", "Young Adult", "Children", "Adults Only",
-             name="audience"),
+            "All Ages", "Research",
```
EdwinGuzman: This also threw an error when trying to save the new value in the admin.

leonardr: This makes sense.

classifier/__init__.py
leonardr: This does need some work but I'm not 100% sure what the right answer is. Here's my best guess:

* Theoretically, all_ages_weight should be added to both total_adult_weight and total_juvenile_weight, but that cancels out when they're being compared directly, and its detrimental when we're making other comparisons, so leave it alone.
* If research_weight is larger than _either_ total_adult_weight + all_ages_weight _or_ total_juvenile_weight + all_ages_weight, then the audience is 'Research'
* If all_ages_weight is larger than _either_ total_adult_weight _or_ total_juvenile_weight, then the audience is 'All Ages'

Amy, does this sound right?

aslagle: I agree with the first point, it doesn't make sense to add the all ages weight to other weights if all ages itself is a possible output.

I'm not sure about the second point - if we have some evidence that a book is for YA or Adult but more evidence that it's for research, calling it research means we'll exclude it from other lanes/searches even though it might be of general interest. Maybe we'd rather do that than include things that might not be interesting?

For the rest, I agree but we might be missing an audience for YA books that are also interesting for adults - maybe that's most YA books though.

EdwinGuzman: @aslagle

EdwinGuzman: For the second point, it makes sense to not include something that may not be interesting.

leonardr: I'm thinking specifically of children's books from before... Dr. Seuss, basically. There are a ton of these and there's very strong bibliographic evidence for filing them in the children's section, but apart from a few classics nobody actually does. They're kept in research collections.

There are YA books of interest to adults, but at this point the interested adults know to go find them in the YA section of the library/bookstore. I don't even know what term we'd use to describe those audiences; everything I can think of starts with "YA, but..."

EdwinGuzman: I updated this to include the logic for the second point.

classifier/__init__.py
```diff
@@ -1281,6 +1323,12 @@ def audience(self, genres=[], default_audience=None):
             # combined they do pass the threshold. Go with
             # 'Young Adult' to be safe.
             audience = Classifier.AUDIENCE_YOUNG_ADULT
+        elif (research_weight > (total_adult_weight + all_ages_weight) and
+            research_weight > (total_juvenile_weight + all_ages_weight)):
```
aslagle: I'm still not sure if this is quite what we want - if juvenile weight passes the threshold, we wouldn't hit this elif even if the research weight is much greater.

EdwinGuzman: I added an example in the tests where this occurs and the audience is research. Maybe it's not as accurate of an example?

leonardr: I think this setup is what Amy has in mind:

adult=100
juvenile=150
research=1000

EdwinGuzman: I see. This can work if the research clause is checked before any of the other clauses, but should research have more of a priority than the other weights?

leonardr: Yeah, check the research hypothesis first, and it has to outweigh every other hypothesis. (be higher than the children threshold, and the ya threshold, and the adult threshold, and the all ages threshold)

leonardr:
> Yeah, that case is already covered. I added a high All Ages weight to a test that already checks this to make sure.

The case where Children looks significantly better than Adult, but All Ages looks even better than that, isn't covered. I was able to express my concern as a failing test.

PR #1148, opened 2020-01-15
Allow an Overdrive integration to use their testing server

leonardr:
This is the core portion of https://jira.nypl.org/browse/SIMPLY-2486. It creates a new setting controlling whether a given OverdriveAPI hits the production or testing servers, and it replaces the hard-coded hostnames in endpoint constants with string interpolation variables.

Any time you reference a URL, you should call the endpoint() method to interpolate the appropriate hostname. _do_get or _do_post will call endpoint() for you -- but this will only  work for URLs like TOKEN_ENDPOINT, where the hostname is the _only_ string interpolation variable.

I designed this to make it easy to manually verify that we haven't missed any endpoint() calls.

### DIFF
--- a/overdrive.py
+++ b/overdrive.py
@@ -94,12 +94,12 @@ class OverdriveAPI(object):
     # system as for other hostnames to give a consistent look to the
     # templates.
     for host in HOSTS.values():
-        host['oauth_patron_host'] = "https://oauth-patron.overdrive.com",
-        host['oauth_host'] = "https://oauth.overdrive.com",
+        host['oauth_patron_host'] = "https://oauth-patron.overdrive.com"
+        host['oauth_host'] = "https://oauth.overdrive.com"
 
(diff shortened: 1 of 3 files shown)


overdrive.py
```diff
@@ -164,6 +207,17 @@ def __init__(self, _db, collection):
...
 
+    def endpoint(self, url, **kwargs):
+        """Create the URL to an Overdrive API endpoint.
+
+        :param url: A template for the URL.
+        :param kwargs: Arguments to be interpolated into the template.
+           The server hostname will be interpolated automatically; you
+           don't have to pass it in.
+        """
+        kwargs.update(self.hosts)
```
EdwinGuzman: Would this be considered a kind of merge between the two objects but the result going into `kwargs`?

leonardr: It is, but **kwargs is a new dict every time, so it's fine to modify it, and `self.hosts` won't be modified.

PR #1149, opened 2020-01-24
Improve performance of reapers

leonardr:
I tried to fix https://jira.nypl.org/browse/SIMPLY-2513 by adding a "bulk delete" feature (similar to the "bulk update" used by ScrubberMonitor), and it worked, but it introduced a number of database changes that I think would have been tricky to validate. Since the performance of reapers isn't all that important -- we just can't let them run out of memory -- I fell back to two simpler tricks that improve performance without changing the way the reapers work:

1. Delete items in batches rather than try to pull all deletable items from the database at once.
2. Allow a model class to specify which of its fields contain potentially huge chunks data that should be deferred (using SQLAlchemy `defer`) when doing bulk operations.

The "bulk delete" code is in an earlier revision of this branch, and I left in place the test of `WorkReaper` I designed to verify the cascading deletes, so if we decide to go back to this idea in the future we won't have to start over.

### DIFF
--- a/model/collection.py
+++ b/model/collection.py
@@ -794,9 +794,7 @@ class Collection(Base, HasFullTableCache):
                 _db.commit()
             if work and search_index and not work.license_pools:
                 search_index.remove_work(work)
-                # TODO: ideally we'd delete the Work itself, but
-                # that's another thing we don't normally do, and it
-                # might have side effects we haven't considered.
+                _db.delete(work)
 
         # Now delete the Collection itself.
         _db.delete(self)
--- a/model/work.py
+++ b/model/work.py
@@ -147,15 +147,14 @@ class Work(Base):
     )
 
     # One Work may be associated with many CustomListEntries.
-    custom_list_entries = relationship(
-        'CustomListEntry', backref='work',
-        cascade="all, delete-orphan"
-    )
+    # However, a CustomListEntry may lose its Work without
+    # ceasing to exist.
+    custom_list_entries = relationship('CustomListEntry', backref='work')
 
-    # One Work may have multiple CachedFeeds.
+    # One Work may have multiple CachedFeeds, and if a CachedFeed
+    # loses its Work, it ceases to exist.
     cached_feeds = relationship(
-        'CachedFeed', backref='work',
-        cascade="all, delete-orphan"
+        'CachedFeed', backref='work', cascade="all, delete-orphan"
     )
 
     # One Work may participate in many WorkGenre assignments.
--- a/model/work.py
+++ b/model/work.py
@@ -79,7 +79,7 @@ class WorkGenre(Base):
     __tablename__ = 'workgenres'
     id = Column(Integer, primary_key=True)
     genre_id = Column(Integer, ForeignKey('genres.id'), index=True)
-    work_id = Column(Integer, ForeignKey('works.id'), index=True)
+    work_id = Column(Integer, ForeignKey('works.id', ondelete='CASCADE'), index=True)
     affinity = Column(Float, index=True, default=0)
 
     @classmethod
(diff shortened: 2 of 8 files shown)


model/collection.py
```diff
@@ -794,9 +794,7 @@ def delete(self, search_index=None):
                 _db.commit()
             if work and search_index and not work.license_pools:
                 search_index.remove_work(work)
-                # TODO: ideally we'd delete the Work itself, but
-                # that's another thing we don't normally do, and it
-                # might have side effects we haven't considered.
+                _db.delete(work)
```
leonardr: Since the WorkReaper does exactly this -- deletes Works from the database because they have no license pools -- I no longer see a problem with deleting Works immediately in the CirculationReaper as they lose their license pools.

model/work.py
```diff
@@ -141,13 +141,21 @@ class Work(Base):
     presentation_edition_id = Column(Integer, ForeignKey('editions.id'), index=True)
 
     # One Work may have many associated WorkCoverageRecords.
-    coverage_records = relationship("WorkCoverageRecord", backref="work")
+    coverage_records = relationship(
+        "WorkCoverageRecord", backref="work",
+        cascade="all, delete-orphan"
+    )
```
leonardr: Without this, we can end up with dangling WorkCoverageRecords that have no associated Work.

model/work.py
```diff
@@ -141,13 +141,21 @@ class Work(Base):
...
 
     # One Work may be associated with many CustomListEntries.
+    # However, a CustomListEntry may lose its Work without
+    # ceasing to exist.
```
leonardr: This one tripped me up so I added a clarifying comment. An example would be an item on the NYT best-sellers list. It's associated with a Work in our system, but then we lose our license to the Work and it's deleted. That book is still on the best-sellers list!

model/work.py
```diff
@@ -141,13 +141,21 @@ class Work(Base):
...
     custom_list_entries = relationship('CustomListEntry', backref='work')
 
-    # One Work may have multiple CachedFeeds.
-    cached_feeds = relationship('CachedFeed', backref='work')
+    # One Work may have multiple CachedFeeds, and if a CachedFeed
+    # loses its Work, it ceases to exist.
+    cached_feeds = relationship(
+        'CachedFeed', backref='work', cascade="all, delete-orphan"
+    )
```
leonardr: This one is a little bit tricky because you _can_ have CachedFeeds created with work=None. But if you have a CachedFeed that _was_ associated with a Work, and then you set work=None, you're effectively deleting the feed. The CachedFeed isn't like the NYT best-seller list; it doesn't have an existence independent of the Work.

model/work.py
```diff
@@ -222,6 +230,13 @@ class Work(Base):
     # catalog.
     marc_record = Column(String, default=None)
 
+    # These fields are potentially large and can be deferred if you
+    # don't need all the data in a Work.
+    LARGE_FIELDS = [
+        'simple_opds_entry', 'verbose_opds_entry', 'marc_record',
+        'summary_text',
+    ]
+
```
leonardr: This should speed up the WorkReaper quite a bit.

monitor.py
```diff
@@ -775,10 +778,19 @@ class ReaperMonitor(Monitor):
...
+
+    If your model class has fields that might contain a lot of data
+    and aren't important to the reaping process, put their field names
+    into a list called LARGE_FIELDS and the Reaper will avoid fetching
+    that information, improving performance.
     """
     MODEL_CLASS = None
     TIMESTAMP_FIELD = None
     MAX_AGE = None
+    BATCH_SIZE = 1000
```
leonardr: The batch size should make the memory problem Robert was seeing go away in general, without us having to fine-tune every single reaper.

model/work.py
EdwinGuzman: Isn't it a problem that it's still on that best-sellers list?

model/work.py
EdwinGuzman: So the `defer` for these values tells the query to not load these columns unless if they are directly accessed, right?

leonardr: Right, if that happened they would be fetched in a separate query.

model/work.py
leonardr: Not really; the NYT dataset is just a bunch of information about some books. There's no requirement that we have licenses for any of those books. If we have a Work that corresponds to an item in the NYT dataset, and then we lose the license, the NYT dataset shouldn't change.

Similarly if we start out _not_ having a license for a book in the NYT dataset, and we get a license, the new Work should become associated with the NYT dataset.

PR #1150, opened 2020-01-27
Remove Work.random

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2518 by removing the `Work.random` field and everything to do with it. This field was once very important -- it's how we made sure the works that showed up in grouped feeds were ordered randomly -- but it's no longer used for anything, and the Monitor that chooses new random values for `Work.random` turns out to be a performance hassle.

This code does remove the capability for a `DatabaseBackedWorkList` to order works in random order, which explains why so many seemingly useful tests are being removed, but right now there are no `DatabaseBackedWorkList`s in the system (outside of the tests), and hopefully we will eventually decide that we can get rid of `DatabaseBackedWorkList` altogether.

(BTW we don't need `Work.random` anymore because we now ensure random ordering of grouped feeds by telling Elasticsearch to assign a small random weight to each search result.)

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -1103,7 +1103,6 @@ class CurrentMapping(Mapping):
             'sort_author_keyword' : ['sort_author'],
             'integer': ['series_position', 'work_id'],
             'long': ['last_update_time'],
-            'float': ['random'],
         }
         self.add_properties(fields_by_type)
 
--- a/lane.py
+++ b/lane.py
@@ -3,7 +3,6 @@ from collections import defaultdict
 from nose.tools import set_trace
 import datetime
 import logging
-import random
 import time
 import urllib
 
@@ -603,7 +602,6 @@ class DatabaseBackedFacets(Facets):
         FacetConstants.ORDER_TITLE : Edition.sort_title,
         FacetConstants.ORDER_AUTHOR : Edition.sort_author,
         FacetConstants.ORDER_LAST_UPDATE : Work.last_update_time,
-        FacetConstants.ORDER_RANDOM : Work.random,
     }
 
     @classmethod
--- a/migration/20200127-remove-work-random.sql
+++ b/migration/20200127-remove-work-random.sql
@@ -1,0 +1,6 @@
+-- Remove indexes that haven't been used for a very long time.
+drop index if exists ix_works_audience_target_age_quality_random;
+drop index if exists ix_works_audience_fiction_quality_random;
+
+-- Remove the works.random column itself.
+ALTER TABLE works DROP COLUMN if exists random;
--- a/model/work.py
+++ b/model/work.py
@@ -177,9 +177,6 @@ class Work(Base):
     # The overall current popularity of this work.
     popularity = Column(Float, index=True)
 
-    # A random number associated with this work, used for sampling/
-    random = Column(Numeric(4,3), index=True)
-
     appeal_type = Enum(CHARACTER_APPEAL, LANGUAGE_APPEAL, SETTING_APPEAL,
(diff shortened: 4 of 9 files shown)


external_search.py
```diff
@@ -1103,7 +1103,6 @@ def __init__(self):
             'sort_author_keyword' : ['sort_author'],
             'integer': ['series_position', 'work_id'],
             'long': ['last_update_time'],
-            'float': ['random'],
```
leonardr: I'm not rebuilding the index from scratch, so migrated ES6 indexes will have a float field called 'random' that doesn't get used, but that's fine -- it's optional.

lane.py
```diff
@@ -603,7 +602,6 @@ class DatabaseBackedFacets(Facets):
         FacetConstants.ORDER_TITLE : Edition.sort_title,
         FacetConstants.ORDER_AUTHOR : Edition.sort_author,
         FacetConstants.ORDER_LAST_UPDATE : Work.last_update_time,
-        FacetConstants.ORDER_RANDOM : Work.random,
```
leonardr: This is the bit that stops random order from working for DatabaseBackedLists. We can't sort randomly because there's no longer a database field to sort on.

model/work.py
```diff
@@ -1669,7 +1664,6 @@ def query_to_json_array(query):
              works_alias.c.imprint,
              works_alias.c.permanent_work_id,
              works_alias.c.presentation_ready,
-             works_alias.c.random,
```
leonardr: Here, we no longer add works.random when creating the search document.

leonardr:
Asking for Edwin's review but pinging @aslagle so she knows this is happening.

migration/20200127-remove-work-random.sql
```diff
@@ -0,0 +1,6 @@
+-- Remove indexes that haven't been used for a very long time.
+drop index if exists ix_works_audience_target_age_quality_random;
```
EdwinGuzman: These are the materialized views that were created before?

EdwinGuzman:
The PR branch is failing in Travis.

migration/20200127-remove-work-random.sql
leonardr: These indexes actually predate the materialized view. They're from back when we were generating grouped feeds with direct database queries. They haven't been necessary for many years.

leonardr:
This code is failing nondeterministically -- about 10% of the time.

```
        # By changing the minimum_featured_quality you can control
        # at what point a work is considered 'featured' -- at which
        # point its quality stops being taken into account.
        #
        # An extreme case of this is to set the minimum_featured_quality
        # to 0, which makes all works 'featured' and stops quality
        # from being considered altogether. Basically all that matters
        # is availability.
        all_featured_facets = FeaturedFacets(
            0, random_seed=Filter.DETERMINISTIC
        )
        assert_featured(
            "Works without considering quality",
            worklist, all_featured_facets,
            [self.hq_available, self.hq_available_2,
             self.not_featured_on_list, self.hq_not_available,
             self.featured_on_list],
        )
```

I don't know if this is new. It's certainly suspicious, because I changed the behavior of the search index. But all I did was stop indexing a random number, which you'd think would make things more deterministic rather than less.

`random_seed=Filter.DETERMINISTIC` stops us from _introducing_ a random element into the scoring, but this particular test also relies on Elasticsearch being deterministic internally. Otherwise there's nothing guaranteeing this particular order. So maybe the test itself is flawed: all we should really be testing is whether all the available books show up before all the unavailable books.

EdwinGuzman:
Nice way to fix the issue.

PR #1151, opened 2020-01-28
Move language utility code to its own module within the util package

leonardr:
I started investigating https://jira.nypl.org/browse/SIMPLY-564 and quickly ran into this bit of refactoring that I wanted to get out of the way. The `LanguageCodes` class was one of the first utility classes I wrote, back when all utility code was in `util.py`, and even though we have a `util` package now, that class was still in the `__init__.py` instead of its own module.

This branch moves language code stuff to `util/languages.py`, creates a new `tests/util/` package,  similar to `tests/model/` (since `tests/` was kind of getting full of `test_util_*.py` modules), and moves the tests to `tests/util/test_languages.py`.

There should be no code changes necessary outside of `util`, since `util/__init__.py` imports the classes from `util/languages.py`.


util/languages.py
```diff
@@ -0,0 +1,626 @@
...
+            return all_names[0]
+        return "/".join(all_names)
+
+
+def languages_from_accept(accept_languages):
+    """Turn a list of (locale, quality) 2-tuples into a list of language codes.
+    
+    TODO: parse_accept_language returns named 3-tuples. How is this
+    function actually used?
+    """
```
leonardr: I grepped around and I don't think this function is used at all. There is also no test for it. Can you confirm that I can delet it?

migration/20200127-remove-work-random.sql
```diff
@@ -3,4 +3,4 @@ drop index if exists ix_works_audience_target_age_quality_random;
 drop index if exists ix_works_audience_fiction_quality_random;
 
 -- Remove the works.random column itself.
-ALTER TABLE works DROP COLUMN if exists random;
+ALTER TABLE works DROP COLUMN if exists random CASCADE;
```
leonardr: This is a leftover change from my QA of https://jira.nypl.org/browse/SIMPLY-2518. Our QA server had an old materialized view lying around for some reason. This change makes the column drop also drop the materialized view, improving the robustness of the migration script.

util/languages.py
EdwinGuzman: I think you're right that it may have become obsolete after that change. `languages_from_accept` is being called in core/util/flask_util in `languages_for_request()` but that function isn't being used anywhere and can be deleted, too.

PR #1152, opened 2020-01-29
Added the LanguageNames class and its tests.

leonardr:
This is more precursor work for https://jira.nypl.org/browse/SIMPLY-564. I created a class that contains a regular expression that can match human-readable names of languages, and a dictionary that you can use to map one of those human-readable names to a set of ISO language codes.

### DIFF
--- a/tests/util/test_languages.py
+++ b/tests/util/test_languages.py
@@ -7,10 +7,12 @@ from nose.tools import (
     assert_raises,
 )
 
-from ...util import (
+from ...util.languages import (
     LanguageCodes,
+    LanguageNames,
     LookupTable,
 )
+
 
 class TestLookupTable(object):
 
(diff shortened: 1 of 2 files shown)


tests/util/test_languages.py
```diff
@@ -82,3 +84,90 @@ def test_name_for_languageset(self):
...
+
+    def test_name_re(self):
+        # Verify our ability to find language names inside text.
+        def find(text, expect):
+            match = LanguageNames.name_re.search(text)
+            if not match:
+                return match
+            return match.groups()[0]
+
+            eq_(expect, LanguageNames.find_language(text))
```
EdwinGuzman: I don't see a reference to `find_language` and this probably doesn't run because it's after the `return`.

tests/util/test_languages.py
```diff
@@ -82,3 +84,90 @@ def test_name_for_languageset(self):
...
+        find("books in Italian", "Italian")
+        find("Chinese Cooking", "Chinese")
+        find("500 spanish verbs", "spanish")
+
+        # Only the first language is returned.
+        find("books in japanese or italian", "japanese")
+        find("english-russian dictionary", "english")
```
EdwinGuzman: These tests are specific to the `find` function for this test, right? `LanguageNames.name_re` _could_ be implemented where you want to get all the languages in a string, right?

util/languages.py
```diff
@@ -605,3 +607,100 @@ def name_for_languageset(cls, languages):
...
+            ('castellano', 'es'),
+        ):
+            add(name, alpha)
+        return name_to_codes
+
+    @classmethod
+    def _build_name_re(cls):
+        return re.compile(
+            r"(\b%s\b)" %
+            r"\b|\b".join(cls.name_to_codes.keys()),
```
EdwinGuzman: I'm guessing python doesn't complain that this class variable doesn't exist since it doesn't look at it until the class is first instantiated below? So if the class variables weren't instantiated, this would throw an error, right?

tests/util/test_languages.py
leonardr: Yeah, this is old code.

tests/util/test_languages.py
leonardr: Yeah, you can use the regular expression for whatever you want. This is just checking that it matches the bits of strings we would expect.

With the "english-russian dictionary" case I was trying to show that the regular expression, on its own, will only match one language. I changed the test to show this better.

util/languages.py
leonardr: Yes, that's right.

PR #1155, opened 2020-02-03
Use the Accept-Language header to augment the languages available to the search engine

leonardr:
Previously the Accept-Language headers were only used to provide a fallback language filter if the Lane doesn't specify a language filter (which is basically never). This makes it hard for patrons who don't read English to search for books in their native language. Before that, the Accept-Language headers were used _exclusively_, which led to complaints by patrons who read in both English and their native language.

This isn't a perfect solution, but I think it's pretty good for the scenario we find ourselves in, given that UX improvements aren't going to happen any time soon.

This fixes https://jira.nypl.org/browse/SIMPLY-2530.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -787,9 +787,9 @@ class SearchFacets(FacetsWithEntryPoint):
     def from_request(cls, library, config, get_argument, get_header, worklist,
                      default_entrypoint=None, **extra):
 
-        # Searches against a WorkList that has no particular language
-        # restrictions will use the languages defined in the
-        # Accept-Language header used by the client.
+        # Searches against a WorkList will use the union of the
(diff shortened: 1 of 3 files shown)


EdwinGuzman:
It seems there's a test that is failing in the PR Travis test.

leonardr:
That seems like another nondeterministic test failure along the lines of the one we saw in https://github.com/NYPL-Simplified/server_core/pull/1150.

EdwinGuzman:
Everything else looks good.

leonardr:
It's possible that other assertions in that test will fail at some point, but I've got a pattern for solving them now: test only the specific assertions that need to be checked, rather than checking that books come out of Elasticsearch in a precisely predictable order.

PR #1156, opened 2020-02-03
Add paired attribute

adriana-alter:
Goes with https://github.com/NYPL-Simplified/circulation-web/pull/290


config.py
```diff
@@ -233,6 +233,8 @@ class Configuration(ConfigurationConstants):
           ],
           "default": [x.INTERNAL_NAME for x in EntryPoint.DEFAULT_ENABLED],
           "category": "Lanes & Filters",
+          "format": "narrow",
+          "locked": True
```
EdwinGuzman: Should these values have comments or maybe a location where all possible values are listed so we know what is possible to do with these configurations?

adriana-alter:
@EdwinGuzman Is this also good to go?

PR #1157, opened 2020-02-07
Initial core support for Overdrive audio manifests

leonardr:
This is the core branch for https://jira.nypl.org/browse/SIMPLY-2546.

Most of the work in this branch is boring:

1. Changing constants to be consistent and accurate, rather than let them become or remain confusing.
2. Making sure that our code draws the right conclusions when it sees the Overdrive API advertising titles in certain formats.

The circulation branch is the one that lets patrons actually get links to audio manifest files.

I added a new "DRM system" so we can talk about the access control mechanism used for Overdrive audiobooks. (It's technically an access control mechanism, but we say 'DRM system' for any technical measure that gates an authorized user's access to a book.)

I chose "Libby DRM" as the name of this DRM system because it doesn't have a special media type (unlike ACSM) and it doesn't have an official name, but it's vaguely associated with the mobile app called "Libby". If another term is preferable I can change it easily.

Previously we used the string "Overdrive DRM" to identify the 'DRM system' in place for Overdrive Read and Overdrive Listen. This was inaccurate; the actual 'DRM system' is the fact that the streaming format itself makes it difficult to save an entire book. We already have other constants for this purpose ("Streaming Text" and "Streaming Audio"), so I started using them.

The string "Overdrive DRM" now only refers to the proprietary DRM system associated with the `application/x-od-media` media type. This is used by the mobile app called "Overdrive" and it turns out we won't be using it. However, we will continue to keep track of it, so long as the Overdrive API advertises `application/x-od-media`, just so we keep an accurate picture of the Overdrive API's capabilities.

I added two new content types--media types that a client must understand if it wants to render a book for its user. Both are based on the media type associated with Overdrive manifest documents, which is `application/vnd.overdrive.circulation.api+json`.

Here's the thing: based on the format and content of the manifest file, I believe `application/vnd.overdrive.circulation.api+json` is used for both ebooks and audiobooks. _We do not have a contractual agreement to get ebooks through this system and this code does not attempt to get ebooks through this system._ But I do want to _eventually_ get to a point where this becomes possible. If and when we do get there, I want it to be obvious internally which manifests give you an ebook and which give you an audiobook. So, for the sake of clarity, I've added two made-up profile names to this media type for internal use:

1. `application/vnd.overdrive.circulation.api+json;profile=audiobook` for audiobook manifests
2. `application/vnd.overdrive.circulation.api+json;profile=ebook` for hypothetical ebook manifests

SimplyE will always see `application/vnd.overdrive.circulation.api+json`, but that's fine, because it knows from other pieces of information in the OPDS feed whether a given title is an ebook or an audiobook.

I then added two new delivery mechanisms to correspond to the new system for delivering audiobook manifests, and the prospective (and purely hypothetical) system for delivering ebook manifests.

I wrote a migration script to make appropriate changes to an old database, which would have been configured with the wrong DRM names and whose Overdrive books would lack the new delivery mechanisms. I've tested the migration script on a database that was already migrated, to make sure it doesn't crash. I haven't tested it on a database that needs migration. Even if all else fails, a circulation manager's `overdrive_format_sweep` script  (which runs on a regular basis) will eventually fix all Overdrive titles in the database based on this branch's expanded understanding of Overdrive API data.

The Overdrive API now uses "audiobook-overdrive" to refer to two different delivery mechanisms: Overdrive Listen (which we register as streaming audio format, streaming 'DRM') and audiobook manifest (which we register as `application/vnd.overdrive.circulation.api+json;profile=audiobook`, Libby DRM). Previously we assumed a 1:1 relationship between Overdrive's internal format names and our delivery mechanisms. The relationship is now one-to-many. This translation is handled by `OverdriveAPI.internal_formats`, which yields a sequence of formats and is used in places that previously assumed there was only one relevant format.

In the circulation branch, we use "audiobook-overdrive-manifest" as an alias to distinguish between "audiobook-overdrive, but as an Overdrive Listen link" and "audiobook-overdrive, but as a manifest link". There is no corresponding alias for "ebook-overdrive". For now, it remains fulfillable only as an Overdrive Read link.


migration/20200201-new-overdrive-formats.sql
```diff
@@ -0,0 +1,81 @@
...
+     )
+   from licensepooldeliveries ld join deliverymechanisms d on ld.delivery_mechanism_id=d.id
+    where d.content_type='Streaming Audio' and d.drm_scheme='Overdrive DRM'
+   ;
+ END;
+
+  -- Do the same thing for ebooks that are available streaming through
+  -- Overdrive Read, it's also available through a manifest. Add
+  -- rows to licensepooldeliveries to represent the available
+  -- manifests.
```
leonardr: n.b. There is no implication here that we can or will _retrieve_ this manifest. It's purely prospective, like `application/x-od-media`.

overdrive.py
```diff
@@ -127,11 +128,14 @@ class OverdriveAPI(object):
 
     EVENT_DELAY = datetime.timedelta(minutes=120)
 
-    # The ebook formats we care about.
-    FORMATS = "ebook-epub-open,ebook-epub-adobe,ebook-pdf-adobe,ebook-pdf-open"
+    # The formats we care about.
+    FORMATS = "ebook-epub-open,ebook-epub-adobe,ebook-pdf-adobe,ebook-pdf-open,audiobook-overdrive".split(",")
```
leonardr: The old code worked by accident: we were checking whether a format was a substring of FORMATS rather than that it was in the list FORMATS.

overdrive.py
```diff
@@ -127,11 +128,14 @@ class OverdriveAPI(object):
...
 
     # The formats that can be read by the default Library Simplified reader.
-    DEFAULT_READABLE_FORMATS = set(["ebook-epub-open", "ebook-epub-adobe"])
+    DEFAULT_READABLE_FORMATS = set(
+        ["ebook-epub-open", "ebook-epub-adobe", "ebook-pdf-open",
```
leonardr: DRM-free PDF support was added to SimplyE a while ago; this updates the Overdrive client to notice those files in a library's Overdrive collection.

overdrive.py
```diff
@@ -127,11 +128,14 @@ class OverdriveAPI(object):
...
-    # The ebook formats we care about.
-    FORMATS = "ebook-epub-open,ebook-epub-adobe,ebook-pdf-adobe,ebook-pdf-open"
+    # The formats we care about.
+    FORMATS = "ebook-epub-open,ebook-epub-adobe,ebook-pdf-adobe,ebook-pdf-open,audiobook-overdrive".split(",")
 
     # The formats that can be read by the default Library Simplified reader.
-    DEFAULT_READABLE_FORMATS = set(["ebook-epub-open", "ebook-epub-adobe"])
+    DEFAULT_READABLE_FORMATS = set(
+        ["ebook-epub-open", "ebook-epub-adobe", "ebook-pdf-open", 
+         "audiobook-overdrive"]
```
leonardr: And then this is the new bit.

overdrive.py
```diff
@@ -993,21 +1020,22 @@ def book_info_to_metadata(cls, book, include_bibliographic=True, include_formats
...
+                    for content_type, drm_scheme in internal_names:
+                        if Representation.is_media_type(content_type):
+                            for sample_info in format['samples']:
+                                href = sample_info['url']
+                                links.append(
+                                    LinkData(
+                                        rel=Hyperlink.SAMPLE,
+                                        href=href,
+                                        media_type=content_type
+                                    )
```
leonardr: This isn't relevant at this point, but one change that happened since this code was written was the Overdrive samples switched from being served as overdrive-epub-open to being served as overdrive-ebook or overdrive-audiobook.

PR #1158, opened 2020-02-11
Don't allow a book to become presentation-ready if it has no medium

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2555. I'll do a small circulation branch but it'll probably just be cleanup.

If we don't know whether a book is an ebook or an audiobook, it shouldn't become presentation-ready. This is generally a failure of the import process and indicates some missing code.

In particular, I noticed the `<dcterms:format>` tag (which is required in ODL feeds, and could show up in OPDS feeds) can be used to derive the medium of the work, but our core code ignores it and our circulation code only uses it to detect audiobooks. If the incoming title is an ebook it doesn't do anything.

I also noticed that if a book in an OPDS feed is only available as a PDF, our OPDS importer won't recognize it as an ebook.

This branch makes the OPDS importer more robust and makes it check for a `<dcterms:format>` tag. It makes "no medium" a reason not to set a work as presentation-ready. If a work does end up not presentation-ready, `logging.warn` is used to indicate that there's probably a problem.

### DIFF
--- a/model/edition.py
+++ b/model/edition.py
@@ -13,6 +13,7 @@ from constants import (
     DataSourceConstants,
     EditionConstants,
     LinkRelations,
+    MediaTypes,
 )
 from contributor import (
     Contributor,
@@ -20,7 +21,10 @@ from contributor import (
 )
 from datasource import DataSource
 from identifier import Identifier
-from licensing import LicensePool
+from licensing import (
+    DeliveryMechanism,
+    LicensePool,
+)
 
 from collections import defaultdict
 import logging
@@ -224,6 +228,32 @@ class Edition(Base, EditionConstants):
             # want to put them down as 'author'.
             return []
 
+
+    @classmethod
+    def medium_from_media_type(cls, media_type):
+        """Derive a value for Edition.medium from a media type.
+
+        TODO: It's not necessary right now, but we could theoretically
+        derive this information from some other types such as
+        our internal types for Overdrive manifests.
+
+        :param media_type: A media type with optional parameters
+        :return: A value for Edition.medium.
+        """
+        if not media_type:
+            return None
+
+        for types, conclusion in (
+            (MediaTypes.AUDIOBOOK_MEDIA_TYPES, Edition.AUDIO_MEDIUM),
+            (MediaTypes.BOOK_MEDIA_TYPES, Edition.BOOK_MEDIUM),
+        ):
+            if any(media_type.startswith(x) for x in types):
+                return conclusion
+
+        if media_type.startswith(DeliveryMechanism.ADOBE_DRM):
+            # Adobe DRM is only applied to ebooks.
+            return Edition.BOOK_MEDIUM
+        return None
 
     @classmethod
     def for_foreign_id(cls, _db, data_source,
--- a/model/work.py
+++ b/model/work.py
@@ -1233,6 +1233,15 @@ class Work(Base):
     def set_presentation_ready(
         self, as_of=None, search_index_client=None, exclude_search=False
     ):
+        """Set this work as presentation-ready, no matter what.
+
+        This assumes that we know the work has the minimal information
+        necessary to be found with typical queries and that patrons
+        will be able to understand what work we're talking about.
+
+        In most cases you should call set_presentation_ready_based_on_content
+        instead, which runs those checks.
+        """
         as_of = as_of or datetime.datetime.utcnow()
         self.presentation_ready = True
         self.presentation_ready_exception = None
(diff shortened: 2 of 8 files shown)


model/edition.py
```diff
@@ -225,6 +229,32 @@ def dedupe(l):
...
+            (MediaTypes.BOOK_MEDIA_TYPES, Edition.BOOK_MEDIUM),
+        ):
+            if any(media_type.startswith(x) for x in types):
+                return conclusion
+
+        if media_type.startswith(DeliveryMechanism.ADOBE_DRM):
+            # Adobe DRM is only applied to ebooks.
+            return Edition.BOOK_MEDIUM
+        return None
+
```
leonardr: This method contains code refactored from `OPDSImporter.get_medium_from_links` (so we can also call it on the contents of a `<dcterms:format>` tag) and greatly expanded.

model/work.py
```diff
@@ -1243,24 +1252,29 @@ def set_presentation_ready(
...
         necessarily mean the presentation is complete.
+
         The absolute minimum data necessary is a title, a language,
-        and a fiction/nonfiction status. We don't need a cover or an
-        author -- we can fill in that info later if it exists.
-        """
+        and a medium. We don't need a cover or an author -- we can
+        fill in that info later if it exists.
 
+        TODO: search_index_client is redundant here.
```
leonardr: I didn't want to change the signature in this branch, but we don't need any of this search stuff anymore.

tests/models/test_work.py
```diff
@@ -491,30 +491,55 @@ def test_set_presentation_ready(self):
...
 
+        # This work is presentation ready because it has a title.
         # Remove the title, and the work stops being presentation
         # ready.
         presentation.title = None
         work.set_presentation_ready_based_on_content(search_index_client=search)
         eq_(False, work.presentation_ready)
 
-        # The work has been removed from the search index.
-        eq_([], search.docs.keys())
```
leonardr: The behavior didn't change, but this assertion stopped having any value when we switched from "immediately remove an unready work from the search index" to "set this work's WorkCoverageRecord back to the REGISTERED state so the search index can handle it later." The thing we actually want to test is `assert_record()`.

tests/models/test_work.py
```diff
@@ -491,30 +491,55 @@ def test_set_presentation_ready(self):
...
+        # Remove the medium, and the work stops being presentation ready.
+        presentation.medium = None
+        work.set_presentation_ready_based_on_content(search_index_client=search)
+        eq_(False, work.presentation_ready)
+
+        presentation.medium = Edition.BOOK_MEDIUM
+        work.set_presentation_ready_based_on_content(search_index_client=search)
+        eq_(True, work.presentation_ready)
+
+        # Remove the language, and it stops being presentation ready.
```
leonardr: These bits weren't previously tested.

tests/test_opds_import.py
```diff
@@ -549,12 +551,19 @@ def test_extract_link(self):
...
         ]
         book_links = [
             LinkData(href="url", rel="http://opds-spec.org/image"),
-            LinkData(href="url", rel="http://opds-spec.org/acquisition/", media_type="application/epub+zip"),
+            LinkData(
+                href="url", rel="http://opds-spec.org/acquisition/",
+                media_type=random.choice(
+                    MediaTypes.BOOK_MEDIA_TYPES
+                ) + ";param=value"
+            ),
```
leonardr: This proves that any of the BOOK_MEDIA_TYPES is evidence that the work is BOOK_MEDIUM, even if there's a parameter attached to the media type.

tests/files/opds/book_without_license.opds
```diff
@@ -22,6 +22,7 @@
     <category term="18" scheme="http://schema.org/typicalAgeRange" label="18"/>
     <category term="http://librarysimplified.org/terms/fiction/Fiction" scheme="http://librarysimplified.org/terms/fiction/" label="Fiction"/>
     <dcterms:language>en</dcterms:language>
+    <dcterms:format>application/audiobook+json;protection="http://www.feedbooks.com/audiobooks/access-restriction"</dcterms:format>
```
leonardr: The changes to this file let me use `test_import_book_that_offers_no_license` as an end-to-end test of `extract_medium`. We're able to figure out that this book is an audiobook even though `schema:additionalType` is missing and there are no links. The only way to figure this out is to look at the `<dcterms:format>` tag.

tests/test_opds_import.py
```diff
@@ -748,6 +757,51 @@ def test_extract_messages(self):
...
+        eq_(
+            Edition.AUDIO_MEDIUM,
+            medium("http://bib.schema.org/Audiobook", ebook_type)
+        )
+        eq_(
+            Edition.BOOK_MEDIUM,
+            medium("http://schema.org/EBook", audio_type)
+        )
+
+        # When schema:additionalType is missing or not useful, the
```
EdwinGuzman: In what case would it not be useful?

tests/models/test_work.py
EdwinGuzman: This is related to the comment you made about how the search code is not relevant in `set_presentation_ready_based_on_content` anymore, right?

tests/test_opds_import.py
leonardr: It might be some random URI that we can't map to a medium.

tests/models/test_work.py
leonardr: Yes, exactly.

PR #1159, opened 2020-02-12
Make CachedFeed work more like Credential: centralize the database access code and do the actual refresh in a hook function

leonardr:
This branch eliminates the race condition described in https://jira.nypl.org/browse/SIMPLY-2532
A circulation branch (https://github.com/NYPL-Simplified/circulation/pull/1389) is necessary as well, because the functionality of `CachedFeed.fetch` has changed.

Basically, `CachedFeed.fetch` used to return a 2-tuple. The first item in the tuple was a `CachedFeed` object and the second was a boolean saying whether the `CachedFeed` was "fresh". A fresh feed could be used as-is. A stale feed needed to be updated by the caller and have its timestamp freshened.

This created a race condition: if two callers request the same nonexistent feed, they each get a brand new `CachedFeed` object. Whoever writes to the database last causes an integrity error, since both `CachedFeed`s represent the feed and have the same unique keys.

This branch fixes the race condition by making `CachedFeed.fetch` work more like `Credential.lookup`. It's no longer the callers' responsibility to see whether the feed is 'fresh' and update `CachedFeed.content` and `CachedFeed.timestamp` if it's not. Instead, the caller passes in a function that gets called if, and only if, it's time to generate a new feed.

By itself, this doesn't solve the problem. But moving all the "how do I regenerate this feed?" code into a hook function lets me add _additional_ code to `CachedFeed.fetch` that runs _after_ the hook function has completed. This new code solves the problem by looking up the `CachedFeed` _again_, before modifying `.content` and `.timestamp`.

That way, if someone else created a new `CachedFeed` while I was running the hook function, I'll pick it up on the second lookup and I won't create a second new `CachedFeed` that will cause a database error. I'll just update the `CachedFeed` that was created a fraction of a second earlier. (Or, if it somehow turns out that the preexisting `CachedFeed` was generated earlier, I'll throw away my work and use it instead.)


model/cachedfeed.py
```diff
@@ -73,120 +75,233 @@ class CachedFeed(Base):
...
-            type=type,
-            facets=facets_key,
-            pagination=pagination_key)
-
-        if force_refresh is True:
-            # No matter what, we've been directed to treat this
-            # cached feed as stale.
-            return feed, False
-
-        if max_age is lane.CACHE_FOREVER:
```
leonardr: Most of the code in this method was refactored, but this bit was removed for simplicity. We no longer need to be afraid of a grouped feed that's so expensive we just can't generate it and we have to fall back to generating a paginated feed. It's not good to generate a grouped feed while an HTTP request is happening, and in normal situations we won't do that, but it's not the end of the world -- it just means one request is slow.

model/cachedfeed.py
```diff
@@ -73,120 +75,233 @@ class CachedFeed(Base):
...
+        should_refresh = cls._should_refresh(feed_obj, max_age)
+        if not should_refresh:
+            # This is a cache hit. We found a matching CachedFeed that
+            # had fresh content.
+            return feed_obj
+
+        # This is a cache miss. We need to generate a new feed.
+        feed_data = unicode(refresher_method())
+        generation_time = datetime.datetime.utcnow()
+
```
leonardr: Here's the new code, below.

opds.py
```diff
@@ -578,163 +578,168 @@ def detailed_author(cls, contributor):
 class AcquisitionFeed(OPDSFeed):
 
     FACET_REL = "http://opds-spec.org/facet"
-    NO_CACHE = object()
+    # TODO: Is this still used?
```
leonardr: I think it's not, but I'll check when I do the circulation branch.

lane.py
```diff
@@ -105,6 +105,12 @@ class BaseFacets(FacetConstants):
     This is intended solely for use as a base class.
     """
 
+    # If the use of a certain faceting object has implications for the
+    # type of feed (the way FeaturedFacets always implies a 'groups' feed),
+    # set the type of feed here. This will override any CACHED_FEED_TYPE
+    # associated with the WorkList.
+    CACHED_FEED_TYPE = None
+
```
leonardr: Making sure that CACHED_FEED_TYPE is always present for WorkLists and Facets (even if the value is None) lets me take `cache_type` out of the `CachedFeed.fetch` method signature. We can always figure out the feed type by looking at the WorkList and the Facets.

opds.py
```diff
@@ -578,163 +578,168 @@ def detailed_author(cls, contributor):
...
 
         :param facets: A GroupsFacet object.
 
-        :return: CachedFeed (if use_cache is True) or unicode
+        :return: A Unicode string containing a (potentially cached) OPDS feed.
         """
-        works_and_lanes = None
-        if lane.children:
-            # Since this lane has children, it's reasonable to at least
-            # try to create a groups feed for it.
```
leonardr: I'm pretty sure it's reasonable to do this whether or not the lane has children, but I'm not sure what happens when the lane doesn't have children.

opds.py
```diff
@@ -578,78 +578,61 @@ def detailed_author(cls, contributor):
...
-                if usable:
-                    return cached.content
-            works_and_lanes = lane.groups(
-                _db=_db, facets=facets, search_engine=search_engine,
-                debug=search_debug
-            )
+        annotator = cls._make_annotator(annotator)
+        facets = facets or FeaturedFacets.default(worklist.get_library(_db))
 
-        if not works_and_lanes:
```
leonardr: Removing this code makes it the caller's responsibility to decide whether to call page() or groups(), and I think that's where the responsibility should lie.

lane.py
```diff
@@ -2454,6 +2483,22 @@ def uses_customlists(self):
...
+        internally.
+
+        :param type: The type of feed.
+        """
+        if type == CachedFeed.GROUPS_TYPE:
+            # Generating grouped feeds on the fly for Lanes is not incredibly
+            # expensive, but it's slow enough that we prefer to regenerate
+            # them in the background (using force_refresh=True) rather
+            # than while someone is waiting for an HTTP response.
+            return CachedFeed.CACHE_FOREVER
```
EdwinGuzman: If grouped feeds will be generated in the background, then why return "CACHE_FOREVER"? Or is that used along with force_refresh=True?

leonardr: CACHE_FOREVER is how `_should_refresh` knows not to say "yes, we should refresh", even if the thing in the cache is really really old.

opds.py
EdwinGuzman: But doesn't the `refresh` method here call the method to generate a group feed? Or is that different?

leonardr: No, this method defines the refresh method, which calls `_generate_groups`, an internal method.

EdwinGuzman: Yeah, this makes sense now. I thought I was looking inside the main `fetch` method but see it's the `groups` method.

tests/models/test_cachedfeed.py
```diff
@@ -17,162 +18,540 @@
...
+
+            return "Non-weird content."
+        result3 = m(
+            self._db, wl, facets, pagination, content_cleared_in_background, 0
+        )
+        now = datetime.datetime.utcnow()
+
+        # Again, a brand new CachedFeed.
+        assert result3 != result2
+        assert result3 != result
```
EdwinGuzman: The older CacheFeeds get deleted in the background eventually, right? Where is that happening?

tests/models/test_cachedfeed.py
```diff
@@ -17,162 +18,540 @@
...
+        eq_(u"facets query string", keys.facets_key)
+        eq_(u"pagination query string", keys.pagination_key)
+
+        # Now we can check that feed_type was obtained by passing
+        # `worklist` and `facets` into MockCachedFeed.feed_type.
+        eq_("mock type", keys.feed_type)
+        eq_((lane, MockFacets), MockCachedFeed.feed_type_called_with)
+
+        # When a WorkList is used instead of a Lane, keys.lane_id is None
+        # but keys.unique_id is set to worklist.unique_id.
```
EdwinGuzman: I think you meant `unique_key`.

tests/models/test_cachedfeed.py
leonardr: CachedFeedReaper in core/monitor.py removes all feeds older than 30 days.

leonardr:
> Looks good, but can we later and in-person go into how you debugged that it was the race condition causing the problem?

We can talk in person for sure once I'm well, but the quick version is:

I went to the logs (which we send to nypl.loggly.com) and found one of the 500 errors sent by the web app.

I used Loggly's "view surrounding events" feature to find the nearby stack trace of the IntegrityError that led to the 500 error.

Now that I'm looking at it again, it's clear that the IntegrityError mentioned in the log is for the `representations` table, not the `cachedfeeds` table. But both are real problems; this branch just fixes the problem that would otherwise happen later on.

PR #1160, opened 2020-02-19
Make sure works are removed from ES

jonathangreen:
When works are reaped by the DB reaper script make sure they are also reaped from elastic-search so they don't get out of sync. 

This came out of a problem LYRASIS was seeing where lanes were not displaying a full list of works. This turned out to be because ES was returning works that no longer existed in the database. 

@leonardr

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -859,6 +859,14 @@ class WorkReaper(ReaperMonitor):
     """
     MODEL_CLASS = Work
 
+    def __init__(self, *args, **kwargs):
+        from external_search import ExternalSearchIndex
+        search_index_client = kwargs.pop('search_index_client', None)
+        super(WorkReaper, self).__init__(*args, **kwargs)
+        self.search_index_client = (
+            search_index_client or ExternalSearchIndex(self._db)
+        )
+
     def query(self):
         return self._db.query(Work).outerjoin(Work.license_pools).filter(
             LicensePool.id==None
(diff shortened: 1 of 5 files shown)


monitor.py
```diff
@@ -859,11 +859,24 @@ class WorkReaper(ReaperMonitor):
...
+            search_index_client or ExternalSearchIndex(self._db)
+        )
+
     def query(self):
         return self._db.query(Work).outerjoin(Work.license_pools).filter(
             LicensePool.id==None
         )
 
+    def delete(self, work):
+        """Delete work from elasicsearch and database."""
```
leonardr: Fix this typo only if you're making another change to this branch.

jonathangreen:
@leonardr here is a push that addresses your feedback.

monitor.py
```diff
@@ -859,23 +859,14 @@ class WorkReaper(ReaperMonitor):
...
     MODEL_CLASS = Work
 
-    def __init__(self, *args, **kwargs):
-        from external_search import ExternalSearchIndex
-        search_index_client = kwargs.pop('search_index_client', None)
-        super(WorkReaper, self).__init__(*args, **kwargs)
-        self.search_index_client = (
-            search_index_client or ExternalSearchIndex(self._db)
-        )
-
```
leonardr: Removing this code will work but, my experience is that instantiating an `ExternalSearchIndex` is pretty expensive. Please reinstate this; otherwise the reaper will be really slow.

jonathangreen: This is added back in now.

PR #1161, opened 2020-02-20
Overdrive test

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2585.  This may be more of a band-aid than a longterm solution (I'm still not sure why the kwargs for this test can't be interpolated; everything looks fine!!), but it does work, so if this is time-sensitive...


leonardr:
I put the proposed fix in https://github.com/NYPL-Simplified/server_core/pull/1162, and I'm okay with closing this PR after that one has been merged.

adriana-alter:
Sounds good; sorry I didn't get to this!

PR #1162, opened 2020-02-24
Allow SearchFacets to take a sort order field so search results can be sorted by title or author

leonardr:
This is a support branch for https://jira.nypl.org/browse/SIMPLY-2538 and https://jira.nypl.org/browse/SIMPLY-2539. It also includes a small fix for https://jira.nypl.org/browse/SIMPLY-2585.

This branch makes SearchFacets a subclass of the normal Facets object. All of the features of Facets are available--you can search for only available books, or only popular books. But the feature we're really interested right now in is the ability to _sort_ search results by title or author.

Naively sorting search results by title creates a situation where your alphabetically-sorted search results include a ton of terrible results that would normally be at the bottom. So this branch also starts using the Elasticsearch `min_score` feature. If `Filter.min_score` is set, then only works that score higher than that number are included in the search results. The `SearchFacets` class knows how to set appropriate values for `min_score`, as well as all the functionality it inherits from `Facets` -- sorting by fields like title or author, showing all books vs. only currently available books, etc.

You can customize all this stuff using the same HTTP query string parameters as `Facets`, and you can also specify a custom `min_score` using the `min_score` query string parameter. So a URL might look like `http://localhost:6500/search?q=cozy+mystery&order=author&min_score=1000`

While testing, I noticed that a good value for `min_score` depends on what type of search you're running, so I think Adriana will need to add some kind of configuration control on the client side. Maybe a slider that runs from 'more results' (min_score=0) to "fewer results" (min_score=2000).

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -354,9 +354,7 @@ class Facets(FacetsWithEntryPoint):
         return config.default_facet(facet_group_name)
 
     @classmethod
-    def from_request(cls, library, config, get_argument, get_header, worklist,
-                     default_entrypoint=None, **extra):
-        """Load a faceting object from an HTTP request."""
+    def _values_from_request(cls, config, get_argument, get_header):
         g = Facets.ORDER_FACET_GROUP_NAME
         order = get_argument(g, cls.default_facet(config, g))
         order_facets = cls.available_facets(config, g)
@@ -365,7 +363,6 @@ class Facets(FacetsWithEntryPoint):
                 _("I don't know how to order a feed by '%(order)s'", order=order),
                 400
             )
-        extra['order'] = order
 
         g = Facets.AVAILABILITY_FACET_GROUP_NAME
         availability = get_argument(g, cls.default_facet(config, g))
@@ -375,7 +372,6 @@ class Facets(FacetsWithEntryPoint):
                 _("I don't understand the availability term '%(availability)s'", availability=availability),
                 400
             )
-        extra['availability'] = availability
 
         g = Facets.COLLECTION_FACET_GROUP_NAME
         collection = get_argument(g, cls.default_facet(config, g))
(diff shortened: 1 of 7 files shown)


lane.py
```diff
@@ -767,16 +779,40 @@ def scoring_functions(self, filter):
...
+class SearchFacets(Facets):
     """A Facets object designed to filter search results.
 
     Most search result filtering is handled by WorkList, but this
     allows someone to, e.g., search a multi-lingual WorkList in their
     preferred language.
     """
 
-    def __init__(self, entrypoint=None, media=None, languages=None, **kwargs):
-        super(SearchFacets, self).__init__(entrypoint, **kwargs)
```
leonardr: The big change I had to make is changing the `SearchFacets` constructor so that it's compatible with the `Facets` constructor. Previously you could just pass in an EntryPoint; now the first argument is supposed to be a Library.

I don't think this is going to be a huge deal. Outside of tests, the only time we should create a SearchFacets object is in `facets_from_request`.

EdwinGuzman: But now it's just `**kwargs` that get passed for `**constructor_kwargs`, right?

lane.py
```diff
@@ -767,16 +779,43 @@ def scoring_functions(self, filter):
...
+        kwargs.setdefault('library', None)
+        kwargs.setdefault('collection', None)
+        kwargs.setdefault('availability', None)
+        order = kwargs.setdefault('order', None)
+
+        if order:
+            default_min_score = self.DEFAULT_MIN_SCORE
+        else:
+            default_min_score = None
+        self.min_score = kwargs.pop('min_score', default_min_score)
```
EdwinGuzman: `pop` in this case because `Facets` doesn't know about `min_score`, right?

lane.py
leonardr: Yes, we add library, collection, availability, and order to the kwargs (all of which are required by the Facets constructor) if there is no current value, and then pass them on into the Facets constructor.

adriana-alter:
Still going through the code, but an issue as I try to hook this up to the front end: ElasticSearch is throwing an error like so: `I don't understand the availability term 'all'`.  I'm not sure where this is coming from; the interface isn't sending any information about availability...

Also, the ElasticSearch docs say that the official name for relevance is "_score", but passing that in from the front end throws an `I don't know how to order a feed by '_score'` error.  Off the top of your head, is there a different attribute I should be using?

PR #1165, opened 2020-02-26
Make it possible to cache a Representation under some URL other than the one used to make the HTTP request

leonardr:
This is core work for https://jira.nypl.org/browse/SIMPLY-2532.

Basically, our NoveList integration needs to be able to make a request to `http://novelist/book?credentials=abc`, but store the results under `http://novelist/book`. This lets us share cached NoveList data among different libraries on the same system, even though different libraries will have different NoveList credentials.

The existing solution modifies `Representation.url` after the database row has been created. The problem is that `Representation.url` isn't just a random field; it's part of a unique index. If multiple patrons are requesting the same NoveList data at the same time, this causes a race condition which makes spurious 500 errors show up in the logs. Multiple threads all believe they have a brand-new database row, then they all try to set its `Representation.url` to the same value and only one of them succeeds.

This solution allows the caller to pass in a hook function that converts "the URL used in the HTTP request" to "the URL stored in Representation.url". This lets us know the value we want to use `Representation.url` ahead of time, allowing us to look up an existing cached row if it exists. If there is no matching row, then `Representation.url` has the correct value right from the start, greatly reducing the risk of a race condition.

### DIFF
--- a/external_search.py
+++ b/external_search.py
@@ -2354,7 +2354,7 @@ class Filter(SearchBase):
         # At this point we know we have a specific list of audiences.
         # We're either going to return that list as-is, or we'll
         # return that list plus ALL_AGES.
-        with_all_ages = as_is + [Classifier.AUDIENCE_ALL_AGES]
+        with_all_ages = list(as_is) + [Classifier.AUDIENCE_ALL_AGES]
 
         if Classifier.AUDIENCE_ALL_AGES in as_is:
             # ALL_AGES is explicitly included.
--- a/model/resource.py
+++ b/model/resource.py
@@ -673,17 +673,59 @@ class Representation(Base, MediaTypes):
     def get(cls, _db, url, do_get=None, extra_request_headers=None,
             accept=None, max_age=None, pause_before=0, allow_redirects=True,
             presumed_media_type=None, debug=True, response_reviewer=None,
-            exception_handler=None):
+            exception_handler=None, url_normalizer=None):
         """Retrieve a representation from the cache if possible.
         If not possible, retrieve it from the web and store it in the
         cache.
+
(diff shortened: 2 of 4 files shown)


external_search.py
```diff
@@ -2354,7 +2354,7 @@ def audiences(self):
         # At this point we know we have a specific list of audiences.
         # We're either going to return that list as-is, or we'll
         # return that list plus ALL_AGES.
-        with_all_ages = as_is + [Classifier.AUDIENCE_ALL_AGES]
+        with_all_ages = list(as_is) + [Classifier.AUDIENCE_ALL_AGES]
```
leonardr: This is an unrelated change I saw while testing the NoveList integration. I don't know where this happens, but at some point `Filter.audiences` is set to a `set`, and appending a list to it causes an error. But there's no reason why a `set` shouldn't be allowed here.

model/resource.py
```diff
@@ -871,16 +920,19 @@ def record_exception(cls, representation, exception, traceback):
...
+
         def do_post(url, headers, **kwargs):
             kwargs.update({'data' : data})
-            return cls.simple_http_post(url, headers, **kwargs)
+            return original_do_get(url, headers, **kwargs)
 
         return cls.get(
             _db, url, do_get=do_post, max_age=max_age,
-            response_reviewer=response_reviewer
+            response_reviewer=response_reviewer, **kwargs
```
leonardr: I needed to make this change to make NoveListAPI.lookup testable. I need to be able to specify a custom `do_get` for the test, even though `post` _also_ creates a custom `do_get`. I do this by having `post` _wrap_ any custom `do_get` it receives in a new, custom function.

PR #1166, opened 2020-02-26
Change s3 storage settings so Access Key and Secret Key are not required

jonathangreen:
S3 recommended practice is that IAM roles be used to give EC2 instances permission to modify S3 buckets instead of configuring a static Access Key and Secret Key. The boto3 library we are using for s3 access is [able to authenticate](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#configuring-credentials) using a IAM role, but only if the Access Key and Secret Key are not passed into the library constructor. 

This makes access key and secret key optional and does not pass them into Boto3 if they are not set so that other authentication mechanisms can be used for s3 buckets. 

Tagging: @leonardr

### DIFF
--- a/s3.py
+++ b/s3.py
@@ -93,8 +93,11 @@ class S3Uploader(MirrorUploader):
     }
 
     SETTINGS = [
-        { "key": ExternalIntegration.USERNAME, "label": _("Access Key"), "required": True },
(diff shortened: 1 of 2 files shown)


leonardr:
In conversation with Jonathan I learned that LYRASIS configures the IAM role on the Amazon EC2 instances to which they deploy.

leonardr (approved):

PR #1170, opened 2020-03-27
Cache control infrastructure

leonardr:
This is the core branch for https://jira.nypl.org/browse/SIMPLY-2633. It introduces a subclass of the Flask `Response` class, which contains some extra logic making it easy to calculate values for the `Cache-Control` and `Expires` headers. There are also convenience subclasses `OPDSFeedResponse` and `OPDSEntryResponse` which offer additional defaults for OPDS feeds and entries.

Here's what you can do with our `Response` subclasses that you would have to do manually with a normal Flask `Response`:

* Set a `max_age` number which controls how long the client (or an intermediary such as a CDN) can cache 
* Set a `private` flag which controls whether or not intermediaries such as web proxies can cache this document. 
* Manually set the Content-Type header to the appropriate subtype of the Atom media type.

In core, the most important use of `Response` is `CachedFeed.fetch`, which is the bit of code that figures out how long an OPDS feed needs to be cached _internally_. As of this branch, OPDS feeds are cached on the client side (via the `Cache-Control` header) for the same amount of time and using the same logic. To pass the "how long should this be cached?" knowledge out of `fetch`, I changed `fetch` to return a `Response` object instead of the `CachedFeed` object itself. The old behavior is still available (you pass in `raw=True`), but it's now only used in tests.

With one exception, all methods in `opds.py` that used to return OPDS feeds or entries now return `Response` objects instead. This is a bigger change than absolutely necessary--we could get most of the performance benefits just by changing the methods that call `CachedFeed.fetch` but it lets me simplify the circulation code quite a bit and it avoids bugs caused by forgetting which methods return `Response` and which don't. This also let me get rid of the `feed_response` and `entry_response` utility functions, which were used in circulation to turn OPDS feeds into Flask `Response` objects. Now the feeds are `Response` objects from the beginning.

### DIFF
--- a/app_server.py
+++ b/app_server.py
@@ -13,7 +13,6 @@ from flask import url_for, make_response
 from flask_babel import lazy_gettext as _
 from io import BytesIO
 from util.flask_util import problem
-from util.opds_writer import OPDSFeed
 from util.problem_detail import ProblemDetail
 import traceback
 import logging
--- a/app_server.py
+++ b/app_server.py
@@ -21,7 +21,7 @@ from opds import (
     AcquisitionFeed,
     LookupAcquisitionFeed,
 )
-from util.flask_util import OPDSFeedResponselike
+from util.flask_util import OPDSFeedResponse
 from util.opds_writer import (
     OPDSFeed,
     OPDSMessage,
@@ -320,7 +320,7 @@ class URNLookupController(object):
             self._db, "Lookup results", this_url, handler.works, annotator,
             precomposed_entries=handler.precomposed_entries,
         )
-        return OPDSFeedResponselike(opds_feed).response
+        return OPDSFeedResponse(opds_feed)
 
     def process_urns(self, urns, **process_urn_kwargs):
         """Process a number of URNs by instantiating a URNLookupHandler
@@ -354,7 +354,7 @@ class URNLookupController(object):
             self._db, urn, this_url, works, annotator,
             precomposed_entries=handler.precomposed_entries
         )
-        return OPDSFeedResponselike(opds_feed).response
+        return OPDSFeedResponse(opds_feed)
 
 
 class URNLookupHandler(object):
--- a/app_server.py
+++ b/app_server.py
@@ -66,41 +66,6 @@ def load_lending_policy(policy):
                         audience)
     return policy
 
-def feed_response(feed, acquisition=True, cache_for=None):
-    """Create a response for an OPDS feed.
-
-    For a feed created through WorkList, you shouldn't need this, because
-    WorkList methods generally return a flask_utils.Responselike.
-    """
-    if acquisition:
-        content_type = OPDSFeed.ACQUISITION_FEED_TYPE
-    else:
-        content_type = OPDSFeed.NAVIGATION_FEED_TYPE
-    return _make_response(feed, content_type, cache_for)
-
-def entry_response(entry, cache_for=None):
-    content_type = OPDSFeed.ENTRY_TYPE
-    return _make_response(entry, content_type, cache_for)
-
-def _make_response(content, content_type, cache_for):
-    if isinstance(content, etree._Element):
-        content = etree.tostring(content)
-    elif not isinstance(content, (bytes, unicode)):
-        content = unicode(content)
-
-    if isinstance(cache_for, int):
-        # A CDN should hold on to the cached representation only half
-        # as long as the end-user.
-        client_cache = cache_for
-        cdn_cache = cache_for / 2
-        cache_control = "public, no-transform, max-age=%d, s-maxage=%d" % (
-            client_cache, cdn_cache)
-    else:
-        cache_control = "private, no-cache"
-
-    return make_response(content, 200, {"Content-Type": content_type,
-                                        "Cache-Control": cache_control})
-
 def load_facets_from_request(
         facet_config=None, worklist=None, base_class=Facets,
         base_class_constructor_kwargs=None, default_entrypoint=None
--- a/app_server.py
+++ b/app_server.py
@@ -69,8 +69,8 @@ def load_lending_policy(policy):
 def feed_response(feed, acquisition=True, cache_for=None):
     """Create a response for an OPDS feed.
 
-    NOTE: You probably want to use flask_utils.Responselike instead,
-    and it's possible that this can be deprecated.
+    For a feed created through WorkList, you shouldn't need this, because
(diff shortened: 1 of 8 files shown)


app_server.py
```diff
@@ -349,7 +320,7 @@ def work_lookup(self, annotator, route_name='lookup', **process_urn_kwargs):
             self._db, "Lookup results", this_url, handler.works, annotator,
             precomposed_entries=handler.precomposed_entries,
         )
-        return feed_response(opds_feed)
+        return OPDSFeedResponselike(opds_feed).response
```
leonardr: In a future branch we could change these to return a Responselike, but since this isn't the main focus of the branch I decided to go through ResponseLike to Response and leave the contract -- that you call these functions and get a Response -- unaltered.

model/cachedfeed.py
```diff
@@ -82,13 +83,15 @@ class CachedFeed(Base):
 
     @classmethod
     def fetch(cls, _db, worklist, facets, pagination, refresher_method,
-              max_age=None
+              max_age=None, raw=False, **response_kwargs
```
leonardr, line 86: A lot of these methods now take **response_kwargs. max_age is treated specially here, even though it ends up being passed in to the Response constructor, because it's _also_ used to determine the maximum age of a feed in the database cache.

util/flask_util.py
```diff
@@ -16,4 +22,125 @@ def problem(type, status, title, detail=None, instance=None, headers={}):
...
+        headers['Cache-Control'] = cache_control
+
+        return headers
+
+
+class OPDSFeedResponse(Response):
+    """A convenience specialization of Response for typical OPDS feeds."""
+    def __init__(self, response=None, status=None, headers=None, mimetype=None,
+                 content_type=None, direct_passthrough=False, max_age=None,
+                 private=None):
```
leonardr, line 128: This method is really verbose, compared to OPDSEntryResponse, because I wanted to make it really easy for anything that used to create a Response object to create an OPDSFeedResponse instead, whether it was using positional arguments or keyword arguments. This isn't as necessary for OPDSEntryResponse, which is only used in a couple places, so I wrote the simpler version for that class that requires keyword arguments.

opds.py
```diff
@@ -783,8 +790,13 @@ def from_query(cls, query, _db, feed_name, url, pagination, url_fn, annotator):
         creating an OPDS feed for a custom list and not cached.
 
         TODO: This is used by the circulation manager admin interface.
-        Investigating replacing the code that uses this so that it uses
-        the search index.
+        Investigate changing the code that uses this to use the search
+        index -- this is inefficient and creates an alternate code path
+        that may harbor bugs.
+
+        TODO: This cannot currently return OPDSFeedResponse because the
```
EdwinGuzman, line 797: What page in the admin is the feed generated from this method used?

leonardr: It looks like it's just CustomListsController.custom_list. The search engine can find all the books on a custom list with no problem. The issue might be that when you're adding books to a custom list, you don't have time to wait around for the search engine to be updated; you need the most up-to-date version of the list.

tests/test_opds.py
```diff
@@ -1102,12 +1107,20 @@ def test_groups_feed(self):
...
         )
-        parsed = feedparser.parse(cached_groups)
+
+        # The result is an OPDSFeedResponse object. The 'private'
+        # argument, unused by groups(), was passed along into the
+        # constructor.
+        assert isinstance(cached_groups, OPDSFeedResponse)
+        eq_(private, cached_groups.private)
+
+        parsed = feedparser.parse(cached_groups.data)
```
EdwinGuzman, line 1123: `.data` since `cached_groups` is the Response and not the content, right?

leonardr: Right.

tests/test_opds.py
```diff
@@ -1282,20 +1303,71 @@ def make_page():
...
+        # Verify the ability to convert an AcquisitionFeed object to an
+        # OPDSFeedResponse containing the feed.
+        feed = AcquisitionFeed(self._db, "feed title", [], TestAnnotator)
+        response = feed.as_response(max_age=101)
+        eq_(200, response.status)
+        assert isinstance(response, OPDSFeedResponse)
+        assert '<title>feed title</title>' in response.data
+        eq_(101, response.max_age)
+
+    def as_response(self):
```
EdwinGuzman, line 1349: Are these two method names correct? They're the same and don't start with `test_`.

tests/util/test_flask_util.py
```diff
@@ -0,0 +1,129 @@
...
+        eq_(True, response.direct_passthrough)
+
+        # Response.headers is tested in more detail below.
+        headers = response.headers
+        eq_("value", headers['Header'])
+        assert 'Cache-Control' in headers
+        assert 'Expires' in headers
+
+    def test_headers(self):
+        # First test cases where the response should not private and not cached.
```
EdwinGuzman, line 39: I think you meant `should be private and not cached`?

util/flask_util.py
```diff
@@ -16,4 +22,125 @@ def problem(type, status, title, detail=None, instance=None, headers={}):
...
+        # Don't modify the underlying dictionary; it came from somewhere else.
+        headers = dict(headers)
+
+        # Set Cache-Control based on max-age.
+        if self.private:
+            private = "private"
+        else:
+            private = "public"
+        if self.max_age and isinstance(self.max_age, int):
+            # A CDN should hold on to the cached representation only half
```
EdwinGuzman, line 101: Why is this the case?

tests/test_opds.py
leonardr: Yeah, these tests weren't running at all.

util/flask_util.py
leonardr: CDN cache times and client cache times are additive. Let's say the CDN cache time is 30 minutes and the client cache time is 30 minutes. When the client requests a document, it might get a document that's 29 minutes out of date. Then it holds on to the document for 30 minutes. At the end of this process, the client is using a document that's nearly an hour out of date, even though the cache time is only 30 minutes.

But the whole point of the CDN cache is to stop client traffic from hitting the origin server. So this problem is unavoidable if we want the CDN to be useful. It's a tradeoff we have to figure out how to make. Setting the CDN cache time to 1/2 of the client cache time is a rule of thumb that helps mitigate the problem while keeping the CDN useful. There's nothing special about the 1/2 rule, but I don't have any ideas for coming up with a better coefficient.

PR #1174, opened 2020-04-22
Facets filter worklist

tdilauro:
## JIRA Ticket

https://jira.nypl.org/browse/SIMPLY-2706

## What does this PR do?

Currently, SimplyE and other clients sometimes display works that aren't available even though "Available Now” is selected. This happens when the search index and database are out of sync and can lead to user confusion and, thus, poor UX.

This PR prevents propagation of <entry>s for these unavailable works to the client when the `availability` facet is now. It does so, by applying constraints corresponding to the active facets to the database query. Because these changes act more broadly on the currently active facets, it may ameliorate future search/database sync UX issues, as well.

## Change description

These changes
- factor the constraint-based facet filtering from `DatabaseBackedFacets.modify_database_query` to a new `Facets.modify_database_query` and
- propagate the active `Facets` object to `DatabaseBackedWorklist.works_from_database`, which uses it to filter the database query.

## How should this be tested?

Added test to ensure that availability faceting results in correct filtering. Most of the functionality is already tested in:
- `tests/test_lane.py:TestDatabaseBackedFacets.test_modify_database_query`

Existing tests have been updated:
- Add optional facet to some method signatures;
- Assign a more specific superclass to a mock; and
- Remove a no-longer applicable assertion.

I have performed manual live testing to demonstrate that unavailable items are no longer propagated.

Can also add tests to ensure faceting object is properly  propagated, if needed. 

## Interested parties

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1609,7 +1609,7 @@ class WorkList(object):
             query_string=None, filter=filter, pagination=pagination,
             debug=debug
         )
-        return self.works_for_hits(_db, hits)
+        return self.works_for_hits(_db, hits, facets=facets)
 
     def filter(self, _db, facets):
         """Helper method to instantiate a Filter object for this WorkList.
@@ -1634,7 +1634,7 @@ class WorkList(object):
         """
         return filter
 
-    def works_for_hits(self, _db, hits):
+    def works_for_hits(self, _db, hits, facets=None):
         """Convert a list of search results into Work objects.
 
         This works by calling works_for_resultsets() on a list
@@ -1646,10 +1646,10 @@ class WorkList(object):
             script fields), WorkSearchResult objects.
         """
 
-        [results] = self.works_for_resultsets(_db, [hits])
+        [results] = self.works_for_resultsets(_db, [hits], facets=facets)
         return results
 
-    def works_for_resultsets(self, _db, resultsets):
+    def works_for_resultsets(self, _db, resultsets, facets=None):
         """Convert a list of lists of Hit objects into a list
         of lists of Work objects.
         """
@@ -1683,11 +1683,14 @@ class WorkList(object):
         # DatabaseBackedWorkList that fetches those specific Works
         # while applying the general availability filters.
         #
+        # If facets were passed in, then they are used to further
+        # filter the list.
+        #
         # TODO: There's a lot of room for improvement here, but
         # performance isn't a big concern -- it's just ugly.
         wl = SpecificWorkList(work_ids)
         wl.initialize(self.get_library(_db))
-        qu = wl.works_from_database(_db)
+        qu = wl.works_from_database(_db, facets=facets)
         a = time.time()
         all_works = qu.all()
 
@@ -1909,7 +1912,7 @@ class WorkList(object):
             filter = Filter.from_worklist(_db, lane, overview_facets)
             queries.append((None, filter, pagination))
         resultsets = list(search_engine.query_works_multi(queries))
-        works = self.works_for_resultsets(_db, resultsets)
+        works = self.works_for_resultsets(_db, resultsets, facets=facets)
 
         for i, lane in enumerate(lanes):
             results = works[i]
@@ -1937,14 +1940,14 @@ class DatabaseBackedWorkList(WorkList):
         lanes can be generated through search engine queries.
 
         :param _db: A database connection.
-        :param facets: A DatabaseBackedFacets object which may put additional
+        :param facets: A Facets object which may place additional
            constraints on WorkList membership.
         :param pagination: A Pagination object indicating which part of
            the WorkList the caller is looking at.
         :param kwargs: Ignored -- only included for compatibility with works().
         :return: A Query.
         """
-        if facets is not None and not isinstance(facets, DatabaseBackedFacets):
+        if facets is not None and not isinstance(facets, Facets):
             raise ValueError(
                 "Incompatible faceting object for DatabaseBackedWorkList: %r" %
                 facets
--- a/lane.py
+++ b/lane.py
@@ -582,6 +582,43 @@ class Facets(FacetsWithEntryPoint):
             else:
                 logging.error("Unrecognized sort order: %s", self.order)
 
+    def modify_database_query(self, _db, qu):
+        """Restrict a query against Work+LicensePool+Edition so that it
+        matches only works that fit the criteria of this Faceting object.
+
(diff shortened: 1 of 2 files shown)


tdilauro:
Pinging @leonardr

lane.py
```diff
@@ -1923,18 +1940,19 @@ def works_from_database(self, _db, facets=None, pagination=None, **kwargs):
...
            the WorkList the caller is looking at.
         :param kwargs: Ignored -- only included for compatibility with works().
         :return: A Query.
         """
-        if facets is not None and not isinstance(facets, DatabaseBackedFacets):
-            raise ValueError(
-                "Incompatible faceting object for DatabaseBackedWorkList: %r" %
+        if facets is not None and not isinstance(facets, Facets):
+            logging.warning(
+                "Ignoring incompatible faceting object for DatabaseBackedWorkList: %r",
```
leonardr, line 1952: As I recall, you changed this from an exception to logging so that `Facets` could be passed in here. Now that `Facets` is explicitly handled, can we change back to raising an exception when (let's say) a `BaseFacets` is used? I think that indicates a programming error and an exception is appropriate.

lane.py
```diff
@@ -1923,18 +1940,19 @@ def works_from_database(self, _db, facets=None, pagination=None, **kwargs):
         lanes can be generated through search engine queries.
 
         :param _db: A database connection.
-        :param facets: A DatabaseBackedFacets object which may put additional
-           constraints on WorkList membership.
+        :param facets: A Facets object which may place additional
+           constraints on WorkList membership. Other objects are ignored.
```
leonardr, line 1944: I see what you mean by this based on the code, but it would be clearer as "Faceting objects are ignored unless they are an instance of Facets or one of its subclasses."

leonardr: Or, if you make the other changes, add a `:raises ValueError` to describe this situation rather than changing the description of `facets`.

lane.py
tdilauro: @leonardr,

We can end up with non-`Facets`-based subclasses of `BaseFacets` here (e.g., `FeatureFacets` and `MARCExporterFacets`). The latter comes through a code path that I know passes `Facets` objects, so not propagating `facets` is not a good option there.

For the former, we could just not propagate `facets` from `WorkList._featured_works_with_lanes`, but the word "presumably" in [this line](https://github.com/NYPL-Simplified/server_core/blob/master/lane.py#L1864) in the docstring gives me pause:
```
:param facets: A faceting object, presumably a FeaturedFacets
```
Anecdotally, I have not seen anything other than `FeatureFacets` coming through this method, but I'm not 100% sure.

These things together are why I thought it would be better to do the filtering once in `DatabaseBackedWorkList.works_from_database` and log the issue. I can also log a traceback here, which would make it possible to see who's doing what at some point in the future.

leonardr: We can also add a `BaseFacets.modify_database_query` implementation that's a no-op -- just returns `qu`. Then we wouldn't have to police what kind of faceting object you pass in. If it's a faceting object that doesn't know how to modify a database query, the query just doesn't get modified.

tdilauro: The no-op `BaseFacets.modify_database_query` is the way to go, I think.

leonardr:
OK, great. My only remaining concern is: some faceting objects work in a search context but not a database context. It's a programming error to pass that kind of faceting object into `works_from_database` -- your faceting object won't do what you want it to do -- which is why I previously checked for that and raised an exception.

Now the exception is gone, so if you pass a `FeaturedFacets` into `works_from_database`, you won't get the results you're looking for -- you should have used `works` -- and you won't get an exception either.

I guess we can resolve this by making those facets raise `NotImplementedError` from their `works_from_database`. That will move the knowledge about which faceting classes support what into the faceting classes themselves.

Basically, I'm now coming around to the position that `BaseFacets.works_from_database` should raise `NotImplementedError`, rather than being a no-op. That way we can get rid of the ad hoc `ValueError` but spot any future programming errors.

The real change made by this branch will be that `Facets` _partially_ implements `modify_database_query` -- you get the filtering functionality but not the sorting -- where it used to be not usable at all.

tdilauro:
As I mentioned in a [previous comment](https://github.com/NYPL-Simplified/server_core/pull/1174#discussion_r414131071):
> We can end up with non-Facets-based subclasses of BaseFacets here (e.g., FeaturedFacets and MARCExporterFacets). The latter comes through a code path that I know passes Facets objects, so not propagating facets is not a good option there.

For example, the table below illustrates one of the common pathways I found by running the test suite. The headings are the type of faceting object and the row below contains the salient entries from the stack.

<table>
<tr>
  <th>Facets</th>
  <th>FeaturedFacets</th>
  <th>MARCExporterFacets</th>
</tr>
<tr>
  <td><pre>
works_for_resultsets, lane.py:1701
works_for_hits, lane.py:1657
works, lane.py:1620
pages, test_external_search.py:1006
test_query_works, test_external_search.py:1022
...
<module>, nosetests:8
  </pre></td>
  <td><pre>
works_for_resultsets, lane.py:1701
works_for_hits, lane.py:1657
works, lane.py:1620
works, test_external_search.py:1888
test_run, test_external_search.py:1904
...
<module>, nosetests:8
  </pre></td>
  <td><pre>
works_for_resultsets, lane.py:1701
works_for_hits, lane.py:1657
works, lane.py:1620
records, marc.py:673
test_records, test_marc.py:522
...
<module>, nosetests:8
  </pre></td>
</tr>
</table>

I don't think we want to throw exceptions for objects that are not`Facets` (or a sublclass thereof) because this will result in poor UX (e.g., CM pops up a modal error that prevents the user from completing the task). So, if we don't let the `BaseFacets` or one of its subclasses handle `works_from_database`, then we will have to stop propagation of the faceting object somewhere further down the stack.

In short, I think no-oping `works_from_database` could just be behavior, rather than representing a programming error, and would allow centrally managing that behavior over time.

leonardr:
OK, I think I see it now. `works_for_resultsets` always called `works_from_database`, because that's how we get the actual Work ORM objects. Previously this wasn't a problem because we didn't propagate the `Facets` object from `works_for_resultsets` into `works_from_database`. An incompatible `Facets` object in `works_from_database` could only be a programming error.

But now we _do_ propagate the `Facets` object, so incompatible `Facets` objects will show up in `works_from_database` all the time. Changing it to a no-op is the best solution.

PR #1175, opened 2020-05-05
Allow faceting objects a say in determining the maximum cache age for a feed

leonardr:
This branch is part of the code to address https://jira.nypl.org/browse/SIMPLY-2582. It gives the faceting object a say in whether and for how long a CachedFeed should be cached.

If different parts of the system have different opinions on max_age, the order of precedence is:

* The `max_age` argument to `CachedFeed.fetch`
* The faceting object (new)
* The `WorkList`

With this in place, it's possible to bypass the cache on any OPDS feed controller that calls `FacetsWithEntryPoint.from_request`  by adding max_age=0 to the query string. That's the new feature in this branch, and I'll be taking advantage of it in an upcoming circulation branch.

When this feature is active, OPDS feeds are generated without consulting the database (which we normally do to see if there's already a usable feed) and they're served without writing them to the database (which we normally do to save time in the future). The second part is because the OPDS feeds generated while ignoring the cache might look different from feeds generated some other way. In particular, my circulation branch ignores CDNs when generating no-cache OPDS feeds, since CDNs add an extra layer of caching and the whole point of SIMPLY-2582 is to give people a cache-free look at the data.

`max_age=0` in a query string gets transformed into the constant IGNORE_CACHE in `Facets.max_cache_age`. ("Cache for zero seconds" and "ignore the cache entirely" mean slightly different things, but it's pretty close -- I don't know if allowing a special constant for max_age in the query string is worth the additional complexity.)

Once 'don't use the cache' gets into the system, it's propagated through any links generated using the faceting object -- see the implementations of `navigate` and `items`.

This only affects tests, but `CachedFeed.fetch(raw=True, max_age=IGNORE_CACHE)` is a contradiction in terms. If you're ignoring the database cache altogether, you can't retrieve or create a `CachedFeed` object. So `CachedFeed.fetch(max_age=IGNORE_CACHE)` ignores the `raw` argument and always returns an `OPDSFeedResponse`.


lane.py
```diff
@@ -269,9 +289,17 @@ def _from_request(
         if isinstance(entrypoint, ProblemDetail):
             return entrypoint
         entrypoint, is_default = entrypoint
+
+        max_cache_age = get_argument(
+            Facets.MAX_CACHE_AGE_NAME, None
+        )
+        max_cache_age = cls.load_max_cache_age(max_cache_age)
+        if isinstance(max_cache_age, ProblemDetail):
```
EdwinGuzman, line 297: The `load_max_cache_age` method can return `None`, but is that an instance of a `ProblemDetail`?

EdwinGuzman: I see how the tests make `load_max_cache_age` return a ProblemDetail, but I don't see the actual function returning a ProblemDetail, just for a value or None.

leonardr: No, in the current version `load_max_cache_age` will never return `ProblemDetail`. But when calling code that loads data from the HTTP request I find it useful to consistently use this pattern:

```
loaded = load_data()
if isinstance(loaded, ProblemDetail):
   return loaded
````

That way if I change `load_data` later so it _does_ return a ProblemDetail, the caller is already ready and tested. In this case we might change `load_max_cache_age` to return a ProblemDetail (rather than None) if you pass in something that's not a number.

PR #1177, opened 2020-05-27
Checking collection for the title during opds import

swandog30:
Ticket: https://jira.nypl.org/projects/SIMPLY/issues/SIMPLY-2780

This branch will check the licensepools table during an opds import to see if the title that is being imported is already known to the system, but not in the collection that is being imported.  It uses the item identifier, collection id, and data source name in the licensepool lookupquery.  I'm not sure if data source name is needed here - or if this query will cover all situations, but this fixed the issue I was having when importing two collections that had significant title overlap.

### DIFF
--- a/model/licensing.py
+++ b/model/licensing.py
@@ -203,6 +203,18 @@ class LicensePool(Base):
         )
 
     @classmethod
+    def exists(self, _db, identifier, collection_id, data_source_name):
+        """Determine if we have a license pool for the given id, collection, and datasource"""
+        from datasource import DataSource
+
+        q = _db.query(LicensePool).join(DataSource).filter(LicensePool.collection_id == collection_id,
+                LicensePool.identifier_id == identifier.id,
+                DataSource.name == data_source_name,
+                LicensePool.data_source_id == DataSource.id)
+
+        return _db.query(q.exists()).scalar()
+
+    @classmethod
     def for_foreign_id(self, _db, data_source, foreign_id_type, foreign_id,
                        rights_status=None, collection=None, autocreate=True):
         """Find or create a LicensePool for the given foreign ID."""
--- a/opds_import.py
+++ b/opds_import.py
@@ -1911,6 +1911,17 @@ class OPDSImportMonitor(CollectionMonitor, HasSelfTests):
                 )
                 return True
 
+            # The imported title might be known already but not in the collection we are importing,
+            # check for that condition
+            has_pool = LicensePool.exists(self._db, identifier, self.collection_id, self.importer.data_source_name)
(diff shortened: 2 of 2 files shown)


model/licensing.py
```diff
@@ -202,6 +202,18 @@ def __repr__(self):
             self.licenses_reserved, self.patrons_in_hold_queue
         )
 
+    @classmethod
+    def exists(self, _db, identifier, collection_id, data_source_name):
```
leonardr, line 206: Defining a data model method to do a certain kind of lookup is fine, but the method should return the object looked up rather than a boolean whether or not the lookup succeeded. That way a caller has the looked-up object in case they want to do something with it.

It's also going to cause confusion to have a data model method that takes a combination of data model objects (Identifier), IDs (collection_id), and names (data_source_name).

Here's an alternate implementation that takes an Identifier, a Collection, and a DataSource -- the caller can look up those objects as necessary.

```
def for_collection(self, _db, identifier, collection, data_source):
    return get_one(_db, LicensePool, identifier=identifier, collection=collection, data_source=data_source)
```

And at this point it's a small enough amount of code that you might not need to even add a data model method.

opds_import.py
```diff
@@ -1911,6 +1911,17 @@ def identifier_needs_import(self, identifier, last_updated_remote):
...
+        # check for that condition
+        has_pool = LicensePool.exists(self._db, identifier, self.collection_id, self.importer.data_source_name)
+        if not has_pool:
+            # The collection doesnt have this title. Importing it will create a LicensePool
+            self.log.info(
+                "Counting %s as new because it is not in the collection.",
+                identifier
+            )
+            return True
+
```
leonardr, line 1924: This makes sense; in fact I think this check should go first. If there's no LicensePool for this book in the collection then there's no need to check for an import record or an import failure -- if it happened, it clearly didn't work.

leonardr (changes requested):
I think the [test failures](https://travis-ci.com/github/NYPL-Simplified/server_core/builds/168452482) happen because you catch some of the test setups cheating: they think that if they add coverage records for all the books in a feed, the OPDS importer will think there's no new content in the feed. That's exactly the case your code catches. Those tests need to be changed to also add LicensePools for the books they "imported".

We don't have a unit test of `identifier_needs_import`, but now would be a good time to write one -- it would be simpler than a test that says you can't import a book from an OPDS feed if there's already a LicensePool for that book in the Collection you're importing into.

swandog30:
Closing this since I unfortunately let this languish.  I'll test against a newer release and resubmit if needed and at an appropriate time.

PR #1178, opened 2020-06-12
Fix relevance sort error message

tdilauro:
## Description

This branch changes `Facets.modify_search_filter` so that it treats a `Facets.order` of "relevant" the same way it already treats a Falsy one (i.e., don't set the filter order and don't log the error
```
Unrecognized sort order: relevance
```

## Motivation and Context

JIRA: https://jira.nypl.org/browse/SIMPLY-2821

Relevance score descending is the default ElasticSearch sort order, but does not map to an ElasticSearch field name. The absence of a search filter order indicates this default, but Facets.modify_search_filter only tests for valid explicit sort orders from FacetConstants.SORT_ORDER_TO_ELASTICSEARCH_FIELD_NAME. If it doesn't find one of the valid ones, it complains, leaves the order unset, and continues. When "relevance" is specified as the order, we get the right behavior, but the irritating error message along with it).

## How Has This Been Tested?

Added new test class to ensure that error message is emitted only when an invalid sort order is specified.

## Checklist:

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -396,6 +396,9 @@ class Facets(FacetsWithEntryPoint):
     Despite the generic name, this is only used in 'page' type OPDS
     feeds that list all the works in some WorkList.
     """
+
+    ORDER_BY_RELEVANCE = "relevance"
+
     @classmethod
     def default(cls, library, collection=None, availability=None, order=None,
                 entrypoint=None):
@@ -647,7 +650,10 @@ class Facets(FacetsWithEntryPoint):
 
         filter.availability = self.availability
         filter.subcollection = self.collection
-        if self.order:
+
+        # No order and relevance order both signify the default and,
+        # thus, either should leave `filter.order` unset.
+        if self.order and self.order != self.ORDER_BY_RELEVANCE:
             order = self.SORT_ORDER_TO_ELASTICSEARCH_FIELD_NAME.get(self.order)
             if order:
                 filter.order = order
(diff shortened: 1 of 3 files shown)


tdilauro:
@leonardr Let me know if you'd like me to do the logging testing a different way (e.g., with a fixture) or if you want me to move the `LocalLogCaptureHandler` to `testing.py`.

leonardr:
I don't want you to spend a lot of time getting this working, but a context manager similar to warnings.catch_warnings would be good to reuse.

```
with warnings.catch_warnings(record=True) as w:
   issue_warning() 
   assert "warning" in w
```

I searched around and there are a few ways of doing this, but nothing seems to be built into nosetests. If you want to make a context manager then that's something I'd like to see in testing; otherwise I think this can wait until we migrate to Python 3.

tdilauro:
It would be super simple to convert what I have to a context manager on the order of

```
    with LocalLogCaptureHandler(logging.root) as log_catcher:
        issue_warning()
        assert "warning" not in log_catcher.records["warning"]
```

I could take a quick at warnings to see what would be needed there.

tdilauro:
@leonardr Would this work for you for the time being?

```
with LogCaptureHandler(some_logger) as logs:
    issue_warning()
    do_not_issue_error()
    assert "warning" in logs.warning
    assert "error" not in logs.error
```

leonardr:
Yeah, that looks very promising.

tdilauro:
I reorged this branch a bit to make it more comprehensible and force pushed it.

PR #1179, opened 2020-06-18
Add support for choosing arbitrary S3 regions and using virtual hosted-style S3 URLs

vbessonov:
## Description

This PR adds support for choosing arbitrary S3 regions and using virtual hosted-style S3 URLs instead of legacy path-style URLs.
`S3Uploader` uses `us-east-1` and virtual hosted-style URLs by default.
However, administrators can change both of these settings in the admin interface.

Also, another important change is getting rid of `urllib.quote_plus` and using `urllib.quote` to encode URLs. It replaces spaces in the part of URLs to `%20` instead of `+` according to [the standard](https://web.archive.org/web/20151218094722/http://blog.lunatech.com/2009/02/03/what-every-web-developer-must-know-about-url-encoding). 
Please note, that modern browsers treat `+` and `%20` equally and they can be used interchangeably. However, `%20` is more preferred.

## Motivation and Context

Buckets created after September 30, 2020, will support only virtual hosted-style requests. Path-style requests will continue to be supported for buckets created on or before this date. For more information, see [Amazon S3 Path Deprecation Plan – The Rest of the Story](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/).

## How Has This Been Tested?

I tested that the admin interface of Circulation Manager works correctly with all new settings and created additional unit tests covering the new functionality.

## Checklist:

### DIFF
--- a/cdn.py
+++ b/cdn.py
@@ -19,7 +19,7 @@ def cdnify(url, cdns=None):
         return url
     scheme, netloc, path, query, fragment = urlparse.urlsplit(url)
 
-    if 'amazonaws.com' in netloc:
+    if netloc.endswith('amazonaws.com'):
         # This is a URL like "http://bucket.s3.region.amazonaws.com/foo".
         # It's equivalent to "http://bucket/foo".
         # i.e. treat the bucket name as the netloc.
--- a/cdn.py
+++ b/cdn.py
@@ -19,8 +19,8 @@ def cdnify(url, cdns=None):
         return url
     scheme, netloc, path, query, fragment = urlparse.urlsplit(url)
 
-    if netloc == 's3.amazonaws.com':
-        # This is a URL like "http://s3.amazonaws.com/bucket/foo".
+    if 'amazonaws.com' in netloc:
+        # This is a URL like "http://bucket.s3.region.amazonaws.com/foo".
         # It's equivalent to "http://bucket/foo".
         # i.e. treat the bucket name as the netloc.
         #
--- a/s3.py
+++ b/s3.py
@@ -411,7 +411,7 @@ class S3Uploader(MirrorUploader):
     def bucket_and_filename(cls, url, unquote=True):
         scheme, netloc, path, query, fragment = urlsplit(url)
 
-        if 'amazonaws.com' in netloc:
+        if netloc.endswith('amazonaws.com'):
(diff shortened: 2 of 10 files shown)


cdn.py
```diff
@@ -19,8 +19,8 @@ def cdnify(url, cdns=None):
         return url
     scheme, netloc, path, query, fragment = urlparse.urlsplit(url)
 
-    if netloc == 's3.amazonaws.com':
-        # This is a URL like "http://s3.amazonaws.com/bucket/foo".
+    if 'amazonaws.com' in netloc:
```
tdilauro, line 22: Do you want to use `endswith()` here?

vbessonov: Totally makes sense, changed it

tdilauro:
@vbessonov Do you have any tests to ensure that any already configured "path-style" S3 buckets will continue to work? There is at least one configured in a production CM.

s3.py
```diff
@@ -161,29 +241,69 @@ def __init__(self, integration, client_class=None):
...
+        :rtype: string
+        """
+        key = path
+
+        # NOTE: path can be an empty string meaning that
+        # we need to generate a URL pointing at the root directory of the bucket.
+        # However, boto3 doesn't allow us to pass the key as an empty string.
+        # As a workaround we set it to a dummy string and later remove it from the generated URL
+        if not path:
+            key = 'dummy'
```
leonardr, line 261: I get why this is necessary but I'd feel better if this dummy key included a frivolous Unicode character, like `dummy☃`. I could see someone creating a bucket actually called `dummy` for testing purposes.

leonardr: I guess it doesn't matter too much since you're never sending `dummy` to the server, so a preexisting bucket by that name won't cause a problem.

vbessonov:
> @vbessonov Do you have any tests to ensure that any already configured "path-style" S3 buckets will continue to work? There is at least one configured in a production CM.

@tdilauro, yes, you can look at `TestS3Uploader.test_url` for example. There are cases `custom_http_url_and_path_without_slash`, `custom_http_url_and_path_with_slash`, `custom_http_url_and_path_without_slash`, `custom_http_url_and_path_with_slash`. And there are similar cases for other methods as well

PR #1183, opened 2020-06-30
Create a separate boto client for generating unsigned URLs

vbessonov:
## Description

This PR changes S3Uploader initialisation logic, it creates two boto3 clients:
- the first one **without** authentication used for generating unsigned URLs
- the second one **with** authentication for working with S3: uploading files, etc.

## Motivation and Context

This PR fixes the authentication issue in S3Uploader caused by [PR 1179](https://github.com/NYPL-Simplified/server_core/pull/1179).
Steps to reproduce the issue:
1. Create a MARC or ONIX collection
2. Try to import it using `bin/directory_import` with valid S3 credentials
In the output you'll see an authentication error.

## How Has This Been Tested?

I tested it by importing an ONIX collection into S3 with `bin/directory_import` script using the steps described above.

## Checklist:

### DIFF
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -128,10 +128,25 @@ class TestS3Uploader(S3UploaderTest):
         # attribute on the S3Uploader object.
         eq_('a transform', uploader.url_transform)
 
-    def test_empty_string(self):
+    @parameterized.expand([
+        (
+            'empty_credentials',
+            None,
+            None
+        ),
+        (
+            'empty_string_credentials',
+            '',
+            ''
+        ),
+        (
+            'empty_string_credentials',
+            'username',
+            'password'
+        )
+    ])
+    def test_initialization(self, name, username, password):
         # Arrange
-        username = 'username'
-        password = 'password'
         settings = {'username': username, 'password': password}
         integration = self._external_integration(
             ExternalIntegration.S3, goal=ExternalIntegration.STORAGE_GOAL, settings=settings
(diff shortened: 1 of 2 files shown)


leonardr:
There's one test failure that seems pretty minor. Looks good otherwise.

tests/test_s3.py
```diff
@@ -129,7 +130,9 @@ def test_instantiation(self):
 
     def test_empty_string(self):
         # Arrange
-        settings = {'username': '', 'password': ''}
+        username = 'username'
+        password = 'password'
+        settings = {'username': username, 'password': password}
```
leonardr, line 135: What is the `empty_string` here if not the credentials? Should this be called something like `test_unsigned_uploader`?

vbessonov: Renamed the test method and added additional cases

tests/test_s3.py
```diff
@@ -128,10 +128,25 @@ def test_instantiation(self):
...
+            None,
+            None
+        ),
+        (
+            'empty_string_credentials',
+            '',
+            ''
+        ),
+        (
+            'empty_string_credentials',
```
leonardr, line 143: The code change makes sense but I'm still confused about this test name -- the credentials are `username` and `password`, not empty strings. I think you just need to rename this test?

vbessonov: I renamed the test to `test_initialization` to reflect that it tests different cases. What do you think would be a good name?

leonardr: test_initialization is fine for the name of the function. I'm talking about the value of the variable `name`. When you're parameterizing the test you set `name="empty_string_credentials"`, `username=""`, `password=""`. That makes sense to me. Then you set `name="empty_string_credentials"` again,  but with `username="username"` and `password="password"`. That's the part I don't get.

vbessonov: Sorry, didn't notice that the second case had the same name, changed it to `non_empty_string_credentials`

leonardr:
Great, it makes sense now.

PR #1184, opened 2020-07-01
Add support for S3 presigned URLs

vbessonov:
## Description

This PR adds support for using S3 presigned URLs required by [PR 1458](https://github.com/NYPL-Simplified/circulation/pull/1458).

## Motivation and Context

This functionality by required by [PR 1458](https://github.com/NYPL-Simplified/circulation/pull/1458).

Please refer to [this document](https://docs.google.com/document/d/1W1QVpkhvaR63FD179PNAzncGNo6Yx2V2mwOhU0vMRv4/edit#) for more details.

## Checklist:


tests/test_s3.py
```diff
@@ -1003,6 +1003,35 @@ def complete(self):
...
+        ('default_expiration_parameter', None, int(S3Uploader.S3_DEFAULT_PRESIGNED_URL_EXPIRATION)),
+        ('empty_expiration_parameter', {S3Uploader.S3_PRESIGNED_URL_EXPIRATION: 100}, 100)
+    ])
+    def test_sign_url(self, name, expiration_settings, expected_expiration):
+        # Arrange
+        region = 'us-east-1'
+        bucket = 'bucket'
+        filename = 'filename'
+        url = 'https://{0}.s3.{1}.amazonaws.com/{2}'.format(bucket, region, filename)
+        expected_url = url + '?AWSAccessKeyId=KEY&Expires=1&Signature=S'
```
tdilauro, line 1011: Minor, but instead of testing that the query params are present and this order, could you instead test that they (and no others) are present and have the correct values?

metadata_layer.py
```diff
@@ -861,10 +868,10 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
         # Mirror it.
         mirror.mirror_one(representation, mirror_url)
 
-        # If we couldn't mirror an open access link representation, suppress
+        # If we couldn't mirror an open/protected access link representation, suppress
         # the license pool until someone fixes it manually.
         if representation.mirror_exception:
-            if pools and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+            if pools and link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.GENERIC_OPDS_ACQUISITION]:
```
tdilauro, line 874: This list is used in multiple places, so might be useful to put it in a variable for DRY/maintainability.

leonardr: I agree, and  core/model/constants.py#LinkRelations is the appropriate place for this.

First, GENERIC_OPDS_ACQUISITION should be added to the MIRRORED list, which contains the link relations that a MirrorUploader might be expected to handle.

By definition, the list you're working with here is the intersection of  MIRRORED (should be run through a MirrorUploader) and CIRCULATION_ALLOWED (a book or book-like thing that can be circulated). So you can calculate it as a new list called something like MIRRORED_BOOKS, or you can just check both lists.

metadata_layer.py
```diff
@@ -791,12 +795,12 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
...
 
         # If we couldn't fetch this representation, don't mirror it,
-        # and if this was an open access link, then suppress the associated
+        # and if this was an open/protected access link, then suppress the associated
         # license pool until someone fixes it manually.
         # The license pool to suppress will be either the passed-in model_object (if it's of type pool),
         # or the license pool associated with the passed-in model object (if it's of type edition).
         if representation.fetch_exception:
-            if pools and link.rel == Hyperlink.OPEN_ACCESS_DOWNLOAD:
+            if pools and link.rel in [Hyperlink.OPEN_ACCESS_DOWNLOAD, Hyperlink.GENERIC_OPDS_ACQUISITION]:
```
leonardr, line 803: This code is better than the alternative but I don't think it will ever be activated. It looks like this code is using `Representation.get` to retrieve a book by its URL before mirroring. I wouldn't expect that to work for books that are non-open-access; instead I'd expect the book to be retrieved some other way and stored in LinkData.content. Am I missing something?

model/collection.py
```diff
@@ -744,9 +742,9 @@ def restrict_to_ready_deliverable_works(
         if not show_suppressed:
             query = query.filter(LicensePool.suppressed==False)
 
-        # Only find books with available licenses.
+        # Only find books with available licenses or books from self-hosted collections using MirrorUploader
         query = query.filter(
-                or_(LicensePool.licenses_owned > 0, LicensePool.open_access)
+            or_(LicensePool.licenses_owned > 0, LicensePool.open_access, LicensePool.self_hosted)
```
leonardr, line 747: In core/external_search.py:Filter#build there is equivalent code for putting this restriction into a query against the Elasticsearch index. Something similar happens in `universal_nested_filters. To get this to work, `open_access` is defined as a boolean field of the "licensepools" sub-document, and "self_hosted" should be added there as well. 

Basically, you should be add a self-hosted book to the data sets defined in `core/tests/test_external_search:TestExternalSearchWithWorks` and `TestFacetFilters`, and verify that the self-hosted book shows up in normal queries. I expect that such a book would not show up in queries now. 

Comprehensive documentation of our search index is [here](https://github.com/NYPL-Simplified/Simplified/wiki/Elasticsearch); I can also answer any questions you have. Once this branch lands I'll update the wiki page.

model/configuration.py
```diff
@@ -58,25 +63,47 @@ class ExternalIntegrationLink(Base, HasFullTableCache):
...
+        {
+            'key': OPEN_ACCESS_BOOKS_KEY,
+            'type': OPEN_ACCESS_BOOKS,
+            'description_type': 'free books',
+            'label': 'Open Access Books Mirror'
+        },
+        {
+            'key': PROTECTED_ACCESS_BOOKS_KEY,
+            'type': PROTECTED_ACCESS_BOOKS,
+            'description_type': 'self-hosted books',
```
leonardr, line 82: Open access books are also self-hosted, so how about calling this "self-hosted, commercially licensed books" or similar?

metadata_layer.py
leonardr: I see that you already modified MIRRORED -- which makes sense since otherwise you'd get some test failures -- so the first part of my comment is moot.

model/work.py
```diff
@@ -1173,11 +1166,13 @@ def active_license_pool(self):
                     # We have an unlimited source for this book.
                     # There's no need to keep looking.
                     break
+            elif p.self_hosted:
+                active_license_pool = p
+                break
```
leonardr, line 1171: It will almost never matter but I don't think `break` is appropriate here. If there is a self-hosted non-open-access edition of a book, but also an open-access edition, then the open-access edition should win because it comes with fewer strings attached.

tests/test_s3.py
leonardr: Another way of doing this would be to guarantee a deterministic order by sorting the keys when generating the URL. These keys are sorted but I think this code is in a library we don't control, so I don't know if that's guaranteed.

leonardr:
Overall this looks really good, my main concern is that these books will probably be excluded from a lot of Elasticsearch queries. Let me know if you have any questions about the Elasticsearch code; I've tried to document it in a lot of detail but I know it's confusing and it can be difficult to diagnose problems because the most common failure case is the query just doesn't run.

leonardr:
Re: Elasticsearch, an easier solution would be to change `Work.to_search_document` so that the Elasticsearch field `licensepools.available` is always true if `LicensePool.self_hosted` is true. That should resolve all the problems I was concerned about, and you'd only have to add something to the test of `Work.to_search_document`.

The reason we index `licensepools.open_access` separately is that a user might specifically want to limit search results to open-access titles. Nobody's going to be limiting search results to self-hosted titles, so we can just treat it as another way a book might become "available", and not tell Elasticsearch the specifics.

tests/test_scripts.py
```diff
@@ -2644,7 +2644,8 @@ def replacement_policy(cls, uploader):
         eq_(Mock.mock_policy, policy)
         # There should be two MirrorUploaders for each purpose.
```
tdilauro, line 2645: It would be helpful to clarify this comment while you're at it. I think you meant:
```
    # There should be two MirrorUploaders, one for each purpose.
```

Sorry I missed that on the original commit.

tests/test_s3.py
vbessonov: This test uses `expected_url` as a return value of the mocked `generate_presigned_url` method:
```python
s3_uploader.client.generate_presigned_url = MagicMock(return_value=expected_url)
```
It's done to not use the real `boto3` in the test and this guarantees that `sign_url` will return exactly this value

metadata_layer.py
vbessonov: I added `SELF_HOSTED_BOOKS` to `Hyperlink` class and started using it

model/configuration.py
vbessonov: I changed the `description_type` as requested

tests/test_scripts.py
vbessonov: Updated the comment

model/work.py
vbessonov: Removed the `break` statement

model/collection.py
vbessonov: I added an additional condition to `Work.to_search_document` using `LicensePool.self_hosted`:
```python
licensepools = select(
            [
                LicensePool.id.label('licensepool_id'),
                LicensePool.data_source_id.label('data_source_id'),
                LicensePool.collection_id.label('collection_id'),
                LicensePool.open_access.label('open_access'),
                LicensePool.suppressed,
                or_(
                    LicensePool.self_hosted,
                    LicensePool.licenses_available > 0,
                ).label('available'),
                (LicensePool.licenses_owned > 0).label('licensed'),
                work_quality_column,
                Edition.medium,
                func.extract(
                    "EPOCH",
                    LicensePool.availability_time,
                ).label('availability_time')
            ]
        ).where(
            and_(
                LicensePool.work_id==work_id_column,
                work_presentation_edition_id_column==Edition.id,
                or_(
                    LicensePool.open_access,
                    # LicensePool.self_hosted,
                    LicensePool.licenses_owned>0,
                ),
            )
        ).alias("licensepools_subquery")
```
However, I'm not completely sure whether I need to add `LicensePool.self_hosted` to **where** statement

metadata_layer.py
vbessonov: I reverted my changes in both places in this method. Are there any other changes required?

leonardr:
A couple follow-up comments based on your changes and things I remembered after coming back overnight:

* A work needs to be reindexed whenever the value of `LicensePool.self_hosted` changes, because the value of `licensepools.available` may also change. You can do this by annotating `licensepool_open_access_change` in model/listeners.py to listen for changes to `self_hosted` as well as `open_access`, and renaming the function to something more generic like `licensepool_storage_status_change`. We don't have test coverage for the `open_access` listener but there are some other listener examples you can copy if you want to add coverage for this stuff.

* `LicensePool.deliverable` needs to be changed to take `self_hosted` into account.

* I think you should uncomment `# LicensePool.self_hosted,`. I would expect a self-hosted book  have `.licenses_owned` set to zero, because self-hosted titles don't have enumerated licenses. Your `test_self_hosted_books_are_available_by_default` passes because you set `.licenses_available` to zero, but a LicensePool created for a unit test will have `.licenses_owned=1`, and that should also be set to zero. At that point the test will fail until you uncomment that line.

vbessonov:
@leonardr, I made all the required changes according to your last comment except one: I couldn't add a test for `licensepool_open_access_change` using `Mock`. Also, I tried to use tests from `test_listeners.py` as an example using `Mock` instead of a custom class. Are there any other examples I could use?

leonardr:
Here's some pseudocode for a test. I wrote it for LicensePool.self_hosted but you can paramaterize it to cover LicensePool.open_access as well.

```
work = self._work(with_license_pool=True)
[pool] = work.licensepools

# Clear out any WorkCoverageRecords created as the work was initialized.
work.coverage_records = []

# Make it look like the work was indexed earlier.
WorkCoverageRecord.add_for(work, operation=WorkCoverageRecord.UPDATE_SEARCH_INDEX_OPERATION

# Change the field
pool.self_hosted = True
self._db.commit # Pretty sure this is necessary, but try it without.

# The coverage record has been removed, indicating that the work needs to be reindexed.
eq_([], work.coverage_records)

# Then verify that if the field is 'set' to its existing value, this doesn't happen.
WorkCoverageRecord.add_for(work, operation=WorkCoverageRecord.UPDATE_SEARCH_INDEX_OPERATION
pool.self_hosted = True
eq_(1, len(work.coverage_records))
```

Let me know if you have problems getting something like this to work. Since

vbessonov:
@leonardr, I added tests according to what we discussed

vbessonov:
@leonardr, added an additional check for `.status=REGISTERED`

tdilauro:
Looks like [commit 5fa7f56](https://github.com/NYPL-Simplified/server_core/pull/1184/commits/5fa7f56cb8ba8b023c41af0cadb97aa990ec7944) and subsequent changes in this PR break some tests in at the [HEAD of the Circulation Manager master branch](https://github.com/NYPL-Simplified/circulation/tree/38f0dc218a9177f75c8035f2d42cc03a535a9444) and I don't see anything in the currently open [CM PRs](https://github.com/NYPL-Simplified/circulation/pulls) that addresses this. The following are failing:

- `admin/controller/test_collections.py:TestCollectionSettings.test_collections_get_collection_protocols`
- `admin/controller/test_controller.py:TestSettingsController.test__mirror_integration_settings`
- `admin/controller/test_metadata_services.py:TestMetadataServices.test_process_get_with_self_tests`

```
======================================================================
FAIL: tests.admin.controller.test_collections.TestCollectionSettings.test_collections_get_collection_protocols
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/timmo/.PyEnvs/circulation-nQEl9FH5/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/Shared/src/lyrasis/circulation/tests/admin/controller/test_collections.py", line 81, in test_collections_get_collection_protocols
    eq_("Books Mirror", books_mirror['label'])
AssertionError: 'Books Mirror' != l'Open Access Books Mirror'

======================================================================
FAIL: tests.admin.controller.test_controller.TestSettingsController.test__mirror_integration_settings
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/timmo/.PyEnvs/circulation-nQEl9FH5/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/Shared/src/lyrasis/circulation/tests/admin/controller/test_controller.py", line 2350, in test__mirror_integration_settings
    eq_(settings[1]["label"], "Books Mirror")
AssertionError: l'Open Access Books Mirror' != 'Books Mirror'

======================================================================
FAIL: tests.admin.controller.test_metadata_services.TestMetadataServices.test_process_get_with_self_tests
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/timmo/.PyEnvs/circulation-nQEl9FH5/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/Shared/src/lyrasis/circulation/tests/admin/controller/test_metadata_services.py", line 107, in test_process_get_with_self_tests
    "Exception getting self-test results for metadata service Test: Metadata Wrangler improperly configured."
AssertionError: None != 'Exception getting self-test results for metadata service Test: Metadata Wrangler improperly configured.'
```

These tests pass for [Server Core commit 6d7e05a8](https://github.com/NYPL-Simplified/server_core/tree/6d7e05a815c2abd8f73e2f40585163765f2a4d14), but are failing at at [commit da39db95](https://github.com/NYPL-Simplified/server_core/tree/da39db95990c77c417e5aae2f9a12acc0cf78cde).

vbessonov:
@tdilauro, I fixed `test_collections_get_collection_protocols` and `test__mirror_integration_settings` in [PR 1458](https://github.com/NYPL-Simplified/circulation/pull/1458)

tdilauro:
Great! Thanks! Hopefully that will get merged soon.

PR #1185, opened 2020-07-08
Add integration tests for S3Uploader using MinIO

vbessonov:
## Description

This PR add integration tests using MinIO as an S3 replacement.

It's based on [PR 1184](https://github.com/NYPL-Simplified/server_core/pull/1184) so it has to be merged first.

## Motivation and Context

There should be a way to perform integration testing of S3Uploader without using real AWS account.

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,10 @@ env:
   global:
     - SIMPLIFIED_TEST_DATABASE="postgres://simplified_test:test@localhost:5432/simplified_core_test"
     - ES_VERSION="6.3.2" ES_DOWNLOAD="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VERSION}.tar.gz"
-    - MINIO_ENDPOINT_URL=http://localhost:9000
+    - MINIO_HOST=localhost:9000
+    - MINIO_ENDPOINT_URL=http://${MINIO_HOST}
     - MINIO_USER=minioadmin
     - MINIO_PASSWORD=minioadmin
-    - MINIO_HOST=localhost:9000
 
 services:
   - postgresql
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ env:
   global:
     - SIMPLIFIED_TEST_DATABASE="postgres://simplified_test:test@localhost:5432/simplified_core_test"
     - ES_VERSION="6.3.2" ES_DOWNLOAD="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VERSION}.tar.gz"
+    - MINIO_ENDPOINT_URL=http://localhost:9000
+    - MINIO_USER=minioadmin
+    - MINIO_PASSWORD=minioadmin
 
 services:
   - postgresql
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ cache: pip
 
 before_install:
   - pip install --upgrade pip
-  - pip install "setuptools>=18.5"
+  - pip install "setuptools>=44.1.0"
   - docker pull minio/minio
   - docker run -d -p 9000:9000 minio/minio server /data
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
     - MINIO_ENDPOINT_URL=http://localhost:9000
     - MINIO_USER=minioadmin
     - MINIO_PASSWORD=minioadmin
+    - MINIO_HOST=localhost:9000
 
 services:
   - postgresql
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ before_install:
   - pip install "setuptools>=18.5"
   - docker pull minio/minio
   - docker run -d -p 9000:9000 minio/minio server /data
-  - sleep 10
 
 install:
   - wget ${ES_DOWNLOAD}
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
 
 services:
   - postgresql
+  - docker
 
 language: python
 
@@ -25,6 +26,8 @@ cache: pip
 before_install:
   - pip install --upgrade pip
   - pip install "setuptools>=18.5"
+  - docker pull minio/minio
+  - docker run -d -p 9000:9000 minio/minio server /data
   - sleep 10
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,10 @@ env:
   global:
     - SIMPLIFIED_TEST_DATABASE="postgres://simplified_test:test@localhost:5432/simplified_core_test"
     - ES_VERSION="6.3.2" ES_DOWNLOAD="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VERSION}.tar.gz"
-    - MINIO_HOST=localhost:9000
-    - MINIO_ENDPOINT_URL=http://${MINIO_HOST}
-    - MINIO_USER=minioadmin
-    - MINIO_PASSWORD=minioadmin
+    - SIMPLIFIED_MINIO_HOST=localhost:9000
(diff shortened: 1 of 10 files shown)


jonathangreen:
I like this, I think doing some testing against a real S3 API will be really helpful. Especially in the CI environment where its easy enough to have additional services running.  

I think it would be a good idea to preserve a way to run the tests locally without having a run a Minio environment. Maybe we can use nose test [attributes](https://nose.readthedocs.io/en/latest/plugins/attrib.html) to do so? 

We could set two attributes: `integration` and `minio` so that you can run the tests with `nosetests -a '!integration'` to run all the tests except those requiring external integration? Thoughts @leonardr or @tdilauro?

It would be nice to allow minio as in the storage configuration options as well, so when testing during local development you could just use minio to test S3 integrations, but that can be left for another time.

tests/test_s3.py
```diff
@@ -104,6 +107,94 @@ def _create_s3_uploader(
         return uploader_class(integration, client_class=client_class)
 
 
+class S3UploaderIntegrationTest(S3UploaderTest):
+    MINIO_PORT = 9000
+    MINIO_ENDPOINT_URL = 'http://localhost:{0}'.format(MINIO_PORT)
+    MINIO_USER = 'minioadmin'
+    MINIO_PASSWORD = 'minioadmin'
```
tdilauro, line 114: I don't think these should be hard-coded, since they might conflict with a developers environment and would require tracking down and changing code to get tests running. You could make it more visible and with an environment variable that's documented alongside `SIMPLIFIED_TEST_DATABASE`, which is used in a similar manner.

leonardr:
Adding a Docker instance to our CI process is a great idea, but we absolutely need to be able to run the tests without having Docker or MinIO. I'm also a little bit concerned about the extra time this adds to the CI process.

We have a situation like this already: we have Elasticsearch integration tests in core/tests/test_external_search.py that check for a running server and drop out if they don't find one. Using test attributes would definitely be a more elegant way of doing this, so I'm in favor of using Jonathan's suggestion here and doing something similar in core/tests/test_external_search.py.

I'd like to have "no integration tests" be the default, but it doesn't look like that's possible with nose attributes, so it's not a big deal.

As for the extra time, it looks to be about 5 seconds, but there's also a `sleep 10` command in `before_install` that I don't think is doing anything. I believe that was originally intended to give the Elasticsearch server time to come up, but we're now handling that with arguments to the `wget` command.

Try taking out the `sleep` command and see if everything still works; that will save us 5 seconds net and we'll have a Docker instance we can use for future integration tests.

s3.py
```diff
@@ -102,6 +102,12 @@ class S3AddressingStyle(Enum):
 class S3Uploader(MirrorUploader):
     NAME = ExternalIntegration.S3
 
+    # AWS S3 host
+    S3_HOST = 'amazonaws.com'
+
+    # MinIO host used in integration tests (we assume that it's always using port 9000)
+    MINIO_HOST = 'localhost:9000'
```
leonardr, line 109: I think the hostname could also be pulled from an environment variable.

tdilauro: > We could set two attributes: `integration` and `minio` so that you can run the tests with `nosetests -a '!integration'` to run all the tests except those requiring external integration? Thoughts @leonardr or @tdilauro?

Yeah, this makes sense to me. It would be very important to me to be able to easily run the tests without any of the integrations and I think this approach would work nicely.

tests/test_s3.py
vbessonov: I replaced hard-coded credentials with environment variables

s3.py
vbessonov: I replaced the hard-coded with an environment variable

vbessonov:
I would love to add MinIO as a separate storage service and there are several ways to achieve it:
- add `ENDPOINT_URL` setting to `S3Uploader` settings and use its value while instantiating a boto3 client
- subclass `S3Uploader` and add `ENDPOINT_URL` there
I think the latter would be more obvious and user friendly.

However, there is one problem. Method `cdnify` defined in `cdn.py` checks whether the URL is S3-based or uses a custom domain name. It works perfectly in the case of S3 because it's enough to check whether the hostname ends with `amazonaws.com` which is defined as a static constant. But it will not work in the case of custom MinIO instance because it can have an arbitrary endpoint URL and we need to get an instance of `ExternalIntegration` to get the real URL.
It means we need to change `cdnfy`: 
1. We need to extend a list of parameters and pass a `Collection` instance
2. Using the `Collection` instance we need to fetch set up mirrors and then fetch their real endpoint URLs
@leonardr, does it make sense?

leonardr:
Is it possible for a single circulation manager to use both S3_HOST and SIMPLIFIED_MINIO_HOST? Or is it more like SIMPLIFIED_MINIO_HOST is an alternate value for S3_HOST?

If there's only one variable here, let's imagine that you make "the host the CM should use to upload files through an S3-like interface" into a site-wide database setting rather than using a constant (or two constants). You could then create a method like `Configuration.cdns()` which is run once, to set up the configuration. That way you don't have to have a database connection lying around every time you want to call `cdnify`.

vbessonov:
@jonathangreen, @tdilauro, could you answer the last comment left by Leonard? I think that it's very possible to have both S3 and MinIO integrations set up and even have several instances of each of them

leonardr:
In that case I think the solution is simpler (though potentially annoying): we need to remove all special treatment of S3 from `cdnify`. Given a URL, we can't make any assumptions about whether or not the URL represents something uploaded to "S3". The same URL might represent an "S3" URL in one context but not in another, and by the time we get into `cdnify` we've lost that context.

The S3-specific code in cdnify just tries to normalize "https://s3.amazonaws.com/bucket/file/" to "https://bucket/file/". If we can guarantee that those old-style URLs are no longer being generated, then we could remove the S3-specific code from `cdnify` altogether.

IMO it's also pretty silly to put a CDN in front of S3, and I think if `cdnify` sometimes fails to do this, it's not the worst thing in the world.

cdn.py
```diff
@@ -1,33 +1,65 @@
...
-from nose.tools import set_trace
 
 from config import Configuration, CannotLoadConfiguration
-from s3 import S3Uploader
+from core.mirror import MirrorUploader
+from core.model import ExternalIntegration
+from core.s3 import S3Uploader
+
+
+def _get_mirrors():
```
leonardr, line 10: As we discussed earlier, you can make cdn.py forget about S3 entirely and remove this code.

s3.py
```diff
@@ -540,12 +603,56 @@ def multipart_upload(self, representation, mirror_to, upload_class=MultipartS3Up
...
+        if callable(client_class):
+            client_class = functools.partial(client_class, endpoint_url=endpoint_url)
+        else:
+            self.client = client_class
+
+        super(MinIOUploader, self).__init__(integration, client_class, host)
+
+
+# MirrorUploader.implementation will instantiate an S3Uploader
+# for storage integrations with protocol 'Amazon S3'.
```
leonardr, line 645: This comment was copy-and-pasted and should be updated.

leonardr:
This looks pretty good; I'll take another look once #1184 is merged.

s3.py
vbessonov: Updated the comment

cdn.py
vbessonov: Removed the S3-related logic from `cdnfy`

vbessonov:
@leonardr, I updated the PR with the latest changes from the master

.travis.yml
```diff
@@ -11,9 +11,14 @@ env:
   global:
     - SIMPLIFIED_TEST_DATABASE="postgres://simplified_test:test@localhost:5432/simplified_core_test"
     - ES_VERSION="6.3.2" ES_DOWNLOAD="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VERSION}.tar.gz"
```
leonardr, line 13: At one point I saw a version of this(?) branch that also moved Elasticsearch into Docker. I think that's a great idea and hope you do a follow-up.

vbessonov: It was [Jonathan's PR](https://github.com/NYPL-Simplified/server_core/pull/1186) but it hasn't been merged yet

PR #1186, opened 2020-07-10
Change Elasticsearch integration tests to use attributes

jonathangreen:
## Description

This PR: 
- Changes the Elasticsearch Integration tests to use an attribute instead of just exiting if Elasticsearch is not found.
- Adds an environment variable for the location the tests should look for Elasticsearch, but falls back to the location that was hard coded before. 
- Runs the Elasticsearch instance used for testing in a docker container, instead of downloading and installing it. 

This is in line with the changes for testing with MinIO in #1185. 

## Motivation and Context

This is a follow-on to the changes in #1185, bringing those changes to the ES integration tests.

## Checklist:

[] I have updated the documentation accordingly.
[x] All new and existing tests passed.


testing.py
```diff
@@ -1083,7 +1084,10 @@ class ExternalSearchTest(DatabaseTest):
     to ensure that it works well overall, with a realistic index.
     """
 
+    SIMPLIFIED_TEST_ELASTICSEARCH = os.environ.get('SIMPLIFIED_TEST_ELASTICSEARCH', u'http://localhost:9200')
```
vbessonov, line 1087: Do we need to set value of `SIMPLIFIED_TEST_ELASTICSEARCH` in .travis.yml?

leonardr: I think we should be explicit, but 9200 is the default Elasticsearch port so I'm not surprised this works without it.

testing.py
```diff
@@ -1084,7 +1084,10 @@ class ExternalSearchTest(DatabaseTest):
...
```
tdilauro, line 1087: It'll be nice to have this. I just ran into the non-default ES port issue last night while running tests with the docker containers for only one of the hosts of a Shared ODL dev environment. The ES host was on port 9210. Adding this to my env files now in anticipation.

testing.py
tdilauro: Yes, I agree. I think we should be as clear as possible whenever we reasonably can.

jonathangreen: Good call @vbessonov, I updated the travis file.

jonathangreen:
Resolved the conflict with this PR and set `SIMPLIFIED_TEST_ELASTICSEARCH` in the `.travis.yml` file, so I think this should be ready to go.

PR #1187, opened 2020-07-10
Fix travis setup issue

jonathangreen:


jonathangreen:
This does not seem needed since the build passed

PR #1188, opened 2020-07-15
Let Credential.lookup's caller force refresher method to run

tdilauro:
## Description

This branch allows the caller of `Credential.lookup` to force the `refresher_method` to run, regardless of other conditions.

## Motivation and Context

There are a couple cases in which it is useful to trigger the `refresher_method` unconditionally:
- Updating an unexpired temporary credential.
- Updating a permanent credential.

## How Has This Been Tested?

- Added new test.
- Tested against both circulation manager and server core test suites.

## Checklist:


tdilauro:
Yes, RBdigital. At least in the near term. It's primarily to let me generalize all the RBD credential logic and gather it into one place. Also, because the lifetime of the bearer token is so long (24 hours), I'm worried that if it gets invalidated, then the user could be locked out for a very long time. So I want to have an easy way to refresh it before it expires.

tdilauro:
This is done and ready to merge.

PR #1190, opened 2020-08-05
Store the last time the patron successfully synced their loans with external APIs

leonardr:
This is the core component of the work for https://jira.nypl.org/browse/SIMPLY-2922. It adds a field "last_loan_activity_sync" to the `patrons` table which tracks the last time we successfully synced our model of a patron's loans and holds with the sources of truth -- external vendor APIs.

This field is cleared out for a given patron whenever a loan or hold is created for that patron. (It's also cleared out when a loan or hold is removed, but that's handled in circulation.) It's also cleared out after fifteen minutes, no matter what -- this lets us quickly notice cases where the patron uses a vendor website or app to conduct a transaction.

### DIFF
--- a/migration/20200702-add-licensepool-self_hosted-field.sql
+++ b/migration/20200702-add-licensepool-self_hosted-field.sql
@@ -1,1 +1,11 @@
-ALTER TABLE licensepools ADD COLUMN self_hosted BOOLEAN;
+DO $$ 
+ BEGIN
+  -- Add the 'self_hosted' column
+  BEGIN
+   ALTER TABLE licensepools ADD COLUMN self_hosted BOOLEAN;
+  EXCEPTION
+   WHEN duplicate_column THEN RAISE NOTICE 'column licensepools.self_hosted already exists, not creating it.';
+  END;
+ END;
+$$;
+
--- a/model/licensing.py
+++ b/model/licensing.py
@@ -806,7 +806,7 @@ class LicensePool(Base):
             # This action creates uncertainty about what the patron's
             # loan activity actually is. We'll need to sync with the
             # vendor APIs.
-            patron.last_loan_activity_sync = None
+            patron_or_client.last_loan_activity_sync = None
         else:
             # An IntegrationClient can have multiple loans, so this always creates
             # a new loan rather than returning an existing loan.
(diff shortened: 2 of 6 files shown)


migration/20200702-add-licensepool-self_hosted-field.sql
```diff
@@ -1 +1,11 @@
...
+ BEGIN
+  -- Add the 'self_hosted' column
+  BEGIN
+   ALTER TABLE licensepools ADD COLUMN self_hosted BOOLEAN;
+  EXCEPTION
+   WHEN duplicate_column THEN RAISE NOTICE 'column licensepools.self_hosted already exists, not creating it.';
+  END;
+ END;
+$$;
+
```
leonardr, line 11: This is an unrelated change. While writing the migration script I noticed that the most recent other migration script didn't have the safety code we use to make sure the migration process doesn't crash if given a database that was already partially migrated.

model/licensing.py
```diff
@@ -800,7 +800,13 @@ def loan_to(self, patron_or_client, start=None, end=None, fulfillment=None, exte
...
             loan, is_new = get_one_or_create(
                 _db, Loan, patron=patron_or_client, license_pool=self,
-                create_method_kwargs=kwargs)
+                create_method_kwargs=kwargs
+            )
+
+            # This action creates uncertainty about what the patron's
+            # loan activity actually is. We'll need to sync with the
+            # vendor APIs.
+            patron_or_client.last_loan_activity_sync = None
```
leonardr, line 809: The loan_to method didn't have a unit test of its own so I added a basic one while testing this.

model/licensing.py
```diff
@@ -820,7 +826,12 @@ def on_hold_to(self, patron_or_client, start=None, end=None, position=None, exte
         start = start or datetime.datetime.utcnow()
         if isinstance(patron_or_client, Patron):
             hold, new = get_one_or_create(
-                _db, Hold, patron=patron_or_client, license_pool=self)
+                _db, Hold, patron=patron_or_client, license_pool=self
+            )
+            # This action creates uncertainty about what the patron's
+            # loan activity actually is. We'll need to sync with the
+            # vendor APIs.
+            patron_or_client.last_loan_activity_sync = None
```
leonardr, line 834: Ditto for on_hold_to.

PR #1192, opened 2020-08-23
For URLs, use only path component to guess content type.

tdilauro:
## Description

This branch adds a method that uses only the `path` component of a URL when attempting to guess the content type based on the "file" extension inferred from URL. Also updated some code where a simpler method meant for files was being used for URLs.

## Motivation and Context

URLs with query strings don't end with a type-hinting "file" extension that can help us guess at the associated content type. Unless we base our guess on the URL's path, we can end up failing to guess or even being mislead by the end of the query string.

## How Has This Been Tested?

- Added new tests to verify that only the `path` component of the URL is used for guessing the content type.
- Ran test suite.
- Ran [Circulation Manager](https://github.com/NYPL-Simplified/circulation) test suite against this submodule branch.

## Checklist:

- [X] All tests in [Circulation Manager](https://github.com/NYPL-Simplified/circulation) test suite passed.

### DIFF
--- a/tests/models/test_resource.py
+++ b/tests/models/test_resource.py
@@ -199,6 +199,8 @@ class TestRepresentation(DatabaseTest):
         m_url = Representation.guess_url_media_type_from_path
         jpg_file = "file.jpg"
         zip_file = "file.ZIP"
+        zip_file_rel_path = "relatively/pathed/file.zip"
+        zip_file_abs_path = "/absolutely/pathed/file.zIp"
 
         eq_(Representation.JPEG_MEDIA_TYPE, m_file(jpg_file))
         eq_(Representation.ZIP_MEDIA_TYPE, m_file(zip_file))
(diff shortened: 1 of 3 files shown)


tests/models/test_resource.py
```diff
@@ -195,18 +195,37 @@ def test_mirrorable_media_type(self):
...
-        m = Representation.guess_media_type
+        m_file = Representation.guess_media_type
+        m_url = Representation.guess_url_media_type_from_path
+        jpg_file = "file.jpg"
+        zip_file = "file.ZIP"
 
-        eq_(Representation.JPEG_MEDIA_TYPE, m("file.jpg"))
-        eq_(Representation.ZIP_MEDIA_TYPE, m("file.ZIP"))
+        eq_(Representation.JPEG_MEDIA_TYPE, m_file(jpg_file))
+        eq_(Representation.ZIP_MEDIA_TYPE, m_file(zip_file))
```
leonardr, line 204: Add a test where the filename is an absolute path to a file on disk; I'm concerned that running a file path through `urlparse` might lose the data we're looking for.

tdilauro: Added.

tdilauro:
This is ready to merge, if you're good with the additional tests.

PR #1193, opened 2020-08-23
Don't propagate missing properties to manifests.

tdilauro:
## Description

This branch improves [Web Publication Manifest](https://github.com/readium/webpub-manifest) generation by avoiding propagation of "missing" properties (those that are `None`) into the manifest's JSON serialization, where they would produce `null`s.

## Motivation and Context

Some clients don't treat a `null` manifest property as if that property had not been specified. The presence of properties with `null` values may cause errors in those clients.

This PR depends on [Circulation Manager PR 1479](https://github.com/NYPL-Simplified/circulation/pull/1479), which fixes an overly narrow test. That has been merged and now all CM tests pass.

- https://jira.nypl.org/browse/SIMPLY-3004

## How Has This Been Tested?

- Added new tests to verify that the missing values were not propagated.
- Ran test suite.
- Ran [Circulation Manager](https://github.com/NYPL-Simplified/circulation) test suite against this submodule branch.

## Checklist:

- [X] Ran full [Circulation Manager](https://github.com/NYPL-Simplified/circulation) test suite and the following tests failed:

### DIFF
--- a/tests/util/test_web_publication_manifest.py
+++ b/tests/util/test_web_publication_manifest.py
@@ -74,6 +74,22 @@ class TestManifest(object):
             dict['resources']
         )
 
+    def test_null_properties_not_propagated(self):
+        manifest = Manifest()
+        additional_properties = dict(extra="value", missing=None)
+
+        manifest.add_link("http://foo/", "self", **additional_properties)
+        manifest.add_reading_order("http://foo/", "text/html", "Chapter 1", **additional_properties)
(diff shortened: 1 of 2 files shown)


tdilauro:
Thanks! You are correct. All tests passing now in both repos, so this is ready to merge.

tdilauro:
Didn't notice that I had left this in draft status. Fixed now, so really ready to merge.

PR #1194, opened 2020-08-24
Add LCP support

vbessonov:
## Description

This PR adds LCP support to Circulation Manager.
The main changes:
1. Configuration framework in core/model/configuration.py
2. `S3Uploader`'s settings refactoring
3. `MirrorUploader.mirror_one` signature change

Linked PRs:
- [PR # 1472](https://github.com/NYPL-Simplified/circulation/pull/1472) in `circulation`
- [PR # 125](https://github.com/NYPL-Simplified/circulation-docker/pull/125) in `circulation-docker`
- [PR # 279](https://github.com/NYPL-Simplified/opds-web-client/pull/279) in `opds-web-client`

## Checklist:

### DIFF
--- a/marc.py
+++ b/marc.py
@@ -560,11 +560,13 @@ class MARCExporter(object):
         for integration in integrations:
             # Only add an integration to choose from if it has a 
             # MARC File Bucket field in its settings.
-            [configuration_setting] = [s for s in integration.settings if s.key=="marc_bucket"]
-            if configuration_setting.value:
-                cls.SETTING['options'].append(
-                    dict(key=str(integration.id), label=integration.name)
-                )
+            configuration_settings = [s for s in integration.settings if s.key=="marc_bucket"]
+
+            if configuration_settings:
+                if configuration_settings[0].value:
+                    cls.SETTING['options'].append(
+                        dict(key=str(integration.id), label=integration.name)
+                    )
         
         return cls.SETTING
 
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -866,7 +866,8 @@ class MetaToModelUtility(object):
             )
 
         # Mirror it.
-        mirror.mirror_one(representation, mirror_url)
+        collection = pools[0].collection if pools else None
+        mirror.mirror_one(representation, mirror_to=mirror_url, collection=collection)
 
         # If we couldn't mirror an open/protected access link representation, suppress
         # the license pool until someone fixes it manually.
@@ -896,7 +897,7 @@ class MetaToModelUtility(object):
             if is_new:
                 # A thumbnail was created distinct from the original
                 # image. Mirror it as well.
-                mirror.mirror_one(thumbnail, thumbnail_url)
+                mirror.mirror_one(thumbnail, mirror_to=thumbnail_url, collection=collection)
 
         if link_obj.rel in Hyperlink.SELF_HOSTED_BOOKS:
             # If we mirrored book content successfully, remove it from
--- a/model/collection.py
+++ b/model/collection.py
@@ -1,5 +1,6 @@
 # encoding: utf-8
 # Collection, CollectionIdentifier, CollectionMissing
+from abc import ABCMeta, abstractmethod
 
 from sqlalchemy import (
     Column,
@@ -30,7 +31,7 @@ from sqlalchemy.sql.expression import (
 from configuration import (
     ConfigurationSetting,
     ExternalIntegration,
-)
+    BaseConfigurationStorage)
 from constants import EditionConstants
 from coverage import (
     CoverageRecord,
@@ -744,7 +745,12 @@ class Collection(Base, HasFullTableCache):
 
         # Only find books with available licenses or books from self-hosted collections using MirrorUploader
         query = query.filter(
-            or_(LicensePool.licenses_owned > 0, LicensePool.open_access, LicensePool.self_hosted)
+            or_(
+                LicensePool.licenses_owned > 0,
+                LicensePool.open_access,
+                LicensePool.unlimited_access,
+                LicensePool.self_hosted
+            )
         )
 
         # Only find books in an appropriate collection.
@@ -848,3 +854,74 @@ collections_customlists = Table(
     ),
     UniqueConstraint('collection_id', 'customlist_id'),
 )
+
+
+class HasExternalIntegrationPerCollection(object):
+    """Interface allowing to get access to an external integration"""
+
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def collection_external_integration(self, collection):
+        """Returns an external integration associated with the collection
+
+        :param collection: Collection
+        :type collection: core.model.Collection
+
+        :return: External integration associated with the collection
+        :rtype: core.model.configuration.ExternalIntegration
+        """
+        raise NotImplementedError()
+
+
+class CollectionConfigurationStorage(BaseConfigurationStorage):
+    """Serializes and deserializes values as library's configuration settings"""
+
+    def __init__(self, external_integration_association, collection):
+        """Initializes a new instance of ConfigurationStorage class
+
+        :param external_integration_association: Association with an external integtation
+        :type external_integration_association: HasExternalIntegrationPerCollection
+
+        :param collection: Collection object
+        :type collection: Collection
+        """
+        self._integration_owner = external_integration_association
+        self._collection_id = collection.id
+
+    def save(self, db, setting_name, value):
+        """Save the value as as a new configuration setting
+
+        :param db: Database session
+        :type db: sqlalchemy.orm.session.Session
+
+        :param setting_name: Name of the library's configuration setting
+        :type setting_name: string
+
+        :param value: Value to be saved
+        :type value: Any
+        """
+        collection = Collection.by_id(db, self._collection_id)
+        integration = self._integration_owner.collection_external_integration(collection)
+        ConfigurationSetting.for_externalintegration(
+            setting_name,
+            integration).value = value
+
+    def load(self, db, setting_name):
+        """Loads and returns the library's configuration setting
+
+        :param db: Database session
+        :type db: sqlalchemy.orm.session.Session
+
+        :param setting_name: Name of the library's configuration setting
+        :type setting_name: string
+
+        :return: Any
+        """
+        collection = Collection.by_id(db, self._collection_id)
+        integration = self._integration_owner.collection_external_integration(collection)
+        value = ConfigurationSetting.for_externalintegration(
+            setting_name,
+            integration).value
+
+        return value
--- a/model/configuration.py
+++ b/model/configuration.py
@@ -1,9 +1,12 @@
 # encoding: utf-8
 # ExternalIntegration, ExternalIntegrationLink, ConfigurationSetting
-
+import inspect
 import json
 import logging
+from abc import abstractmethod, ABCMeta
+from contextlib import contextmanager
 
+from enum import Enum
 from flask_babel import lazy_gettext as _
 from sqlalchemy import (
     Column,
@@ -176,6 +179,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     OPDS_FOR_DISTRIBUTORS = u'OPDS for Distributors'
     ENKI = DataSourceConstants.ENKI
     FEEDBOOKS = DataSourceConstants.FEEDBOOKS
+    LCP = DataSourceConstants.LCP
     MANUAL = DataSourceConstants.MANUAL
 
     # These protocols were used on the Content Server when mirroring
@@ -215,6 +219,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     # Integrations with STORAGE_GOAL
     S3 = u'Amazon S3'
     MINIO = u'MinIO'
+    LCP = u'LCP'
 
     # Integrations with CDN_GOAL
     CDN = u'CDN'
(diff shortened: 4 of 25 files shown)


opds.py
```diff
@@ -1,74 +1,49 @@
...
+    defaultdict,
+)
 
 from lxml import etree
+from sqlalchemy.orm.session import Session
 
 from cdn import cdnify
-from config import Configuration
 from classifier import Classifier
+from core.model import ExternalIntegration
```
leonardr, line 13: This import style only works when the file is being imported as part of a submodule called `core`. This causes the import problem you see in the tests. Import this class using the 'from model import' sequence a little lower in the file.

marc.py
```diff
@@ -560,11 +560,13 @@ def get_storage_settings(cls, _db):
...
-                cls.SETTING['options'].append(
-                    dict(key=str(integration.id), label=integration.name)
-                )
+            configuration_settings = [s for s in integration.settings if s.key=="marc_bucket"]
+
+            if configuration_settings:
+                if configuration_settings[0].value:
+                    cls.SETTING['options'].append(
+                        dict(key=str(integration.id), label=integration.name)
+                    )
```
leonardr, line 569: How can it happen that a single integration has multiple values for `marc_bucket`?

metadata_layer.py
```diff
@@ -708,7 +708,7 @@ class MetaToModelUtility(object):
 
     log = logging.getLogger("Abstract metadata layer - mirror code")
 
-    def mirror_link(self, model_object, data_source, link, link_obj, policy):
+    def mirror_link(self, model_object, data_source, link, link_obj, policy, collection_id=None):
```
leonardr, line 711: Since the other arguments to this method are data model objects, `collection_id` should also be a data model object called `collection`.

Also, if `model_object` is a `LicensePool` then the only acceptable value for `collection` is `model_object.collection`. Anything else is wrong and should result in an exception or at least be ignored. That is, if a LicensePool is in collection A then it never makes sense to mirror its resources to collection B.

marc.py
vbessonov: It can't. This piece of code checks whether this configuration setting exists.
It's required because I changed the number of settings in `LCPMirror` defined in `circulation`. It doesn't have `marc_bucket` setting

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
         if value is None:
             value = cls.EXCLUDED_AUDIO_DATA_SOURCES_DEFAULT
         return value
+
+
+class ExternalIntegrationOwner(object):
```
leonardr, line 779: "Owner" sounds like a business-level relationship, the way a library "owns" a license to a book. Since this is a mixin class I recommend calling it `HasExternalIntegration`, similar to `HasSelfTests` in `selftest.py`.

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+    """Interface allowing to get access to an external integration"""
+
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def external_integration(self, db):
+        """Returns an external integration owned by this object
```
leonardr, line 786: This is less important in a docstring than in the name of the class, but "associated with" is more accurate than "owned by". A collection doesn't own an S3 integration; the two are tied together.

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+        raise NotImplementedError()
+
+
+class CollectionExternalIntegrationOwner(object):
+    """Interface allowing to get access to an external integration"""
+
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def collection_external_integration(self, db, collection_id):
```
leonardr, line 803: Our general rule is that, whenever possible, we pass around data model objects rather than database session + row ID. You can always get the database session from a data model object with `Session.object_session(obj)`.

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+        :param db: Database session
+        :type db: sqlalchemy.orm.session.Session
+
+        :return: External integration owned by this object
+        :rtype: core.model.configuration.ExternalIntegration
+        """
+        raise NotImplementedError()
+
+
+class CollectionExternalIntegrationOwner(object):
```
leonardr, line 797: Since this is like `HasExternalIntegration` for things that can have a different `ExternalIntegration` for every `Collection`, I suggest calling it `HasExternalIntegrationPerCollection`.

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+    NUMBER = 'number'
+
+    def to_control_type(self):
+        """Converts the value to a attribute type understandable by circulation-web
+
+        :return: String representation of attribute's type
+        :rtype: string
+        """
+        # NOTE: For some reason, circulation-web converts "text" into <text> so we have to turn it into None
+        # In this case circulation-web will use <input>
```
leonardr, line 966: I looked into this but couldn't find the cause. @adriana-alter can you look into this? Is circulation-web creating a `<text>` tag (or maybe `<textarea>`?) when it hears about a configuration setting of type "text"?

vbessonov: I suspect [this line](https://github.com/NYPL-Simplified/circulation-web/blob/c1eaead8a53dab419d760ee9af6f9eacdd6932ef/src/components/ProtocolFormField.tsx#L64)

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+        :type collection_id: int
+
+        :return: External integration owned by the collection
+        :rtype: core.model.configuration.ExternalIntegration
+        """
+        raise NotImplementedError()
+
+
+class BaseConfigurationStorage(object):
+    """Serializes and deserializes values as library's configuration settings"""
```
leonardr, line 819: Here and in `ConfigurationStorage` and `CollectionConfigurationStorage`, the term "library" is used in a way I think is inaccurate. A `ConfigurationSetting` _may_ be associated with an `ExternalIntegration` and a `Library`, but that's not what happens in any of these cases:

* This is an abstract class
* `ConfigurationStorage` is used to store settings on an `ExternalIntegration` alone
*  `CollectionConfigurationStorage` is used to store settings on an `ExternalIntegration` and a `Collection`.

Take out the term "library's"; it would only be appropriate in a class called `LibraryConfigurationStorage`.

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+        :return: Dictionary containing option metadata in the SETTINGS format
+        :rtype: Dict
+        """
+        return {
+            'key': self.key,
+            'label': self.label
+        }
+
+
+class ConfigurationMetadataOwner(object):
```
leonardr, line 1031: I'd prefer something like `HasConfigurationSettings`.

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+        self._category = category
+
+    def __get__(self, owner_instance, owner_type):
+        """Returns a value of the setting
+
+        :param owner_instance: Instance of the owner, class having instance of ConfigurationMetadata as an attribute
+        :type owner_instance: Optional[ConfigurationMetadataOwner]
+
+        :param owner_type: Owner's class
+        :type owner_type: Optional[Type]
```
leonardr, line 1107: `owner_type` isn't used and shouldn't be necessary since you can call `type(owner_instance)`.

leonardr: Never mind -- this is part of the `__get__` contract.

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+            ConfigurationAttribute.DEFAULT.value: self.default,
+            ConfigurationAttribute.OPTIONS.value:
+                [option.to_settings() for option in self.options]
+                if self.options
+                else None,
+            ConfigurationAttribute.CATEGORY.value: self.category
+        }
+
+
+class ConfigurationBucket(ConfigurationMetadataOwner):
```
leonardr, line 1242: Actually since this is the base mixin class, the one we will be seeing throughout  the code, _it_ should be called `HasConfigurationSettings`.

opds.py
```diff
@@ -1538,6 +1513,16 @@ def acquisition_link(cls, rel, href, types):
             indirect_types = []
         link = cls.link(rel, href, initial_type)
         indirect = cls.indirect_acquisition(indirect_types)
+
+        if active_loan and \
+                active_loan.license_pool.collection.protocol == ExternalIntegration.LCP and \
+                active_loan.patron.external_identifier:
+            hashed_passphrase = AtomFeed.makeelement(
```
leonardr, line 1520: Add a reference in a comment to
https://readium.org/lcp-specs/notes/lcp-key-retrieval.html#including-a-hashed-passphrase-in-an-opds-1-catalog
so it's clear this isn't something we made up.

model/configuration.py
```diff
@@ -769,3 +774,567 @@ def excluded_audio_data_sources(cls, _db):
...
+                    [option.to_settings() for option in options_attribute]
+                    if options_attribute
+                    else None,
+                ConfigurationAttribute.CATEGORY.value: category_attribute
+            })
+
+        return settings
+
+
+class ConfigurationFactory(object):
```
leonardr, line 1320: I'm not sure about this; note to myself to come back once I see how it's used.

s3.py
```diff
@@ -131,81 +128,156 @@ class S3Uploader(MirrorUploader):
...
-         },
+        S3UploaderConfiguration.access_key.to_settings(),
+        S3UploaderConfiguration.secret_key.to_settings(),
+        S3UploaderConfiguration.book_covers_bucket.to_settings(),
+        S3UploaderConfiguration.open_access_content_bucket.to_settings(),
+        S3UploaderConfiguration.protected_access_content_bucket.to_settings(),
+        S3UploaderConfiguration.marc_file_bucket.to_settings(),
+        S3UploaderConfiguration.s3_addressing_style.to_settings(),
+        S3UploaderConfiguration.s3_presigned_url_expiration.to_settings(),
+        S3UploaderConfiguration.url_template.to_settings()
```
leonardr, line 280: I thought the purpose of `ConfigurationMetadata.get_configuration_metadata` was to avoid needing to maintain this kind of code.

model/licensing.py
```diff
@@ -206,6 +206,10 @@ def __repr__(self):
             self.licenses_reserved, self.patrons_in_hold_queue
         )
 
+    @property
+    def unlimited_access(self):
+        return self.licenses_owned is None
+
```
leonardr, line 212: This looks like you've added another situation where a book is considered available to be checked out. `Work.to_search_documents` needs to be changed so that an unlimited access book always counts as "available" and "licensed". Otherwise books of this sort won't show up when the end-user is filtering on books that are "available now".

model/configuration.py
vbessonov: Here is a small example of how it's used:
```
with self._configuration_factory.create(
                self._configuration_storage, db, LCPEncryptionConfiguration) as configuration:

            if self._lcpencrypt_exists_locally(configuration):
                result = self._run_lcpencrypt_locally(file_path, identifier, configuration)

                return result
```

The general idea is to move instantiation of `ConfigurationBucket` descendants out of the client code. `ConfigurationBucket` descendants have not to be created directly because `ConfigurationBucket`'s constructor accepts a database session and storing it may lead to the problems (like accessing session aware objects using a stale session). Instead `ConfigurationFactory` abstracts this process and provides a way to instantiate `ConfigurationBucket` descendants using a context manager

model/configuration.py
adriana-alter: I believe so.  Should be no problem to fix, but, just to be totally clear: what do we want it to do instead?

s3.py
vbessonov: `ConfigurationMetadata.get_configuration_metadata` returns all setting attributes and `ConfigurationBucket.to_settings` returns them in SETTINGS format. In the begging I was thinking to use just `ConfigurationBucket.to_settings` but it seems that the order in which setting attributes is not guaranteed, they somehow got mixed by `inspect.get_members`.
There are two solutions:
1. Mention all the attributes explicitly the way it's done now
2. Add `index` attribute to `ConfigurationMetadata` and sort by it

model/configuration.py
vbessonov: I think it should be `<input type="text">`

opds.py
vbessonov: Haven't noticed, fixed it, thank you!

model/configuration.py
vbessonov: Changed it to `HasExternalIntegration`

model/configuration.py
vbessonov: Changed signature to the following:
```
def collection_external_integration(self, collection):
```

model/configuration.py
vbessonov: Changed it to `HasExternalIntegrationPerCollection`

model/configuration.py
vbessonov: Changed it accordingly

model/configuration.py
vbessonov: Changed it to `HasConfigurationSettings`

model/configuration.py
vbessonov: Changed it to `ConfigurationGrouping`

opds.py
vbessonov: Added a comment

model/configuration.py
vbessonov: Changed the comment accordingly

model/configuration.py
adriana-alter: Will do!  Can it go into the next sprint, or is it time-sensitive?

marc.py
leonardr: I see, so a single integration might have _less_ than one value for `marc_bucket`.

model/configuration.py
leonardr: I think next sprint should be fine. @vbessonov add a TODO to change this back after Adriana makes the change in circulation-web, file a ticket in JIRA and assign it to her.

model/configuration.py
leonardr: OK, I don't think this is how I'd do it but it makes sense.

s3.py
leonardr: Let's stick with #1 for now. IMO this is the biggest benefit of the whole `ConfigurationGrouping` thing so I'm a little upset that it's not working, but we can fix it later.

s3.py
```diff
@@ -14,6 +14,8 @@
 from enum import Enum
 from flask_babel import lazy_gettext as _
 
+from core.model.configuration import ConfigurationOption, ConfigurationGrouping, ConfigurationMetadata, \
+    ConfigurationAttributeType
```
leonardr, line 18: Same import problem as I mentioned before. See below for `from model import` example.

model/configuration.py
vbessonov: Sure, created [SIMPLY-3008](https://jira.nypl.org/browse/SIMPLY-3008) assigned it to Adriana and mentioned it in a TODO in the code.

metadata_layer.py
vbessonov: Removed `collection_id`, added code for fetching collection dynamically:
```
# Mirror it.
collection = pools[0].collection if pools[0].collection else None
mirror.mirror_one(representation, mirror_to=mirror_url, collection=collection)
```

s3.py
vbessonov: I added `index` to `ConfigurationMetadata` and replaced `S3.SETTINGS` with `S3UploaderConfiguration.to_settings()`

model/configuration.py
```diff
@@ -1097,6 +1038,12 @@ def __init__(self, key, label, description, type, required=False, default=None,
         self._options = options
         self._category = category
 
+        if index is not None:
+            self._index = index
+        else:
+            ConfigurationMetadata._counter += 1
+            self._index = ConfigurationMetadata._counter
+
```
leonardr, line 1046: This is a good solution.

leonardr:
This looks good to merge once the circulation branch is ready as well.

vbessonov:
I rebased onto `master` and squashed all the commits into a single one so it doesn't pollute the history

model/collection.py
```diff
@@ -745,7 +745,12 @@ def restrict_to_ready_deliverable_works(
 
         # Only find books with available licenses or books from self-hosted collections using MirrorUploader
         query = query.filter(
-            or_(LicensePool.licenses_owned > 0, LicensePool.open_access, LicensePool.self_hosted)
+            or_(
+                LicensePool.licenses_owned > 0,
+                LicensePool.open_access,
+                LicensePool.unlimited_access,
```
leonardr, line 751: I'm amazed that this works, since `unlimited_access` is not a database field, but I checked and it does work.

leonardr:
This is ready for merge after I'm able to release 3.0.11

model/collection.py
vbessonov: It works because of `@hybrid_property` decorator, it makes it act like a real database field

model/licensing.py
vbessonov: Added additional logic as requested

PR #1195, opened 2020-09-04
Always go to the database to find out whether the site configuration has been updated; caching this information locally is false economy

leonardr:
This branch changes the default `timeout` for `site_configuration_last_update` to zero. Previously this value was `None`, meaning that the actual value was controlled by a database setting. But acquiring the database setting caused potentially serious warnings -- conceivably even causing [SIMPLY-2983](https://jira.nypl.org/browse/SIMPLY-2983).

Here's what I'm more certain about: any value for `timeout` other than zero caused problems on any site with multiple app servers: specifically problems like [SIMPLY-2991](https://jira.nypl.org/browse/SIMPLY-2991) and [SIMPLY-2514](https://jira.nypl.org/browse/SIMPLY-2514). Setting the value to zero makes these problems go away with a minimal effect on site performance. So I think we should peg this value to zero; that's what this branch does.

This branch takes a minimally invasive approach -- the configuration setting is still around, as is the code to fetch its value, but we never actually fetch it. That's so we can change back easily if this turns out to be a mistake.

The other change I made was renaming the `timeout` variable in `site_configuration_has_changed` to `cooldown`. Previously we had two related timeouts: `site_configuration_last_update` had a `timeout` that controlled how often we went to the database to check a certain Timestamp, and `site_configuration_has_changed` had a `timeout` (now called `cooldown`) that controlled how often we wrote new values to that Timestamp. I don't think `cooldown` is behind any SIMPLY-2991-like problems, because the amounts of time are so small -- less than a second -- but theoretically we could peg `cooldown` to zero with minimal effect on performance, since configuration changes are so rare. It shouldn't make a difference if we end up writing to that Timestamp two or three times in a single request.

### DIFF
--- a/config.py
+++ b/config.py
@@ -519,14 +519,22 @@ class Configuration(ConfigurationConstants):
         :return: a datetime object.
 
         """
+
         now = datetime.datetime.utcnow()
 
+        # NOTE: Currently we never check the database (because timeout is
+        # never set to None). This code will hopefully be removed soon.
         if _db and timeout is None:
             from model import ConfigurationSetting
             timeout = ConfigurationSetting.sitewide(
                 _db, cls.SITE_CONFIGURATION_TIMEOUT
             ).int_value
+
         if timeout is None:
+            # NOTE: this only happens if timeout is explicitly set to
+            # None _and_ no database value is present. Right now that
+            # never happens because timeout is never explicitly set to
+            # None.
             timeout = 60
 
         last_check = cls.instance.get(
--- a/config.py
+++ b/config.py
@@ -478,6 +478,10 @@ class Configuration(ConfigurationConstants):
 
     # A sitewide configuration setting controlling *how often* to check
     # whether the database configuration has changed.
+    #
+    # NOTE: This setting is currently not used; the most reliable
+    # value seems to be zero. Assuming that's true, this whole
+    # subsystem can be removed.
     SITE_CONFIGURATION_TIMEOUT = 'site_configuration_timeout'
 
     # The name of the service associated with a Timestamp that tracks
--- a/config.py
+++ b/config.py
@@ -495,7 +495,7 @@ class Configuration(ConfigurationConstants):
 
     @classmethod
     def site_configuration_last_update(cls, _db, known_value=None,
-                                       timeout=None):
+                                       timeout=0):
         """Check when the site configuration was last updated.
 
         Updates Configuration.instance[Configuration.SITE_CONFIGURATION_LAST_UPDATE].
@@ -508,8 +508,9 @@ class Configuration(ConfigurationConstants):
 
         :param timeout: We will only call out to the database once in
             this number of seconds. If we are asked again before this
-            number of seconds elapses, we will assume site configuration
-            has not changed.
+            number of seconds elapses, we will assume site
+            configuration has not changed. By default, we call out to
+            the database every time.
 
         :return: a datetime object.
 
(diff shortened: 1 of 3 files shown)


leonardr:
@EdwinGuzman I'm going to ask @jonathangreen to do the formal review of this branch but I'd like you to take a look at well.

jonathangreen:
This code looks okay to me. I think the minimally invasive approach is a bit confusing, since at first glance it still looks like the default timeout is set for 60, but we never get to that code because we are passing in `0` instead of `None`. https://github.com/NYPL-Simplified/server_core/blob/3dceb9aa1f2c614daa4acf149b09e479e691580d/config.py#L529-L530

I'm getting my test environment setup and I'm going to make sure this stops the warning from appearing, then I will approve.

jonathangreen (approved):
Clarifying comments look good to me. 👍🏻

I gave this a test in the LYRASIS test environment. I was running a script against the circulation managers to constantly change lane order. Before the change I was seeing the warning consistently. For ~100 lane order changes I'd see the warning ~4 times. After this patch I ran though 300 lane order changes and didn't see any warnings in the logs.

tdilauro:
@leonardr FYI. This change broke a test in circulation manager:
- `tests.test_controller.TestBaseController.test_library_for_request_reloads_settings_if_necessary`

```
2020-09-09 11:38:11,094:root:INFO:config.py:Connecting to database: postgres://simplified:***@localhost:5432/simplified_circ_test
2020-09-09 11:38:12,470:root:INFO:config.py:Connecting to database: postgres://simplified:***@localhost:5432/simplified_circ_test
{"host": "Mule.local", "name": "root", "level": "INFO", "timestamp": "2020-09-09T15:38:14.565178", "app": "simplified", "message": "Permanent work ID for 4718: Quite British/Bull, John -> quite british/bull john/book -> dceb1c57-5042-edc8-1879-1e00243a2824 (was None)", "filename": "edition.py"}
{"host": "Mule.local", "name": "root", "level": "INFO", "timestamp": "2020-09-09T15:38:14.618500", "app": "simplified", "message": "Presentation changed for Edition Quite British (by John Bull, pub=None, ident=Gutenberg ID/2001, pwid=dceb1c57-5042-edc8-1879-1e00243a2824, language=eng, cover=None)", "filename": "edition.py"}
{"host": "Mule.local", "name": "Mock external search index", "level": "INFO", "timestamp": "2020-09-09T15:38:14.734530", "app": "simplified", "message": "Created 1 search documents in 0.02 seconds", "filename": "external_search.py"}
{"host": "Mule.local", "name": "Mock external search index", "level": "INFO", "timestamp": "2020-09-09T15:38:14.734604", "app": "simplified", "message": "Uploaded 1 search documents in  0.00 seconds", "filename": "external_search.py"}
{"host": "Mule.local", "name": "Mock external search index", "level": "INFO", "timestamp": "2020-09-09T15:38:14.734660", "app": "simplified", "message": "Successfully indexed 1 documents, failed to index 0.", "filename": "external_search.py"}
E
======================================================================
ERROR: tests.test_controller.TestBaseController.test_library_for_request_reloads_settings_if_necessary
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/timmo/.PyEnvs/circulation-nQEl9FH5/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/Shared/src/lyrasis/circulation/tests/test_controller.py", line 1186, in test_library_for_request_reloads_settings_if_necessary
    model.site_configuration_has_changed(self._db, timeout=0)
TypeError: site_configuration_has_changed() got an unexpected keyword argument 'timeout'

----------------------------------------------------------------------
Ran 1 test in 2.303s

FAILED (errors=1)
```

PR #1197, opened 2020-09-15
Peg feedparser to a version that supports Python 2.

leonardr:
Feedparser 6.0.0 was just released, and drops support for Python 2. This branch pegs our feedparser to a version that supports Python 2.

We're getting pretty close to being able to migrate to Python 3, so hopefully this won't keep happening for much longer.


tdilauro (approved):
LGTM

PR #1198, opened 2020-09-15
Adding qa_feed function to return a feed of works based on filters.

EdwinGuzman:
@leonardr still a WIP but can you review this and https://github.com/NYPL-Simplified/circulation/pull/1490?

I noticed when trying to use a `SearchFacets` object that it applied only when calling `query_works_multi` so only the last set of facets combination was applied. Instead, I'm adding a filter which seems to return what is specified. One thing missing is that it's on a per library basis rather than including all potential libraries in the database.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1997,6 +1997,43 @@ class WorkList(object):
                 yield work, lane
 
 
+class JackpotWorkList(WorkList):
+    def initialize(self, library, filter_list):
+        super(JackpotWorkList, self).initialize(library)
+        self.library = library
+        self.filter_list = filter_list
+
+    def groups(self, _db, include_sublanes=True, facets=None,
+               search_engine=None, debug=False, **response_kwargs):
+        """
+        :param facets: A FeaturedFacets object that may restrict the works on view.
+        :param search_engine: An ExternalSearchIndex to use when
+            asking for the featured works in a given WorkList.
+        :param debug: A debug argument passed into `search_engine` when
+            running the search.
+        :yield: A sequence of (Work, WorkList) 2-tuples, with each
+            WorkList representing the child WorkList in which the Work is
+            found.
+        """
+        pagination = Pagination(size=5)
+        queries = []
+        for (collection, medium, availability) in self.filter_list:
+            from external_search import Filter
+            facets = Facets.default(self.library, availability=availability)
+            filter = Filter.from_worklist(_db, self, facets)
+            # filter = Filter(collections=collection, media=medium, languages=["eng"])
+            # filter.availability = availability
+            queries.append((None, filter, pagination))
+
+        resultsets = list(search_engine.query_works_multi(queries))
+        l = Lane()
+        results = l.works_for_resultsets(_db, resultsets)
+
+        flat_list = [item for sublist in results for item in sublist]
+
+        yield None, flat_list
+
+
 class DatabaseBackedWorkList(WorkList):
     """A WorkList that can get its works from the database in addition to
     (or possibly instead of) the search index.
--- a/lane.py
+++ b/lane.py
@@ -2016,22 +2016,25 @@ class JackpotWorkList(WorkList):
             found.
         """
         pagination = Pagination(size=5)
-        queries = []
         for (collection, medium, availability) in self.filter_list:
+            # Set up getting the filter to use for the es query.
             from external_search import Filter
             facets = Facets.default(self.library, availability=availability)
             filter = Filter.from_worklist(_db, self, facets)
-            # filter = Filter(collections=collection, media=medium, languages=["eng"])
-            # filter.availability = availability
-            queries.append((None, filter, pagination))
+            filter.media = medium
 
-        resultsets = list(search_engine.query_works_multi(queries))
-        l = Lane()
-        results = l.works_for_resultsets(_db, resultsets)
+            query = (None, filter, pagination)
 
-        flat_list = [item for sublist in results for item in sublist]
+            # Create a Worklist for every query
+            wl = WorkList()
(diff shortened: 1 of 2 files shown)


leonardr:
I was thinking of a different way to do this, but now that I see what you've done I'm not sure which way is less work.

My idea was to create a `WorkList` for every lane you want in the grouped feed, then make them all children of one big `Worklist`, and call `Worklist.groups` on the big `Worklist`. That would make the Elasticsearch query and assemble the results into the form you need to create a grouped feed.

Now that I look closer there are two problems with my idea:

1) `Worklist.groups` will end up making a separate query for every child of the `Worklist`, because it's only optimized to use `query_works_multi` on `Lane` objects and you'd be creating regular `Worklist`s. So it won't be as fast as it could be.

2) `WorkList.groups` assumes that all of the child `Worklist`s are using the same `Facets` object, and you're mixing different facets. Specifically, your 'lanes' have different availability settings: some are for books that are currently available and some are for books that are not. That information is controlled by the `Facets` object. (Setting `filter.availability` won't do anything; I assume that's placeholder code.)

Since we're not super concerned with performance (and we can optimize later) this might still be the way to go. You can call `WorkList.groups` twice, once with a `Facets` that says "only books that are available" and once with a `Facets` that says "only books that are not available". Or maybe we can allow a `WorkList` to contain a `Facets` object which is always used in its query, overriding any other `Facets` object.

To avoid confusion will follow up with a second comment about the code you actually wrote.

leonardr:
After going over your code again I like your solution a lot and I think it's really close to working, so forget about my solution, at least for now. I see just two issues:

1. the `filter.availability` thing I mentioned earlier. For "currently available" you need to pass in a regular `Facets` object that contains this information. For "not currently available" it's a bit tricky since there's no other place where we're specifically looking for books that are _not_ available. However this can wait for a follow-up branch since the testers' most crucial need is for books they can check out and open. (I think the easiest way to do this would be to create an `UnavailableTitlesFacets` subclass of `Facets` that is hard-coded to find only unavailable books.)

2. You're creating a regular `AcquisitionFeed` and we eventually want to get a grouped feed. This can also wait for a follow-up branch, but look at `AcquisitionFeed.groups`. You'll eventually want to plug into that system. Possibly by defining a special `JackpotWorklist` class that defines `groups` using code like the code you have in this PR. I think this is simple enough that you can put it in this branch rather than waiting for a follow-up, if only so you don't have to rewrite the tests.

leonardr:
"per-library basis" is exactly what we want -- the QA scripts are going to be run in the context of a specific library with patron credentials for that library. So you've got that part right.

EdwinGuzman:
I haven't tackled the Filter/Facets and availability issue yet, but I created the `JackpotWorkList` and I'm using it with `AcquisitionFeed.groups`. One thing I can't figure out is what `JackpotWorkList` should return. The `flat_list` contains all the works from the es query that was run, but I'm missing the next step for that.

leonardr:
The idea is that one test case becomes one query which becomes one "sublane" of the grouped feed. So try creating a different WorkList for each (collection, medium, availability) tuple and yielding that along with each Work in the corresponding result set.

That would then render in SimplyE like:

Overdrive - Book - now 
[lane of books]

Overdrive - Book - unavailable
[lane of books]

Overdrive - Audio - now 
[lane of books]

Overdrive - Audio - unavailable
[lane of books]

* WorkLists can be completely transient -- they don't need to correspond to anything in the database and you can just create them on the fly.
* However you might need to set `MAX_CACHE_AGE=CachedFeed.IGNORE_CACHE` for this WorkList and the other ones you create to populate it, since we don't want this being stored in the `cachedfeeds` table.
* The names of the WorkLists can just be string representations of the 3-tuples that created them.
* Each WorkList should have the jackpost worklist (`self`) as its `.parent`. I don't know if that makes a difference but just in case.

EdwinGuzman:
The last feedback was helpful and made sense.  I couldn't add jackpotWorkList as the parent to each workList, however. This is a result of the feed from my library (using `circulation-patron-web`) but not showing the other collections and lanes:

Screen Shot 2020-09-29 at 5 03 30 PM

lane.py
```diff
@@ -1997,6 +1997,46 @@ def _featured_works_with_lanes(
...
+        :yield: A sequence of (Work, WorkList) 2-tuples, with each
+            WorkList representing the child WorkList in which the Work is
+            found.
+        """
+        pagination = Pagination(size=5)
+        for (collection, medium, availability) in self.filter_list:
+            # Set up getting the filter to use for the es query.
+            from external_search import Filter
+            facets = Facets.default(self.library, availability=availability)
+            filter = Filter.from_worklist(_db, self, facets)
```
leonardr, line 2023: I don't think this is a reason to delay the merge any longer, but you're taking in a `facets` argument that's being ignored. I think eventually we might want to use the `facets` argument but for now can you add a comment to the docstring that says the value is ignored?

If you want to try actually using `facets`, you can do something like `these_facets = facets.navigate(availability=availability)` inside the for loop; I think that should work.

leonardr (approved):
This is ready to merge; the `facets` thing can wait if you don't have time.

leonardr (changes requested):
Wait, I totally missed the fact that there are no tests here. Do you have time to write them? Otherwise I'll take this branch over. The QA team really needs this feed.

EdwinGuzman:
I don't think I'll have time to write them from now until Friday, so if you can take over that'd be great. Thanks :\

leonardr:
This code was merged into https://github.com/NYPL-Simplified/circulation/pull/1490 and reworked.

PR #1199, opened 2020-09-21
Explicitly set version numbers for all Python dependencies

leonardr:
This branch is the core portion of https://jira.nypl.org/browse/SIMPLY-3100. All Python dependencies are set to the current latest releases that support Python 2. This will eliminate random breakage in the future caused by dependency releases, especially releases that drop Python 2 support.

I came up with this list by creating a fresh virtual environment and using `pip freeze`.

```
virtualenv env
source env/bin/activate
pip install --use-feature=2020-resolver -r requirements.txt # This is he old requirements.txt
pip freeze > requirements.txt # This is the new one
```

Then I edited the new requirements.txt to preserve comments found in the old one.

In both this branch and the circulation branch I also had to remove this line inserted by `pip freeze`:

```
pkg-resources==0.0.0
```

This looks like [a bug in Ubuntu](https://github.com/pypa/pip/issues/4022).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ elasticsearch-dsl==6.4.0
 enum34==1.1.10
 feedparser==5.2.1
 Flask==1.1.2
-Flask-Babel==2.0.0
+Flask-Babel==1.0.0
 funcsigs==1.0.2
 futures==3.3.0
 fuzzywuzzy==0.18.0 # For author name manipulations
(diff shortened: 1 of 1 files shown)


EdwinGuzman (approved):
`requirements.txt` is case insensitive? There are some changes in the new file but it probably doesn't matter.

leonardr:
Python package names in PyPI are case-insensitive. but they have a preferred capitalization which `pip freeze` prints out.

PR #1200, opened 2020-09-21
Add a database index for loan's external_identifier field

vbessonov:
## Description

This PR adds a new database index for `loans`'s `external_identifier` field.

## Motivation and Context

This index is required in the case of LCP when we are not able to use a library name as a part of endpoint URLs and have to use LCP license ID (loan's external identifier) to fetch a library: loan -> patron -> library.

## Checklist:

- [x] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.

### DIFF
--- a/migration/20200925-add-index-on-loan-external_identifier.sql
+++ b/migration/20200925-add-index-on-loan-external_identifier.sql
@@ -1,0 +1,1 @@
+CREATE index ix_loan_external_identifier on loans (external_identifier);
--- a/model/patron.py
+++ b/model/patron.py
@@ -338,6 +338,8 @@ class Loan(Base, LoanAndHoldMixin):
         start = self.start or datetime.datetime.utcnow()
         return start + default_loan_period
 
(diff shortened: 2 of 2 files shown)


leonardr (changes requested):
This needs a migration script. Look at 20191022-unique-credentials.sql for an example. Your implementation will be simpler since there's only one index and it's not required that the external identifier be unique.

vbessonov:
@leonardr, I added a migration creating an index on `loans.external_identifier`.
The example you provided was creating indices using BEGIN/EXCEPTION statements but I din't use them because I thought it's not required in this case.

leonardr:
Please do add the BEGIN...EXCEPT thing. We want to make migration scripts idempotent as much as possible. Sometimes migration script suites are run multiple times if there was a crash or error during migration, and this makes it smoother.

vbessonov:
@leonardr, thank you for adding the safety to code to the migration. Could we merge this PR now?

PR #1201, opened 2020-09-30
Support code for "no access outside your root lane" feature

leonardr:
This is a core support branch for https://jira.nypl.org/browse/SIMPLY-3125.

* The unsupported "lending policy" feature is removed and reimplemented by expanding a feature that _is_ supported -- patron root lanes.
* I added a cache for the result of `Library.has_root_lanes`, a somewhat complicated database query that is now going to be called on almost every incoming circulation manager request.
* I added new methods `Work.age_appropriate_for_patron` and `Patron.work_is_age_appropriate`. These methods (more precisely, the helper method `Patron.age_appropriate_match`) implement a reasonable interpretation of what it means for a work (which has an audience and a target age) to be "age-appropriate" for a patron (whose root lane may have _multiple_ audiences as well as a target age). I needed both methods because when we're calculating Worklist visibility (below) we need to answer a _general_ age-appropriateness question, without having any specific Work in mind.
* I added a new method, `WorkList.accessible_to` and `Lane.accessible_to`, which control whether a WorkList is accessible to a given patron. Basically, WorkLists are accessible to a patron unless they _might_ contain age-inappropriate material. (This lets us give Open eBooks access to related-books lanes such as "books by the same author" lanes) Lanes are subject to the same rules, but they must _also_ be in scope from the patron's root lane.


app_server.py
```diff
@@ -52,20 +52,6 @@ def cdn_url_for(*args, **kwargs):
...
-        logging.info("Lending policy: %s", policy)
-        policy = json.loads(policy)
-    for external_type, p in policy.items():
-        if Patron.AUDIENCE_RESTRICTION_POLICY in p:
-            for audience in p[Patron.AUDIENCE_RESTRICTION_POLICY]:
-                if not audience in Classifier.AUDIENCES:
-                    raise ValueError(
-                        "Unrecognized audience in lending policy: %s" %
-                        audience)
-    return policy
```
leonardr, line 68: This unsupported feature has been removed, and reimplemented by expanding a feature that _is_ supported: patron root lanes.

model/patron.py
```diff
@@ -216,25 +213,6 @@ def works_on_loan_or_on_hold(self):
         loans = self.works_on_loan()
         return set(holds + loans)
 
-    def can_borrow(self, work, policy):
```
leonardr, line 219: This is the only place where the lending policy was applied. Logic around whether or not a patron can borrow a book now lives in circulation, although the main component piece of that logic -- whether a real or hypothetical work is age-appropriate for a patron -- is now in Patron.

model/patron.py
```diff
@@ -300,6 +278,48 @@ def synchronize_annotations(self, value):
                 _db.delete(annotation)
         self._synchronize_annotations = value
 
+    @property
+    def root_lane(self):
+        """Find the Lane, if any, to be used as the Patron's root lane.
```
leonardr, line 287: This code was moved in from circulation. It's a bit tricky because the patron is not directly associated with a Lane; instead, a Lane may be associated with a _list_ of patron types.

model/work.py
```diff
@@ -781,6 +781,120 @@ def language_code(self):
...
+            # book in the system.
+            return True
+
+        # The patron can interact with this Work if any of the audiences
+        # in their root lane (in conjunction with the root lane's target_age)
+        # would match the Work.
+        return any(
+            self.age_appropriate_for(audience, root.target_age)
+            for audience in root.audiences
+        )
```
leonardr: This is complicated but it's necessary to handle real cases like audiences=[Children, Young Adult], target_age=[9,13].

model/work.py
```diff
@@ -781,6 +781,120 @@ def language_code(self):
...
+
+        if (reader_audience == Classifier.AUDIENCE_YOUNG_ADULT
+            and self.audience in (Classifier.AUDIENCES_YOUNG_CHILDREN)):
+            # A YA reader can see any children's title.
+            return True
+
+        if (reader_audience in (Classifier.AUDIENCES_YOUNG_CHILDREN)
+            and self.audience == Classifier.AUDIENCE_YOUNG_ADULT):
+            # Children cannot see any YA title.
+            return False
```
leonardr: Keep in mind that this method will be called once for every audience in the lane's `.audiences`. Consider the case of 
audiences=[Children, Young Adult], target_age=[9,13], with a YA book with target_age=(12,15).

The first call will be (Children, [9,13]) and it will fail, because children cannot access YA titles even if the age range overlaps. The second call will be (Young Adult, [9,13]) and it will succeed, because a YA reader can access a YA title if the age range overlaps.

requirements.txt
```diff
@@ -8,6 +8,7 @@ click==7.1.2
 elasticsearch==6.8.1
 elasticsearch-dsl==6.4.0
 enum34==1.1.10
+expiringdict==1.2.1
```
leonardr, line 11: This is the same version of expiringdict used in circulation.

model/patron.py
```diff
@@ -300,6 +279,176 @@ def synchronize_annotations(self, value):
...
+            # Multiple root lanes for a patron indicates a
+            # configuration problem, but we shouldn't make the patron
+            # pay the price -- just pick the first one.
+            logging.error(
+                "Multiple root lanes found for patron type %s.",
+                patron.external_type
+            )
+        return lanes[0]
+
+    def work_is_age_appropriate(self, work_audience, work_target_age):
```
leonardr, line 324: Note to myself on Tuesday: This is where I left off reviewing the code and making sure everything has test coverage.

leonardr: Thank you, past self.

tests/test_opds.py
```diff
@@ -92,7 +92,11 @@ def test_authors(self):
         edition.add_contributor(
             "Jonathan Frakes", Contributor.NARRATOR_ROLE
         )
-        author, contributor = Annotator.authors(None, edition)
+        author, contributor = sorted(
+            Annotator.authors(None, edition),
+            key=lambda x: x.tag
+        )
+
```
leonardr, line 99: This change is unrelated. I got a spurious test failure because the tags come out of `authors` in a nondeterministic order.

classifier/__init__.py
```diff
@@ -86,8 +86,8 @@ class Classifier(object):
     # "All ages" actually means "all ages with reading fluency".
     ALL_AGES_AGE_CUTOFF = 8
 
-    AUDIENCES_JUVENILE = [AUDIENCE_CHILDREN, AUDIENCE_YOUNG_ADULT,
-                          AUDIENCE_ALL_AGES]
+    AUDIENCES_YOUNG_CHILDREN = [AUDIENCE_CHILDREN, AUDIENCE_ALL_AGES]
```
vbessonov, line 89: Why do we need to have `AUDIENCE_ALL_AGES` in `AUDIENCES_YOUNG_CHILDREN` and in `AUDIENCES_JUVENILE`?

model/patron.py
```diff
@@ -300,6 +282,188 @@ def synchronize_annotations(self, value):
...
+           Classifier, representing the general reading audience to
+           which the reader belongs.
+
+        :param reader_age: A number or 2-tuple representing the age or
+           age range of the reader.
+        """
+        if reader_audience is None:
+            # A patron with no particular audience restrictions
+            # can see everything.
+            return True
```
vbessonov, line 391: Would it make sense to add logs with the `DEBUG` level so later it would be easier to troubleshoot this code?

model/library.py
```diff
@@ -292,6 +298,36 @@ def enabled_facets_setting(self, group_name):
...
+        # (or vice versa) might not see the change take effect without
+        # a server restart.
+        value = Library._has_root_lane_cache.get(self.id, None)
+        if value is None:
+            from ..lane import Lane
+            _db = Session.object_session(self)
+            root_lanes = _db.query(Lane).filter(
+                Lane.library==self
+            ).filter(
+                Lane.root_for_patron_type!=None
```
vbessonov, line 325: [PEP 8 Style Guide](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html) recommends to use `is` to compare something to `None`

vbessonov (changes requested):
It looks really good, I just proposed a couple of minor changes

classifier/__init__.py
leonardr: AUDIENCE_ALL_AGES is for books that should be available to young children, young adults, and adults. So it needs to appear in every list of audiences -- it's also in AUDIENCES_ADULT.

model/library.py
leonardr: That's true, but this isn't really performing a comparison; it's creating a SQLAlchemy `BinaryExpression` object that gets turned into SQL like `Lane.root_for_patron_type is not NULL`. So using != is correct in this case.

model/patron.py
leonardr: Good idea.

vbessonov (approved):

PR #1205, opened 2020-10-25
Add OPDS 2.0 import support

vbessonov:
## Description

This PR adds OPDS 2.0 support to Circulation Manager:
- `OPDS2Importer`
- `OPDS2ImportMonitor`

## Motivation and Context

[SIMPLY-2901](SIMPLY-2901)

## Checklist:


opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+    ):
+        """Initialize a new instance of OPDS2Importer class.
+
+        :param db: Database session
+        :type db: sqlalchemy.orm.session.Session
+
+        :param collection: Circulation Manager's collection.
+            LicensePools created by this OPDS2Import class will be associated with the given Collection.
+            If this is None, no LicensePools will be created -- only Editions.
+        :type collection: sqlalchemy.orm.session.Session
```
leonardr: This type should be `Collection`

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+
+        :param content_modifier: A function that may modify-in-place representations (such as images and EPUB documents)
+            as they come in from the network.
+        :type content_modifier: Callable
+
+        :param map_from_collection: Identifier mapping
+        :type map_from_collection: Dict
+
+        :param mirrors: A dictionary of different MirrorUploader objects for different purposes
+        :type mirrors: List[MirrorUploader]
```
leonardr: Should be a Dict.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+        """
+        self._logger.debug(
+            u"Started extracting subject metadata from {0}".format(encode(subject))
+        )
+
+        subject_metadata_list = []
+
+        if is_string(subject):
+            subject_metadata = SubjectData(
+                type=None, identifier=None, name=subject, weight=1
```
leonardr: I don't like the idea of creating a subject with no type. Should this be Subject.TAG as well?

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+        medium = self._extract_medium(publication, derived_medium)
+
+        publisher = first_or_default(publication.metadata.publishers)
+        if publisher:
+            publisher = publisher.name
+
+        published = publication.metadata.published
+        subjects = self._extract_subjects(publication.metadata.subjects)
+        contributors = (
+            self._extract_contributors(publication.metadata.authors)
```
leonardr: Whenever you call _extract_contributors you should pass in a role constant corresponding to the part of the metadata you're extracting.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+
+        contributor_metadata_list = []
+
+        if is_string(contributor):
+            contributor_metadata = ContributorData(
+                sort_name=None,
+                display_name=contributor,
+                family_name=None,
+                wikipedia_name=None,
+                roles=None,
```
leonardr: I assume this is the situation where the OPDS feed lists a single string as the contributor. When you call _extract_contributors I mention that you should pass in a default role; this is where the default role should be used.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+            )
+
+            contributor_metadata_list.append(contributor_metadata)
+        elif isinstance(contributor, core_ast.Contributor):
+            contributor_metadata = ContributorData(
+                sort_name=contributor.sort_as,
+                display_name=contributor.name,
+                family_name=None,
+                wikipedia_name=None,
+                roles=contributor.roles if contributor.roles else None,
```
leonardr: Again, the default role should be used here -- it looks like when parsing an OPDS 2 feed we always know what the role is supposed to be.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+
+        self._logger.debug(
+            u"Finished extracting contributor metadata from {0}: {1}".format(
+                encode(subject), encode(subject_metadata_list)
+            )
+        )
+
+        return subject_metadata_list
+
+    def _extract_contributors(self, contributor):
```
leonardr: I think this should take a default role.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+        """Extract a list of ContributorData objects from the webpub-manifest-parser's contributor.
+
+        :param contributor: Parsed contributor object
+        :type contributor: Union[
+            str,
```
leonardr: I kind of assumed the parser would always give us a list of `core_ast.Contributor` objects, complete with role, and we wouldn't have to handle the "string" case here.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+
+        :return: Link metadata
+        :rtype: LinkData
+        """
+        self._logger.debug(
+            u"Started extracting link metadata from {0}".format(encode(link))
+        )
+
+        # FIXME: It seems that OPDS 2.0 spec doesn't contain information about rights so we use the default one.
+        rights_uri = RightsStatus.rights_uri_from_string("")
```
leonardr, line 222: If it helps, `rights_uri` is only relevant for actual copies of the book -- basically when the link relation is http://opds-spec.org/acquisition/open-access. In that case the default rights_uri can be GENERIC_OPEN_ACCESS. Maybe http://opds-spec.org/acquisition/ is also relevant for self-hosted books that are not open-access; in that case the default rights_uri can be IN_COPYRIGHT.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+            )
+            image_links.append(cover_link)
+
+        if len(sorted_raw_image_links) > 1:
+            cover_link = self._extract_link(
+                sorted_raw_image_links[1],
+                feed_self_url,
+                default_link_rel=Hyperlink.THUMBNAIL_IMAGE,
+            )
+            image_links.append(cover_link)
```
leonardr, line 330: I see what you're doing -- pick the biggest image to be the main one and a smaller version to be the thumbnail -- but I don't think this will work in general. There's no guarantee that these images have the same media type, or that the second-largest image isn't far too large to use as a thumbnail.

At the same time I don't know of a better solution. Ideally the CM would import the biggest image and make its own thumbnail, but there's no guarantee that the CM has any way of mirroring thumbnails, since that's an optional feature.

We have a few constants for the ideal size and aspect ratio for a thumbnail image:

```
    IDEAL_COVER_ASPECT_RATIO = 2.0/3
    IDEAL_IMAGE_HEIGHT = 240
    IDEAL_IMAGE_WIDTH = 160
```

So put in a TODO to fix this up using those constants. Instead of using the second-largest image as a thumbnail, find the image that would make the best thumbnail because of its dimensions, media type, and aspect ratio.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+                    nested_acquisition_object = acquisition_object
+
+                    while nested_acquisition_object.child:
+                        nested_acquisition_object = first_or_default(
+                            acquisition_object.child
+                        )
+
+                    drm_scheme = DeliveryMechanism.NO_DRM
+
+                    if acquisition_object.type == DeliveryMechanism.ADOBE_DRM:
```
leonardr: Can we avoid this if/else tree by making a set in DeliveryMechanism containing all the known DRM types?

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+            )
+        )
+
+        media_types_and_drm_scheme = []
+
+        if link.properties:
+            if (
+                link.properties.availability.state
+                == opds2_ast.OPDS2AvailabilityType.AVAILABLE.value
+            ):
```
leonardr, line 400: I see this in your example feed in python-webpub-manifest parser but I don't think this is part of OPDS 2 or RWPM. What does this mean, and why do you ignore format information if this information is missing or its value is other than 'available'?

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+                media_types_and_drm_scheme.append((link.type, DeliveryMechanism.NO_DRM))
+
+        self._logger.debug(
+            u"Finished extracting media types and a DRM scheme from {0}: {1}".format(
+                encode(link), encode(media_types_and_drm_scheme)
+            )
+        )
+
+        return media_types_and_drm_scheme
```
leonardr, line 441: Note to myself: stopping here for the evening.

requirements.txt
```diff
@@ -55,3 +55,4 @@ uWSGI==2.0.19.1
 # watchtower is for Cloudwatch logging integration
 watchtower==0.8.0
 Werkzeug==1.0.1
+webpub-manifest-parser
```
vbessonov, line 58: As for now I uploaded `webpub-manifest-parser` to my personal PyPi account, once it's merged I'll move it to NYPL's account

opds2_import.py
vbessonov: It's [a part](https://github.com/opds-community/drafts/blob/master/schema/properties.schema.json#L80) of the OPDS 2.0 schema. I found a lot of examples of using `availability` property in the Pret Numerique feed.

For example, the example below show that the book is available:
```
{
      "type": "application/opds-publication+json",
      "rel": "http://opds-spec.org/acquisition/borrow",
      "href": "http://montreal.pretnumerique.ca/v1/media/feedhttps-www-feedbooks-com-item-3123893-9782811221126-epub/loans_activation.opds2?_resource_id=5d4da62423579414ddd09c68",
      "properties": {
        "availability": {
          "state": "available"
        },
        "indirectAcquisition": [{
          "type": "application/vnd.adobe.adept+xml",
          "child": [{
            "type": "application/epub+zip"
          }]
        }]
      }
    }
```

The other example below shows that the books is not available until `2020-10-23T17:03:36.178Z`:
```{
      "type": "application/opds-publication+json",
      "rel": "http://opds-spec.org/acquisition/borrow",
      "href": "http://montreal.pretnumerique.ca/v1/bookings.opds2?_resource_id=5f775a842357946dbe59f5af\u0026medium_id=ENQC109242-9782897059637-pdf",
      "properties": {
        "availability": {
          "state": "unavailable",
          "until": "2020-10-23T17:03:36.178Z"
        },
        "indirectAcquisition": [{
          "type": "application/vnd.adobe.adept+xml",
          "child": [{
            "type": "application/pdf"
          }]
        }, {
          "type": "application/vnd.readium.lcp.license.v1.0+json",
          "child": [{
            "type": "application/pdf"
          }]
        }]
      }
    }
```

Since `availability` property is optional, I also add another condition allowing to add formats of acquisition links not having this property.

leonardr: I see, it's in the schema but not written up in the spec yet. I created these options for OPDS 1.2 but had to put it in an extension so I'm glad it'll be in the core spec for 2.0.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+            + self._extract_contributors(publication.metadata.editors)
+            + self._extract_contributors(publication.metadata.artists)
+            + self._extract_contributors(publication.metadata.illustrators)
+            + self._extract_contributors(publication.metadata.letterers)
+            + self._extract_contributors(publication.metadata.pencilers)
+            + self._extract_contributors(publication.metadata.colorists)
+            + self._extract_contributors(publication.metadata.inkers)
+            + self._extract_contributors(publication.metadata.narrators)
+            + self._extract_contributors(publication.metadata.contributors)
+            + self._extract_contributors(publication.metadata.imprints)
```
leonardr: `imprints` should be treated similarly to `publishers`: the first value encountered should become the 'imprint' metadata value.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+
+        last_opds_update = publication.metadata.modified
+
+        identifier, _ = Identifier.parse_urn(self._db, publication.metadata.identifier)
+        identifier_data = IdentifierData(
+            type=identifier.type, identifier=identifier.identifier
+        )
+
+        # FIXME: There are no measurements in OPDS 2.0
+        measurements = []
```
leonardr, line 569: This is fine; if the information isn't there you don't need to provide it.

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+
+        # FIXME: There is no series information in OPDS 2.0
+        series = None
+        series_position = None
```
leonardr, line 573: This seems like something we need to work out with Hadrien, probably using https://schema.org/CreativeWorkSeries.

vbessonov: Actually, I found an example in Feedbooks feed:
```
"belongsTo": {
          "collection": "The Age of Madness",
          "series": [{
            "name": "The Age of Madness",
            "position": 2,
            "links": [{
              "type": "application/opds+json",
              "href": "https://catalog.feedbooks.com/series/32576.json?lang=en"
            }]
          }]
        },
```

opds2_import.py
```diff
@@ -0,0 +1,830 @@
...
+        # FIXME: It seems that OPDS 2.0 spec doesn't contain information about rights so we use the default one
+        rights_uri = RightsStatus.rights_uri_from_string("")
+
+        circulation_data = CirculationData(
+            default_rights_uri=rights_uri,
+            data_source=data_source_name,
+            primary_identifier=identifier_data,
+            links=links,
+            licenses_owned=LicensePool.UNLIMITED_ACCESS,
+            licenses_available=LicensePool.UNLIMITED_ACCESS,
```
leonardr, line 584: Up to this point open-access books have had licenses_owned=0 and licenses_available=0. I agree that now that we have UNLIMITED_ACCESS, it makes more sense to use that instead. The only place I know this will cause problems is in the dashboard, where we already know we have a problem with the -1 value for unlimited-access books (I filed https://jira.nypl.org/browse/SIMPLY-3238 for this; it's not urgent).

In the future I'd like to see this value standardized for open-access works obtained through other means; this would mean changes to other modules + a migration script.

leonardr:
This is looking pretty good and since this is critical path for you, and not for me, I'm fine if you want to merge it before addressing everything I mentioned.

The only things I'd really like to see resolved before the merge are the `imprint` thing and the assignment of roles to `Contributor`s based on which collection they were in in the OPDS 2 feed. Those are the two items where if you import a feed using the current code, you'd probably want to delete the collection and re-import it after improving this code.

opds2_import.py
vbessonov: Fixed it

opds2_import.py
vbessonov: Updated the type, thank you

opds2_import.py
vbessonov: Updated the code accordingly

opds2_import.py
vbessonov: Added `default_role` parameter to `_extract_contributors`

opds2_import.py
vbessonov: Changed it to `roles=contributor.roles if contributor.roles else default_role`

opds2_import.py
vbessonov: Yes, you're right, I updated the parser so now it always returns `core_ast.Contributor`

opds2_import.py
vbessonov: Added the comment

opds2_import.py
vbessonov: Updated the code accordingly

opds2_import.py
vbessonov: Fixed it

opds2_import.py
vbessonov: Updated the code

opds2_import.py
vbessonov: Sure, added `KNOWN_DRM_TYPES` to `DeliveryMechanism`

opds2_import.py
vbessonov: Changed the code accordingly

vbessonov:
@leonardr, I fixed most of the issues

opds2_import.py
```diff
@@ -144,72 +137,63 @@ def _extract_subjects(self, subject):
...
-                encode(subject), encode(subject_metadata_list)
+            u"Finished extracting subjects metadata: {0}".format(
+                encode(subject_metadata_list)
             )
         )
 
         return subject_metadata_list
 
-    def _extract_contributors(self, contributor):
+    def _extract_contributors(self, contributors, default_role=None):
```
leonardr: You don't end up using this, but `contributors.role` can't be null in the database. The default role here should be AUTHOR_ROLE.

leonardr (approved):
Are you comfortable merging this now?

vbessonov:
@leonardr, yes, I think it's ready to merge now. I'll working on a new PR adding a ProQuest integration, if anything missing I'll add it there

opds2_import.py
vbessonov: Changed `default_role`

model/contributor.py
```diff
@@ -75,6 +75,10 @@ class Contributor(Base):
     PHOTOGRAPHER_ROLE = u"Photographer"
     TRANSLATOR_ROLE = u"Translator"
     ILLUSTRATOR_ROLE = u"Illustrator"
+    LETTERER_ROLE = u"Letterer"
+    PENCILER_ROLE = u"Penciler"
+    COLORIST_ROLE = u"Colorist"
+    INKER_ROLE = u"Inker"
```
leonardr, line 81: The test failure is because you added role types without also adding [the corresponding MARC codes](https://www.loc.gov/marc/relators/relaterm.html) to Contributor.MARC_ROLE_CODES. The only one I can find a direct analog for is Colorist, which has the role code "clr".

[In this comment,](https://github.com/readium/webpub-manifest/issues/26#issuecomment-479429879) @HadrienGardeur recommends treating the other roles as simply "contributor". So use the MARC code "ctb" for the other three.

vbessonov:
@leonardr, I updated the new MARC codes

opds2_import.py
HadrienGardeur: Yes this is covered in RWPM and OPDS 2.0: https://readium.org/webpub-manifest/contexts/default/#collections--series

PR #1206, opened 2020-10-27
Make it possible to pass a Pagination object into WorkList.groups

leonardr:
This branch makes it possible to pass a Pagination object into `WorkList.groups`. The `size` associated with the Pagination object restricts the maximum size of each individual sublane, without restricting the size of the overall feed.

This operates in conjunction with a circulation branch (https://github.com/NYPL-Simplified/circulation/pull/1512) that uses this feature to limit the size of the "jackpot" feed.


lane.py
```diff
@@ -1907,18 +1913,23 @@ def _groups_for_lanes(
...
-        target_size = library.featured_lane_size
-
-        # We ask for a few extra works for each lane, to reduce the
-        # risk that we'll end up reusing a book in two different
-        # lanes.
-        ask_for_size = max(target_size+1, int(target_size * 1.10))
-        # TODO: we're reusing this pagination object, which means
-        # page_loaded will be called multiple times. Could this be a
-        # problem?
-        pagination = Pagination(size=ask_for_size)
```
leonardr, line 1921: This TODO became a warning in the docstring.

lane.py
```diff
@@ -1994,12 +2004,13 @@ def _done_with_lane(lane):
...
                 for x in lane.groups(
-                    _db, include_sublanes=False, facets=facets,
+                    _db, include_sublanes=False,
+                        pagination=pagination, facets=facets,
                 ):
                     yield x
 
     def _featured_works_with_lanes(
-        self, _db, lanes, facets, pagination, search_engine, debug=False
+        self, _db, lanes, pagination, facets, search_engine, debug=False
```
leonardr, line 2013: This makes the method signatures consistent -- pagination always comes before facets.

tests/test_lane.py
```diff
@@ -4172,8 +4182,26 @@ def works(self, _db, facets, *args, **kwargs):
...
+        # If no pagination object is passed in (the most common case),
+        # a new Pagination object is created based on the featured lane
+        # size for the library.
+        groups = list(
+            parent._groups_for_lanes(self._db, relevant, queryable, None,
+                                     facets)
+        )
+
+        (ignore1, pagination, ignore2) = parent._featured_works_with_lanes_called_with
+        assert isinstance(pagination, Pagination)
```
vbessonov, line 4248: Should we care about consistency and try to use `eq_` as much as possible? For example, does it make sense to replace this code with the following?
```python
eq_(True, isinstance(pagination, Pagination))
```

opds.py
```diff
@@ -587,20 +589,21 @@ def groups(cls, _db, title, url, worklist, annotator,
...
-                _db, title, url, worklist, annotator, facets,
-                search_engine, search_debug
+                _db=_db, title=title, url=url, worklist=worklist,
+                annotator=annotator, pagination=pagination, facets=facets,
+                search_engine=search_engine, search_debug=search_debug
             )
 
         return CachedFeed.fetch(
-            _db, worklist=worklist, facets=facets, pagination=None,
+            _db=_db, worklist=worklist, facets=facets, pagination=pagination,
```
vbessonov: Does it make sense to change the order in which parameters are passed and pass `pagination` before `facets`?

lane.py
```diff
@@ -1907,18 +1913,23 @@ def _groups_for_lanes(
...
+            # No pagination object was provided. Our target size is
+            # the featured lane size, but we'll ask for a few extra
+            # works for each lane, to reduce the risk that we end up
+            # reusing a book in two different lanes.
+            target_size = library.featured_lane_size
+
+            # We ask for a few extra works for each lane, to reduce the
+            # risk that we'll end up reusing a book in two different
+            # lanes.
+            ask_for_size = max(target_size+1, int(target_size * 1.10))
```
vbessonov, line 1929: Could you tell when it's possible?

vbessonov (approved):
I added a couple of minor suggestions related to the code style and a question about need to ask for extra works for each lane.

opds.py
leonardr: Good idea.

lane.py
leonardr: Sure, let's say you have a lane for "Romance" and a lane for "Fantasy". Since a book can be classified in multiple genres, it might happen that we randomly pick the same "romance fantasy" book for both lanes.

tests/test_lane.py
leonardr: I'm not terribly concerned about this.

PR #1207, opened 2020-10-28
Don't show breadcrumbs for lanes between the library root and a lane that's a root for some patron type.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-3129. It changes `AcquisitionFeed.add_breadcrumbs` to omit breadcrumbs for lanes that are between the library root and a lane that is the root lane for a specific patron type.

If there is no patron-specific root lane between the lane being breadcrumbed and the library root, the code should work exactly as before. Effectively, this means that Open eBooks and The SimplyE Collection are the only sites currently affected.

If an entry point is selected, the selection is always presumed to have happened in the second breadcrumb.

To use Open eBooks as an example (and ignoring the fact that Open eBooks doesn't have selectable entrypoints), let's say your patron type puts your root lane as "Early Grades", you've navigated from there to the "Early Grades > Animals" lane, and you have the "Audio" point selected.

Breadcrumbs before:

All books > Audio > Early Grades > Animals

Breadcrumbs after:

Early Grades > Audio > Animals

Note that this means you may start at "All books", click into "Early Grades", and lose your breadcrumb back to "All books". Fixing this will take an extra effort since we don't currently vary the contents of most feeds based on the authenticated patron.

Since this is a complex feature that I've made more complex, I put a lot of work into refactoring the test so that a) test assertions are easy to write and easy to understand, b) the test makes an exact assertion about the URL of every breadcrumb, not just an assertion about the title plus some general assertions about some of the URLs.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -1410,12 +1410,31 @@ class AcquisitionFeed(OPDSFeed):
     def add_breadcrumbs(self, lane, include_lane=False, entrypoint=None):
         """Add list of ancestor links in a breadcrumbs element.
 
+        :param lane: Add breadcrumbs from up to this lane.
+        :param include_lane: Include `lane` itself in the breadcrumbs.
+        :param entrypoint: The currently selected entrypoint, if any.
+
         TODO: The switchover from "no entry point" to "entry point" needs
         its own breadcrumb link.
         """
-        # Ensure that lane isn't top-level before proceeding
 
-        entrypointQuery = ("?entrypoint=" + entrypoint.INTERNAL_NAME) if entrypoint != None else ""
+        # The breadcrumbs may be end up being cut off by a
+        # patron-type-specific root lane.
+        stop_at_root_lane = False
+        full_parentage = list(lane.parentage)
+        usable_parentage = []
+        for ancestor in lane.parentage:
+            usable_parentage.append(ancestor)
+            if isinstance(ancestor, Lane) and ancestor.root_for_patron_type:
+                # Root lane for a specific patron type.
+                stop_at_root_lane = True
+                break
+
+
+        if entrypoint is None:
+            entrypoint_query = ""
+        else:
+            entrypoint_query = "?entrypoint=" + entrypoint.INTERNAL_NAME
         annotator = self.annotator
         if annotator.lane_url(lane) != annotator.default_lane_url():
             breadcrumbs = AtomFeed.makeelement("{%s}breadcrumbs" % AtomFeed.SIMPLIFIED_NS)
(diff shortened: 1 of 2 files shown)


opds.py
```diff
@@ -1409,43 +1414,98 @@ def show_current_entrypoint(self, entrypoint):
...
         its own breadcrumb link.
         """
-        # Ensure that lane isn't top-level before proceeding
+        if entrypoint is None:
+            entrypoint_query = ""
+        else:
+            entrypoint_query = "?entrypoint=" + entrypoint.INTERNAL_NAME
+
+        # Breadcrumbs for lanes may be end up being cut off by a
+        # patron-type-specific root lane. If so, that lane -- not the
```
EdwinGuzman, line 1430: What is a patron type in this situation? It's also in the description but I'm not sure what it refers to.

leonardr: It's the value of `Patron.external_type`.

EdwinGuzman (approved):

opds.py
```diff
@@ -1432,15 +1432,19 @@ def add_breadcrumbs(self, lane, include_lane=False, entrypoint=None):
         site_root_lane = None
         usable_parentage = []
         if lane is not None:
-            for ancestor in lane.parentage:
+            for ancestor in [lane] + list(lane.parentage):
```
leonardr, line 1435: While testing against Open eBooks I discovered that my original implementation doesn't work if you ask for breadcrumbs for a lane that is, itself the root lane for some patron type. `lane.parentage` doesn't include the lane itself, so we never ran the `ancestor.root_for_patron_type` against the lane we were looking at.

opds.py
```diff
@@ -1494,16 +1499,6 @@ def add_breadcrumbs(self, lane, include_lane=False, entrypoint=None):
...
-        # Include link to the lane itself.
-        # For search, breadcrumbs include the searched lane
-        if include_lane:
-            breadcrumbs.append(
-                AtomFeed.link(
-                    title=lane.display_name,
-                    href=annotator.lane_url(lane) + entrypoint_query
-                )
-            )
-
```
leonardr, line 1447: We don't need this code anymore since `include_lane` now makes sure the lane itself goes into `usable_parentage`. It's then handled like all the other lanes as we iterate over `usable_parentage`.

tests/test_opds.py
```diff
@@ -2050,6 +2050,8 @@ def title(x):
         # the breadcrumbs will be start at that lane -- we won't see
         # the sublane's parent or the library root.
         sublane.root_for_patron_type = ["ya"]
+        assert_breadcrumbs([], sublane)
+
```
leonardr, line 2054: This is the test that was missing. Previously hitting a patron root lane would give you an "All books" breadcrumb when it should have given you no breadcrumbs at all.

EdwinGuzman (approved):

PR #1208, opened 2020-11-01
Add ProQuest integration

vbessonov:
## Description

This PR contains changes required for `ProQuestOPDS2Importer` implemented in [PR # 1520](https://github.com/NYPL-Simplified/circulation/pull/1520).

## Motivation and Context

[SIMPLY-3251](https://jira.nypl.org/browse/SIMPLY-3251)

## Checklist:

### DIFF
--- a/model/configuration.py
+++ b/model/configuration.py
@@ -182,6 +182,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     FEEDBOOKS = DataSourceConstants.FEEDBOOKS
     LCP = DataSourceConstants.LCP
     MANUAL = DataSourceConstants.MANUAL
+    PROQUEST = DataSourceConstants.PROQUEST
 
     # These protocols were used on the Content Server when mirroring
     # content from a given directory or directly from Project
@@ -523,6 +524,7 @@ class ExternalIntegration(Base, HasFullTableCache):
             if include_secrets or not setting.is_secret:
                 lines.append(explanation)
         return lines
+
 
 class ConfigurationSetting(Base, HasFullTableCache):
     """An extra piece of site configuration.
@@ -882,6 +884,8 @@ class ConfigurationAttributeType(Enum):
     TEXTAREA = 'textarea'
     SELECT = 'select'
     NUMBER = 'number'
+    LIST = 'list'
+    MENU = 'menu'
 
     def to_control_type(self):
         """Converts the value to a attribute type understandable by circulation-web
@@ -909,6 +913,7 @@ class ConfigurationAttribute(Enum):
     DEFAULT = 'default'
     OPTIONS = 'options'
     CATEGORY = 'category'
+    FORMAT = 'format'
 
 
 class ConfigurationOption(object):
@@ -1035,7 +1040,9 @@ class ConfigurationMetadata(object):
             default=None,
             options=None,
             category=None,
-            index=None):
+            format=None,
+            index=None
+    ):
         """Initializes a new instance of ConfigurationMetadata class
 
         :param key: Setting's key
@@ -1070,6 +1077,7 @@ class ConfigurationMetadata(object):
         self._default = default
         self._options = options
         self._category = category
+        self._format = format
 
         if index is not None:
             self._index = index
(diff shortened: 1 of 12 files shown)


model/work.py
```diff
@@ -874,10 +874,9 @@ def calculate_presentation_edition(self, policy=None):
         )
         return changed
 
-
     def calculate_presentation(
         self, policy=None, search_index_client=None, exclude_search=False,
-        default_fiction=None, default_audience=None
+        default_fiction=None, default_audience=Classifier.AUDIENCE_ADULT
```
vbessonov: @leonardr, I set `default_audience` to `Classifier.AUDIENCE_ADULT` instead of implementing [SIMPLY-2610](https://jira.nypl.org/browse/SIMPLY-2610)

leonardr: I think this is OK. Since this code was written we have had many incoming feeds with no audience information and the implication was always that the works are for adults.

leonardr: Christina pushed back a little in https://jira.nypl.org/browse/SIMPLY-2610; I'm waiting for her response. It may be better to simply implement 2610.

Another temporary way to solve this problem would be to continue to set audience=None for these books but to change the way books are placed into lanes. Specifically, we could say that a lanes whose audiences include AUDIENCE_ADULT or AUDIENCE_RESEARCH should also include any books that have no audience set.

The real problem we're trying to address is not that these books don't have an audience set (that's the most accurate conclusion -- we don't have the information) but that they don't show up in the default lanes.

model/credential.py
```diff
@@ -89,14 +85,40 @@ class Credential(Base):
     IDENTIFIER_FROM_REMOTE_SERVICE = "Identifier Received From Remote Service"
 
     @classmethod
-    def lookup(self, _db, data_source, type, patron, refresher_method,
+    def _filter_out_credential(cls, credential, allow_persistent_token):
```
leonardr: Putting this in a helper method is good; can you rename this to something like `_filter_invalid_credential`? Currently it sounds like the credential is _always_ filtered out.

model/configuration.py
```diff
@@ -182,6 +182,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     FEEDBOOKS = DataSourceConstants.FEEDBOOKS
     LCP = DataSourceConstants.LCP
     MANUAL = DataSourceConstants.MANUAL
+    PROQUEST = DataSourceConstants.PROQUEST
```
leonardr, line 185: You also need to add this to `DataSource.well_known_sources`. This isn't as important as it used to be but it tells the system two important things:

1. ProQuest offers actual books you can read, not just metadata (`offers_licenses`)
2. ProQuest books are always identified by their ProQuest Doc ID (`primary_identifier_type`)

`offers_metadata_lookup` should be false. `refresh_rate` is no longer used so just set it to None.

util/__init__.py
```diff
@@ -533,3 +535,48 @@ def parse(cls, amount):
             amount = amount[1:]
         return Money(amount, currency)
 
+
+def is_string(value):
```
leonardr: Can you put this in `util/string_helpers.py`? We have some other Python compatibility functions in there.

leonardr:
I suspect there may be some new files missing from this PR; there's nothing really Proquest-specific, there are no new tests, and you define an interface class that has no implementations.

model/work.py
leonardr: Based on the conversation with Christina in https://jira.nypl.org/browse/SIMPLY-2610, I don't think users will accept a change like this which adds metadata that was not provided by the vendor, the administrator, or a librarian.

My preference would be to resolve this by resolving SIMPLY-2610.

leonardr: Let me make this stronger: based on Christina's feedback I think you need to implement SIMPLY-2610 instead of setting this default.

vbessonov: Sure, will do

vbessonov: I created [PR # 1214 in server_core](https://github.com/NYPL-Simplified/server_core/pull/1214) and merged it to this branch.

model/credential.py
vbessonov: Updated the method's name

model/configuration.py
vbessonov: Added `ProQuest` to `DataSource.well_known_sources`

util/__init__.py
vbessonov: Moved `is_string` to `util/string_helpers.py`

leonardr (approved):

PR #1209, opened 2020-11-04
Apply lane visibility rules to the top-level WorkList

leonardr:
This fixes a problem I found doing QA of https://jira.nypl.org/browse/SIMPLY-3125: the lane visibility rules are not applied to the top-level WorkList, because it's not a Lane.

This poses a small problem: all authenticated patrons can see the top-level WorkList, but a patron with a root lane should not be able to see it.

This branch does more work than is strictly necessary to solve this problem because I introduced a new type of WorkList -- TopLevelWorkList -- that interacts with Lane, and I wanted to make sure that a TopLevelWorkList appears to be an ancestor of every Lane in the same Library.

I'm concerned about deploying this change to production because the currently deployed https://catalog.openebooks.us/ seems to depend on the top-level grouped feed being accessible to all authenticated users. My original idea was to make that feed redirect when appropriate (like the IndexController does) rather than deny access. This implementation turned out to be much easier, but if we want to deploy to Open eBooks production before updating the web catalog, I should probably take another look at the redirect idea.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -1492,7 +1492,7 @@ class WorkList(object):
     def is_self_or_descendant(self, ancestor):
         """Is this WorkList the given WorkList or one of its descendants?
 
-        :param target: A WorkList.
+        :param ancestor: A WorkList.
         :return: A boolean.
         """
         for candidate in [self] + list(self.parentage):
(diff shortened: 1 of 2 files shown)


EdwinGuzman (approved):
Looks good but the PR test is falling. Looks like it may just need to be retriggered?

PR #1211, opened 2020-11-05
Add an availability facet 'not_now' which finds only books that are _not_ currently available.

leonardr:
This is the core portion of https://jira.nypl.org/browse/SIMPLY-3236. It adds a new availability facet which finds books that are not currently available. This facet cannot be configured for use in normal sites, or used in normal `Facets` objects, because patrons don't want to search only books that are not currently available. Instead, the `JackpotFacets` class defined in https://github.com/NYPL-Simplified/circulation/pull/1521 allows it as a possibility.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -672,18 +672,27 @@ class Facets(FacetsWithEntryPoint):
         # Apply any superclass criteria
         qu = super(Facets, self).modify_database_query(_db, qu)
 
+        available_now = or_(
+            LicensePool.open_access == True,
+            LicensePool.self_hosted == True,
+            LicensePool.licenses_available > 0
+        )
+
         if self.availability == self.AVAILABLE_NOW:
-            availability_clause = or_(
-                LicensePool.open_access == True,
-                LicensePool.licenses_available > 0
-            )
+            availability_clause = available_now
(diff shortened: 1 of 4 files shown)


lane.py
```diff
@@ -672,18 +672,27 @@ def modify_database_query(self, _db, qu):
         # Apply any superclass criteria
         qu = super(Facets, self).modify_database_query(_db, qu)
 
+        available_now = or_(
+            LicensePool.open_access == True,
+            LicensePool.self_hosted == True,
```
EdwinGuzman, line 677: What does it mean to be "self_hosted" and that it's "available now"?

EdwinGuzman (approved):
This looks good but just a minor question .

lane.py
leonardr: `self_hosted` books are books that are not open access, but were uploaded to a content server under the control of the library and are available without restriction to authenticated patrons. Right now that means they were uploaded to an LCP server and made available through an LCP collection.

This comes out of Viacheslav's work on LCP. As part of that work he changed a number of places where we assumed a book was "available" only because it's open access or because there are a certain number of licenses owned. This is a place that we missed originally.

PR #1214, opened 2020-11-08
Make default audience a configurable setting for collection

vbessonov:
## Description

This PR makes default audience a configurable setting for collections.

## Motivation and Context

[SIMPLY-2610](https://jira.nypl.org/browse/SIMPLY-2610)

## Checklist:

### DIFF
--- a/model/work.py
+++ b/model/work.py
@@ -85,6 +85,7 @@ class WorkGenre(Base):
     def __repr__(self):
         return "%s (%d%%)" % (self.genre.name, self.affinity*100)
 
+
 class Work(Base):
     APPEALS_URI = "http://librarysimplified.org/terms/appeals/"
 
@@ -874,6 +875,17 @@ class Work(Base):
         )
         return changed
 
+    def _get_default_audience(self):
+        """Return the default audience.
+
+        :return: Default audience
+        :rtype: Optional[str]
+        """
+        for license_pool in self.license_pools:
+            if license_pool.collection.default_audience:
+                return license_pool.collection.default_audience
+
+        return None
 
     def calculate_presentation(
         self, policy=None, search_index_client=None, exclude_search=False,
@@ -889,6 +901,8 @@ class Work(Base):
         * The best available summary for the work.
         * The overall popularity of the work.
         """
+        if not default_audience:
+            default_audience = self._get_default_audience()
 
         # Gather information up front so we can see if anything
         # actually changed.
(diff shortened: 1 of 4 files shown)


vbessonov:
@leonardr, what kind of values can be set in this configuration setting? Will it be a string or a list of strings?

model/work.py
```diff
@@ -874,6 +875,17 @@ def calculate_presentation_edition(self, policy=None):
...
 
+    def _get_default_audience(self):
+        """Return the default audience.
+
+        :return: Default audience
+        :rtype: Optional[str]
+        """
+        for license_pool in self.license_pools:
+            if license_pool.collection.default_audience:
+                return license_pool.collection.default_audience
```
leonardr, line 886: Since there can be multiple LicensePools for the same work from different collections, this will choose one at random. I was picturing a solution that went into `Metadata.apply` and added some extra information to the Identifier while we know which data source we're talking about. But that doesn't actually remove the random element; it just creates multiple pieces of (potentially conflicting) evidence which the classifier needs to sort out later.

So I'm OK with this solution overall. I don't think the random element will come up in a real case.

leonardr (approved):

leonardr:
> @leonardr, what kind of values can be set in this configuration setting? Will it be a string or a list of strings?

It will be a single optional value. The list of possible values is found in core.classifier.Classifier.AUDIENCES, and leaving the value blank ("No default audience") should be the default choice.

The configuration for this setting would probably go into OPDSImporter.BASE_SETTINGS so it will be a configuration option for both OPDS1 and OPDS2. You might also want it in ODL imports -- in that case store the configuration as a constant somewhere and include it from OPDSImporter and ODLAPI.

leonardr (changes requested):
Un-doing my approval since the configuration for this setting still needs to be added to the API classes.

vbessonov:
@leonardr, I added the configuration setting to `OPDSImporter`

tests/models/test_work.py
```diff
@@ -399,6 +399,25 @@ def test_calculate_presentation_sets_presentation_ready_based_on_content(self):
         work.calculate_presentation()
         eq_(True, work.presentation_ready)
 
+    def test_calculate_presentation_uses_default_audience_set_as_collection_setting(self):
+        default_audience = u"Audience"
```
leonardr: One more change: use a real audience here, like Classifier.AUDIENCE_ADULT. Eventually I'd like to make `Work.audience` an enumerated type and I'd rather not have tests where it's set to out-of-band strings.

leonardr (approved):

PR #1218, opened 2020-11-17
Make sure 'available' and 'licensed' are always booleans when turning a Work into a search document

leonardr:
I was not able to reproduce this problem in a unit test scenario, but it was easy to reproduce against NYPL's QA server while doing QA on https://jira.nypl.org/browse/SIMPLY-3236. The only thing I can think of is that it might depend on which version of Postgres you're running.

The problem happens when a book is not available through a given LicensePool: in this case Work.to_search_document gives "null" instead of "false" for "licensepools.availability". Previously this wasn't a problem because we only ever looked for books that were available -- i.e. had a "true" value for licensepools.availability. Thanks to 3236, we now look for books that are _not_ available, and Elasticsearch doesn't treat "null" as a "false" value -- it doesn't seem possible to filter on a null value at all.

This branch changes the SQL so that we are always guaranteed to get either True or False for 'availability' and 'licensed'. I believe there is a "falsey" value being returned in some cases which SQLAlchemy turns into null instead of False. But I don't know when this value is being returned as opposed to False.

### DIFF
--- a/migration/20201117-remove-all-search-coverage.sql
+++ b/migration/20201117-remove-all-search-coverage.sql
@@ -1,0 +1,3 @@
+-- Remove all WorkCoverageRecords pertaining to the search index. This
+-- will force a complete reindex on the next run of bin/search_index_refresh.
+delete from workcoveragerecords where operation='update-search-index';
--- a/model/work.py
+++ b/model/work.py
@@ -1545,6 +1545,13 @@ class Work(Base):
         # The work quality field is stored in the main document, but
         # it's also stored here, so that we can apply a nested filter
         # that combines quality with other fields found only in the subdocument.
+
+        def explicit_bool(label, t):
+            # Ensure we always generate True/False instead of
+            # True/None. Elasticsearch can't filter on null values.
+            return t.label(label)
+            return case([(t, True)], else_=False).label(label)
+
         licensepools = select(
             [
                 LicensePool.id.label('licensepool_id'),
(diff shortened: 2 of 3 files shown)


tests/models/test_work.py
```diff
@@ -1117,7 +1117,9 @@ def assert_time_match(python, postgres):
             eq_(pool.suppressed, match['suppressed'])
             eq_(pool.data_source_id, match['data_source_id'])
 
+            assert isinstance(match['available'], bool)
             eq_(pool.licenses_available > 0, match['available'])
+            assert isinstance(match['licensed'], bool)
```
leonardr, line 1122: These assertions don't fail even when the bug is present, but I figured it can't hurt to add them.

migration/20201117-remove-all-search-coverage.sql
```diff
@@ -0,0 +1,3 @@
+-- Remove all WorkCoverageRecords pertaining to the search index. This
+-- will force a complete reindex on the next run of bin/search_index_refresh.
+delete from workcoveragerecords where operation='update-search-index';
```
EdwinGuzman, line 3: Why does this need a reindex?

model/work.py
```diff
@@ -1545,23 +1545,37 @@ def query_to_json_array(query):
         # The work quality field is stored in the main document, but
         # it's also stored here, so that we can apply a nested filter
         # that combines quality with other fields found only in the subdocument.
+
+        def explicit_bool(label, t):
+            # Ensure we always generate True/False instead of
+            # True/None. Elasticsearch can't filter on null values.
+            return t.label(label)
+            return case([(t, True)], else_=False).label(label)
```
EdwinGuzman, line 1552: You're not getting to line 1553 because of the return statement in the previous line.

leonardr: Ugh, that's my test code trying to reproduce the error. It needs to go.

migration/20201117-remove-all-search-coverage.sql
leonardr: There's a bunch of books in the index that have been indexed with available=null instead of available=false. (Not _all_ books, but a lot of them.) Reindexing them with the new code will make the same books indexed with available=false.

EdwinGuzman (approved):

PR #1220, opened 2020-12-01
Add DSL based on Python grammar

vbessonov:
## Description

This PR adds DSL based on Python grammar.

## Motivation and Context

[SIMPLY-3334](https://jira.nypl.org/browse/SIMPLY-3334)

## Checklist:

### DIFF
--- a/tests/python_expression_dsl/test_evaluator.py
+++ b/tests/python_expression_dsl/test_evaluator.py
@@ -1,0 +1,195 @@
+from nose.tools import assert_raises, eq_
+from parameterized import parameterized
+
+from ...python_expression_dsl.evaluator import (
+    DSLEvaluationError,
+    DSLEvaluationVisitor,
+    DSLEvaluator,
+)
+from ...python_expression_dsl.parser import DSLParseError, DSLParser
+
+
+class Subject(object):
+    """Dummy object designed for testing DSLEvaluator."""
+
+    def __init__(self, attributes):
+        """Initialize a new instance of Subject.
+
+        :param attributes: List of attributes
+        :type attributes: List[str]
+        """
+        self._attributes = attributes
+
+    @property
+    def attributes(self):
+        """Return the list of attributes.
+
+        :return: List of attributes
+        :rtype: List
+        """
+        return self._attributes
+
+    def get_attribute_value(self, index):
+        """Dummy method to test method invocation.
+
+        :param index: Attribute's index
+        :type index: int
+
+        :return: Attribute's value
+        :rtype: Any
+        """
+        return self._attributes[index]
+
+
+class TestDSLEvaluator(object):
+    @parameterized.expand(
+        [
+            ("incorrect_expression", "?", None, None, None, DSLParseError),
+
+            ("numeric_literal", "9", 9),
+            ("numeric_float_literal", "9.5", 9.5),
+
+            ("unknown_identifier", "foo", None, None, None, DSLEvaluationError),
+            ("known_identifier", "foo", 9, {"foo": 9}),
+            (
+                "unknown_nested_identifier",
+                "foo.bar",
+                None,
+                {"foo": 9},
+                None,
+                DSLEvaluationError,
+            ),
+            ("known_nested_identifier", "foo.bar", 9, {"foo": {"bar": 9}}),
+            ("known_nested_identifier", "foo.bar.baz", 9, {"foo": {"bar": {"baz": 9}}}),
+            ("known_nested_identifier", "foo.bar[0].baz", 9, {"foo": {"bar": [{"baz": 9}]}}),
+            (
+                "identifier_pointing_to_the_object",
+                "'eresources' in subject.attributes",
+                True,
+                {"subject": Subject(["eresources"])},
+            ),
+
+            ("simple_negation", "-9", -9),
+            ("simple_parenthesized_expression_negation", "-(9)", -(9)),
+            ("parenthesized_expression_negation", "-(9 + 3)", -(9 + 3)),
+            ("slice_expression_negation", "-(arr[1])", -12, {"arr": [1, 12, 3]}),
+
+            ("addition_with_two_operands", "9 + 3", 9 + 3),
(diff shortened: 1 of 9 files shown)


leonardr:
You can fix the imports that are breaking your tests by replacing "core.util" with "..util". See core/credential.py for an example.

tests/python_expression_dsl/test_evaluator.py
```diff
@@ -0,0 +1,138 @@
...
+            ("triple_inversion", "not not not 9 < 3", not not not 9 < 3),
+
+            ("conjunction", "9 == 9 and 3 == 3", 9 == 9 and 3 == 3),
+            ("disjunction", "9 == 3 or 3 == 3", 9 == 3 or 3 == 3),
+
+            ("simple_parenthesized_expression", "(9 + 3)", (9 + 3)),
+            ("arithmetic_parenthesized_expression", "2 * (9 + 3) * 2", 2 * (9 + 3) * 2),
+
+            ("slice_expression", "arr[1] == 12", True, {"arr": [1, 12, 3]}),
+        ]
```
leonardr, line 168: These tests are very concise and readable. There are three additional types of tests I'd like to see:

* expressions that contain multiple variable references, e.g. `arr[1] + arr[2]` and `value1 * value2`
* expressions that attempt to use built-in python functions like `print(var)` and `sum(arr)`
* expressions that attempt to invoke methods on variables, like `stringvar.upper()`

I don't expect any of these tests to show anything weird, but it'll prove that your DSL does the right thing.

tests/python_expression_dsl/test_evaluator.py
```diff
@@ -0,0 +1,138 @@
...
+            ("unknown_identifier", "foo", None, None, DSLEvaluationError),
+            ("known_identifier", "foo", 9, {"foo": 9}),
+            ("unknown_nested_identifier", "foo.bar", None, {"foo": 9}, DSLEvaluationError),
+            ("known_nested_identifier", "foo.bar", 9, {"foo": {"bar": 9}}),
+            (
+                "identifier_pointing_to_the_object",
+                "'eresources' in subject.attributes",
+                True,
+                {"subject": Subject(["eresources"])},
+            ),
```
leonardr, line 70: I'm a little bit concerned about this one. If there's a data model object in the context, attribute access will give you access to basically any other data model object in the system. 

Is it possible to have attribute access only work in the case of a nested identifier? Do you need this to work for the SAML use case you're considering?

leonardr (changes requested):

tests/python_expression_dsl/test_evaluator.py
vbessonov: I added a couple of test cases:
```python
            ("complex_slice_expression", "arr[1] + arr[2]", 15, {"arr": [1, 12, 3]}),

            ("method_call", "str.upper()", "HELLO WORLD", {"str": "Hello World"}),
            ("builtin_function_call", "min(1, 2)", min(1, 2)),
```
You can find expressions that contain multiple variable references in cases covering associative, commutative, and distributive laws.

tests/python_expression_dsl/test_evaluator.py
vbessonov: Using a DSL expression you can only get access to only what's inside the context which can be either a dictionary or a Python object. The test case above emulates the scenario I'm planning to implement in SAML enhancements PR. In that case the DSL engine will have access only to `api.saml.metadata.model.SAMLSubject`, all of its nested properties and some built-in Python functions:
```python
    BUILTIN_FUNCTIONS = {
        "abs": abs,
        "all": all,
        "any": any,
        "len": len,
        "max": max,
        "min": min,
    }
```

leonardr:
A general note: I would like to see this eventually put into a separate project on PyPI, I think it has value outside of this project.

tests/python_expression_dsl/test_evaluator.py
leonardr: My concern is with the nested properties and method calls. If you put a SQLAlchemy data model object into the context, then executing `obj.delete()` will delete that object from the database when the request completes. If nested properties also work, executing something like `obj.patron.delete()` wil delete some _other_ object from the database.

Maybe there is no solution apart from code review of anything that uses this system to make sure it doesn't put unsafe objects into the context. I was hoping we could avoid that by limiting the DSL's expressive capability to something closer to XPath.

leonardr:
Basically, if you think there's no solution to the safety problem short of careful review of what gets put in the context, then we can merge this. If you have other ideas I'd like to consider them.

tests/python_expression_dsl/test_evaluator.py
vbessonov: My main intent was to make the DSL engine almost as much powerful as Python so future users could write filter expression of any complexity but I do see your point. We can try to run DSL expression in a [sandbox](https://stackoverflow.com/questions/3068139/how-can-i-sandbox-python-in-pure-python) by, for example, executing them in a separate `exec` context.

vbessonov: The other thing that we can do is to make a deep copy of context objects by serializing and deserializing them before putting into the context - it will allow to unbind them from CM's execution context, SQLAlchemy session, etc.

leonardr:
Per a discussion on Slack, this is OK to merge and use in the SAML branch with two changes:

1. Method invocations should only be allowed for built-in objects like stirngs.
2. The `evaluate` method should deepcopy its `context` before evaluation.

vbessonov:
@leonardr, I added requested changes:

1. By default method invocation is allowed only for built-in classes (float, int, str). Method invocation for other classes should be allowed explicitly by sending a list with their names to `DSLEvaluationVisitor` constructor or to `DSLEvaluator.evaluate` method.
2. All the context objects are deep copied inside `DSLEvaluationVisitor.context` property.

leonardr (approved):
This looks quite solid now. Still hoping a) we can eventually solve this problem with a more standard language and b) this can eventually be moved into a standalone Python project.

PR #1221, opened 2020-12-01
Update the Overdrive client to use v2 of the availability API

leonardr:
This is the core portion of a fix to https://jira.nypl.org/browse/SIMPLY-3302 -- using v2 of the Overdrive availability API instead of v1.

The main change is that in the v2 response, the number we're looking for depends on whether we're trying to update a normal Overdrive account or an Overdrive Advantage account.

I've resisted the urge to refactor this whole thing, but that does need to be done: I filed https://jira.nypl.org/browse/SIMPLY-3343 to track that work and explain why.

### DIFF
--- a/overdrive.py
+++ b/overdrive.py
@@ -279,13 +279,18 @@ class OverdriveAPI(object):
     def advantage_library_id(self):
         """The library ID for this library, as we should look for it in
         certain API documents served by Overdrive.
+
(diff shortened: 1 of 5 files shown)


tdilauro (approved):
These changes look good and I appreciate the judicious comments and TODOs to aid future work.

PR #1222, opened 2020-12-03
Add new optional MeasurementReaper

jonathangreen:
## Description

This adds a new database reaper to reap the measurements table. This reaper is enabled by default, but can be disabled with a configuration setting. 

## Motivation and Context

On some of the LYRASIS servers the measurements table is taking up 100s of GB of space. After running this reaper and letting the database vacuum complete, these tables are taking up only 100mb. 

The large measurements table is slowing down our instances and putting unnecessary stress on our database server. It seems likely there are other issues that is the root, causing so many measurements to be recorded. These will be addressed separately, but this will clean up old unused measurements. 

Speaking to @leonardr he would like to keep old measurements for future analysis purposes, so I've added a configuration setting to disable this reaper. The reaper is enabled by default however, as I think most people will want to clean up old measurements.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -10,11 +10,13 @@ from sqlalchemy.sql.functions import func
 from sqlalchemy.sql.expression import (
     or_,
     and_,
-    outerjoin,
+    outerjoin, delete,
 )
 
 import log # This sets the appropriate log format and level.
+from model.configuration import ConfigurationSetting
 from config import Configuration
+
 from coverage import CoverageFailure
 from metadata_layer import TimestampData
 from model import (
@@ -37,6 +39,7 @@ from model import (
     Timestamp,
     Work,
     WorkCoverageRecord,
+    Measurement,
 )
 
 
@@ -771,10 +774,10 @@ class ReaperMonitor(Monitor):
     REGISTRY = []
 
     def __init__(self, *args, **kwargs):
-        self.SERVICE_NAME = "Reaper for %s.%s" % (
-            self.MODEL_CLASS.__name__,
-            self.TIMESTAMP_FIELD
-        )
+        self.SERVICE_NAME = "Reaper for %s" % self.MODEL_CLASS.__name__
+        if self.TIMESTAMP_FIELD is not None:
+            self.SERVICE_NAME += ".%s" % self.TIMESTAMP_FIELD
+
         super(ReaperMonitor, self).__init__(*args, **kwargs)
 
     @property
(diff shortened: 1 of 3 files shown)


monitor.py
```diff
@@ -901,6 +904,30 @@ def delete(self, collection):
 ReaperMonitor.REGISTRY.append(CollectionReaper)
 
 
+class MeasurementReaper(ReaperMonitor):
+    """Remove measurements that are not the most recent"""
+    MODEL_CLASS = Measurement
+    MEASUREMENT_REAPER_ENABLED = 'MeasurementReaper.enabled'
```
leonardr: This looks like an old name for the sitewide setting that you ended up not using.

leonardr (approved):
This looks fine apart from a potential dangling constant. I filed https://jira.nypl.org/browse/SIMPLY-3350 to track this work.

monitor.py
jonathangreen: Good catch, thanks! Decided to change where it was defined.

jonathangreen:
@leonardr updated with constant removed

tests/test_monitor.py
```diff
@@ -35,7 +36,7 @@
     Subject,
     Timestamp,
     Work,
-    WorkCoverageRecord,
+    WorkCoverageRecord, Measurement, get_one_or_create, ConfigurationSetting,
```
tdilauro, line 39: Can these go on separate lines in alpha order? #nit

tests/test_monitor.py
```diff
@@ -62,7 +63,7 @@
     SweepMonitor,
     TimelineMonitor,
     WorkReaper,
-    WorkSweepMonitor,
+    WorkSweepMonitor, MeasurementReaper,
```
tdilauro, line 66: Can these go on separate lines in alpha order? #nit

monitor.py
```diff
@@ -37,6 +39,7 @@
     Timestamp,
     Work,
     WorkCoverageRecord,
+    Measurement,
```
tdilauro, line 42: Can you move this into alpha order for consistency? #nit

monitor.py
```diff
@@ -10,11 +10,13 @@
 from sqlalchemy.sql.expression import (
     or_,
     and_,
-    outerjoin,
+    outerjoin, delete,
```
tdilauro, line 13: Can these go on separate lines in alpha order? #nit

tdilauro (approved):
This looks good, aside from a couple nit picky things I noted.

jonathangreen:
@tdilauro I wonder if some of those nit comments might a good start to a broader discussion on coding standards in the project. Some of the import lists weren't already in alphabetical order, but if that is the expectation, which I think is totally reasonable, maybe we should have some formal coding standards / contributing guidelines?

leonardr:
I don't comment on the appearance of code unless it's _really_ bad, because in my experience it really hurts team velocity. However stuff like "imports in alphabetical order" is definitely good practice and I'd be interested in seeing an automatic linter put into our pipeline to take care of this stuff. 

Is anyone interested in looking into this? Ideally I'd like something similar to Github's dependabot that puts up no-brainer PRs.

tdilauro:
I wasn't trying to go too big with those nits and I wouldn't have even submitted them if I had realized the PR had already been merged. I actually groaned out loud and felt a little bad when I saw the new PR to address them.

That said, I don't think it would be a bad idea to introduce some coding standards. During the move to Python 3 seems like a natural time to address some of this more systematically and incrementally.

I'd definitely be willing to consider incorporating `pylint`, `flake8`, or even `black` into my own workflows. It would be great to eventually get to a pre-commit hook approach to ensure that things are cleaned up, so that they don't have to be part of the human code review.

vbessonov:
We can try to reuse some of the work I did in [python-webpub-manifest-parser](https://github.com/NYPL-Simplified/python-webpub-manifest-parser/pull/2/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485). It already has support for isort, black, flake8, and pylint which are run by Python 3.6 using tox but the whole project is written using Python 2.

PR #1223, opened 2020-12-03
Address nits from #1222

jonathangreen:
Here is a small PR to address the nits from @tdilauro on #1222

There is a lot of noise in the diff, but its all just moving around imports. This is the result from running PyCharm's "optimize imports" which is supposed to follow Python's [PEP 8 section on import style](https://www.python.org/dev/peps/pep-0008/#imports).

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -1,46 +1,37 @@
-from nose.tools import set_trace
 import datetime
-import os
 import logging
-import time
 import traceback
+
 from sqlalchemy.orm import defer
-from sqlalchemy.sql import select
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
Looks good.

vbessonov:
What do you think about using `isort` to sort and optimize imports?

leonardr:
I think running `isort` on this code for a one-time fix is a good start towards instituting something more automated. We did something similar a while ago to remove trailing spaces at the end of lines.

PR #1225, opened 2020-12-09
Make it possible to search for books that are in _some_ series without caring about _which_ series

leonardr:
This is a support branch for https://jira.nypl.org/browse/SIMPLY-3235. It makes it possible to create a `Filter` object that matches only books that are in _some_ series. The series can be anything, but the 'series' field must have been indexed and must be a non-empty string.

I did not implement the opposite feature -- using `series=False` to find books that _don't_ belong to a series -- because there's no current use case for it and implementing new Elasticsearch filtering features is always a pain, both to do at first and to maintain.

### DIFF
--- a/tests/test_external_search.py
+++ b/tests/test_external_search.py
@@ -497,6 +497,7 @@ class TestExternalSearchWithWorks(EndToEndSearchTest):
         self.ya_romance.presentation_edition.subtitle = (
             "Modern Fairytale Series, Volume 7"
         )
+        self.ya_romance.presentation_edition.series = "Modern Fairytales"
 
         self.no_age = _work()
         self.no_age.summary_text = "President Barack Obama's election in 2008 energized the United States"
@@ -705,7 +706,7 @@ class TestExternalSearchWithWorks(EndToEndSearchTest):
         expect(self.moby_dick, "moby", classics)
 
         # This finds books that belong to _some_ series.
-        expect(self.moby_dick, "", Filter(series=True))
+        expect([self.moby_dick, self.ya_romance], "", Filter(series=True))
 
         # Find results based on genre.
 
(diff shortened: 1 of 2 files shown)


tests/test_external_search.py
```diff
@@ -497,10 +497,15 @@ def populate_works(self):
...
             "Modern Fairytale Series, Volume 7"
         )
+        self.ya_romance.presentation_edition.series = "Modern Fairytales"
 
         self.no_age = _work()
         self.no_age.summary_text = "President Barack Obama's election in 2008 energized the United States"
 
+        # Set the series to the empty string rather than None -- this isn't counted
+        # as the book belonging to a series.
+        self.no_age.presentation_edition.series = ""
```
vbessonov, line 507: It's not very clear to me what should happen in this case: according to condition in `external_search` a new condition matching empty series will be added to the chain. Isn't it what `series=False` should do?

vbessonov (approved):

tests/test_external_search.py
leonardr: As background, this test is creating a pretty large number of books, indexing them in Elasticsearch, and then seeing which books show up in the search results when different search terms and filters are provided.

Setting series=True in the Filter constructor will add two extra clauses to the Elasticsearch query:

1. The 'series' field on a book must be indexed: `Exists(field="series")`
2. The indexed value must not be identical to the empty string: `Bool(must_not=[Term(**{"series.keyword": ""})])`

This sets up a test of that second clause. Elasticsearch indexes a field only if the value is not null, and it treats the empty string as a non-null value. Without that second clause, the `no_age` book will incorrectly show up in the test run with the filter `Filter(series=True)`, because Elasticsearch will notice that the 'series' field was indexed.

Basically, this is testing that the empty string isn't treated as a real "series".

leonardr: If I did add support for `series=False` then this same book would we used in a second assertion: we would check that it _does_ show up in a query with `series=False`, even though its `series` field is set to a non-null value.

vbessonov: Now it's clear, thank you for the clarification.

PR #1227, opened 2020-12-18
Fix OPDS importer to handle improved DPLA/Feedbooks OPDS 1.x+ODL feed

arielmorelli:
## Description

This PR add:
* the possibility of use the \<dcterm:identifier\> instead of \<id\>
* use custom accept header instead of the system default

## Motivation and Context

[SIMPLY-3392](https://jira.nypl.org/browse/SIMPLY-3392)

## How Has This Been Tested?

* New unit tests were add.
* Was used the real Feedbook XML URL for tests: https://market.feedbooks.com/api/libraries/harvest.atom
* To run the database, elasticsearch and minio were used in a docker-compose environment. The complete env can be found [here](https://github.com/arielmorelli/dev_env_for_circulation/tree/master)

## Checklist:

### DIFF
--- a/model/configuration.py
+++ b/model/configuration.py
@@ -269,7 +269,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     CUSTOM_ACCEPT_HEADER = u"custom_accept_header"
 
     # If want to use an identifier different from <id>, use this config.
-    PRIMARY_IDENTIFIER_SOURCE = u"PRIMARY_IDENTIFIER_SOURCE"
+    PRIMARY_IDENTIFIER_SOURCE = u"primary_identifier_source"
     DCTERMS_IDENTIFIER = u"first_dcterms_identifier"
 
     _cache = HasFullTableCache.RESET
--- a/model/configuration.py
+++ b/model/configuration.py
@@ -265,6 +265,13 @@ class ExternalIntegration(Base, HasFullTableCache):
     USERNAME = u"username"
     PASSWORD = u"password"
 
+    # If the request should use a custom headers, put it here.
+    CUSTOM_ACCEPT_HEADER = u"custom_accept_header"
+
+    # If want to use an identifier different from <id>, use this config.
+    CUSTOM_IDENTIFIER = u"custom_identifier"
+    DCTERMS_IDENTIFIER = u"first_dcterms_identifier"
+
     _cache = HasFullTableCache.RESET
     _id_cache = HasFullTableCache.RESET
 
@@ -487,6 +494,22 @@ class ExternalIntegration(Base, HasFullTableCache):
     @password.setter
     def password(self, new_password):
         return self.set_setting(self.PASSWORD, new_password)
+
+    @hybrid_property
+    def custom_accept_header(self):
+        return self.setting(self.CUSTOM_ACCEPT_HEADER).value
+
+    @custom_accept_header.setter
+    def custom_accept_header(self, new_custom_accept_header):
+        return self.set_setting(self.CUSTOM_ACCEPT_HEADER, new_custom_accept_header)
+
+    @hybrid_property
+    def custom_identifier(self):
+        return self.setting(self.CUSTOM_IDENTIFIER).value
+
+    @custom_identifier.setter
+    def custom_identifier(self, new_custom_identifier):
+        return self.set_setting(self.CUSTOM_IDENTIFIER, new_custom_identifier)
 
     def explain(self, library=None, include_secrets=False):
         """Create a series of human-readable strings to explain an
--- a/model/configuration.py
+++ b/model/configuration.py
@@ -269,7 +269,7 @@ class ExternalIntegration(Base, HasFullTableCache):
     CUSTOM_ACCEPT_HEADER = u"custom_accept_header"
 
     # If want to use an identifier different from <id>, use this config.
-    CUSTOM_IDENTIFIER = u"custom_identifier"
+    PRIMARY_IDENTIFIER_SOURCE = u"PRIMARY_IDENTIFIER_SOURCE"
     DCTERMS_IDENTIFIER = u"first_dcterms_identifier"
 
     _cache = HasFullTableCache.RESET
@@ -504,12 +504,13 @@ class ExternalIntegration(Base, HasFullTableCache):
         return self.set_setting(self.CUSTOM_ACCEPT_HEADER, new_custom_accept_header)
 
     @hybrid_property
-    def custom_identifier(self):
-        return self.setting(self.CUSTOM_IDENTIFIER).value
+    def primary_identifier_source(self):
+        return self.setting(self.PRIMARY_IDENTIFIER_SOURCE).value
 
-    @custom_identifier.setter
-    def custom_identifier(self, new_custom_identifier):
-        return self.set_setting(self.CUSTOM_IDENTIFIER, new_custom_identifier)
+    @primary_identifier_source.setter
+    def primary_identifier_source(self, new_primary_identifier_source):
+        return self.set_setting(self.PRIMARY_IDENTIFIER_SOURCE,
+                                new_primary_identifier_source)
 
     def explain(self, library=None, include_secrets=False):
         """Create a series of human-readable strings to explain an
--- a/opds2_import.py
+++ b/opds2_import.py
@@ -823,7 +823,8 @@ class OPDS2Importer(OPDSImporter):
 
         return dates
 
-    def extract_feed_data(self, feed, feed_url=None, custom_identifier=None):
+    def extract_feed_data(self, feed, feed_url=None,
+                          primary_identifier_source=None):
         """Turn an OPDS 2.0 feed into lists of Metadata and CirculationData objects.
 
         :param feed: OPDS 2.0 feed
--- a/opds2_import.py
+++ b/opds2_import.py
@@ -823,8 +823,7 @@ class OPDS2Importer(OPDSImporter):
 
         return dates
 
-    def extract_feed_data(self, feed, feed_url=None,
-                          primary_identifier_source=None):
+    def extract_feed_data(self, feed, feed_url=None):
         """Turn an OPDS 2.0 feed into lists of Metadata and CirculationData objects.
 
         :param feed: OPDS 2.0 feed
--- a/opds2_import.py
+++ b/opds2_import.py
@@ -823,7 +823,7 @@ class OPDS2Importer(OPDSImporter):
 
         return dates
 
-    def extract_feed_data(self, feed, feed_url=None):
+    def extract_feed_data(self, feed, feed_url=None, custom_identifier=None):
         """Turn an OPDS 2.0 feed into lists of Metadata and CirculationData objects.
 
         :param feed: OPDS 2.0 feed
--- a/opds_import.py
+++ b/opds_import.py
@@ -576,7 +576,8 @@ class OPDSImporter(object):
     def __init__(self, _db, collection, data_source_name=None,
                  identifier_mapping=None, http_get=None,
                  metadata_client=None, content_modifier=None,
-                 map_from_collection=None, mirrors=None
+                 map_from_collection=None, mirrors=None,
+                 primary_identifier_source=None
     ):
         """:param collection: LicensePools created by this OPDS import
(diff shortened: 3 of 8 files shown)


tests/test_opds_import.py
```diff
@@ -533,6 +535,24 @@ def test_extract_metadata(self):
...
 
+    def test_use_dcterm_identifier_as_id(self):
+        data_source_name = "Data source name " + self._str
+        importer = OPDSImporter(
+            self._db, collection=None, data_source_name=data_source_name
+        )
+        metadata, failures = importer.extract_feed_data(
+            self.feed_with_id_and_dcterms_identifier,
+            custom_identifier=ExternalIntegration.DCTERMS_IDENTIFIER
+        )
```
leonardr, line 549: This is missing the opposite test: importing this same feed with custom_identifier set to None. In that case `<id>` should be the primary identifier for both books.

model/configuration.py
```diff
@@ -265,6 +265,13 @@ class ExternalIntegration(Base, HasFullTableCache):
     USERNAME = u"username"
     PASSWORD = u"password"
 
+    # If the request should use a custom headers, put it here.
+    CUSTOM_ACCEPT_HEADER = u"custom_accept_header"
+
+    # If want to use an identifier different from <id>, use this config.
+    CUSTOM_IDENTIFIER = u"custom_identifier"
```
leonardr: Calling this CUSTOM_IDENTIFIER makes it sound like the configuration value should be _one specific identifier_. I recommend calling it something like PRIMARY_IDENTIFIER_SOURCE to avoid confusion -- this is the place you look for the identifier, not the identifier itself.

opds_import.py
```diff
@@ -532,6 +532,29 @@ class OPDSImporter(object):
             "label": _("Password"),
             "description": _("If HTTP Basic authentication is required to access the OPDS feed (it usually isn't), enter the password here."),
         },
+        {
+            "key": ExternalIntegration.CUSTOM_ACCEPT_HEADER,
+            "label": _("Custom accept header"),
+            "required": False,
+            "description": _("Some servers expect an accept header to decide which file to send. You can use */* if the server doesn't expect anything. The default values if left blank is: 'application/atom+xml;profile=opds-catalog;kind=acquisition, application/atom+xml;q=0.9, application/xml;q=0.8, */*;q=0.1'"),
```
leonardr: Rather than mentioning this complex default in the description, make it the default value as part of the configuration. This will let you get rid of the `_get_accept_header` method altogether.

opds_import.py
```diff
@@ -727,7 +750,7 @@ def _parse_identifier(self, identifier):
         """
         return parse_identifier(self._db, identifier)
 
-    def import_from_feed(self, feed, feed_url=None):
+    def import_from_feed(self, feed, feed_url=None, custom_identifier=None):
```
leonardr: Other pieces of configuration like `data_source_name` and `mirrors` are configured in the `OPDSImporter` constructor; this should be moved to the constructor as well. (And it should be called something like `primary_identifier_source` -- same ideas as my earlier comment.)

Among other things, this means you won';t have to pass this variable from one method to another--it'll be an instance variable.

opds_import.py
```diff
@@ -929,7 +952,18 @@ def extract_feed_data(self, feed, feed_url=None):
...
+            external_identifier = None
+            if custom_identifier == ExternalIntegration.DCTERMS_IDENTIFIER:
+                dcterms_ids = xml_data_dict.get('dcterms_identifiers', [])
+                if len(dcterms_ids) > 0:
+                    external_identifier, ignore = Identifier.for_foreign_id(
+                            self._db, dcterms_ids[0].type, dcterms_ids[0].identifier)
+
+            if external_identifier is None:
+                external_identifier, ignore = Identifier.parse_urn(self._db, id)
+
```
leonardr, line 990: If there are multiple <dcterms:identifier> tags, I would expect _all_ of them to be picked up and converted into IdentifierData objects. That's because we may need those identifiers later on, for things like finding book covers, or determining that two different books are the same literary work.

The metadata layer will treat the _first_ IdentifierData in a list as the primary identifier--the one you're creating with this code. I would create a helper method that handles the `identifier_mapping`, `identified_failures`, `identifier_obj` stuff, and then call it for every item in the the `dcterms_identiifer` list, regardless of the value of `custom_identiifer`.

Then, if `custom_identifier` is _not_ set to `DCTERMS_IDENTIFIER`, I would parse the `<id>`, run it through the helper method, and insert the resulting `IdentifierData` at the _beginning_ of the list, making it the primary.

Basically, everything in `dcterms_identifier` should become an `IdentifierData`, with the `<id>`  _maybe_ becoming an `IdentifierData` as well.

tests/files/opds/feed_with_id_and_dcterms_identifier.opds
```diff
@@ -0,0 +1,52 @@
...
+<summary>Book 1 summary.</summary>
+<category label="Fiction" term="FBFIC000000" scheme="http://www.feedbooks.com/categories"/>
+<link type="text/html" rel="http://opds-spec.org/acquisition/buy" href="https://book.uri/buy">
+  <opds:price currencycode="USD">01.01</opds:price>
+</link>
+</entry>
+<entry schema:additionalType="http://schema.org/Ebook">
+<title>Book 2</title>
+<id>https://root.uri/2</id>
+<dcterms:identifier xsi:type="dcterms:URI">urn:ISBN:9781468316438</dcterms:identifier>
```
leonardr, line 36: If and when you add code to handle multiple `dcterms:identifier` items, modify this document to include more than one.

opds_import.py
leonardr: Looking more closely at the code , I see that that `dcterms_identifiers` already contains a bunch of `IdentifierData` objects. Those objects will be passed in to the `Metadata` constructor as the `identifiers` argument, so the imported book will in fact end up with a number of associated `Identifier`s. So that part is already fine; I was wrong about what this code is doing.

All you need to do is change what ends up as 'primary_identiifer'. Right now you do that by replacing `external_identifier`. That's a little awkward because an `IdentifierData` is becoming an `Identifier` and then immediately becoming an `IdentifierData` again. But it works fine.

So this could use a comment to explain that you're seeing which of the IdentifierData objects should become the primary identifier, but other than that no code changes are necessary.

leonardr (changes requested):
This is nearly there. There are some issues with the way the configuration settings are conveyed, and a minor issue where method signatures are changed when they don't need to be. But the problem I thought was the most serious turned out to be something that can be addressed with an explanatory comment.

tests/files/opds/feed_with_id_and_dcterms_identifier.opds
leonardr: You don't need to add that code -- it should already happen -- but modify the test data, and then change the test to show that all the identifiers are present.

opds_import.py
arielmorelli: I totally agree about changing the description to get smaller one and use a default value.
Without running a script to populate the db with the values for the existing collections we cannot get rid of the `_get_accept_headers` method, so I’ll let the method existing to keep the backwards compatibility.

opds_import.py
arielmorelli: I was trying to do this, but I just use the same logic of the custom accept header. I'm changing to use the same logic of `data_source_name` and `mirrors` now.

opds_import.py
arielmorelli: Once we just can take the `dcterms:identifier` from the metadata (limitations on the feedparse library), I couldn't use the value before transform in a `IdentifierData`.
I'm commenting this trying to make clear.

Also, I discovery that <id> was being ignored, i'm fixing it too.

opds_import.py
leonardr: You could write a database migration script for this and add it to this branch, but it shouldn't be necessary: the default value will be used if a collection has no ConfigurationSetting object for this setting.

Try it out, but setting a default and removing the code should make everything work as it did before.

opds_import.py
leonardr: I see. Since the `OPDSImporter` constructor takes a `collection` as an argument, you can pull the information from the collection configuration, as you do in `OPDSImportMonitor` with the custom accept header, and avoid adding any new arguments to any method signatures.

opds_import.py
arielmorelli: I ran 2 tests: without changing the code and removing the `_get_accept_header method`, both with the same database config.

Without changing the code, I got this accept in request header:
`'Accept': 'application/atom+xml;profile=opds-catalog;kind=acquisition,application/atom+xml;q=0.9,application/xml;q=0.8,*/*;q=0.1'`

Removing the `_get_accept_header` method, I got this accept header:
`'Accept': '*/*' `

It will probably break many things that are working. I think is better not touching it right now. What you think, @leonardr ?

tests/files/opds/feed_with_id_and_dcterms_identifier.opds
arielmorelli: My fault, totally forgot about this test case.

opds_import.py
leonardr: I don't see how you would end up with the `*/*` Accept header unless it was stored in your database configuration. But the branch works as it is so I agree it's not necessary to remove that code right now.

arielmorelli:
@leonardr can you check again, please? I resolved everything we talked about in this PR.

opds_import.py
```diff
@@ -955,11 +967,25 @@ def extract_feed_data(self, feed, feed_url=None, custom_identifier=None):
...
-                            self._db, dcterms_ids[0].type, dcterms_ids[0].identifier)
+                            self._db, dcterms_ids[0].type, dcterms_ids[0].identifier
+                    )
+                    # the external identifier will be add later, so it must be removed at this point
+                    new_identifiers = dcterms_ids[1:]
+                    # Id must be in the identifiers with lower weight.
+                    id_type, id_identifier = Identifier.type_and_identifier_for_urn(id)
+                    id_weight = (dcterms_ids[-1].weight-0.1
+                                            if dcterms_ids[-1].weight > 0.2
+                                            else 0.1)
```
leonardr: I think it's OK to give all the identifiers the same weight.

tests/test_opds_import.py
```diff
@@ -535,23 +536,69 @@ def test_extract_metadata(self):
...
+            '9781683351504',
+            '9780312939458',
+        ]
+        for target_identifier in expected_identifier:
+            found = False
+            for entry in book_3.identifiers:
+                if entry.identifier == target_identifier:
+                    found = True
+                    break
+            ok_(found, "Cannot find %s" % target_identifier)
```
leonardr: This is fine, but for next time, make two lists and compare them directly, rather than writing a test that has nested loops -- it's easier to read. Since the ordering of the lists isn't relevant, convert them to sets and compare the sets. So something like this:

```
eq_(
  set([x.identifier for x in book_3.identifiers]),
  set(expected_identifier)
)
```

If there's a failure the error message will make clear the difference between the expected result and the actual.

See this example from elsewhere in test_opds_import.py:
```
        # The pools have presentation editions.
        eq_(set(["The Green Mouse", "Johnny Crow's Party"]),
            set([x.presentation_edition.title for x in pools_g]))
```

leonardr: This would also verify that no unexpected identifiers snuck into `book_3.identifiers`. That's not a huge concern here, but this test only verifies that `book_3.identifiers` contains every expected identifier, not that the expected and actual group is identical.

opds2_import.py
```diff
@@ -823,7 +823,8 @@ def extract_last_update_dates(self, feed):
 
         return dates
 
-    def extract_feed_data(self, feed, feed_url=None, custom_identifier=None):
+    def extract_feed_data(self, feed, feed_url=None,
+                          primary_identifier_source=None):
```
leonardr: The final argument is no longer used in this method; take it out altogether to avoid changing the method signature.

leonardr (changes requested):
Make all the identifiers the same weight, and take out the now-unnecessary final argument to extract_feed_data, and we should be good to go.

leonardr (approved):
This looks great! Congratulations on your first PR on this project!

PR #1228, opened 2020-12-21
Modify OverdriveAPI.make_link_safe to also replace "v1" links with "v2" links when appropriate

leonardr:
This branch adapts an existing method designed for "fixing" links we get from Overdrive to handle a new case: a case where Overdrive gives us a link to v1 of the availability API, and we need v2. Continuing to use the v1 link causes https://jira.nypl.org/browse/SIMPLY-3370.

[The corresponding circulation branch](https://github.com/NYPL-Simplified/circulation/pull/1540) calls  `make_link_safe` on the actual link that turns out to be a v1 link.

### DIFF
--- a/overdrive.py
+++ b/overdrive.py
@@ -533,8 +533,9 @@ class OverdriveAPI(object):
         """
         parts = list(urlparse.urlsplit(url))
         parts[2] = urllib.quote(parts[2])
+        endings = ("/availability", "/availability/")
         if (parts[2].startswith("/v1/collections/")
-            and parts[2].endswith("/availability")):
+            and any(parts[2].endswith(x) for x in endings)):
             parts[2] = parts[2].replace(
                 "/v1/collections/", "/v2/collections/", 1
             )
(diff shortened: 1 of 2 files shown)


overdrive.py
```diff
@@ -525,11 +525,19 @@ def metadata_lookup_obj(self, identifier):
...
+        The {} part is completely obnoxious and I have complained about it to
         Overdrive.
+
+        The availability part is to make sure we always use v2 of the
+        availability API, even if Overdrive sent us a link to v1.
         """
         parts = list(urlparse.urlsplit(url))
         parts[2] = urllib.quote(parts[2])
+        if (parts[2].startswith("/v1/collections/")
+            and parts[2].endswith("/availability")):
```
vbessonov: Is there any chance the Overdrive URL can end with `/availability/` (with an additional `/` at the end)?

vbessonov (approved):
I just added one comment to make sure that Overdrive can't add an additional `/` at the end.

overdrive.py
leonardr: There are no examples like this in the API documentation, but it's easy to add a little defensive code.

The URL might have a query string, but urlsplit takes care of that.

PR #1229, opened 2021-01-04
Circulation Manager ignores patron authentication state when requesting an OPDS feed's alternate link referencing Complete Catalog Entry Resource

vbessonov:
## Description

This PR drops the OPDS cache usage for permalinks returning catalog entries. Depending on the authentication status of the patron it can add to the catalog entry fulfilment links, revocation links, etc.

## Motivation and Context

[SIMPLY-3354](https://jira.nypl.org/browse/SIMPLY-3354)

## Checklist:

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -1010,7 +1010,7 @@ class AcquisitionFeed(OPDSFeed):
 
     @classmethod
     def single_entry(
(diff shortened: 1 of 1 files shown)


leonardr (approved):
This is a good change, so let's merge it for its own sake, but I suspect you will end up not needing it in your circulation branch.

PR #1230, opened 2021-01-07
Complete OPDS catalog entries must contain self links

vbessonov:
## Description

This PR adds an ability to override `rel` of `application/atom+xml;type=entry;profile=opds-catalog` links: depending on whether the OPDS entry is complete or incomplete `rel` will be set to `self` or `alternate` accordingly.

## Motivation and Context

[SIMPLY-3355](https://jira.nypl.org/browse/SIMPLY-3355)

## Checklist:

- [ ] All new and existing tests passed.

### DIFF
--- a/opds.py
+++ b/opds.py
@@ -63,6 +63,19 @@ class Annotator(object):
 
     opds_cache_field = Work.simple_opds_entry.name
 
+    def is_work_entry_solo(self, work):
+        """Return a boolean value indicating whether the work's OPDS catalog entry is served by itself,
+            rather than as a part of the feed.
+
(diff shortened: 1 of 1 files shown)


leonardr:
See my comment on https://jira.nypl.org/browse/SIMPLY-3355. I believe there's a simple way to resolve this problem while leaving the more difficult work for later.

leonardr (commented):
My requests for terminology changes from https://github.com/NYPL-Simplified/circulation/pull/1547 also apply here.

leonardr (approved):

PR #1233, opened 2021-01-22
Set a default value for licensepools.self_hosted

leonardr:
This branch fixes SIMPLY-3470. I also added unit test coverage for Facets.modify_database_query since we've had a number of problems with its precise semantics.

### DIFF
--- a/lane.py
+++ b/lane.py
@@ -675,7 +675,7 @@ class Facets(FacetsWithEntryPoint):
         available_now = or_(
             LicensePool.open_access == True,
             LicensePool.self_hosted == True,
-            LicensePool.unlimited_access == True,
+            LicensePool.unlimited_access,
             LicensePool.licenses_available > 0
         )
 
@@ -693,7 +693,11 @@ class Facets(FacetsWithEntryPoint):
             # depending on what exactly the wording is.
             availability_clause = LicensePool.open_access == True
         elif self.availability == self.AVAILABLE_NOT_NOW:
-            availability_clause = not_(available_now)
+            # The book must be licensed but currently unavailable.
+            availability_clause = and_(
+                not_(available_now),
+                LicensePool.licenses_owned > 0
+            )
 
         qu = qu.filter(availability_clause)
 
--- a/migration/20200702-add-licensepool-self_hosted-field.sql
+++ b/migration/20200702-add-licensepool-self_hosted-field.sql
@@ -2,7 +2,7 @@ DO $$
  BEGIN
   -- Add the 'self_hosted' column
   BEGIN
-   ALTER TABLE licensepools ADD COLUMN self_hosted BOOLEAN;
+   ALTER TABLE licensepools ADD COLUMN self_hosted BOOLEAN DEFAULT false;
   EXCEPTION
    WHEN duplicate_column THEN RAISE NOTICE 'column licensepools.self_hosted already exists, not creating it.';
   END;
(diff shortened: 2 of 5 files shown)


lane.py
```diff
@@ -675,7 +675,7 @@ def modify_database_query(self, _db, qu):
         available_now = or_(
             LicensePool.open_access == True,
             LicensePool.self_hosted == True,
-            LicensePool.unlimited_access == True,
+            LicensePool.unlimited_access,
```
leonardr, line 678: This doesn't make a difference, but it's consistent with what we do later in this method, and it makes it clear that unlimited_access is a synthetic property rather than a boolean stored in the database.

lane.py
```diff
@@ -693,7 +693,11 @@ def modify_database_query(self, _db, qu):
             # depending on what exactly the wording is.
             availability_clause = LicensePool.open_access == True
         elif self.availability == self.AVAILABLE_NOT_NOW:
-            availability_clause = not_(available_now)
+            # The book must be licensed but currently unavailable.
+            availability_clause = and_(
+                not_(available_now),
+                LicensePool.licenses_owned > 0
+            )
```
leonardr, line 700: This does make a difference; without this code a book that's _never_ going to be available could show up in a feed of  books that are not available _right now_.

migration/20200702-add-licensepool-self_hosted-field.sql
```diff
@@ -2,7 +2,7 @@ DO $$
  BEGIN
   -- Add the 'self_hosted' column
   BEGIN
-   ALTER TABLE licensepools ADD COLUMN self_hosted BOOLEAN;
+   ALTER TABLE licensepools ADD COLUMN self_hosted BOOLEAN DEFAULT false;
```
leonardr, line 5: Just in case anyone hasn't upgraded since July.

migration/20210121-self-hosted-is-required.sql
```diff
@@ -0,0 +1,7 @@
+-- The licensepools.self_hosted field has a default value, but the migration
+-- script that added it to old servers didn't set one. Now old servers have
+-- an invalid 'null' value in that field.
+--
+-- (see also 20200702-add-licensepool-self_hosted-field.sql)
+ALTER TABLE licensepools ALTER COLUMN self_hosted SET DEFAULT false;
+UPDATE licensepools SET self_hosted = false WHERE self_hosted is null;
```
leonardr, line 7: Both of these are idempotent so there's no worry about running them on an already migrated site, and thus no need to wrap them in a SQL try/catch block.

lane.py
EdwinGuzman: What do you mean by synthetic property? Is it like having _any_ value instead of explicitly being a boolean?

lane.py
EdwinGuzman: In what case will a book be stored that will never be available? (If a book is not getting a license, why include it in the db?)

leonardr: This happens when a library licenses a book and then the license expires. We still want to keep track of the book because people may have active loans or annotations on the book. 

"Never" is maybe too strong because the library could always license the book again.

lane.py
leonardr: No, it should always be a boolean. I meant that `LicensePool.unlimited_access` is defined in Python through `@property`, not as a real database field. I'd never seen a Python property used in a SQLAlchemy query before but I verified this works when reviewing Viacheslav's original code.

EdwinGuzman (approved):

PR #1234, opened 2021-01-28
Pin urllib3 to a pre-1.25 version to work around an Axis 360 ACS problem

leonardr:
## Description

This branch pins urllib3 to the final 1.24 version as a way of avoiding https://jira.nypl.org/browse/SIMPLY-3477.

This is a temporary measure until either Axis 360 changes their ACS setup or we figure out a better long-term solution.

## How Has This Been Tested?

Axis 360 fulfillment starts succeeding once urllib3 is downgraded to a 1.24 version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -52,7 +52,8 @@ singledispatch==3.4.0.3
 six==1.15.0
 SQLAlchemy==1.3.19
 textblob==0.15.3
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
Looks good to me, assuming tests pass.

PR #1236, opened 2021-02-03
Use delete instead of delete-orphan to manage the relationship between Library/ExternalIntegration and ConfigurationSetting.

leonardr:
This is an attempt to resolve https://jira.nypl.org/browse/SIMPLY-2983, with potential implications for future work.

Coming back to the "delete-orphan" cascade technique (https://docs.sqlalchemy.org/en/14/orm/cascades.html#delete-orphan) after getting more experience with SQLAlchemy, I'm pretty sure it's not suited for a lot of the places we're using it. delete-orphan is designed for:

> a related object that is “owned” by its parent, with a NOT NULL foreign key, so that removal of the item from the parent collection results in its deletion.

An example of a place where this is suitable is Patron/Annotation. An Annotation cannot exist without a Patron, so setting `annotation.patron = None` is the same as deleting it.

An example of a place where I think we should use "delete" instead of "delete-orphan"  is Library/CirculationEvent. CirculationEvent.library is a foreign key, but it's not a problem if that value is null. I don't think this comes up a lot because we never actually set `CirculationEvent.library = None`, but it's incorrect.

And we should _definitely_ be using "delete" instead of "delete-orphan" in Library/ConfigurationSetting and ExternalIntegration/ConfigurationSetting, because there we have _two_ foreign keys, either or both of which might be null. We also have an ongoing problem where `ConfigurationSetting` objects are randomly deleted in a way that indicates a delete cascade is going haywire.

This branch makes the delete-orphan->delete change for the two relationships where I think this problem is the worst--in fact, where I suspect this problem is causing SIMPLY-2983. Depending on how this goes, I'll do a follow-up branch that changes delete-orphan->delete for the other cases where I think it makes sense.


tests/models/test_configuration.py
```diff
@@ -224,6 +227,32 @@ def test_relationships(self):
...
+        for_integration = ConfigurationSetting.for_externalintegration(
+            self._str, integration
+        )
+
+        # Remove library and external_integration.
+        for_library.library = None
+        for_integration.external_integration = None
+        self._db.commit()
+
+        # That was a weird thing to do, but the ConfigurationSettings
```
EdwinGuzman, line 249: Weird thing to do because this doesn't happen in the code, but can happen through the UI somehow?

EdwinGuzman:
I can probably test this in the UI by creating a Library and ConfigurationSetting and deleting the Library and looking for the ConfigurationSetting in the db?

tests/models/test_configuration.py
leonardr: I don't think it happens at all. And I'd actually prefer the behavior where setting one of those fields to None immediately deletes the ConfigurationSetting. But "delete-orphan" isn't a good way to get that behavior in a situation where there's any legitimate reason for those fields to be None.

leonardr:
> I can probably test this in the UI by creating a Library and ConfigurationSetting and deleting the Library and looking for the ConfigurationSetting in the db?

Deleting the Library will trigger the "delete" part of the cascade, so the ConfigurationSettings will still end up deleted. That's worth a try to verify that it still works, but I've never found a good way to trigger the "delete-orphan" part specifically. I believe "delete-orphan" only happens by mistake.

tests/models/test_configuration.py
leonardr: I think it's weird to do that because the way I feel is, if you want to delete a database object, just delete it. Don't set a foreign key it depends on to a null value.

jonathangreen:
This change looks good to me! Good catch! 

I did a little testing in the UI to see how deletes cascade to see if this will leave any configuration settings around that would have been legitimately deleted before and here is what I found:

Test setup
- New CM for testing
- Add a library
- Add a overdrive collection and add that collection to the library

I started fresh and performed that setup before conducting each of these tests:
1. Disassociate the library from the collection
2. Delete the library
3. Delete the collection, then run `database_reaper`

Look at the settings left in the `configurationsettings` table after each step. With both the new and the old code, the results were the same, so I don't think these results should hold up this PR.

I was a bit surprised by the results though, so maybe another ticket is in order to make sure that `configurationsettings` get cleaned up if they are no longer needed.

> 1. Disassociate the library from the collection

After this the `configurationsettings` referencing the `library` and `externalintegrations` for the collection were still in the table, but their values were set to null. 

> 2. Delete the library

All the `configurationsettings` referencing the `library` are removed.

> 3. Delete the collection, then run `database_reaper`

All of the `configurationsettings` and `externalintegrations` for the collection remain in the database, even though the collection itself is removed.

leonardr:
Thanks, Jonathan, at least #1 and #3 there look like they need some work. I filed https://jira.nypl.org/browse/SIMPLY-3504 to cover that work.

EdwinGuzman (approved):
My testing wasn't as in-depth as Jonathan's but did notice that the results were the same. Approving since there's a ticket for the other work.

PR #1237, opened 2021-02-04
When a library is removed from a collection, its configuration for that collection is completely deleted

leonardr:
This branch partially addresses https://jira.nypl.org/browse/SIMPLY-3504; I also have a circulation branch for it at https://github.com/NYPL-Simplified/circulation/pull/1553.

Basically, if you disassociate a library from a collection, you should call `Collection.disassociate_library`, which takes care of clearing that library's configuration settings for the collection. Deleting a collection disassociates every library from the collection and deletes the corresponding ExternalIntegration.

I don't think this is causing any serious problem, but there is an inconsistency in the way we handle ConfigurationSettings when the things being configured are deleted, and my other PR https://github.com/NYPL-Simplified/server_core/pull/1235 might turn this into a real problem by making it difficult to recreate a collection after deleting it.

### DIFF
--- a/tests/models/test_collection.py
+++ b/tests/models/test_collection.py
@@ -571,6 +571,63 @@ class TestCollection(DatabaseTest):
         # Only the failing identifier is in the query.
         eq_([unresolved_id], result.all())
 
+    def test_disassociate_library(self):
+        # Here's a Collection.
+        collection = self._default_collection
+
+        # It's associated with two different libraries.
+        assert self._default_library in collection.libraries
+        other_library = self._library()
+        collection.libraries.append(other_library)
+
+        # It has an ExternalIntegration, which has some settings.
+        integration = collection.external_integration
+        setting1 = integration.set_setting("integration setting", "value2")
+        setting2 = ConfigurationSetting.for_library_and_externalintegration(
+            self._db, "default_library+integration setting",
+            self._default_library, integration,
+        )
+        setting2.value = "value2"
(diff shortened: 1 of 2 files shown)


tests/models/test_collection.py
```diff
@@ -885,13 +942,11 @@ def remove_work(self, work):
         # has any LicensePools), but not the second.
         eq_([work], index.removed)
 
-        # The ExternalIntegration and its settings are still around,
-        # since multiple Collections can be based on the same
-        # ExternalIntegration.
```
leonardr, line 890: I really don't think this is true. Maybe we thought it might be true when we were writing the Collection/ExternalIntegration code, but we've decided on a system where an Externalintegration is just a convenient way of gathering together some configuration information about a Collection. You could create two Collections with the same credentials (though this sounds like a bad idea) but they would have different underlying ExternalIntegrations.

jonathangreen (commented):
This looks good to me. 👍🏻 I went though the same manual testing as before, and the configurationsettings and externalintegrations in the database are both cleaned up this time. 

I'm wondering if maybe we want to add a DB migration to clean up old orphaned configurationsettings or externalintegrations that may have been left behind in the past. They shouldn't be hurting anything, but they may cause some issues after #1235 lands.

leonardr:
I think we'll be OK. The migration script for #1235 will delete any duplicates, so we'll be left with what looks like some weird sitewide settings (no externalintegration_id or library_id). Going forward, we won't be creating or using any of those fake sitewide settings, so they'll just be ignored.

PR #1238, opened 2021-02-09
Make the OPDS importer update acquisition links

vbessonov:
## Description

This PR fixes the bug related to the OPDS importer not updating acquisition.

I don't think that adding `LicensePool. reset_open_access_download_url` method is a good solution, maybe it's better to create a setter for `LicensePool.best_open_access_link` property.

## Motivation and Context

Steps to reproduce:

1. Import an OPDS feed using the OPDS importer.
2. Update acquisition links.
3. Reimport the feed.
4. Notice that the links were not updated.

## How Has This Been Tested?

@tdilauro, I used the following steps for testing:
1. Restore the database dump:
```
pg_restore -d simplified_circulation_dump local-db-simplified_circ_db.dump
```
2. Create a mock service using [SoapUI](https://www.soapui.org/) served at `http://localhost:8080/test` which returns the same feed except `Album de grilles en fer creux et fer massif` book where:
- I updated all acquisition links by adding `1234567890` suffix to the their end.
- I set `<updated>` tag's value to the current date and time in UTC.
3. Update the URL of the OPDS collection:
```
update collections set external_account_id = 'http://localhost:8080/test' where id = 5;
```
4. Run `opds_import_monitor` script to reimport the updated book. 
_Please note that the whole page where `Album de grilles en fer creux et fer massif` is going to be reimported (this is by design)._
5. Recreate the search index. 
_Please note that the actions described above are required only because the local search index doesn't have information about all the books from the dump which is not the case in production._
- Drop the current index: 
```
curl --location --request DELETE 'localhost:9200/_all'
```
- Truncate all the work coverage records:
```
truncate workcoveragerecords;
```
- Refresh the search index by running `search_index_refresh`.
- Drop the cached feeds:
```
truncate cachedfeeds;
```
6. Open the admin UI. Because I don't know the admin's password I had to reset it by running the following query that sets the admin's password to `servers@lyrasistechnology.org`:
```
update admins set password_hashed = '$2a$12$sdOOruVaDzgVqVF6K4zp5.dyYtcykzIlGpZEA.m8RL3nJHaMPTswu' where id = 19;
```
7. Create a basic authentication provider. It's required for the CM's feed to work correctly, otherwise you won't be able to checkout books.
8. Find `Album de grilles en fer creux et fer massif` book and check it out.
9. Try to download the book and check in the logs or using the debugger that the fulfilment link actually contains `1234567890` suffix at the end.

Also I just found out that the change I made doesn't fully fix the issue related to cover links not being updated. In the case of this particular feed, they remain the same. You can confirm this by looking at the loans feed at `http://localhost:6500/190150/loans` after executing **step 8**.

It looks like that there are a couple of more changes required to fix the cover links issue:
1. Remove `Resource` records corresponding to `Hyperlink` records being deleted in `Metadata.apply` that can be implemented by adding `Hyperlink.delete` and `Resource.delete` methods in the same way as `LicensePoolDeliveryMechanism.delete`.
2. Erase `Edition.cover_full_url` and `Edition.cover_thumbnail_url` in `Metadata.apply` if `ReplacementPolicy.links` is `True`.
3. Reinitialize `Edition.cover_full_url` and `Edition.cover_thumbnail_url` using "new" links specified in `Metadata` object.
Unfortunately, I failed when I tried to implement all those steps and I think it can be implemented in a separate PR once this one is merged.

## Checklist:

### DIFF
--- a/model/licensing.py
+++ b/model/licensing.py
@@ -1125,7 +1125,7 @@ class LicensePool(Base):
         if not self.identifier:
             return
         q = Identifier.resources_for_identifier_ids(
-            _db, [self.identifier.id], open_access
+            _db, [self.identifier.id], open_access, data_source=self.data_source
         )
         for resource in q:
             yield resource
@@ -1222,7 +1222,13 @@ class LicensePool(Base):
             self.data_source, self.identifier, *args, **kwargs
         )
 
+    def reset_open_access_download_url(self):
+        """Reset this license pool's open-access download url."""
+        self._open_access_download_url = None
+
+
 Index("ix_licensepools_data_source_id_identifier_id_collection_id", LicensePool.collection_id, LicensePool.data_source_id, LicensePool.identifier_id, unique=True)
+
 
 class LicensePoolDeliveryMechanism(Base):
     """A mechanism for delivering a specific book from a specific
--- a/metadata_layer.py
+++ b/metadata_layer.py
@@ -1990,21 +1990,23 @@ class Metadata(MetaToModelUtility):
             # If there is a Work associated with the Edition's primary
             # identifier, mark it for recalculation.
 
-            # Any LicensePool will do here, since all LicensePools for
-            # a given Identifier have the same Work.
-            pool = get_one(
-                _db, LicensePool, identifier=edition.primary_identifier,
-                on_multiple='interchangeable'
-            )
-            if pool and pool.work:
-                work = pool.work
-                if work_requires_full_recalculation:
-                    work.needs_full_presentation_recalculation()
-                else:
-                    work.needs_new_presentation_edition()
+            pools = _db.query(
+                LicensePool).filter(LicensePool.identifier == edition.primary_identifier)
+
+            for pool in pools:
+                pool.reset_open_access_download_url()
+
+            for pool in pools:
+                if pool and pool.work:
+                    work = pool.work
+                    if work_requires_full_recalculation:
+                        work.needs_full_presentation_recalculation()
+                    else:
+                        work.needs_new_presentation_edition()
+
+                    break
 
         return edition, work_requires_new_presentation_edition
-
 
     def make_thumbnail(self, data_source, link, link_obj):
         """Make sure a Hyperlink representing an image is connected
--- a/opds_import.py
+++ b/opds_import.py
@@ -829,6 +829,7 @@ class OPDSImporter(object):
         policy = ReplacementPolicy(
             subjects=True,
             links=True,
+            formats=True,
             contributions=True,
             rights=True,
             link_content=True,
--- a/tests/files/opds/content_server_mini_with_different_links.opds
+++ b/tests/files/opds/content_server_mini_with_different_links.opds
@@ -1,0 +1,64 @@
+<feed xmlns:simplified="http://librarysimplified.org/terms/" xmlns:app="http://www.w3.org/2007/app" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:opds="http://opds-spec.org/2010/catalog" xmlns:schema="http://schema.org/" xmlns="http://www.w3.org/2005/Atom" xmlns:bibframe="http://bibframe.org/vocab/">
+  <id>http://localhost:5000/</id>
+  <title>Open-Access Content</title>
+  <updated>2015-01-02T16:56:40Z</updated>
+  <link href="http://localhost:5000/"/>
+  <link href="http://localhost:5000/" rel="self"/>
+  <entry schema:additionalType="http://schema.org/PublicationIssue">
+    <id>urn:librarysimplified.org/terms/id/Gutenberg%20ID/10441</id>
+    <bibframe:distribution bibframe:ProviderName="Gutenberg"/>
+    <title>The Green Mouse</title>
+    <schema:alternativeHeadline>A Tale of Mousy Terror</schema:alternativeHeadline>
+    <author>
+      <name></name>
+      <simplified:sort_name>Chambers, Robert W. (Robert William)</simplified:sort_name>
+    </author>
+    <summary>This is a summary!</summary>
+    <updated>2015-01-02T16:56:40Z</updated>
+    <published>2014-01-02T16:56:40Z</published>
+    <simplified:pwid>6db4cf90-0129-0eaf-410a-502b20a45e67</simplified:pwid>
+    <schema:Rating schema:ratingValue="0.3333" schema:additionalType="http://librarysimplified.org/terms/rel/quality"/>
+    <schema:Rating schema:ratingValue="0.6000"/>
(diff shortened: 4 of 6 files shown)


tdilauro:
I verified this roughly based on the steps above. Got a redirect to the correct to the expected resource, depending on what I had in the acquisition link.

I can also confirm that the cover links are NOT updating. Given your comments above and the fact that you changed the title of this PR, I suspect that you ar changing the scope of this PR. If so, it would be appropriate to change the scope in the description above, as well.

vbessonov:
@tdilauro, I updated the description. If required, I can also create a new branch to avoid having `cover-links` in the branch name.

tdilauro:
> @tdilauro, I updated the description. If required, I can also create a new branch to avoid having `cover-links` in the branch name.

It would probably be good to get "covers" out of the branch name (for anyone perusing the log). However, I think pushing a new branch name effectively deletes the current branch and adds a new one, which would close this PR. @leonardr or another admin for this repo might be able to rename this branch directly in GH, though.

That said, there's not much discussion to transfer to another PR, so I don't think it would be terrible to supersede this PR.

model/licensing.py
```diff
@@ -1368,6 +1374,17 @@ def delete(self):
...
+        # Except deleting this record itself, we also need to remove all linked Hyperlink records.
+        if self.resource_id:
+            from resource import Resource
+            resource = get_one(_db, Resource, id=self.resource_id)
+
+            for hyperlink in resource.links:
+                _db.delete(hyperlink)
+
+            _db.delete(resource)
+
```
leonardr: This doesn't seem right to me. Two different LicensePoolDeliveryMechanisms are allowed to have the same Resource. This is unusual but it could happen if two different OPDS feeds linked to the same open-access book at the same URL.

Can you talk about the problem you're seeing here in the situation when a single OPDS feed changes which Resource it's linking to for a given book?

leonardr: Judging from your earlier comments, I feel like this is code that should be running in `Metadata.apply`, near this comment:

```
# Associate all links with the primary identifier.
```

That code should delete any Hyperlinks that used to be associated with this book's identifier but no longer are. And it shouldn't be necessary to delete the Resource itself (since another book might be using it). Why isn't that code working for you?

vbessonov: I think it's related to the fact that `LicensePool.best_open_access_link` relies on `LicensePool.best_open_access_resource` which in turn relies on `LicensePool.open_access_links` which fetches all resources associated with the pool's identifier which is probably not right.
I'm using the dump of the production system where we have multiple resources from different data sources associated with the same identifier.
It means that the code `Metadata.apply` remains hyperlinks from different data sources untouched, removing only hyperlinks from the current data source. Later, after we reset `LicensePool._open_access_download_url` by calling `reset_open_access_download_url` method, it tries to refresh the acquisition link when `LicensePool.best_open_access_link` gets called during the feed generation. Considering the fact that `LicensePool.open_access_links` fetches all the resources it ends up with wrong not updated resources belonging to different resources and shows the old acquisition link.
The other option here would be to explicitly specify the data source when calling `Identifier.resources_for_identifier_ids` in `LicensePool.open_access_links`. I'll test this approach.

vbessonov: @leonardr, the change I did (got rid of removing resources in `LicensePoolDeliveryMechanism.delete` and refined filtration of resources by the data source in `LicensePool.open_access_links`) surprisingly worked in the local environment I recreated from the production dump but it caused `TestOPDSImporter.test_importer_updates_acquisition_and_cover_links` test to fail.
The problem is that after the second import we end up with two resources: old one and a new one. And even filtration by data source doesn't help.

leonardr: This sounds like things are working -- open_access_links should be getting all resources associated with that identifier, which might include the same book from multiple sources. Maybe `best_open_access_resource` is just reaching the wrong decision in your situation? What are the different data sources you have providing the same book? What makes one better than another?

vbessonov: There is nothing wrong about data sources, somehow we just ended up with multiple data sources basically having the same resources pointing to the same URLs:
```
- Data source # 1 <- hyperlink # 1 <- resource # 1 with `http://books.com/book1.epub`
- Data source # 2 <- hyperlink # 2 <- resource # 2 with `http://books.com/book1.epub`
...
- Data source # N <- hyperlink # N <- resource # N with `http://books.com/book1.epub`
```

After we run `opds_import_monitor` for a particular collection (let's say related to data source # 2) we end up with the following state in the database (note that only resource # 2's URL has been updated):
```
- Data source # 1 <- hyperlink # 1 <- resource # 1 with `http://books.com/book1.epub`
- Data source # 2 <- hyperlink # 2 <- resource # 2 with `http://books.com/book1234567890.epub`
...
- Data source # N <- hyperlink # N <- resource # N with `http://books.com/book1.epub`
```
`LicensePool. best_open_access_resource` without filtration by a data source fetches the first resource and we end up with the wrong, not updated URL `http://books.com/book1.epub` instead of `http://books.com/book1234567890.epub`.

vbessonov:
@tdilauro, I created a SQL script for fixing this issue by updating the database:
```
# Delete all the hyperlinks related to the specific data source.
# It will force CM to use newly imported links.
delete
from
  hyperlinks as h
using
  resources as r
where
  h.resource_id = r.id
  and r.data_source_id = 47
;

# Swipe out fulfillment info because the URL of the acquisition link has changed and it has to be updated.
update loans as l
set
  fulfillment_id = null
from licensepools as lp
where
  l.license_pool_id = lp.id
  and lp.data_source_id = 47
;

# Delete fulfillment objects because the URL of the acquisition link has 
# changed and they have to be updated
delete from licensepooldeliveries
where
  data_source_id = 47
;

# Reset open_access_download_url so because the URL of the acquisition link has changed and it has to be updated.
update licensepools
set
  open_access_download_url = null
where
  data_source_id = 47
;

# Reset cover URLs to force them to be updated during the next import.
update editions
set
  cover_full_url = null,
  cover_thumbnail_url = null
where
  data_source_id = 47
;

# Make sure that the new cover and acquisition links have been updated.
select
  i.id,
  h.data_source_id,
  h.rel,
  r.url
from identifiers as i
inner join hyperlinks as h on i.id = h.identifier_id
inner join resources as r on h.resource_id = r.id
where
  i.identifier = 'urn:x-internet-archive:ebooks-app:item:albumdegrillesen00gand'
;
```
After you run the script, you can reimport the feed by running `opds_import_monitor` using `--force` flag.

@leonardr, do you think it's a valid approach?

model/licensing.py
leonardr: Deleting the old resources will work as a one-time measure, but it sounds like this problem can happen in the normal course of events, so we shouldn't rely on this technique.

You were able to look at this data and know that the first resource was wrong and the second resource was correct. Is there a way to encode that knowledge into `best_open_access_resource`?  Maybe we could add a rule that if there are two indistinguishable resources, the one that was imported later should take precedence.

PR #1241, opened 2021-02-25
Make basic info settings sys-admin-only

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3556 and https://jira.nypl.org/browse/SIMPLY-3557; it turns out that librarians and library managers weren't supposed to have access to the fields in the "Basic Information" section after all.


leonardr (approved):
This makes sense since changing this URL can screw up the SimplyE integration.

PR #1242, opened 2021-03-02
CollectionType knows if it is a self-hosted type.

tdilauro:
## Description

Adds `is_self_hosted` method to `CollectionType` enum.

## Motivation and Context

Centralize logic for which collection types are self hosted.

## How Has This Been Tested?

Ran new and existing tests.

## Checklist:
- N/A I have updated the documentation accordingly.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1779,6 +1779,15 @@ class CollectionType(Enum):
     def __str__(self):
         return self.name
 
+    @property
+    def is_self_hosted(self):
+        """Whether this collection type is self-hosted.
+
+        :return: Boolean asserting whether to treat this CollectionType as self-hosted.
+        :rtype: bool
+        """
+        return self in (self.OPEN_ACCESS, self.PROTECTED_ACCESS)
+
 
 class CollectionInputScript(Script):
     """A script that takes collection names as command line inputs."""
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -2111,6 +2111,27 @@ class TestConfigureLaneScript(DatabaseTest):
         eq_(expect, output.getvalue())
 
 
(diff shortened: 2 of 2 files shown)


scripts.py
```diff
@@ -1779,6 +1779,15 @@ class CollectionType(Enum):
...
         return self.name
 
+    @property
+    def is_self_hosted(self):
+        """Whether this collection type is self-hosted.
+
+        :return: Boolean asserting whether to treat this CollectionType as self-hosted.
+        :rtype: bool
+        """
+        return self in (self.OPEN_ACCESS, self.PROTECTED_ACCESS)
```
vbessonov, line 1789: I have a small concern about it. This is right in the context of `DirectoryImportScript` because we know that we import the content into Circulation Manager's storage, and it can be considered as self-hosted, but in other cases, OA content can be hosted by third parties and cannot be considered as self-hosted.
CollectionType is currently used only in `DirectoryImportScript`, so we can neglect that, but I'm just afraid that this enumeration will be used somewhere else, and it will bring some confusion.
For example, you can see [here](https://github.com/NYPL-Simplified/server_core/blob/develop/model/work.py#L1193) that we really differentiate OA and self-hosted as two separate cases.
Maybe we can fix it by just moving `CollectionType` enumeration inside `CollectionInputScript`.

vbessonov (commented):
I left a comment to express my concern regarding considering OA collections to be self-hosted.

scripts.py
tdilauro: Ah, I see. Thank you! If this should not be generalized, then I will withdraw this PR and go back to the original logic.

tdilauro:
Withdrawing.

PR #1243, opened 2021-03-05
Pass the feed URL into assert_importable_content, not an unrelated value

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-3588. The OPDS importer resolves relative links when doing a real import, but not when running a self-test. It looks like the signature of OPDSImporter.follow_one_link changed so that the first item of the return value is a list of "next" links, not the URL of the original page. That's the value we were passing in to `assert_importable_content`, and since it wasn't a URL (it was an empty list), we weren't able to resolve relative links.

Fortunately, we have the URL of the original request -- we used it to _make_ the original request -- so it's an easy fix to just pass that information in to `assert_importable_content`.

### DIFF
--- a/tests/test_opds_import.py
+++ b/tests/test_opds_import.py
@@ -2286,7 +2286,7 @@ class TestOPDSImportMonitor(OPDSImporterTest):
             # First we will get the first page of the OPDS feed.
             def follow_one_link(self, url):
                 self.follow_one_link_called_with.append(url)
-                return (url, "some content")
+                return ([], "some content")
 
         feed_url = self._url
         self._default_collection.external_account_id = feed_url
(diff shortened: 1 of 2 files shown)


tests/test_opds_import.py
```diff
@@ -2286,7 +2286,7 @@ class Mock(OPDSImportMonitor):
             # First we will get the first page of the OPDS feed.
             def follow_one_link(self, url):
                 self.follow_one_link_called_with.append(url)
-                return (url, "some content")
+                return ([], "some content")
```
leonardr, line 2289: The test didn't fail before because the old(?) behavior was still being mocked.

EdwinGuzman (approved):
Approving but the PR Travis test is failing.

leonardr:
The PR build is failing because it exceeds a Docker Hub rate limit. I filed https://jira.nypl.org/browse/SIMPLY-3592 to deal with this. I feel like we don't need the PR build given that we run the tests on every push. What do you think?

EdwinGuzman:
Yeah, both builds are essentially the same so we may not need the specific PR branch build. I ran into this issue yesterday but pushed another commit about 10 minutes later and it worked. If there's a 100 pull limit every 6 hours, I'm not sure if it's actually being applied.

leonardr:
I shut off Travis notifications on PRs for server_core, let's see what that does.

PR #1244, opened 2021-03-08
Python 3 Update

EdwinGuzman:
Fixes [SIMPLY-2215](https://jira.nypl.org/browse/SIMPLY-2215).

The goal for this is to fully convert the codebase to python 3 (3.6 which seems stable on Travis). This is used as a git submodule so Docker is not needed for this repo.
* Ran the codebase through `2to3`.
* Updated all `TODO python3` comments
* Removed dependency on `six` and `futures`.
* Removed `native_string` helper function.
* A common update was moving from `urlparse` to `urllib.parse`.
* Updated how `metaclass` is declared.
* Had to update some map/filter functions to list comprehensions.
* Had to fix comparisons where one variable could be `None` and it was expecting a number.
* Updated lxml `tounicode` to `tostring` and slightly updated the call for the updated package version.

Missing todos:
* A migration script to update the `marc_record` `Work` column to binary strings. Moving from `String` to `Binary` in the database works but as discussed with Leonard, this may be a big migration.

Although this PR is using python 3, there are still updates I discussed with @nballenger that are not part of this PR but want to include in subsequent PRs such as:
* linting
* moving from nosetests to pytest
* code coverage tool


python_expression_dsl/evaluator.py
```diff
@@ -20,7 +20,7 @@
     Visitor,
 )
 from .parser import DSLParser
-
+import types
```
EdwinGuzman, line 23: @vbessonov Do you mind looking over this file I updated to make it work in python 3? One test was failing because it was looking for a `module` type as a built in class. Then, in line 366, `im_class` is not available anymore so I updated that to check `.__self__.__class__`.

classifier/__init__.py
```diff
@@ -92,7 +92,7 @@ class Classifier(object):
     AUDIENCES = set([AUDIENCE_ADULT, AUDIENCE_ADULTS_ONLY, AUDIENCE_YOUNG_ADULT,
                      AUDIENCE_CHILDREN, AUDIENCE_ALL_AGES, AUDIENCE_RESEARCH])
     AUDIENCES_NO_RESEARCH = [
-        x for x in AUDIENCES if x != AUDIENCE_RESEARCH
+        x for x in AUDIENCES if x != "Research"
```
leonardr: Can we avoid spelling out the constant twice?

config.py
```diff
@@ -216,6 +216,7 @@ class Configuration(ConfigurationConstants):
     # Level 2 settings can be modified only by library managers and system admins (i.e. not by librarians).  Level 3 settings can be changed only by system admins.
     # If no level is specified, the setting will be treated as Level 1 by default.
     ALL_ACCESS = 1
+    global SYS_ADMIN_OR_MANAGER
```
leonardr: ???

classifier/__init__.py
EdwinGuzman: I forgot to point a few lines out I was having trouble updating. Here, `AUDIENCE_RESEARCH` is not defined in the scope of the list comprehension. It's fine if it's used without the `if` statement but I couldn't figure out how to use the variable otherwise.

coverage.py
```diff
@@ -393,7 +393,7 @@ def process_batch_and_handle_results(self, batch):
         num_ignored = 0
         records = []
 
-        unhandled_items = set(batch)
+        unhandled_items = batch
```
leonardr: I'm a little concerned about this change because I think it is possible for a batch to contain the same item twice. If that happens, whatever service is processing the items will probably return only a single status for that item. We'll call `unhandled_items.remove(item.obj)` once and `item` will still be in `unhandled_items` once we're done, making it look like the service just ignored that item.

If it's not possible to turn this into a set in Python 3, can we build a de-duped list instead?

config.py
EdwinGuzman: This is used later on line 291. Without the `global` I get a `name 'SYS_ADMIN_OR_MANAGER' is not defined`. I wasn't sure about this because `SYS_ADMIN_ONLY` is used in a similar way but that doesn't throw an error.

log.py
```diff
@@ -35,17 +33,18 @@ def only_native_strings(s):
             we don't want to try to interpolate an incompatible type; it
             could lead to a UnicodeDecodeError.
             """
-            if isinstance(s, (bytes, unicode)):
-                s = native_string(s)
+            if isinstance(s, bytes):
+                s = s.decode("utf-8")
             return s
+        message = only_native_strings(record.msg)
```
leonardr: This was originally designed for forward-compatibility with Py3. Now that we're moving over it could be simplified.

marc.py
```diff
@@ -589,7 +589,7 @@ def create_record(cls, work, annotator, force_create=False, integration=None):
         record = None
         existing_record = getattr(work, annotator.marc_cache_field)
         if existing_record and not force_create:
-            record = Record(data=existing_record.encode("utf8"), force_utf8=True)
+            record = Record(data=existing_record, force_utf8=True)
```
leonardr: You already mentioned this in your description, but this change and the one below are potentially a problem.

model/coverage.py
```diff
@@ -66,7 +66,7 @@ def not_covered(cls, count_as_covered=None,
         """
         if not count_as_covered:
             count_as_covered = cls.DEFAULT_COUNT_AS_COVERED
-        elif isinstance(count_as_covered, basestring):
+        elif isinstance(count_as_covered, str):
```
leonardr: I'm a little worried about stuff like this because if you pass in a bytestring it won't be covered anymore. I don't anticipate that actually happening, though. Is the general attitude that bytestrings are so rare in Python 3 that it won't come up?

model/identifier.py
```diff
@@ -859,6 +853,9 @@ def __eq__(self, other):
             return False
         return (self.type, self.identifier) == (other.type, other.identifier)
 
+    def __hash__(self):
+        return self.id
+
```
leonardr, line 858: I'm a little worried about the difference between `__eq__` (where two objects are the same if they're the same identifier) and `__hash__` (where two objects are the same if they're the same database row). Do you remember why you had to add this?

EdwinGuzman: Yeah, there are two tests that fail without this, `TestWork.test_reject_covers` and `TestEdition.test_equivalent_identifiers`. The error is `TypeError: unhashable type: 'Identifier'`. It looks like I can test for equivalency without putting the Identifiers in a `set(...)` but the `test_reject_covers` might need more than that to fix the test.

opds.py
```diff
@@ -1071,13 +1071,13 @@ def single_entry(
         if raw or entry is None:
             return entry
         if isinstance(entry, OPDSMessage):
-            entry = unicode(entry)
+            # entry = entry.decode("utf-8")
```
leonardr: What's going on here?

EdwinGuzman: This was left over when I was debugging which I'll remove. In some tests, `entry` was a byte string but I resolved that (I just don't remember exactly how).

coverage.py
EdwinGuzman: I'll check this again.

marc.py
EdwinGuzman: Right, now the data in `marc_record` will always be a byte string so the encoding and decoding is removed.

coverage.py
```diff
@@ -772,7 +772,7 @@ def _data_source_for_registration(cls, _db, data_source, autocreate=False):
             return DataSource.lookup(_db, cls.DATA_SOURCE_NAME)
         if isinstance(data_source, DataSource):
             return data_source
-        if isinstance(data_source, basestring):
+        if isinstance(data_source, str):
```
vbessonov: Are we sure that `data_source` is always a Unicode string here? Can it be a byte string, a`bytes` object containing ASCII characters?

Python-Future has a [couple of recommendations](https://python-future.org/compatible_idioms.html#basestring) regarding how to translate such a Python 3 compatible code.

log.py
EdwinGuzman: Maybe the name of the function is confusing because of the conversion, but there are occasions where the message or arguments can be a byte string so this is still useful. Maybe changing it to `use_string(...)` makes more sense?

config.py
EdwinGuzman: I removed this but moved the constant variables to `ConfigurationConstants` so I could call `ConfigurationConstants.SYS_ADMIN_OR_MANAGER`.

model/identifier.py
EdwinGuzman: I removed this and updated related tests in the last commit.

model/coverage.py
EdwinGuzman: Right, it's mostly the case when reading and writing to files that a byte string will come up.

coverage.py
EdwinGuzman: I'm not sure why it failed before but I added it back now.

model/identifier.py
EdwinGuzman: I added the `__hash__` function back in but it's now using the `type` and `identifier` instead of just the `id`. This is from what Jonathan did in his python 3 branch. This fixes the two tests I mentioned above but then others that I didn't realize failed when this function was removed.

classifier/__init__.py
leonardr: What if you defined RESEARCH_AUDIENCES as a set containing a single item and then made AUDIENCES_NO_RESEARCH the difference between the two sets? clearly AUDIENCES is in scope here.

config.py
leonardr: Does it help to put these constants into `ConfigurationConstants`? They should probably be there anyway.

model/identifier.py
leonardr: what if you implemented hash as `hash((self.type, self.identifier))`?

config.py
```diff
@@ -60,7 +60,13 @@ class ConfigurationConstants(object):
...
     ENABLED_FACETS_KEY_PREFIX = "facets_enabled_"
     DEFAULT_FACET_KEY_PREFIX = "facets_default_"
-
+    
+    # The "level" property determines which admins will be able to modify the setting.  Level 1 settings can be modified by anyone.
+    # Level 2 settings can be modified only by library managers and system admins (i.e. not by librarians).  Level 3 settings can be changed only by system admins.
+    # If no level is specified, the setting will be treated as Level 1 by default.
+    ALL_ACCESS = 1
+    SYS_ADMIN_OR_MANAGER = 2
+    SYS_ADMIN_ONLY = 3
```
leonardr, line 67: I'm still interested in seeing these moves to ConfigurationConstants but this is fine if that doesn't work.

leonardr: Wait, never mind, that's exactly what you did.

model/identifier.py
leonardr: And, again, my suggestion is exactly what you (and Jonathan) did.

python_expression_dsl/evaluator.py
vbessonov: @EdwinGuzman, I checked it and it seems it works correctly, thanks!

vbessonov (approved):

vbessonov:
I haven't finished walking through the whole PR yet, I wanted to add some comments but accidentally submitted an approval.

coverage.py
leonardr: The only way a bytestring data source might come in here is if it's part of an HTTP request, e.g. when someone is setting up a new OPDS import. I don't know whether we automatically decode form submissions as they come in.

I'm fine with replacing `isinstance(x, basestring)` with `isinstance(x, (bytes, str))` anywhere to avoid this situation.

EdwinGuzman: Yeah, I'll make that update replacing `isinstance(x, basestring)` to `isinstance(x, (bytes, str))`.

EdwinGuzman:
This PR now includes the `pytest` update done by @jonathangreen . I switched from python 2.7 to python 3.6 in the `tox.ini` file and updated a few files with minor updates (the biggest/tedious part of this was fixing merge conflicts).

classifier/bisac.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, result, *ruleset):
                     % special_variables[rule]
                 )
 
-            if isinstance(rule, basestring):
+            if isinstance(rule, (bytes, str)):
```
jonathangreen, line 93: Should we be allowing bytes here? Seems like we're dealing with strings and should only allow `str`? I know this was discussed before, but I think in Python 3 its dangerous to treat bytes and strings the same. Byes don't have all the string methods. Later in this function we call `Lowercased(rule)` will this work with bytes?

There may be some errors from these at first, but I'm of the opinion we should be making sure everything is a string, and treating any bytes that are floating around as an error, unless they specifically need to be bytes.

log.py
jonathangreen: I think renaming the function to something more clear makes sense. The function and comment about native string implementation in this version of python is a bit confusing in a code base just targeting python 3.

My generally though I wonder if we should be treating logging bytes as an error here, instead of trying to correct for it? It seems unsafe to generally assume that bytes are utf-8 encoded in Python 3.

marc.py
jonathangreen: Is there a reason to be storing the marc records as bytes rather then a string? Do we expect other encodings to be stored?

model/work.py
```diff
@@ -220,7 +220,7 @@ class Work(Base):
     # A precalculated MARC record containing metadata about this
     # work that would be relevant to display in a library's public
     # catalog.
-    marc_record = Column(String, default=None)
+    marc_record = Column(Binary, default=None)
```
jonathangreen: We need a database migration for this right? 

Since `String` and `Binary` are treated differently at the DB schema level this will create a column in a newly initialized database that is different from an existing one.

classifier/bisac.py
leonardr: I agree about only accepting strings here. All the matching rules are present in the Python code so we should be able to guarantee they're strings.

model/work.py
jonathangreen: You can ignore this comment. I missed that it was a todo in the PR description.

marc.py
leonardr: I'm suspicious of the idea that a MARC record can ever be a Unicode string; it's a binary format. So I'm not sure what that decode() was doing in the first place.

If we're generating some subset of MARC that can always be a Unicode string, then I think it's fine to store it as Unicode and avoid some complexity on the database side.

classifier/bisac.py
leonardr: More generally, I'm OK with starting with code like this and then taking another pass to try and eliminate as many of these as possible.

marc.py
jonathangreen: MARC is a binary format, but one of the [specified character encodings](https://www.loc.gov/marc/specifications/speccharucs.html) for that binary format is Unicode (specifically UTF-8 encoded unicode). I believe the `force_utf8` flag we are passing when creating the record sets the [marc record to the unicode encoding](https://gitlab.com/pymarc/pymarc/-/blob/master/pymarc/record.py#L103-104). It is output as a byte string, since MARC is a binary format, but as the code is written I believe we [always a unicode record](https://gitlab.com/pymarc/pymarc/-/blob/master/pymarc/record.py#L387-390). So we could either store it as bytes as it is here or we could store it as a string.

I'm not sure which one is best. 
Pros: 
 - Storing it as bytes would allow us to support other MARC encodings in the future.
 - Storing as bytes it more true to the format since it is a binary format.

Cons:
 - Requires a (potentially large) DB migration.
 - Makes it hard to query against the marc record, if we ever want to make DB queries against them.
 
 Another possibility would be to store it in the database in some intermediate form. It looks like pymarc [offers a json serialization](https://gitlab.com/pymarc/pymarc/-/blob/master/pymarc/record.py#L448) that can be [round-tripped](https://gitlab.com/pymarc/pymarc/-/blob/master/pymarc/marcjson.py#L63). We could potentially store MARC as JSON in the DB.

leonardr: I see! That simplifies things a lot. Let's start storing our MARC records in the database as Unicode. I think that will solve the problem for the medium term. I don't want to do that DB migration as part of the Python 3 conversion, and I'd rather not do it at all.

EdwinGuzman: I'll revert back to unicode and update any relevant code. Should we make a ticket for future MARC-record related db/code updates?

leonardr: I think if we decide to stick with Unicode, we don't have to do any more work until we (for whatever reason) need to store in the database a MARC record that can't be represented as Unicode. We don't have that use case yet, so no need for a ticket.

PR #1245, opened 2021-03-08
Replace nosetests with pytest

jonathangreen:
## Overview

In anticipation of the work to move from Python 2 to Python 3 I want to make sure that the CI system in place can smoothly handle this transition. During the initial testing with Python 3 we had some issues with the nose log capture handler.

Looking into `nosetests`, I noticed it had its last update in 2016 and has recommended that it not be used for new projects since 2015.
> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

## Work

Pytest seems widely used and supported, and has some support for running nose style tests. This PR includes all the work to replace nosetests with pytest and to run the new pytest tests with github actions rather then Travis CI.

There are a lot of changes here so I've tried to break them out into commits that might make reviewing these changes easier. Each one has a fairly descriptive commit message. 

There are 2 broad categories of work here: 
- Replace nosetests with pytest. This involved a lot of code changes, but very few changes to the actual test logic. In order to support pytest:
  -  the main package level fixture had to be moved into a nose style fixture. 
  - The `setup` and `teardown` class level fixtures had to be renamed to `setup_method` and `teardown_method`.
  - The nose assertion functions had to be replaced with plain python assertions. This work was largely automated with [nose2pytest](https://github.com/pytest-dev/nose2pytest). Although some hand cleanup of cases it couldn't automatically transform had to be done. 
- Replace travisci with github actions & [tox](https://tox.readthedocs.io/en/latest/index.html). You can see github actions [running the tests](https://github.com/lyrasis/simplye-server-core/actions/runs/633783224) though CI workflow in the [lyrasis repository](https://github.com/lyrasis/simplye-server-core).
  - Replace the `travis.yml` file with a `.github/workflows/test.yml` file
  - Add a tox configuration. This is very basic right now, but will become much more useful when testing with Python 2 & 3.
  - Change around the requirements files, so there is a requirements-dev with development dependencies.

## Testing

If you have all the services used by the tests setup you can simply run the `tox` command and it will run the tests. 

To test with all the services in docker containers:
```bash
# Start the containers for testing
docker run -d -p 9005:5432/tcp --name db -e POSTGRES_USER=simplified_test -e POSTGRES_PASSWORD=test -e POSTGRES_DB=simplified_circulation_test postgres:9.6
docker run -d -p 9006:9200/tcp --name es -e discovery.type=single-node elasticsearch:6.8.6
docker run -d -p 9007:9000/tcp --name minio -e MINIO_ACCESS_KEY=simplified -e MINIO_SECRET_KEY=12345678901234567890 bitnami/minio:latest

# Add elasticsearch plugin
docker exec es elasticsearch-plugin install -s analysis-icu
docker restart es

# Set environment variables
export SIMPLIFIED_TEST_DATABASE="postgres://simplified_test:test@localhost:9005/simplified_circulation_test"
export SIMPLIFIED_TEST_ELASTICSEARCH="http://localhost:9006"
export SIMPLIFIED_TEST_MINIO_ENDPOINT_URL="http://localhost:9007"
export SIMPLIFIED_TEST_MINIO_USER="simplified"
export SIMPLIFIED_TEST_MINIO_PASSWORD="12345678901234567890"

# Run tox
tox
```

## Additional Work

There is some open work to do before this PR could be merged: 
- Update circulation to follow the same pattern here using pytest for testing. 
- Get server_core running against both Python 2 and Python 3 building on the work done here. 

I wanted to get this initial work wrapped up in a PR to allow some time to review / discuss the changes before those follow up PRs. @leonardr @tdilauro @vbessonov review or feedback about the approach here would be great.


leonardr:
FYI @EdwinGuzman  has a Python 3 port branch in progress here: https://github.com/NYPL-Simplified/server_core/pull/1244. He hasn't moved from nosetests to pytest yet but I know that's something he and @nballenger were planning to work on.

Since this isnt' directly related to py3 I think the best strategy is to merge this into develop and then merge develop into Edwin's feature branch.

.github/workflows/test.yml
```diff
@@ -0,0 +1,40 @@
+name: Test Server Core
+on: [push, pull_request]
```
leonardr, line 2: Edwin and I have been discussing this and we don't think it's necessary to run these tests on pull_request given that we also run them on push. Getting rid of the pull_request run would save resources and help us avoid https://jira.nypl.org/browse/SIMPLY-3592.

tests/classifiers/test_ddc.py
```diff
@@ -23,42 +22,42 @@ def test_audience(self):
...
+        assert False == fic("616.9940092")
+        assert False == fic(615)
+        assert False == fic(800)
+        assert False == fic(814)
 
     def test_classification(self):
         def c(identifier):
             i, name = DDC.scrub_identifier(identifier)
-            eq_(name, None)
+            assert name == None
```
nballenger, line 60: It would be nice to have comparisons to `None`|`True`|`False` be flake8 compliant, which would mean using `assert name is None` instead of `assert name == None`.

.github/workflows/test.yml
jonathangreen: I guess it depends on the contribution workflow. 

I don't have access to directly push branches into these repos, so my contributions (this one for instance) would only trigger a CI run if it were triggered on a PR.

jonathangreen: Looking for a solution from github there is a thread about this problem [here](https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662). 

The suggested solution is: 
```yaml
on:
  push:
    branches:
    - develop
  pull_request:
    branches:
    - develop
``` 

Would it work to do something like this? CI always runs on PR, or on any commit to develop branch?

testing.py
```diff
@@ -82,53 +80,6 @@
 import mock
 import inspect
 
-from nose.plugins.attrib import attr
-
-def package_setup():
```
leonardr, line 87: Note to myself to be on the lookout for where this setup code went to.

tests/test_external_search.py
```diff
@@ -3782,7 +3782,7 @@ def validate_sort_order(filter, main_field):
         # added to the collection. This requires an aggregate function
         # and potentially a nested filter.
         f.order = added_to_collection
-        first_field = validate_sort_order(f, sort_field)
+        first_field = validate_sort_order(f, added_to_collection)
```
jonathangreen, line 3785: This was using the [`sort_field`](https://github.com/NYPL-Simplified/server_core/pull/1245/commits/574f2334dcd3906c2972b16d983388ae9e28c564#diff-354b38fb624ea8ce96bb2b285df0b7f1a70ebce9e2932a8c4e8485856a8de8d9R3760) loop variable. I think this was just a copy paste error.

tests/test_opds2_import.py
```diff
@@ -198,7 +198,7 @@ def test(self):
             moby_dick_delivery_mechanism.delivery_mechanism.drm_scheme)
         assert (
             MediaTypes.EPUB_MEDIA_TYPE ==
-            moby_dick_delivery_mechanism.delivery_mechanism.media_type)
+            moby_dick_delivery_mechanism.delivery_mechanism.content_type)
```
jonathangreen, line 201: I think this is also just a mistake in the tests. This is referencing the loop variable [`media_type` set in licensing.py](https://github.com/NYPL-Simplified/server_core/blob/develop/model/licensing.py#L1495).

tests/test_opds_import.py
```diff
@@ -2013,36 +2072,51 @@ def test_resources_are_mirrored_on_import(self):
...
-            'http://root/working-cover-image'
+        # Both items were requested
+        assert epub10441['url'] in http.requests
+        assert epub10557['url'] in http.requests
+
+        # The import process requested each remote resource in the feed. The thumbnail
+        # image was not requested, since we never trust foreign thumbnails. The order they
+        # are requested in is not deterministic, but after requesting the epub the images
+        # should be requested.
+        index = http.requests.index(epub10441['url'])
```
jonathangreen, line 2086: There are a lot of changes to this test because during an OPDS import we iterate [over a dictionary of metadata values](https://github.com/NYPL-Simplified/server_core/blob/develop/opds_import.py#L779), that means that the order of requests we see in these tests can change depending on the order of that iteration. So we have to take into account that we don't know which book from the test feed will be imported first.

tests/classifiers/test_bisac.py
```diff
@@ -25,100 +22,100 @@ class TestMatchingRule(object):
...
     def test_fiction_match(self):
         rule = MatchingRule(True, fiction, "art")
-        eq_(None, rule.match("juvenile nonfiction", "art"))
-        eq_(None, rule.match("art"))
-        eq_(True, rule.match("juvenile fiction", "art"))
-        eq_(True, rule.match("fiction", "art"))
+        assert None == rule.match("juvenile nonfiction", "art")
+        assert None == rule.match("art")
+        assert True == rule.match("juvenile fiction", "art")
+        assert True == rule.match("fiction", "art")
```
leonardr, line 83: Looking at the Pytest website it does seem clear that the syntactic sugar has gone away, but how bad is it? When the tests are running and one of these assertions fail, do we just get a bare AssertionError or does it have info about what went wrong?

tests/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+                # This is the first time running these tests
+                # on this server, and the tables don't exist yet.
+                pass
+            else:
+                raise
+
+    yield
+
+    if 'TESTING' in os.environ:
+        del os.environ['TESTING']
```
leonardr: Ok, here it is. I'll come back and look at this after I finish the rest of the review.

tests/classifiers/test_bisac.py
jonathangreen: I find the output from pytest very nice. Here is an example of an [old failed test run](https://github.com/lyrasis/simplye-server-core/runs/1990635874?check_suite_focus=true#step:8:1507). 

There are some technical details in the [pytest documentation](https://docs.pytest.org/en/stable/assert.html#assertion-introspection-details) about how they achieve the introspection into what went wrong.

tests/models/test_circulationevent.py
```diff
@@ -190,7 +186,7 @@ def test_uniqueness_constraints_no_library(self):
 
         # Reuse the timestamp and you get an IntegrityError which ruins the
         # entire transaction.
-        assert_raises(
+        pytest.raises(
```
nballenger, line 189: Seems like if we're switching to pytest it makes sense to change these calls to the context manager form:

```Python
with pytest.raises(SomeException) as exc:
    ...
    call_that_raises_SomeException()
assert exc.value == "some message"
```

The old form isn't deprecated, but the docs do pretty strongly suggest going with the context manager. I'm also not sure you can 100% count on just subbing in `pytest.raises` for `assert_raises`. Maybe that's safe, but the declared signatures of the functions are different, so I'm not sure they always treat the contents of `*args` and `**kwargs` exactly the same.

nballenger (commented):
I'm not familiar enough with the codebase to make subject area comments, so I'm mostly looking at pytest usage stuff.

tests/classifiers/test_ddc.py
jonathangreen: I totally agree it would be nice to use the `is` operator here. I think there are a lot of improvements we could make to these tests to better utilize pytest. It would be great to convert a lot of the fixtures to use the pytest fixture system as well.

For this pass though the code though, since there there are a LOT of assertions, I used the [nose2pytest](https://github.com/pytest-dev/nose2pytest) tooling which made these changes, rather then updating the code by hand. 

The old [`eq_` function](https://github.com/nose-devs/nose/blob/7c26ad1e6b7d308cafa328ad34736d34028c122a/nose/tools/trivial.py#L28) we were calling was doing this same direct comparison.
```python
def eq_(a, b, msg=None):
    """Shorthand for 'assert a == b, "%r != %r" % (a, b)
    """
    if not a == b:
        raise AssertionError(msg or "%r != %r" % (a, b))
``` 

Maybe we could leave further pytestification of the code to future work?

tests/models/test_circulationevent.py
jonathangreen: This construct also came from [nose2pytest](https://github.com/pytest-dev/nose2pytest). In their [documentation](https://github.com/pytest-dev/nose2pytest#status) they mention:
> Some Nose functions can be handled via a global search-replace, so a fixer was not a necessity:
    - assert_raises: replace with pytest.raises

So these were all replaced using a find / replace strategy. I did some basic testing, and it is working: the tests fail if the exception isn't raised or if a different exception is raised. 

Similar to my other comment, I'd like to keep this PR as a mostly automated update to pytest, and have a future effort to tackle some of the additional work that could be done fully utilizing pytest.

tests/classifiers/test_ddc.py
nballenger: Oh definitely--I don't think it's a blocker at all.

tests/models/test_circulationevent.py
nballenger: Works for me!

tests/test_external_search.py
leonardr: Agreed.

tests/test_opds.py
```diff
@@ -449,24 +443,40 @@ def setup(self):
                       self.romance]
         )
 
+    def _assert_xml_equal(self, a, b):
+        # Compare xml is the same, we use etree to canonicalize the xml
+        # then compare the canonical versions
+        assert etree.tostring(a, method="c14n2") == \
+            etree.tostring(etree.fromstring(b), method="c14n2")
+
```
leonardr, line 451: Nice.

tests/test_opds2_import.py
leonardr: Yes, Edwin found a problem with the same cause in his py3 port branch. Any reference to DeliveryMechanism.media_type should be content_type instead.

tests/test_opds_import.py
leonardr: This is fine for now but how about sorting that dictionary to guarantee a reliable order?

.github/workflows/test.yml
leonardr: I see what you're saying about the permissions, so we need to keep the PR tests. I'd like the tests to run _before_ we push to develop though. Running the test afterwards seems like closing the barn door after the horse has left.

Let's leave it as is for now; we might be able to reduce the number of times we hit Docker, or we can just run the test authenticated as a user with a higher rate limit.

jonathangreen: Agreed about seeing code tested before it hits the develop branch. 

If the workflow to get code into the develop branch always involves merging a pull request, then I think this solution holds. The code will be tested before its merged by the pull request invocation of the tests, and then tests will be run on the commits to develop after the PR is merged.

The place where tests might not get run is if code is being merged into develop, but not going through the PR process. I don't think this is ever the case, but perhaps some internal NYPL development goes in this way?

leonardr: No, we never push to develop without a PR.

I guess the sticking point for me is, when I push a _development_ branch I want the tests to run while I'm writing the PR, rather than waiting til the PR exists. That way by the time I'm done drafting the PR I can see the test failures. I don't see a way of doing that without duplicative work, though.

tests/classifiers/test_bisac.py
leonardr: Yeah, that looks great.

tests/conftest.py
leonardr: After looking up fixture scopes in https://docs.pytest.org/en/stable/fixture.html this makes sense.

leonardr:
Having gone through this enormous diff I agree that this is the way forward. I'd like to merge this after these two changes:

* A similar branch for circulation. (metadata and library_registry can come later)
* Changes to the status checks used by this repo to protect the branches. I'm not sure how this will work -- presumably merging this branch will create new status checks for the Github Actions. Should we manually verify that the tests pass, force-merge the branch, and then change the status checks to use the Github Actions going forward?

tests/test_opds_import.py
tdilauro: Starting with 3.7, maintaining dictionary order is part of the language. And as an implementation detail CPython has done this since 3.5. I think we should be okay, so long as we are not directly using the hash. We could also run the tests with a fixed value for PYTHONHASHSEED, if needed. Or am I not understanding the issue here correctly?

leonardr: Agreed that this won't be a problem in Python 3, but until we move over unpredictable iterations are a problem. Sorting the dictionary could be an alternative to making a lot of other changes to the test.

But Jonathan already made those changes so it's not a big deal given the short lifespan of this code.

jonathangreen:
Re: changes to the status checks. 
I *think* actions would run on this branch if it were coming from inside of the repo. If you want to push this branch into the NYPL repo, or give me access to, we could test that theory. Then we can see tests run though CI and pass.

leonardr:
You're right. I copied your code into a branch on the server_core repo and the checks do run:
https://github.com/NYPL-Simplified/server_core/pull/1246/checks?check_run_id=2086702626#step:6:58

There are a lot of test failures, because the SIMPLIFIED_TEST_DATABASE environment variable isn't set at the point when the checks are run. If you can suggest a fix for that and push it to this branch, I'll mirror it and try again.

I'll use https://github.com/NYPL-Simplified/server_core/pull/1246 to test things out, but I'll do a final merge of this branch to preserve credit and revision history.

jonathangreen:
@leonardr I think there are a few files missing from what you pushed to #1246 and that is why its failing. 

I believe you should be able to push the same commits from this PR into a branch on this repository like this (this assumes your origin remote is NYPL-simplified):
```
# Remove local pytest branch
git branch -D pytest
# Pull this PR into a branch called pytest
git fetch origin pull/1245/head:pytest
# Push it up to the NYPL repo, overwriting the existing pytest branch
git push --force origin pytest
```

leonardr:
Oh, that's great -- I didn't know about the syntax for fetching from a PR. Now all the tests run with a single failure:
https://github.com/NYPL-Simplified/server_core/runs/2089258674?check_suite_focus=true

tests/conftest.py
```diff
@@ -1,43 +1,17 @@
...
-    # This will make sure we always connect to the test database.
-    os.environ['TESTING'] = 'true'
-
-    # Ensure that the log configuration starts in a known state.
-    LogConfiguration.initialize(None, testing=True)
-
-    # Drop any existing schema. It will be recreated when
-    # SessionManager.initialize() runs.
-    #
-    # Base.metadata.drop_all(connection) doesn't work here, so we
```
jonathangreen: Any idea why this didn't work beofre? It does work for me now in my testing, and it seems like an easier strategy, so I changed this to use `Base.metadata.drop_all()`. I can roll this back if that is undesirable.

testing.py
```diff
@@ -1521,3 +1521,32 @@ def raise_for_status(self):
         implemented by real requests Response objects.
         """
         pass
+
+
+@pytest.fixture(autouse=True, scope="session")
```
jonathangreen, line 1526: Move this back to testing.py, and use it as a pytest plugin, so it can be pulled into circulation.

tests/conftest.py
```diff
@@ -1,43 +1,17 @@
...
+# This is kind of janky, but we import the session fixture
+# into these tests here. Plugins need absolute import paths
+# and we don't have a package structure that gives us a reliable
+# import path, so we construct one.
+# todo: reorg core file structure so we have a reliable package name
+from os.path import abspath, dirname, basename
+
+# Pull in the session_fixture defined in core/testing.py
+# which does the database setup and initialization
+pytest_plugins = ["{}.testing".format(basename(dirname(dirname(abspath(__file__)))))]
```
jonathangreen, line 11: I don't really like this, but it does do the job. If anyone has any suggestions for something cleaner here. I can give them a try.

Pytest can only pull in plugins from an absolute import. Because the folder name can change for core, depending what its checked out as, or if its being pulled into circulation etc. 

This lets you run the tests on core from within circulation, or when checked out individually, by figuring out what the package name should be based on filename.

tests/conftest.py
leonardr: I think it didn't drop the materialized views because they weren't counted as 'tables'. Now we no longer use materialized views so it should be fine.

tests/conftest.py
leonardr: I don't have a better idea--the issue you describe has definitely caused us problems in the past.

jonathangreen:
It looks like actions are now running on this branch. 🎉 And the tests are passing so that is a win.

I brought this up to date with the latest commits to develop and pushed in a few other changes that were necessary in getting pytest running the circulation tests.

leonardr:
After talking it over with Edwin (to coordinate with his Python 3 work) I've disabled the Travis integration for server_core and this branch is good to merge. Long live pytest!

PR #1249, opened 2021-03-18
Update Python 3 CI Configuration

jonathangreen:
I wanted to make a suggestion for how we setup CI for Python 3. I pulled this file out of the Python 3 branch I had going, and applied it to @EdwinGuzman Python 3 branch. 

This PR:
- Sets up CI testing to run against multiple Python versions and simplifies the CI config
- Uses tox-docker to simplify the process of testing against containers
- Adds some notes to the readme about different options you can pass to tox

@EdwinGuzman I made this as PR onto your Python 3 branch, since its really just a small add on to that branch. If you want me to target another branch, or wait until the Python 3 PR is merged, just let me know and I'll change the PR.

This PR is showing that branch failing against Python 3.9. I can remove 3.9 from the CI config if we don't want to test against 3.9 for now, but it might be good to support Python 3.9 so I've left it in there for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,11 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
+    # We want to run on external PRs, but not on our own internal PRs as they'll be run
+    # by the push to the branch. This prevents duplicated runs on internal PRs.
+    # Some discussion of this here: 
+    # https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,7 @@ jobs:
     name: Run Core Tests
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,9 @@ jobs:
   test-core:
     name: Run Core Tests
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -23,6 +23,89 @@ Should you need to work on the core alone, use a traditional git workflow:
...
+flag. 
+
+Test Python 3.8
+```
+tox -e py38
+```
+
+### Docker
+
+If you install `tox-docker` tox will take care of setting up all the service containers necessary to run the unit tests
```
EdwinGuzman, line 62: This means that `tox-docker` is optional, right? If it's not installed, will `tox` run without issues? I haven't tested this yet but there are docker configs in the tox.ini file and I'm not sure if it'll just ignore those configs.

EdwinGuzman:
This looks great and pointing to the `py3-base` branch is fine. I'll test it out soon. I don't have any concerns but was just confused the first time around because I have instances of postgres and elastic search running locally and they conflicted with the docker-based examples and ports (since they expected specific ports which are not the default).

README.md
jonathangreen: Yes. `tox-docker` is optional. Tox just ignores that config if you don't have docker installed.

The [`envlist` line in tox.ini](https://github.com/NYPL-Simplified/server_core/pull/1249/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R2) does mean that tox will try to run with the docker configuration by default (if you don't pass any environment flags to it). We could change this to not default to running with docker if that is preferable.

I like `tox-docker` as a solution here, since it makes it fairly easy to run the checks the same way as they are run in CI if something is failing in GH actions, but not failing for you locally. It also makes it really easy to run the tests if you are new to the codebase.

.github/workflows/test.yml
```diff
@@ -5,6 +5,15 @@ jobs:
...
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+    # We want to run on external PRs, but not on our own internal PRs as they'll be run
+    # by the push to the branch. This prevents duplicated runs on internal PRs.
+    # Some discussion of this here: 
+    # https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```
jonathangreen, line 16: I think this is a reasonable solution to the problem of actions running twice that [@leonardr mentioned in his comment here](https://github.com/NYPL-Simplified/server_core/pull/1245/files/be15b56c90463e17d6881b5f0272d4eda55fef73#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88).

README.md
EdwinGuzman: Yeah, I agree about using `tox-docker`. Maybe you can say that installing `tox-docker` is preferable when testing locally so that it parallels GH actions.

EdwinGuzman:
The setup works well for me for python 3.6 and 3.7 but not 3.8 or 3.9. I don't have python 3.8 or 3.9 installed on my machine and that's probably why. I get

```
ERROR:  py38-docker: InterpreterNotFound: python3.8
ERROR:  py39-docker: InterpreterNotFound: python3.9
```

though I wasn't expecting this.

jonathangreen:
That makes sense, I see a similar error when testing with versions that aren't on my system. I can add a bit to the readme about that as well. Tox will look for python versions on your system to test with, but it won't actually go install a new version. 

If your on OSX, I'm using pyenv to install all the different python versions to test against. Then it runs though them all. If you want to test with just the python versions you have installed you can do something like: 
`tox -e "py{36,37}-docker"`

So tox just tests those versions locally.

EdwinGuzman:
Great, that works. Do you mind adding that info into the readme?

It seems there's a minor issue with python 3.9. Let's leave the config there and see if I can get that fixed.

jonathangreen:
@EdwinGuzman updated the README with that extra info.

EdwinGuzman (approved):

EdwinGuzman:
Thanks!

tdilauro:
> It seems there's a minor issue with python 3.9. Let's leave the config there and see if I can get that fixed.

@EdwinGuzman One issue with 3.9 may be that base64.decodestring/encodestring were [removed](https://docs.python.org/3.9/whatsnew/3.9.html#removed).

> base64.encodestring() and base64.decodestring(), aliases deprecated since Python 3.1, have been removed: use base64.encodebytes() and base64.decodebytes() instead. (Contributed by Victor Stinner in bpo-39351.)

EdwinGuzman:
Yes, I'm trying to update those functions to `decodebytes` and `encodebytes` but I'm running into some issues. I encountered this problem in `circulation` because I don't think `decodestring` or `encodestring` are used in `server_core`.

EdwinGuzman:
This last commit makes all the test pass in 3.9 - https://github.com/NYPL-Simplified/server_core/commit/e00243a992bd60e6fe79c7d5f79466614799dc85 

Looks like `etree.getchildren()` was deprecated and only gives warnings in 3.6-3.8. In 3.9 it throws an error.

PR #1250, opened 2021-03-18
Add name and short_name settings

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3613; as per [this spike](https://jira.nypl.org/browse/SIMPLY-3565), I'm adding `name` and `short_name` to the list of fields that the library create/edit form in the admin interface dynamically generates (rather than the current inexplicable set-up, whereby the front end hard-codes those two fields).


leonardr (approved):
This looks great; I am approving on the assumption that you have a corresponding front-end PR.

adriana-alter:
Yes--[here](https://github.com/NYPL-Simplified/circulation-web/pull/341#pullrequestreview-615932000)!

PR #1253, opened 2021-03-22
Add ConfigurationMetadata.to_bool

vbessonov:
## Description

This PR adds `ConfigurationMetadata.to_bool` allowing to convert ConfigurationMetadata instance's values to boolean scalars (True/False).

## Motivation and Context

`ConfigurationMetadata.to_bool` resembles `ConfigurationSetting.bool_value`. I couldn't make it a property because it already has `__get__` magic method returning its internal value in a raw format.

## Checklist:

### DIFF
--- a/model/configuration.py
+++ b/model/configuration.py
@@ -3,38 +3,24 @@
 import inspect
 import json
 import logging
-from abc import abstractmethod, ABCMeta
+from abc import ABCMeta, abstractmethod
 from contextlib import contextmanager
-
 from enum import Enum
+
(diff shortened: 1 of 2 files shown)


leonardr (approved):
This looks fine. Even though this can't be a property, how would you feel about calling this `bool_value` so it matches the property?

tdilauro (approved):
This looks good. I agree with @leonardr that making this property-like would be nice to match the way we handle settings elsewhere, e.g.: `settings.setting(...).value`.

PR #1254, opened 2021-03-23
Feature/collection monitor command line args

tdilauro:
## Description

This branch makes some enhancements focused on collection monitors. At the highest level, these changes allow us to:
- run collection monitors with a subset of collections and
- understand to which collection a monitor's log message applies.

Here's a little more detail:
- The monitor's logger provides a prefix that includes the name of the monitor's collection, if a collection is associated with the monitor.
- The `CollectionMonitor.all` method now takes a `collections` parameter to help its (mostly script runner) collaborators control which collections will have monitors generated for them.
  - If none specified, then all collections will be returned, in an order determined by their timestamps.
  - If one or more collections are specified, then a monitor will be run for each, in the order the collections are specified.
- `RunCollectionMonitorScript`s now parses arguments during initialization and accepts zero or more collection names. Those names are transformed to their associated collections and passed on to `CollectionMonitor.all`.

## Motivation and Context

- It is currently difficult or impossible to determine to which collection a monitor's log messages refer.
- Recently, we have needed to run monitors either against specific collections or  against all collections in a controlled manner.

https://jira.nypl.org/browse/SIMPLY-3416

## How Has This Been Tested?

Tested some existing monitors to verify that the new collection name details were present in log messages.
Created new tests and, in addition to running this repo's test suite, I ran that of the current `circulation` repo `develop` branch with `core` updated to this `server_core` commit.

## Checklist:

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -34,6 +34,19 @@ from model import (
 from model.configuration import ConfigurationSetting
 
 
+class CollectionMonitorLogger(logging.LoggerAdapter):
+    """Prefix log messages with a collection, if one is present."""
+
+    def __init__(self, logger, extra):
+        self.logger = logger
+        self.extra = extra
+        collection = self.extra.get('collection', None)
+        self.log_prefix = '[{}] '.format(collection.name) if collection else ''
+
+    def process(self, msg, kwargs):
+        return '{}{}'.format(self.log_prefix, msg), kwargs
+
+
 class Monitor(object):
     """A Monitor is responsible for running some piece of code on a
     regular basis. A Monitor has an associated Timestamp that tracks
@@ -99,7 +112,10 @@ class Monitor(object):
     @property
     def log(self):
         if not hasattr(self, '_log'):
-            self._log = logging.getLogger(self.service_name)
+            self._log = CollectionMonitorLogger(
+                logging.getLogger(self.service_name),
+                {'collection': self.collection},
+            )
         return self._log
 
     @property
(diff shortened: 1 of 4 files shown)


monitor.py
```diff
@@ -122,6 +122,9 @@ def initial_start_time(self):
             return self.default_start_time
         return datetime.datetime.utcnow()
 
+    @property
+    def message_prefix(self):
+        return '[{}] '.format(self.collection.name) if self.collection else ''
```
leonardr: Since a `Monitor` is associated with a `collection`, I think this would go more easily into the name of the logger (set in `Monitor.log`). That would get you the result you want without having to change the actual logging calls. Theoretically it would also make it possible to use the collection name in log filtering rules.

leonardr (commented):
Everything looks right apart from a quibble about where the message prefix should go.

monitor.py
tdilauro: I'll dig deeper into the logger approach. I started to go down that path, but was in a hurry. But I think it is a better way to go, so I'll find a little more time for this. Thanks.

tdilauro:
I pulled this back into draft to prevent it getting merged before I look into an inconsistent test result in `circulation` based on this PR.

tdilauro:
@leonardr I was able to move the prefix into a the logger. So, this is set from my perspective, pending your review.

leonardr (approved):
This is fine, but not what I imagined; I imagined you would put the collection name in the name of the logger, creating a distinct logger for each collection.  This solves the problem of keeping log_prefix out of every logging call, so it's all good.

PR #1256, opened 2021-04-02
UTC Datetime Update

EdwinGuzman:
Helps resolve [SIMPLY-3663](https://jira.nypl.org/browse/SIMPLY-3663).

I originally updated the syntax for all `utcnow` and `utcfromtimestamp` functions to `now` and `fromtimestamp`, respectively with the appropriate params. I decided to go with the `pytz.UTC` value rather than `datetime.timezone.utc` because the standard library recommends using `pytz`. I then noticed that even though timezone-aware dates were being created and inserted into the database, that the value queried from the database lost its timezone (from test failures).

I then updated db timestamp columns to be `Column(DateTime(timezone=True), nullable=False)`  and this worked. But here's where I wasn't sure if this needs a migration. I can't be sure if older data without a timestamp actually causes an issue when it's being read. I tried a migration script that has `ALTER TABLE cachedfeeds ALTER COLUMN "timestamp" SET DATA TYPE timestamptz;`, for example. This returns timezone-aware datetime objects from the database but running the script potentially crashed my app server. With new databases, running `./core/bin/initialize_database` returned `Illegal instruction: 4`. Will need help on this next week.


EdwinGuzman:
Update: I found that the `Column(DateTime(timezone=True), ...)` crashes the circulation and metadata wrangler flask apps but they don't give any good errors or warnings (with and without the migration update). It looks like postgres stores dates in UTC regardless of whether the python datetime object is naive or aware, so maybe the best approach right now is to manually "convert" the date from the db into an aware object through `.replace(tzinfo=pytz.UTC)`. So far this is working and fixes broken tests, but it's a much more tedious approach and we'd have to remember this whenever we make any updates.

leonardr:
What do the errors and warnings look like, even if they're not helpful? I would like to make sure that datetimes always come out of the database in a timezone-aware form.

The code is fine as far as it goes but I can tell it's going to cause problems keeping everything straight. What do you think about introducing some helper functions to at least make it easier?

`now()` -- shortcut for datetime.datetime.now(tz=pytz.UTC)
`utc(datetime)` -- shortcut for datetime.replace(tzinfo=pytz.UTC)

I wish there was some way to just stop naive datetimes from coming into the system at all...

EdwinGuzman:
> What do the errors and warnings look like, even if they're not helpful? I would like to make sure that datetimes always come out of the database in a timezone-aware form.

There are no warnings or errors. After running `python app.py` and going to the localhost page, the server just stops. If I use `flask run`, then an `Illegal instruction: 4` error appears before the server stops. I tried debugging to see what happens and it looks like there are still some naive/aware dates being compared and the server crashes rather than giving an error.

> What do you think about introducing some helper functions to at least make it easier?

I was thinking about this but it's easy to not use the helper function and write the native implementation. It made more sense to me to write it out every time so I get used to the syntax rather than relying on a helper function but I can change this, it's not a big deal.

Without `Column(DateTime(timezone=True))`, I have to make updates like the following (since dates coming from the db are not aware types) to fix tests and it's a lot more tedious:

```
feed_obj, is_new = get_one_or_create(_db, cls, **kwargs)
if feed_obj.timestamp:
    feed_obj.timestamp = feed_obj.timestamp.replace(tzinfo=pytz.UTC)
```

So I'm going to try again with the db migration in place again and keep debugging to see where there are still naive/aware comparisons being made.

tdilauro:
I think the helpers could be handy, but `now()` could be confusing, since someone might assume that it was imported from `datetime`. Care should be taken to avoid that confusion.

EdwinGuzman:
Updates:
* Added helper functions in `/util/datetime_helpers.py`.
* Added a migration script to update the data type column for dates and convert existing data.

I think something is inherently wrong with my setup where the app fails to run after its initial run. I thought it had something to do with this branch, the migration script, and aware and naive dates being compared, but this issue happens on new installs on other branches.

* branches tested: `develop` and `python3` for the circulation manager
* steps done: create a new `simplified_circulation_dev` database, run the app, create a library, add elasticsearch, check the group feed on `:6500/library/groups`
* Stop the app and then run the app again -- the server stops and no error is logged.

This makes me think something else is wrong on my local machine and makes this branch hard to test. I'm going to keep digging my issue but if possible, @leonardr can you try this branch and the migration script on one of your circ managers?

analytics.py
```diff
@@ -1,12 +1,13 @@
...
 import importlib
 import contextlib
-import datetime
 import os
 from collections import defaultdict
+from sqlalchemy.orm.session import Session
+
 from .model import ExternalIntegration
 from .config import CannotLoadConfiguration
-from sqlalchemy.orm.session import Session
```
EdwinGuzman, line 9: I made misc cosmetic changes for package imports throughout the app in this branch.

leonardr:
Running the migration script shows migrations for a couple of fields that I don't think are in the model:

```
ERROR:  column "created" does not exist
LINE 1: update credentials set created = (created at time zone 'utc'...
                                          ^
ERROR:  column "updated" does not exist
LINE 1: update credentials set updated = (updated at time zone 'utc'...
```

That's not pertinent to what you're asking though; I'm still investigating.

leonardr:
I'm able to run the `python3` circulation branch with this core branch without an issue. Let's work this out interactively.

EdwinGuzman:
The migration script issue was bad copying and pasting but that's fixed now.

model/listeners.py
```diff
@@ -74,9 +80,9 @@ def _site_configuration_has_changed(_db, cooldown=1):
             _db = Session.object_session(_db)
 
         # Update the timestamp.
-        now = datetime.datetime.utcnow()
+        now = utc_now()
         earlier = now-datetime.timedelta(seconds=cooldown)
-        sql = "UPDATE timestamps SET finish=:finish WHERE service=:service AND collection_id IS NULL AND finish<=:earlier;"
+        sql = "UPDATE timestamps SET finish=(:finish at time zone 'utc') WHERE service=:service AND collection_id IS NULL AND finish<=(:earlier at time zone 'utc');"
```
EdwinGuzman, line 83: This turned out to be a big problem.

When using this branch, I created a library "library1" successfully but when I went to `:6500/library1/groups`, the page had an internal server error. After lots of digging around, in `reload_settings_if_changed` the time check was off and `self.load_settings()` was never called. This meant that the app didn't know about this new library even though it was successfully created and was in the database.

model/listeners.py
```diff
@@ -61,8 +64,11 @@ def site_configuration_has_changed(_db, cooldown=1):
 
 def _site_configuration_has_changed(_db, cooldown=1):
     """Actually changes the timestamp on the site configuration."""
-    now = datetime.datetime.utcnow()
+    now = utc_now()
     last_update = Configuration._site_configuration_last_update()
+    if last_update:
+        last_update = to_utc(last_update)
```
leonardr: It should be the responsibility of `._site_configuration_last_update()` to make sure the datetime it's returning is timezone-aware.

scripts.py
```diff
@@ -2540,12 +2538,14 @@ def update_timestamps(self, migration_file):
         if self.overall_timestamp is None:
             return
 
+        if self.overall_timestamp.finish is not None:
+            finish_timestamp = to_utc(self.overall_timestamp.finish)
```
leonardr: TimestampInfo should be responsible for guaranteeing that .start and .finish are timezone-aware and UTC, if they exist.

tests/test_s3.py
```diff
@@ -774,45 +776,45 @@ def test_cover_image_url(
             'marc',
             'SHORT',
             'Lane',
-            datetime.datetime(2020, 1, 1, 0, 0, 0),
-            'https://marc.s3.amazonaws.com/SHORT/2020-01-01%2000%3A00%3A00/Lane.mrc'
+            datetime_utc(2020, 1, 1, 0, 0, 0),
+            'https://marc.s3.amazonaws.com/SHORT/2020-01-01%2000%3A00%3A00%2B00%3A00/Lane.mrc'
```
leonardr, line 780: I wasn't expecting this but I guess it's better than the alternative. I need to check whether it will affect backwards compatibility though.

leonardr: I think we're fine; MARC files uploaded prior to this change will have the old URL stored in their Resource objects.

EdwinGuzman: Do those Resource objects need to be converted if they are then read at a later point?

leonardr: I don't think so, since we only mirror these resources once. (If we mirrored it again it would be for a different date and the URL would be different anyway.)

PR #1258, opened 2021-05-11
All CachedFeeds are served as privately cacheable for libraries that don't show everything to all patrons

leonardr:
This branch fixes https://jira.nypl.org/browse/OE-25 by changing the caching rules for all CachedFeeds served for libraries that direct patrons to different lanes based on their patron type. For those libraries, all CachedFeeds are treated as private. 

"Private" here refers to a concept defined in util/flask_util.py:Response as a way of bundling together a few features of HTTP. These feeds are not necessarily "private" in the sense that their contents are secret; "private" here only refers to the rules about when the feeds can be cached. Most of this branch is tweaking what exactly "private" means:
 
Previously, private resources were served with `Cache-Control: private`, meaning they can only be cached by the final recipient, not by intermediaries such as CDNs. This is still true.

As of this branch, private resources are also served with `Vary: Authorization`, meaning they can only be cached by the credentials that requested them. This is the piece that (I hope) actually fixes OE-25.

And as of this branch, private resources are _not_ served with `Cache-Control: s-maxage`. That told intermediaries how long to cache the representation, which conflicted with the `Cache-Control: private` directive.

### DIFF
--- a/util/flask_util.py
+++ b/util/flask_util.py
@@ -92,18 +92,28 @@ class Response(FlaskResponse):
         # Don't modify the underlying dictionary; it came from somewhere else.
         headers = dict(headers)
 
-        # Set Cache-Control based on max-age.
+        # Set headers based on privacy settings and maximum age.
         if self.private:
             private = "private"
+
+            # A private resource should be re-requested, rather than
+            # retrieved from cache, if the authorization credentials
+            # change from those originally used to retrieve it.
+            headers['Vary'] = 'Authorization'
         else:
             private = "public"
         if self.max_age and isinstance(self.max_age, int):
-            # A CDN should hold on to the cached representation only half
-            # as long as the end-user.
             client_cache = self.max_age
-            cdn_cache = self.max_age / 2
-            cache_control = "%s, no-transform, max-age=%d, s-maxage=%d" % (
-                private, client_cache, cdn_cache
+            if self.private:
+                # A private resource should not be cached by
+                # intermediaries at all.
+                s_maxage = ""
+            else:
+                # A public resource can be cached by intermediaries
+                # for half as long as the end-user can cache it.
+                s_maxage = ", s-maxage=%d" % self.max_age / 2
+            cache_control = "%s, no-transform, max-age=%d,%s" % (
+                private, client_cache, s_maxage
             )
 
             # Explicitly set Expires based on max-age; some clients need this.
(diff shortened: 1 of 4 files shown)


leonardr:
@kristojorg, take a look at the HTTP side of this and verify that this makes sense.

EdwinGuzman (approved):
This makes sense to me. What is still a bit confusing is why this setting `private` to true depends on the library having root lanes. In what scenario can a library provide a feed that is not based on different lanes for a patron type?

leonardr:
Open eBooks is the only library where different patron types get assigned different root lanes. If you sign in as an Early Grades patron, you should never be shown a Middle Grades feed -- even if (this is the point of OE-25) there was a Middle Grades patron who used the device before you and cached a Middle Grades feed in the private browser cache.

With, let's say, NYPL, we have different patron types, but when you log in, everyone always gets sent the same top-level feed. So there's no issue with a cache having a copy of a feed that a patron shouldn't be shown.

Theoretically we could say that people with juvenile cards a) get sent to a special "juvenile" set of lanes and b) can't access any books that won't fit in those lanes, and at that point we'd get something more like Open eBooks, but we haven't set that policy at NYPL.

util/flask_util.py
```diff
@@ -92,18 +92,28 @@ def _headers(self, headers={}):
...
+            if self.private:
+                # A private resource should not be cached by
+                # intermediaries at all.
+                s_maxage = ""
+            else:
+                # A public resource can be cached by intermediaries
+                # for half as long as the end-user can cache it.
+                s_maxage = ", s-maxage=%d" % (self.max_age / 2)
+            cache_control = "%s, no-transform, max-age=%d%s" % (
+                private, client_cache, s_maxage
```
kristojorg, line 116: I might be following this logic incorrectly, but just to be certain, when `self.private == True`, the `cache-control` header should have `private`, and when `self.private == False` then `cache-control` doesn't need the `private` directive.

kristojorg (commented):
Looks good, just the one comment

util/flask_util.py
leonardr: Yes, the code starting here sets `private` to either "private" or "public" depending on the value of `self.private`:

https://github.com/NYPL-Simplified/server_core/pull/1258/files/075647968fee78d86d21931fe233b9420593a56f#diff-40292c475f9d61e929ebbc06204673b0d7e5e58acf523ce1b00d18b17ae6f089R95

Then whatever string is in `private` gets put in the Cache-Control header. So `private` could be more accurately called `private_status` or something.

PR #1259, opened 2021-05-25
Simplify UTC timestamp migration, run as single statement transactions

nballenger:
I've changed two things:

1. I set the `migration/20210405-update-datetime-columns.sql` file up to use one `ALTER` per table, and removed the `UPDATE` statements because the conversion is actually accomplished during the table alteration.
2. I added some logic to `scripts.py` in the `DatabaseMigrationScript` class, aimed at running individual statements in a migration file within their own transactions. That only triggers if somewhere in the .sql file the string `SIMPLYE_MIGRATION_TRANSACTION_PER_STATEMENT` appears.

I don't have the circ manager set up locally in a way that makes this easy to test, so anything you spot would be good to know about before I build this and try it on QA.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -2491,7 +2491,7 @@ class DatabaseMigrationScript(Script):
         # Update timestamp for the migration.
         self.update_timestamps(migration_filename)
 
-    def _extract_statements_from_sql_file(filepath):
+    def _extract_statements_from_sql_file(self, filepath):
         """
         From an SQL file, return a python list of the individual statements.
 
--- a/scripts.py
+++ b/scripts.py
@@ -2052,6 +2052,8 @@ class DatabaseMigrationScript(Script):
     # There are some SQL commands that can't be run inside a transaction.
     TRANSACTIONLESS_COMMANDS = ['alter type']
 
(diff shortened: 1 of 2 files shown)


leonardr (approved):
The code makes sense. Tomorrow I'll spend some time setting up a database that needs this migration, and make sure the SQL works. I can't test the performance characteristics though.

nballenger:
Thanks--I think as long as it runs successfully that's fine, we'll just have to see how it fares at scale on the QA db.

scripts.py
```diff
@@ -2483,6 +2491,34 @@ def _run_migration(self, migration_path, timestamp):
         # Update timestamp for the migration.
         self.update_timestamps(migration_filename)
 
+    def _extract_statements_from_sql_file(filepath):
```
leonardr: This is missing its `self`.

leonardr (changes requested):
While running this for real I found a minor syntax error. After fixing it, the migration ran and I ran spot checks to verify that the fields were actually migrated.

leonardr (approved):

PR #1260, opened 2021-05-27
Fix two Python 3 conversion errors in Overdrive code

leonardr:
This branch fixes two Python 3 conversion problems in Overdrive code:

* When generating the Authorization header, keys and secrets like `foo` were being encoded as `b'foo'`, leading to authentication failures.
* In one case, Unicode strings were converted to JSON before processing, but bytestrings (which we actually get from Overdrive) were being left alone, causing a crash.

In addition to the unit tests, I ran the overdrive_format_sweep, overdrive_monitor_recent, and overdrive_new_titles scripts using the QA database. Without the first fix in place, all three scripts would fail immediately. Without the second fix in place, a couple of these scripts (overdrive_monitor_recent and overdrive_format_sweep, I think) would fail pretty quickly.

### DIFF
--- a/overdrive.py
+++ b/overdrive.py
@@ -346,7 +346,7 @@ class OverdriveAPI(object):
     @property
     def token_authorization_header(self):
         s = b"%s:%s" % (self.client_key, self.client_secret)
-        return base64.standard_b64encode(s).strip()
+        return "Basic " + base64.standard_b64encode(s).strip()
 
     def token_post(self, url, payload, headers={}, **kwargs):
         """Make an HTTP POST request for purposes of getting an OAuth token."""
--- a/overdrive.py
+++ b/overdrive.py
@@ -1,4 +1,3 @@
-
 import datetime
 import isbnlib
 import os
@@ -439,7 +438,8 @@ class OverdriveAPI(object):
         )
         return self.make_link_safe(url)
 
-    def _get_book_list_page(self, link, rel_to_follow='next'):
+    def _get_book_list_page(self, link, rel_to_follow='next',
+                            extractor_class=None):
         """Process a page of inventory whose circulation we need to check.
 
         Returns a 2-tuple: (availability_info, next_link).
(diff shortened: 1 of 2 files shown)


overdrive.py
```diff
@@ -344,12 +343,15 @@ def get(self, url, extra_headers, exception_on_401=False):
         else:
             return status_code, headers, content
 
+    @property
+    def token_authorization_header(self):
+        s = b"%s:%s" % (self.client_key, self.client_secret)
+        return "Basic " + base64.standard_b64encode(s).strip()
+
```
leonardr, line 350: `token_post` itself gets mocked by the `MockOverdriveAPI`. I moved this to a separate method so I could test it in isolation.

overdrive.py
```diff
@@ -445,24 +448,20 @@ def _get_book_list_page(self, link, rel_to_follow='next'):
            one book.
         `next_link` is a link to the next page of results.
         """
+        extractor_class = extractor_class or OverdriveRepresentationExtractor
         # We don't cache this because it changes constantly.
         status_code, headers, content = self.get(link, {})
-        if isinstance(content, str):
+        if isinstance(content, (bytes, str)):
```
leonardr, line 454: Here's the other bug fix. It's now the same as code we have several times elsewhere in this file.

This is some of the first code I wrote for the project, back in 2014, and these methods for extracting information from Overdrive API responses aren't tested very well, nor designed to be testable.

tests/test_overdrive.py
```diff
@@ -329,6 +338,50 @@ def test_advantage_differences(self):
...
+            (url, headers, body) = self.api.requests.pop()
+            assert len(self.api.requests) == 0
+            assert url == "http://first-page/"
+
+            # The extractor was used to extract a link to the page
+            # with rel="some-rel".
+            #
+            # Note that the Python data structure (`original_data`) is passed in,
+            # regardless of whether the mock response body is a Python
+            # data structure, a bytestring, or a Unicode string.
```
leonardr, line 374: This bit tests the specific bug fix I introduced.

PR #1262, opened 2021-06-02
Convert bytestring ConfigurationSetting values to Unicode before storing in the database

leonardr:
This branch fixes bugs found during in-person testing of qa-circulation.librarysimplified.org using the SimplyE mobile app.

* Setting `ConfigurationSetting.value` to the bytestring `foo` converted it to its Python representation, `b'foo'`, rather than converting it to the Unicode string `foo`.

I looked through the code for other places where we encrypt or decrypt things, and didn't find any other problems like this.

### DIFF
--- a/model/configuration.py
+++ b/model/configuration.py
@@ -733,7 +733,9 @@ class ConfigurationSetting(Base, HasFullTableCache):
 
     @value.setter
     def value(self, new_value):
-        if new_value is not None:
+        if isinstance(new_value, bytes):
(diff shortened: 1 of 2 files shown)


EdwinGuzman (approved):
I either missed this or relied on `str` updating the value from a byte string to unicode.

leonardr:
yeah, unfortunately str doesn't work that way unless you specify an encoding.

PR #1266, opened 2021-06-22
Split up database migration statements into groups that can be executed in parallel.

leonardr:
This branch reorders the SQL statements in the big timezone migration script and adds advice from Ashok Kumar, NYPL's DBA, about executing the statements in two parallelized batches to reduce the time this migration will take.

We're going to try this out in an upcoming migration, and I'm OK with holding off merging this until we see how well it works, but I wanted to keep a record of Ashok's advice in our code base for the benefit of other folks.

### DIFF
--- a/migration/20210604-reverse-update-datetime-columns.sql
+++ b/migration/20210604-reverse-update-datetime-columns.sql
@@ -2,6 +2,11 @@
 -- It should only be uncommented in the event that you need to deploy this fix
 -- via the build system. Otherwise, you should run the commands by hand.
 --
+-- These statements are grouped into the same batches as
+-- 20210405-update-datetime-columns.sql, and it may be possible to
+-- save time by running the batches in parallel, as described in that
+-- file.
+--
 -- Force the core migration script to run each command in this file as an individual transaction:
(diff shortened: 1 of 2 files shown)


migration/20210604-reverse-update-datetime-columns.sql
```diff
@@ -2,20 +2,22 @@
 -- It should only be uncommented in the event that you need to deploy this fix
 -- via the build system. Otherwise, you should run the commands by hand.
 --
+-- These statements are grouped into the same batches as
+-- 20210405-update-datetime-columns.sql, and it may be possible to
+-- save time by running the batches in parallel, as described in that
+-- file.
+--
```
nballenger, line 9: Possibly also worth noting (or not) that these statements are all likely to run quite a bit faster than the forward migration. I think it's because these are shrinking the column storage to go from timestamp with tz back to timestamp without tz instead of having to allocate new space. I definitely saw faster ALTER completion times in the rollback we did for OE.

nballenger (approved):
Thanks for breaking these out, it'll be helpful when we run them.

PR #1267, opened 2021-09-01
SIMPLY-3873 Update shebang lines for scripts to py3

nballenger:
Matching circulation PR here: https://github.com/NYPL-Simplified/circulation/pull/1633


leonardr (approved):
Most of these migration scripts could just be removed, but this is good for now.

PR #1270, opened 2021-09-20
Handle the case where there are multiple items in heaviest_child.

leonardr:
This PR fixes https://jira.nypl.org/browse/SIMPLY-3858 by changing the way a sequence of (GenreData : (GenreData, int)) tuples is sorted. GenreData objects aren't comparable, so the correct solution is to sort by the ints.

### DIFF
--- a/classifier/__init__.py
+++ b/classifier/__init__.py
@@ -1491,7 +1491,10 @@ class WorkClassifier(object):
         #    print("", parent, genre, weight)
         made_it = False
         while not made_it:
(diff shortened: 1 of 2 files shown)


nballenger:
Colon went missing after a for loop, pushed up a commit that put it back to hopefully get the tests running successfully.

nballenger (approved):
Looks good to me, tests are all green, fix makes sense to the degree that I understand the classification system.

PR #1272, opened 2021-11-15
Adds class method to Configuration for getting static dir

nballenger:
As part of the Docker refactor, I'm making a consolidated place for static files.

### DIFF
--- a/config.py
+++ b/config.py
@@ -471,10 +471,10 @@ class Configuration(ConfigurationConstants):
         """
         Locate the static resources for this installation.
 
-        Default location is /simplye_static. To use a different location, set the value
+        Default location is /simplified_static. To use a different location, set the value
         of the SIMPLIFIED_STATIC_DIR environment variable.
         """
-        default_static_dir = '/simplye_static'
+        default_static_dir = '/simplified_static'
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -466,6 +466,37 @@ def database_url(cls):
...
 
+    @classmethod
+    def static_resources_dir(cls):
+        """
+        Locate the static resources for this installation.
+
+        Default location is /simplye_static. To use a different location, set the value
+        of the SIMPLIFIED_STATIC_DIR environment variable.
+        """
+        default_static_dir = '/simplye_static'
```
leonardr: Very minor thing but it's branding related so I have to be picky -- call this `simplified_static` instead. "SimplyE" is the name of a mobile app, not the generic term that should be used here.

leonardr (approved):
This makes sense -- what sort of static files are we talking about? The crontab?

nballenger:
It's the admin app js and css, font files, etc. Also the files in `resources` and `resources/images`.

config.py
nballenger: Good catch. I'll go through and change it in the various places I've used it.

PR #1275, opened 2022-01-12
OE-317 Check authentication_flow_document type

keithbauer:
This first checks if the authentication flow document is a list or a single entry. From there it checks the document's `type` as it was originally.

This change is needed since `BasicAuthenticationProvider` is sending two authentication documents.

### DIFF
--- a/util/authentication_for_opds.py
+++ b/util/authentication_for_opds.py
@@ -8,8 +8,9 @@ class OPDSAuthenticationFlow(object):
     FLOW_TYPE = None
 
     def authentication_flow_document(self, _db):
-        """Convert this object into a dictionary that can be used in the
-        `authentication` list of an Authentication For OPDS document.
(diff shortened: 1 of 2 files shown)


util/authentication_for_opds.py
```diff
@@ -12,13 +12,21 @@ def authentication_flow_document(self, _db):
         `authentication` list of an Authentication For OPDS document.
```
leonardr: The docstring still says this will return one dictionary, but it might now return a single dictionary or a list of them. I'd either change the docstring to make this explicit or make this method always yield a sequence of documents (at which point it should probably be renamed to `authentication_flow_documents`).

You also need to update tests/test_authentication_for_opds.py to test a mock flow that defines `_authentication_flow_document` to return two different flows documents.

leonardr (approved):

PR #1278, opened 2022-02-09
Simply-4025 Convert tests to Pytest

keithbauer:
## Description

Convert tests to use pytest fixtures.

## Motivation and Context

[SIMPLY-4025](https://jira.nypl.org/browse/SIMPLY-4025)
[SIMPLY-4024](https://jira.nypl.org/browse/SIMPLY-4024)

## Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/testing.py
+++ b/testing.py
@@ -1535,8 +1535,8 @@ def session_fixture():
 
     # Drop any existing schema. It will be recreated when
     # SessionManager.initialize() runs.
-    engine = SessionManager.engine()
-    Base.metadata.drop_all(engine)
+    # engine = SessionManager.engine()
+    # Base.metadata.drop_all(engine)
 
     yield
 
--- a/testing.py
+++ b/testing.py
@@ -1524,7 +1524,7 @@ class MockRequestsResponse(object):
         """
         pass
 
-"""
+
 @pytest.fixture(autouse=True, scope="session")
 def session_fixture():
     # This will make sure we always connect to the test database.
@@ -1535,14 +1535,14 @@ def session_fixture():
 
     # Drop any existing schema. It will be recreated when
     # SessionManager.initialize() runs.
-    engine = SessionManager.engine()
-    Base.metadata.drop_all(engine)
+    #engine = SessionManager.engine()
+    #Base.metadata.drop_all(engine)
 
     yield
 
     if 'TESTING' in os.environ:
         del os.environ['TESTING']
-"""
+
 
 def pytest_configure(config):
     # register our custom marks with pytest
--- a/testing.py
+++ b/testing.py
@@ -1524,7 +1524,7 @@ class MockRequestsResponse(object):
         """
         pass
 
-
+"""
 @pytest.fixture(autouse=True, scope="session")
 def session_fixture():
     # This will make sure we always connect to the test database.
@@ -1535,14 +1535,14 @@ def session_fixture():
 
     # Drop any existing schema. It will be recreated when
     # SessionManager.initialize() runs.
-    # engine = SessionManager.engine()
-    # Base.metadata.drop_all(engine)
+    engine = SessionManager.engine()
+    Base.metadata.drop_all(engine)
 
     yield
 
     if 'TESTING' in os.environ:
         del os.environ['TESTING']
-
+"""
 
 def pytest_configure(config):
     # register our custom marks with pytest
--- a/testing.py
+++ b/testing.py
@@ -1533,11 +1533,6 @@ def session_fixture():
     # Ensure that the log configuration starts in a known state.
     LogConfiguration.initialize(None, testing=True)
 
-    # Drop any existing schema. It will be recreated when
-    # SessionManager.initialize() runs.
-    #engine = SessionManager.engine()
-    #Base.metadata.drop_all(engine)
-
     yield
 
     if 'TESTING' in os.environ:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -106,7 +106,7 @@ def create_collection():
     """
     Returns a constructor function for creating a Collection.
     """
-    def _create_collection(db_session, name=None, protocol=ExternalIntegration.OPDS2_IMPORT,
+    def _create_collection(db_session, name=None, protocol=ExternalIntegration.OPDS_IMPORT,
                            external_account_id=None, url=None, username=None,
                            password=None, data_source_name=None):
         name = name or f"Good Reads #{random.randint(1, 9999)}"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -108,7 +108,7 @@ def create_collection():
     def _create_collection(db_session, name=None, protocol=ExternalIntegration.OPDS2_IMPORT,
                            external_account_id=None, url=None, username=None,
                            password=None, data_source_name=None):
-        name = name or "Good Reads"
+        name = name or f"Good Reads #{random.randint(1, 9999)}"
         collection, _ = get_one_or_create(db_session, Collection, name=name)
         collection.external_account_id = external_account_id
         integration = collection.create_external_integration(protocol)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -82,9 +82,8 @@ def init_test_db():
         except ProgrammingError:
             pass
 
-    with engine.connect() as conn:
-        Base.metadata.create_all(conn)
     with engine.connect() as connection:
+        Base.metadata.create_all(connection)
         db_session = Session(connection)
 
         # Populate Genres
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,7 +33,8 @@ from ..model.constants import MediaTypes
 from ..model.contributor import Contributor
 from ..model.coverage import (
     CoverageRecord,
-    Timestamp
+    Timestamp,
+    WorkCoverageRecord
 )
 from ..model.customlist import CustomList
 from ..model.datasource import DataSource
@@ -707,3 +708,24 @@ def create_work(create_edition):
         return work
 
     return _create_work
+
+
+@pytest.fixture
+def create_work_coverage_record():
+    """
+    Returns a constructor function for creating a WorkCoverageRecord.
+    """
+    def _create_work_coverage_record(db_session, work, operation=None, status=CoverageRecord.SUCCESS):
+        record, _ = get_one_or_create(
+            db_session, WorkCoverageRecord,
+            work=work,
+            operation=operation,
+            create_method_kwargs = dict(
+                timestamp=utc_now(),
+                status=status,
+            )
+        )
+
+        return record
+
+    return _create_work_coverage_record
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -73,7 +73,7 @@ def db_session(db_engine):
 
 
 @pytest.fixture
-def initialize_datasource_and_genres(db_session):
+def init_datasource_and_genres(db_session):
     # This probably needs a better name
     # This was lifted from core/model/__init__.py::SessionManager::initialize_data
     list(DataSource.well_known_sources(db_session))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -150,23 +150,30 @@ def create_edition(create_licensepool):
                         self_hosted=False, unlimited_access=False):
         id = identifier_id or random.randint(1, 9999)
         id = str(id)
+
         source = DataSource.lookup(db_session, data_source_name)
         if not source:
             # source can't be None, so for now this is the sensible default
             source, _ = get_one_or_create(db_session, DataSource, name=data_source_name)
+
         wr = Edition.for_foreign_id(db_session, source, identifier_type, id)[0]
         if not title:
             title = "Test Book"
         wr.title = str(title)
         wr.medium = Edition.BOOK_MEDIUM
+
         if series:
             wr.series = series
+
         if language:
             wr.language = language
+
         if authors is None:
             authors = "Test Author"
+
         if isinstance(authors, str):
             authors = [authors]
+
         if authors:
             primary_author_name = str(authors[0])
             contributor = wr.add_contributor(primary_author_name, Contributor.PRIMARY_AUTHOR_ROLE)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -646,7 +646,7 @@ def create_representation():
                 repr.mirror_url = "http://example.com/" + str(random.randint(1, 9999))
                 repr.mirrored_at = utc_now()
 
-            return repr
+        return repr
 
     return _create_representation
 
@@ -760,7 +760,7 @@ def create_work_coverage_record():
             db_session, WorkCoverageRecord,
             work=work,
             operation=operation,
-            create_method_kwargs = dict(
+            create_method_kwargs=dict(
                 timestamp=utc_now(),
                 status=status,
             )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,26 +13,41 @@ from sqlalchemy import create_engine
 from sqlalchemy.exc import ProgrammingError
 from sqlalchemy.orm.session import Session
 
-from core.classifier import Classifier
-from core.config import Configuration
-from core.lane import Lane
+from ..classifier import Classifier
+from ..config import Configuration
+from ..lane import Lane
 
-from core.model import Base, classifier, get_one_or_create
-from core.model.admin import Admin
-from core.model.classification import Genre, Subject
-from core.model.collection import Collection
-from core.model.configuration import ExternalIntegration
-from core.model.constants import MediaTypes
-from core.model.contributor import Contributor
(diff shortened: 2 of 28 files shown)


tests/models/test_admin.py
```diff
@@ -1,156 +1,180 @@
...
-        self.admin.add_role(AdminRole.LIBRARIAN, self._default_library)
-        assert False == self.admin.can_see_collection(c1);
-        assert True == self.admin.can_see_collection(c2);
-
-        self.admin.remove_role(AdminRole.LIBRARIAN, self._default_library)
-        assert False == self.admin.can_see_collection(c1);
-        assert False == self.admin.can_see_collection(c2);
+
+def test_password_hashed(db_session, create_admin_user):
+    # GIVEN: An Admin user
```
nballenger: It'd probably be good if these were actual docstrings. [PEP 257](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring) says that's "a string literal that occurs as the first statement in a module, function, class, or method definition," so you have to quote it instead of using single-line comment syntax for it to be an actual docstring. I'm not sure there's an _immediately_ obvious benefit to docstring vs. comment, but the argument for using them is that real docstrings end up in `whatever.__doc__` and can be introspected by various pieces of pre-process and run-time tooling like the [`help()`](https://docs.python.org/3/library/functions.html#help) built-in, IDE parsers, etc. I think inline comments that aren't string literals are just discarded during compilation, so you can't read them at runtime unless you bring in something like [`inspect.getsource()`](https://docs.python.org/3/library/inspect.html#inspect.getsource), and even that module expects there to be docstrings for stuff like `inspect.cleandoc()`.

tests/models/test_admin.py
```diff
@@ -1,156 +1,180 @@
...
+
+    admin = create_admin_user(db_session)
+    assert [] == Admin.with_password(db_session).all()
+
+    # WHEN: A password is set
+    # THEN: The Admin.with_password query contains users
+    admin.password = "password"
+    assert [admin] == Admin.with_password(db_session).all()
+
+    admin2 = create_admin_user(db_session, email="admin2@nypl.org")
```
nballenger: Given that this is open source code run by folks other than NYPL, I wonder if it's worth creating a way to set the domain that'll be used in test email addresses. It only really matters if a test potentially triggers an actual email message (since we don't want to get spammed by outside test runners), but I could see it being useful.

nballenger: The alternative (for tests that need a domain or address whose structure is valid even if it doesn't really resolve) would probably be to go with [`example.com`](https://en.wikipedia.org/wiki/Example.com), or a fake domain using the newer [`.example`](https://en.wikipedia.org/wiki/.example) or [`.test`](https://en.wikipedia.org/wiki/.test) top level domains IANA came up with a while back.

tests/models/test_admin.py
```diff
@@ -1,156 +1,180 @@
...
+    assert None == Admin.authenticate(db_session, "other@nypl.org", "password")
+    assert None == Admin.authenticate(db_session, "example@nypl.org", "password")
+
+def test_roles(db_session, create_admin_user, create_library):
+    # GIVEN: An Admin user and a Library
+    # WHEN:  Changing the Admin user roles
+    # THEN:  The Admin user should have appropriate access based on role
+    admin = create_admin_user(db_session)
+    library = create_library(db_session)
+    assert False == admin.is_system_admin()
```
nballenger: If you're testing for a boolean output, it's a good idea to use the `is` comparator rather than `==`. In the vast majority of cases it won't matter, but it's worth noting that there's some hidden type coercion that happens when you use the double equals with a boolean value. Since the double equals isn't a boolean operator, all boolean values passed to it end up converted to the integers 1 and 0 before comparison happens. So you can get this situation (that's oddly reminiscent of JavaScript's double vs. triple equals):

```Python
❯ python -W ignore
>>> True is True
True
>>> True is 1
False
>>> True == 1
True
```

There's two valid alternatives to `assert myobj.mymethod() == False`:

1. `assert myobj.mymethod() is False`
2. `assert not myobj.mymethod()`

I think most Python sources favor the second method any time you're testing the truth of a single value, and would recommend only using the `is` comparator when you have to determine whether two unknown booleans are the same. That said, I actually prefer the first approach in a test context, because it very clearly states the expected outcome, and doesn't require extra thought on the part of the reader.

Also (and this is largely personal style, but a thing I Have Opinions About), I think it's more readable to have assertions structured left to right as:

```Python
assert actual_value == expected_value     # or whatever comparator is appropriate, if not ==
```

rather than `expected` before `actual`. I don't think there's a definitive rule about this anywhere, so feel free to disagree if you also Have Opinions. FWIW, Guido van Rossum [said in 2010](https://mail.python.org/pipermail/python-dev/2010-December/106954.html) that thinking about it in terms of "expected" and "actual" isn't good, and we should just use "first" and "second". But whatever, that dude's wrong.

tests/models/test_admin.py
```diff
@@ -1,156 +1,180 @@
...
+    admin = create_admin_user(db_session, email="admin@nypl.org")
+    admin.password = "password"
+    other_admin = create_admin_user(db_session, email="other@nypl.org")
+    other_admin.password = "banana"
+
+    assert admin == Admin.authenticate(db_session, "admin@nypl.org", "password")
+    assert None == Admin.authenticate(db_session, "other@nypl.org", "password")
+    assert None == Admin.authenticate(db_session, "example@nypl.org", "password")
+
+def test_roles(db_session, create_admin_user, create_library):
```
nballenger: I think there's an argument for parameterizing this one. Something like:

```Python
@pytest.parameterize(
    "role,sysadmin,sitewide_libmgr,sitewide_lib,libmgr,lib",
    pytest.param(
        None, False, False, False, False, False,
        id="no_role_added"
    ),
    pytest.param(
        AdminRole.SYSTEM_ADMIN, True, True, True, True, True,
        id="system_admin_role"
    ),
    ... # and so on with the other roles
)
def test_roles(
    db_session, create_admin_user, create_library, role, sysadmin,
    sitewide_libmgr, sitewide_lib, libmgr, lib
):
    admin = create_admin_user(db_session)
    library = create_library(db_session)

    if role:
        admin.add_role(AdminRole.SYSTEM_ADMIN)

    assert admin.is_system_admin() is sysadmin
    assert admin.is_sitewide_library_manager() is sitewide_libmgr
    assert admin.is_sitewide_librarian() is sitewide_lib
    assert admin.is_library_manager(library) is libmgr
    assert admin.is_librarian(library) is lib
```

tests/conftest.py
```diff
@@ -5,7 +5,494 @@
...
+    engine.dispose()
+
+
+@pytest.fixture
+def db_session(db_engine):
+    with db_engine.connect() as connection:
+        transaction = connection.begin_nested()
+        session = Session(connection)
+        yield session
+        transaction.rollback()
```
nballenger: I know this seems like the right way to do this, but I have to say some part of me doesn't entirely trust it. Maybe peek at the db after a test run some time just to see if there's anything floating around in there that shouldn't have stuck around between tests.

tests/conftest.py
```diff
@@ -5,7 +5,494 @@
...
+        if authors is None:
+            authors = "Test Author"
+        if isinstance(authors, str):
+            authors = [authors]
+        if authors:
+            primary_author_name = str(authors[0])
+            contributor = wr.add_contributor(primary_author_name, Contributor.PRIMARY_AUTHOR_ROLE)
+            if not contributor.display_name and ',' not in primary_author_name:
+                contributor.display_name = primary_author_name
+            wr.author = primary_author_name
```
nballenger: Stylistic thing that you don't need to agree with--I find it more readable to have an empty line to set off conditional blocks that aren't actually related or part of an if/elif/else construct. This kind of thing:

```Python
x = something()
if x:
    print("yup, x alright")

y = something()
if y:
    print("y is in fact true")

if x and y:
    print("both of them!")
elif x and not y:
    print("the x but not the y")
elif y and not x:
    print("the y but not the x")
else:
    print("neither of them")

return some_other_thing
```

I think it's more of a big deal in code that goes on longer than one vertical screen (~50 lines), because it gives it more visual texture as you scroll so it's easier to find your place quickly.

nballenger (commented):
I think you're on the right track and making great progress, and I really appreciate it. I know it's a bit of a slog, but this is going to be super useful.

Here's my argument for test classes: they give you another level of granularity for running groups of tests. Right now we've got enough weird interference between tests that we can't really treat a partial test run as deterministic, so this doesn't come up much. Once we get these things disentangled though, it's going to be great to just run the tests relevant to what you're working on.

I think there's a reasonable counter-argument that anything you could group by class you could also group by file. However, I do think there are cases where it makes sense to put related groups of tests in the same file, but still be able to easily run specific subsets in isolation. Classes are probably the easiest way to do that, since the pytest runner can take an argument in the `file/path/test_file.py::TestClassName` format.

There's another case where you want to run tests that are related along some axis that the class hierarchy doesn't easily represent, like "run every test that hits an endpoint requiring authentication." For that case (which I think is altogether rarer) there are things like custom pytest marks--but those don't _replace_ the utility of the class-based groupings, they just augment it.

nballenger:
One more argument for using test classes. Pytest fixtures all have a specific [scope](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) that determines how often they get set up and torn down. The scope can be one of:

* `function` (default)
* `class`
* `module`
* `package`
* `session`

If we don't have test classes we're leaving out an entire scope type, so we have fewer hooks for controlling how frequently fixture objects get instantiated. When we're looking for ways to speed up tests, I suspect finding cases where we can expand fixture scope is going to grant some easy wins.

tests/models/test_admin.py
leonardr: There may be some other way around this problem (and it might not even be a problem with pytest), but there are a lot of tests in the test suite that work that way now, because the test suite reports the first line of the docstring instead of the name of the function--making it difficult to locate the failing test.

Again, if this isn't a problem or we can make it not be a problem with pytest, I'm all for string literals.

tests/conftest.py
```diff
@@ -5,7 +5,891 @@
...
+    for table in reversed(Base.metadata.sorted_tables):
+        try:
+            engine.execute(table.delete())
+        except ProgrammingError:
+            pass
+
+    with engine.connect() as conn:
+        Base.metadata.create_all(conn)
+
+    # Avoid CannotLoadConfiguration errors related to CDN integrations.
```
leonardr: There's a chance this code isn't necessary anymore.

tests/conftest.py
leonardr: With https://github.com/NYPL-Simplified/circulation/pull/1691 we learned that there does seem to be code in setup methods that operates outside this nested transaction.

tests/conftest.py
```diff
@@ -5,7 +5,891 @@
...
+            if key in Configuration.instance:
+                del(Configuration.instance[key])
+
+        session.close()
+
+
+@pytest.fixture
+def initalize_data(db_session):
+    # Should all of core/model/__init__.py::SessionManager::initialize_data
+    # be in here instead of splitting out DeliveryMechanisms from DataSource & Genres?
```
leonardr: I think it should. The built-in DataSources and Genres should be in the database from the beginning and any test should be able to rely on their presence. Creating them on a test-by-test basis (which I assume is what happens now) is also a drag on performance.

leonardr: Same for the delivery mechanisms. I'm OK with leaving the other fixtures in place and making them inputs into this one.

Although, since this fixture takes db_session as an input, does that mean these objects are being created inside the database session and they get rolled back when the test is over? It would be faster and conceptually preferable to create these "built-in" database objects with Pytest session scope so they are part of the database that gets reverted to after a test runs.

testing.py
```diff
@@ -1535,8 +1535,8 @@ def session_fixture():
 
     # Drop any existing schema. It will be recreated when
     # SessionManager.initialize() runs.
-    engine = SessionManager.engine()
-    Base.metadata.drop_all(engine)
+    #engine = SessionManager.engine()
+    #Base.metadata.drop_all(engine)
```
leonardr: Before merging, remove this code rather than leave it commented out.

tests/models/test_collection.py
```diff
@@ -810,56 +930,62 @@ def expect(qu, works):
             qu, [overdrive_ebook, self_hosted_lcp_book, unlimited_access_book]
         )
 
-    def test_delete(self):
-        """Verify that Collection.delete will only operate on collections
-        flagged for deletion, and that deletion cascades to all
-        relevant related database objects.
+    def test_delete(self, monkeypatch, db_session, create_collection, create_edition, create_library,
```
leonardr: Is monkeypatch necessary here?

tests/models/test_identifier.py
```diff
@@ -179,131 +214,153 @@ def test_parse_urns(self):
...
         # We have no way of associating ISBNs with license pools.
         # If we try to parse an ISBN URN in a context that only accepts
         # URNs that can have associated license pools, we get an exception.
         isbn_urn = "urn:isbn:1449358063"
+
+        # Delete all DataSources to force the exception
+        rows = db_session.query(DataSource).all()
+        for row in rows:
+            db_session.delete(row)
+        db_session.flush()
```
leonardr, line 285: This shouldn't be necessary. The only way I can see this making sense is if some other piece of code is establishing a DataSource that provides LicensePools for ISBN, and that DataSource isn't being rolled back when the test ended.

tests/models/test_identifier.py
```diff
@@ -660,41 +765,46 @@ def get_entry_dict(entry):
...
+            'non_ascii_type_ascii_identifier_no_title',
+            'non_ascii_type_non_ascii_identifier_no_title',
+            'ascii_type_ascii_identifier_ascii_title',
+            'ascii_type_non_ascii_identifier_ascii_title',
+            'non_ascii_type_ascii_identifier_ascii_title',
+            'non_ascii_type_non_ascii_identifier_ascii_title',
+            'ascii_type_ascii_identifier_non_ascii_title',
+            'ascii_type_non_ascii_identifier_non_ascii_title',
+            'non_ascii_type_ascii_identifier_non_ascii_title',
+            'non_ascii_type_non_ascii_identifier_non_ascii_title'
```
leonardr: This is purely aesthetic, but it's more difficult for me to keep track of the parameters when the IDs are kept separately from the inputs. Can you confirm this setup makes more sense for you?

tests/models/test_library.py
```diff
@@ -1,81 +1,100 @@
...
         # Short name cannot contain a pipe character.
-        def set_to_pipe():
+        with pytest.raises(ValueError):
             library.library_registry_short_name = "foo|bar"
-        pytest.raises(ValueError, set_to_pipe)
 
         # You can set the short name to None. This isn't
         # recommended, but it's not an error.
         library.library_registry_short_name = None
+        assert None == library.library_registry_short_name # Do we need this assert?
```
leonardr: Yeah, I think so -- it lets us verify that this isn't silently implemented as a no-op or something.

tests/models/test_licensing.py
```diff
@@ -502,195 +650,255 @@ def test_update_availability_does_nothing_if_given_no_data(self):
...
+            """
+            edition = create_edition(db_session, identifier_id=id)
+            pool = create_licensepool(
+                db_session,
+                edition,
+                open_access=open_access,
+                data_source_name=data_source_name,
+                with_open_access_download=with_open_access_download
+            )
+            return pool
```
leonardr, line 720: I believe you mentioned this in our call. I'd like to find a way to make this a real fixture but this is OK for now.

tests/models/test_admin.py
leonardr: I agree that example.com is better (makes it clear to the reader the address is a stand-in) but since unit tests should never send actual email, changing this can wait.

tests/models/test_identifier.py
keithbauer: I agree that it can be difficult to keep track of the IDs, especially when there are a lot. I'll update this to use `pytest.param(...)` to keep the ID near the parameters.

leonardr (approved):
Overall this is very good -- putting consistent docstrings on the tests looks great. I think this is all right to merge for the sake of getting our pipeline unblocked. There are more improvements that can be made, but I don't see any problems or regressions, and this is already quite a large branch.

In particular, I agree 100% with Nick that "finding cases where we can expand fixture scope is going to grant some easy wins." Some of the fixtures you have now could be expanded to have session scope. Once we merge core into circulation, this will become more important, since every PR will be running the full test suite and it'll take a lot longer.

Add the comments Nick and I put in this issue to SIMPLY-4025 as additional action items, or resolve them if they're easy (like my comment about removing code rather than commenting it out) and I think this can be merged.

keithbauer:
I've implemented most of the feedback and added an action item comment to SIMPLY-4025 for the rest.

